PR #1860, opened 2014-01-12
Use calendar.timegm when calculating cookie expiration

sigmavirus24:
Fixes #1859

Credit: @lukasa

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -7,6 +7,7 @@ requests.utils imports from here, so be careful with imports.
 """
 
 import time
+import calendar
 import collections
 from .compat import cookielib, urlparse, urlunparse, Morsel
 
@@ -393,8 +394,8 @@ def morsel_to_cookie(morsel):
         expires = time.time() + morsel['max-age']
     elif morsel['expires']:
         time_template = '%a, %d-%b-%Y %H:%M:%S GMT'
-        expires = time.mktime(
-            time.strptime(morsel['expires'], time_template)) - time.timezone
+        expires = calendar.timegm(time.strptime(morsel['expires'],
+                                                time_template))
     return create_cookie(
         comment=morsel['comment'],
         comment_url=bool(morsel['comment']),


Lukasa:
We need to confirm that this will match the actual parsing logic of cookies before we merge this: we don't want to start messing up cookies created in this way.

sigmavirus24:
That's a good point @Lukasa. I'm guessing @gazpachoking might have a good idea. I'm just too tired to think this hard right now =P

Lukasa:
Me too, I'll take a look at this tomorrow.

kennethreitz:
Any update?

sigmavirus24:
Sorry I never got around to that. @Lukasa how about you?

Lukasa:
Uh, no, totally dropped the ball here. I'll put it on my list. =)

sigmavirus24:
@Lukasa how should we make sure this doesn't mess anything up?

Lukasa:
I'm not sure, to be honest. I'm wondering if there's a good way we can confirm our cookie-parsing code works in all locales.

kennethreitz:
Closing due to inactivity.

Lukasa:
Ok, I'm reopening this because I've finally validated that the stdlib does this. That suggests that it's the right thing to do. I'll have to do the merge manually, but I'd like to have this.

@sigmavirus24, one question: could we meaningfully add this to a 2.9.1, or should it wait for 2.10.0?

sigmavirus24:
I think this could be included in 2.9.1

requests/cookies.py
```diff
@@ -393,8 +394,8 @@ def morsel_to_cookie(morsel):
         expires = time.time() + morsel['max-age']
     elif morsel['expires']:
         time_template = '%a, %d-%b-%Y %H:%M:%S GMT'
```
sigmavirus24, line 396: Should the template have "GMT" in it?

Lukasa, line 396: The honest answer is "probably not", to be honest: the RFC really doesn't require it at all, and instead only specifies that the time is to be parsed as UTC.

Lukasa:
\o/ Finally fixed, nearly two years after the original patch was proposed!

PR #2396, opened 2015-01-02
Improving format (whitespace related PEP8).

0xfoo:
A few instances of to much or to few whitespace.


sigmavirus24:
requests does not follow PEP8. Kenneth disagrees with many of the suggestions. You can read more on #1543, #1964, and #2298

PR #2412, opened 2015-01-19
Remove entirely unnecessary and unused bits from requests.compat

sigmavirus24:
A tiny refactor I noticed while looking into a separate bug.


Lukasa:
Fine by me. :cake:

kennethreitz:
Hmmm, interesting.

PR #2415, opened 2015-01-20
Move noncebit to the only place it is used

### LINKED ISSUE #2408: Wrong digest header creation when server has qop="auth,auth-int"
I always get "401 unauthorized" when trying to log in to a server which writes qop="auth,auth-int" in its answer. The problem doesn't exist with Curl.
I debugged the requests code and probably found the problem in auth.py. The function build_digest_header in class HTTPDigestAuth calculates
    noncebit = "%s:%s:%s:%s:%s" % (nonce, ncvalue, cnonce, qop, HA2)
and gives back
    base += ', qop="auth", nc=%s, cnonce="%s"' % (ncvalue, cnonce)
to the server.

So in our client digest calculation we use qop="auth,auth-int", which we take from the server response, and make the server use qop="auth" in its calculation by giving back this header.

While debugging I can log in correctly if I change noncebit on the fly to
    noncebit = "%s:%s:%s:%s:%s" % (nonce, ncvalue, cnonce, "auth", HA2)
This would be my suggestion for the solution. Theqop for the calculation should be the same as we give back in our header.

sigmavirus24:
Since we only allow for "auth" qop-value, hardcode it

Fixes #2408

I want to add tests but I'm not sure a good way to write them. =(

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -124,13 +124,15 @@ class HTTPDigestAuth(AuthBase):
         s += os.urandom(8)
 
         cnonce = (hashlib.sha1(s).hexdigest()[:16])
-        noncebit = "%s:%s:%s:%s:%s" % (nonce, ncvalue, cnonce, qop, HA2)
         if _algorithm == 'MD5-SESS':
             HA1 = hash_utf8('%s:%s:%s' % (HA1, nonce, cnonce))
 
         if qop is None:
             respdig = KD(HA1, "%s:%s" % (nonce, HA2))
         elif qop == 'auth' or 'auth' in qop.split(','):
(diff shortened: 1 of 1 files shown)


berndschultze:
I merged the changes into my auth.py and it worked fine, both for my servers which offer qop="auth" and for those which offer qop="auth,auth-int".
From my point of view the issue #2408 is fixed by this commit.

Lukasa:
Testing this is a nightmare, it's super stateful and you have to patch time and all sorts of weirdness. Let's just not.

sigmavirus24:
I'm really considering spending a weekend or something adding tests around stuff and re-factoring it because we should be able to add tests for this kind of thing.

PR #2427, opened 2015-01-26
Bug fix: field uri in digest authentication should not be empty when enc...

luozhaoyu:
...ounter http redirections.

https://github.com/kennethreitz/requests/issues/2426

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -103,7 +103,8 @@ class HTTPDigestAuth(AuthBase):
         # XXX not implemented yet
         entdig = None
         p_parsed = urlparse(url)
-        path = p_parsed.path
+        #: path is request-uri defined in RFC 2616 which should not be empty
+        path = p_parsed.path or "/"
         if p_parsed.query:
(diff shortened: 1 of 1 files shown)


sigmavirus24:
I haven't read this yet, and I'm certain it really is this simple, but I want to re read the RFC about this regardless to make sure this applies to all cases. I suppose we've never had trouble with this because people were using URIs that had paths when authenticating. It doesn't hurt to be safe though. =)

Thanks for your patience @luozhaoyu

luozhaoyu:
@sigmavirus24 Sure! It is rare because I could only reproduce it when use a proxy and encounter a redirect...

sigmavirus24:
I'm comfortable with this. Sorry for the delay @luozhaoyu. I lost track of this.

Thoughts @Lukasa?

Lukasa:
Fine by me. =)

sigmavirus24:
Thanks @luozhaoyu

luozhaoyu:
@sigmavirus24 You're welcome!

PR #2428, opened 2015-01-27
Docs fixed

sepulchered:
Moved "Custom Headers" part after "More complicated POST requests" and before "POST a Multipart-Encoded File" in quickstart doc for adding headers is mentioned in the last one.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -169,6 +169,21 @@ download, the above is the preferred and recommended way to retrieve the
 content.
 
 
+Custom Headers
+--------------
+
+If you'd like to add HTTP headers to a request, simply pass in a ``dict`` to the
+``headers`` parameter.
+
+For example, we didn't specify our content-type in the previous example::
+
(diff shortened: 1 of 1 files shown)


Lukasa:
I think the correct fix is actually to come up with an example that doesn't require the `data` parameter. Custom headers should come before anything with 'more complicated' in the title. =)

sepulchered:
I see, I just decided that it would be enough because custom headers are not mentioned in "More complicated POST requests".

Lukasa:
Yeah, I can see that, but I think headers are the simpler topic. If you feel strongly about it though I can merge as-is and consider changing it later. =)

sepulchered:
Moved it back and set example to use only headers kwarg without data. :)

sepulchered:
Sorry for not being careful, now it finally seems like I fixed everything that was needed.

kennethreitz:
I think i prefer the existing form of this. Thank you, though!

PR #2434, opened 2015-02-02
Remove urllib3-specific section of iter_chunks

### LINKED ISSUE #2433: iter_lines method will always hold the last response in the server in a buffer
The implementation of the `iter_lines` and `iter_content` methods in `requests` means that when receiving line-by-line data from a server in "push" mode, the latest line received from the server will almost invariably be smaller than the `chunk_size` parameter, causing the final read operation to block.

 A good example of this is the Kubernetes `watch` api, which produces one line of JSON output per event, like this:

    {"type":"ADDED","object":{"kind":"Service","id":"kubernetes-ro","uid":"5718d954-a31a-11e4-8c74-20cf30467e62","creationTimestamp":"2015-01-23T11:10:25-05:00","selfLink":"/api/v1beta1/services/kubernetes-ro","resourceVersion":4,"apiVersion":"v1beta1","namespace":"default","port":80,"protocol":"TCP","labels":{"component":"apiserver","provider":"kubernetes"},"selector":null,"containerPort":0,"portalIP":"10.254.6.100"}}

If you compare the output of:

    import requests
    r = requests.get('http://localhost:8080/api/v1beta1/watch/services',
                     stream=True)

    for line in r.iter_lines():
        print line

With the output of `curl` running against the same URL, you will see
that the output from the Python code lags behind the output seen by
`curl ` by one line.

I was able to work around this behavior by writing my own `iter_lines`
method, which looks like this:

    def iter_lines(fd, chunk_size=1024):
        '''Iterates over the content of a file-like object line-by-line.'''

        pending = None

        while True:
            chunk = os.read(fd.fileno(), chunk_size)
            if not chunk:
                break

            if pending is not None:
                chunk = pending + chunk
                pending = None

            lines = chunk.splitlines()

            if lines and lines[-1]:
                pending = lines.pop()

            for line in lines:
                yield line

        if pending:
            yield(pending)

This works around the problem partly by calling `os.read`, which will
happily return fewer bytes than requested in `chunk_size`.  With the
above routing available, the following code behaves correctly:

    import requests
    r = requests.get('http://localhost:8080/api/v1beta1/watch/services',
                     stream=True)

    for line in iter_lines(r.raw):
        print line

This code will always print out the most recent reply from the server
when it is received.

larsks:
This commit resolves #2433 by using os.read() on self.raw.fileno()
in all cases, and removes the use fo the self.raw.stream() iterator
which would block rather than returning short reads.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -7,6 +7,7 @@ requests.models
 This module contains the primary objects that power Requests.
 """
 
+import os
 import collections
 import datetime
 
@@ -649,24 +650,11 @@ class Response(object):
         available encoding based on the response.
         """
         def generate():
-            try:
-                # Special case for urllib3.
-                try:
-                    for chunk in self.raw.stream(chunk_size, decode_content=True):
-                        yield chunk
-                except ProtocolError as e:
-                    raise ChunkedEncodingError(e)
-                except DecodeError as e:
-                    raise ContentDecodingError(e)
-                except ReadTimeoutError as e:
-                    raise ConnectionError(e)
-            except AttributeError:
-                # Standard file-like object.
-                while True:
-                    chunk = self.raw.read(chunk_size)
-                    if not chunk:
-                        break
-                    yield chunk
+            while True:
+                chunk = os.read(self.raw.fileno(), chunk_size)
+                if not chunk:
+                    break
+                yield chunk
 
             self._content_consumed = True
 


Lukasa:
I wonder if this shouldn't go into urllib3 instead, thereby giving the benefit there too. @shazow? @sigmavirus24?

shazow:
I'd like that.

larsks:
The existing documentation for `urllib3.response.stream` says:

> A generator wrapper for the read() method. A call will block until
> `amt` bytes have been read from the connection or until the
> connection is closed.

So from that perspective things are behaving as expected (it's supposed to block).  So...would the fix be a new method, or a boolean (e.g., `allow_short_reads`), or something else?

shazow:
Hm, actually are we sure that doing something like .fileno() will work on AppEngine etc?

larsks:
@shazow Maybe not?  Looking at the code in `urllib3.response`, `self._fp` is set from `body`, which could be a string type instead of a file pointer.

...oh wait, that's a lie, it couldn't be a string type in that case, but it only checks for a `read` method so it might not have `fileno()`, I guess.

Lukasa:
Mmm, it's not clear that we can safely do this. A potential optimisation, though?

shazow:
I'm open to have it use this method if it's of the right type, if that helps. :)

sigmavirus24:
This breaks decoding of content too, no?

PR #2436, opened 2015-02-03
Fix for failing test: test_connection_error

colindickson:
http://fooobarbangbazbing.httpbin.org does not seem to be an unknown domain anymore.  I've also split up this unit test into two tests since they test slightly different things.

### DIFF
--- a/test_requests.py
+++ b/test_requests.py
@@ -301,11 +301,13 @@ class RequestsTestCase(unittest.TestCase):
         r = s.get(url)
         assert r.status_code == 200
 
-    def test_connection_error(self):
+    def test_connection_error_invalid_domain(self):
         """Connecting to an unknown domain should raise a ConnectionError"""
         with pytest.raises(ConnectionError):
-            requests.get("http://fooobarbangbazbing.httpbin.org")
+            requests.get("http://doesnotexist.google.com")
 
+    def test_connection_error_invalid_port(self):
+        """Connecting to an invalid port should raise a ConnectionError"""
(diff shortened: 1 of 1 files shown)


test_requests.py
```diff
@@ -301,11 +301,13 @@ def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self):
...
+    def test_connection_error_invalid_domain(self):
         """Connecting to an unknown domain should raise a ConnectionError"""
         with pytest.raises(ConnectionError):
-            requests.get("http://fooobarbangbazbing.httpbin.org")
+            requests.get("http://doesnotexist.google.com")
 
+    def test_connection_error_invalid_port(self):
+        """Connecting to an invalid port should raise a ConnectionError"""
         with pytest.raises(ConnectionError):
             requests.get("http://httpbin.org:1")
```
kevinburke, line 312: Huh. This line hangs on my machine and I'm pretty sure I wrote this test. Maybe your local machine is sending back a TCP RST, mine isn't

kevinburke:
LGTM

test_requests.py
kevinburke, line 312: Passes on ci.kennethreitz.org anyway

Lukasa:
Thanks! :cake:

test_requests.py
colindickson, line 312: @kevinburke i will make another PR today which will fix the hang.   (i didn't notice the hang last night on my windows machine, but i do have it on my ubuntu machine.)   cheers.

PR #2437, opened 2015-02-03
quickstart: using a list as a value in query params

tomscytale:
Update docs  - how to pass a list of values in a query


Lukasa:
:cake: :star: Thanks!

PR #2438, opened 2015-02-04
add a timeout value to connection_error_invalid_port test to accelerate failure

colindickson:
As mentioned here https://github.com/kennethreitz/requests/pull/2436 , one of the tests seemed like it was hanging.   In fact, the test was just (correctly) waiting to hit its timeout value.  We now give it a shorter timeout value to accelerate the testing process.


Lukasa:
Yeah, this feels sensible to me. :cake:

PR #2439, opened 2015-02-04
Suppress the warning if verify=False

### LINKED ISSUE #2214: verify=False and requests.packages.urllib3.disable_warnings()
As of 1.9 of `urllib3`, the following warning appears once per invocation:
``` 
/usr/local/lib/python2.7/site-packages/requests-2.4.0-py2.7.egg/requests/packages/urllib3/connectionpool.py:730: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html (This warning will only appear once by default.)
  InsecureRequestWarning)
```

When using `verify=False` would it be useful to also set `requests.packages.urllib3.disable_warnings()`?

I understand that this is a design decision that not everyone might agree with. :)

zaitcev:
Setting verify to False differs from verify=None in being explicitly set,
and therefore the warning being undesirable.

See this for the a proposed alternative that involves importing
implementation details of Requests onto the application:
 https://github.com/kennethreitz/requests/issues/2214#issuecomment-72941896

This fix avoids deep imports and pulling the knowledge of internals of
Requests into the application.

Fixes #2214

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -183,7 +183,12 @@ class HTTPAdapter(BaseAdapter):
             conn.cert_reqs = 'CERT_REQUIRED'
             conn.ca_certs = cert_loc
         else:
-            conn.cert_reqs = 'CERT_NONE'
+            if verify is False:
+                # Forced by the caller
+                conn.cert_reqs = 'CERT_EXPLICITLY_NONE'
+            else:
+                # Happened to be none
+                conn.cert_reqs = 'CERT_NONE'
             conn.ca_certs = None
 
         if cert:
--- a/requests/packages/urllib3/connection.py
+++ b/requests/packages/urllib3/connection.py
@@ -253,7 +253,8 @@ class VerifiedHTTPSConnection(HTTPSConnection):
             match_hostname(cert, self.assert_hostname or hostname)
 
         self.is_verified = (resolved_cert_reqs == ssl.CERT_REQUIRED
-                            or self.assert_fingerprint is not None)
+                            or self.assert_fingerprint is not None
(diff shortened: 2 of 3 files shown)


Lukasa:
Some notes:
1. urllib3 is a separate project that we vendor in its entirety. This PR affects it, and so requires a separate change in that library.
2. 'Explicit' is dependent on who you are. This is 'explicit' to the person who calls requests, but not to the person above them, if any. The warning is not just for the person writing against requests, but for anyone who uses an application that uses requests.

I'm -1 on this change.

sigmavirus24:
`CERT_EXPLICITLY_NONE` means nothing and doesn't add much value here. Given that this is mostly in urllib3 and would rely on acceptance there, I'm closing this until progress has been made there.

zaitcev:
Sorry, but the closing statement is false. CERT_EXPLICITLY_NONE means CERT_NONE set explicitly, and its value is making the distinction between such case and CERT_NONE set explicitly, purely by the lack of certificates.

Lukasa:
@zaitcev The closing statement was: "Given that this is mostly in urllib3 and would rely on acceptance there, **I'm closing this until progress has been made there**." (Emphasis mine.)

sigmavirus24:
@zaitcev `CERT_NONE`, `CERT_REQUIRED` and `CERT_OPTIONAL` are all defined on the `ssl` module and have a real meaning in the context of performing validation. `CERT_EXPLICITLY_NONE` means absolutely nothing in the same context.

PR #2440, opened 2015-02-06
Update to use readthedocs.org instead of rtfd.org

aquarion:
https://toolbelt.rtfd.org currently causes an SSL error because the wildcard SSL cert is for *.readthedocs.org.


sigmavirus24:
Who added this link?

Thanks for fixing this @acquarion!

PR #2441, opened 2015-02-08
Raise explicitly when exclusive parameters are provided

andreif:
It's a common case when one sends multipart request with json meta data. Currently, it is not clear that `json` argument is silently omitted. The same happens when both `data` and `json` are provided.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -422,6 +422,8 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         length = None
 
         if json is not None:
+            if files is not None:
+                raise NotImplementedError('JSON data and files are mutually exclusive.')
             content_type = 'application/json'
             body = json_dumps(json)
 
@@ -456,6 +458,8 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
                         content_type = None
                     else:
                         content_type = 'application/x-www-form-urlencoded'
+                elif data and json is not None:
+                    raise NotImplementedError('JSON data and generic body data are mutually exclusive.')
 
             self.prepare_content_length(body)
 


sigmavirus24:
Thanks for the PR @andreif!

I'm mostly -0.5 on this. I think the right approach is to fix the documentation, since that's clearly what's lacking here (in fact, I could find no examples of using the json parameter actually, and I take full blame for that).

That said, the API for multipart requests seems confusing enough that I think some really good documentation around the `data`, `json`, and `files` parameters is really necessary.

Further, I think if we are going to raise an exception here, it should be a `ValueError`. `NotImplementedError` is usually used for abstract methods that need to be overridden by a subclass. That's not what we expect people to do here. What actually happens here, the problem is that we technically received to many values and can't possibly know what the users want us to do.

andreif:
@sigmavirus24 Well, I wonder if we could take both files and json. It's a common case to upload files with json meta data, so maybe it makes sense to have a nice api for this.

andreif:
It would be nice to write

``` py
r = requests.post('https://www.googleapis.com/upload/drive/v2/files',
                  params={'uploadType': 'multipart'},
                  headers={'Authorization': 'Bearer ' + access_token},
                  json={'title': 'test.txt'},
                  files={'file': open('test.txt', 'rb')})
```

instead of

``` py
r = requests.post('https://www.googleapis.com/upload/drive/v2/files',
                  params={'uploadType': 'multipart'},
                  headers={'Authorization': 'Bearer ' + access_token},
                  files={'file': open('test.txt', 'rb'),
                         None: (None, json.dumps({'title': 'test.txt'}),
                                  'application/json; charset=UTF-8')})
```

andreif:
I could look into it if you are fine with accepting json argument together with files

Lukasa:
@andreif That's a good idea, but sadly the API doesn't render out well. For instance, what happens if instead of uploading one file you're uploading two? And how do you associate them together?

Sadly, JSON in mutlipart/form-data messages is not standardised in any sense, so there's no way we can extend our API for that to be consistently useful. In that situation I'd want to resist the temptation to guess.

However, I'm +0 on throwing exceptions when mutually exclusive parameters are passed, because I don't really think we should fail silently. That said, it's backwards incompatible, and so we need to hold it for at least a minor release.

andreif:
@Lukasa Alright. Should I change to `ValueError` in this case?

sigmavirus24:
@andreif please do change it to a ValueError

andreif:
@sigmavirus24 I took existing raise as an example. Should I change all of them to `ValueError` or should I keep the original as it is?

``` py
raise NotImplementedError('Streamed bodies and files are mutually exclusive.')
```

kennethreitz:
This should be solved with documentation.

PR #2450, opened 2015-02-19
Update README to use Shields badges

iKevinY:
The SVG badges served by [Shields](http://shields.io) are friendlier to retina displays and have a standardized design.

Shields Comparison

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,11 +1,11 @@
 Requests: HTTP for Humans
 =========================
 
-.. image:: https://badge.fury.io/py/requests.png
-    :target: http://badge.fury.io/py/requests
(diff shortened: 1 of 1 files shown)


Lukasa:
No objection from me. @sigmavirus24?

iKevinY:
One thing I noticed is that there's less precision in the download count – not sure if this problematic.

sigmavirus24:
Meh. Download counts are working exactly correctly with PyPI at the moment anyway.

sigmavirus24:
Thanks @iKevinY !

iKevinY:
No worries!

Issue #1155, opened 2013-01-31
Add MultiDict Support


dmckeone:
[Comment by Kenneth Reitz](https://github.com/kennethreitz/requests/pull/1103#issuecomment-12370830): 

> We need to support MultiDict. This is long overdue.

kennethreitz:
+1

sigmavirus24:
Should the values be ordered or not? I have no preference, I'm just wondering if others do.

tlong-dev:
For what? Headers? Query string?

Typically I would say yes either way. Preserving order does not occlude the un-ordered use case whereas the opposite does.

sigmavirus24:
@xealot yeah, that was my opinion. I have something working currently, but it so minimal that I'm not sure it is covers all the corner cases. .... I'll be pushing it to a branch in my repo tonight.

dmckeone:
+1 for ordered support.  The impetus for this feature was discovered when using `werkzeug.datastructures.OrderedMultiDict` with GET/POST parameters, so if ordered support wasn't there it wouldn't hit the use case.

I'm not very familiar with what other MultiDicts are in the Python community, but I think the only thing that is really needed (at least for werkzeug MultiDict) is to pass `items(multi=True)` or `iteritems(multi=True)` when iterating instead of just `items()` or `iteritems()`.  Because Requests used to work with lists of 2-tuples, it seems to support this behaviour almost everywhere, it just needs to iterate slightly differently when breaking up the dictionary into key-value pairs.   If all MultiDicts behaved this way (with the same parameter) then perhaps a solution like:

```
# requests.utils 

def to_key_val_list(value):

    ...

    if isinstance(value, dict):
        try:
            value = value.items(multi=True)
        except TypeError:
            value = value.items()
```

could work to take any dict that behaves like a MultiDict without having to require a separate Requests-only dictionary.

The problem is that this solution would be specific to Werkzeug, so perhaps sigmavirus24's in-progress solution is better;  force it into a known type that is specific to Requests, and then know that it's always that type.  

I'm not familiar enough with what is out there in the Python community, or what is in Requests, to make that judgement call.  However, I do know that the 'behave-as-if' solution mentioned above would be a little more user-friendly because the user would not be required to create a new object instance prior to calling Requests.

sigmavirus24:
[...]

PR #2453, opened 2015-02-20
document combination of repeated response headers

requiredfield:
as explained in https://github.com/kennethreitz/requests/issues/1155#issuecomment-53229867.

motivation:

requests uses a dict for response headers. In a given response, the same header may appear multiple times with different values, but dicts of course allow each key to map to only one value. requests [handles this intelligently](https://github.com/kennethreitz/requests/issues/1155#issuecomment-53229867), but users will not expect this given it's not documented in any of these places:

- http://docs.python-requests.org/en/latest/user/quickstart/#response-headers
- http://docs.python-requests.org/en/latest/api/?highlight=headers#requests.request
- http://docs.python-requests.org/en/latest/api/#requests.Response.headers

(Also, they may (likely) be unfamiliar with this part of http://tools.ietf.org/html/rfc7230#section-3.2, while only actually being familiar with werkzeug and other libraries that use a multidict instead to handle repeated headers.)

I think there's a perfect spot to document this behavior (right after the docs explain how the dict is already special in that it's case-insensitive), and given that #1155 will probably be closed, I figured it's a good time to document this and increase users' future understanding.

Thanks for the great work on requests.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -347,6 +347,15 @@ So, we can access the headers using any capitalization we want::
     >>> r.headers.get('content-type')
     'application/json'
 
+It is also special in that the server could have sent the same header multiple
+times with different values, but requests combines them so they can be
+represented in the dictionary within a single mapping, as per
+`RFC 7230 <http://tools.ietf.org/html/rfc7230#section-3.2>`_:
+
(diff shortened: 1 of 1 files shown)


Lukasa:
:cake: Beautiful. =)

requiredfield:
Thanks for merging. =)

Issue #2476, opened 2015-03-06
iter_lines: Skipping first line using next(..)


plaes:
When trying to skip first line of  `response.iter_lines()` using `next()`, some extra data from following lines is also consumed.

Test script:
```python
import requests
from contextlib import closing

with closing(requests.get('http://httpbin.org/stream/5', stream=True)) as r:
    skip = next(r.iter_lines())
    print 'SKIP: ', skip
    for n, l in enumerate(r.iter_lines(), start=1):
        print 'PROCESS: {}'.format(n), l
```

With following output (line 1 totally missing, and line 2 being partial):
```
SKIP:  {"url": "http://httpbin.org/stream/5", "headers": {"Host": "httpbin.org", "Accept-Encoding": "gzip, deflate", "Accept": "*/*", "User-Agent": "python-requests/2.4.3 CPython/2.7.9 Linux/4.0.0-rc2+"}, "args": {}, "id": 0, "origin": "xxx"}
PROCESS: 1 n.org/stream/5", "headers": {"Host": "httpbin.org", "Accept-Encoding": "gzip, deflate", "Accept": "*/*", "User-Agent": "python-requests/2.4.3 CPython/2.7.9 Linux/4.0.0-rc2+"}, "args": {}, "id": 2, "origin": "xxx"}
PROCESS: 2 {"url": "http://httpbin.org/stream/5", "headers": {"Host": "httpbin.org", "Accept-Encoding": "gzip, deflate", "Accept": "*/*", "User-Agent": "python-requests/2.4.3 CPython/2.7.9 Linux/4.0.0-rc2+"}, "args": {}, "id": 3, "origin": "xxx"}
PROCESS: 3 {"url": "http://httpbin.org/stream/5", "headers": {"Host": "httpbin.org", "Accept-Encoding": "gzip, deflate", "Accept": "*/*", "User-Agent": "python-requests/2.4.3 CPython/2.7.9 Linux/4.0.0-rc2+"}, "args": {}, "id": 4, "origin": "xxx"}
```

sigmavirus24:
This is a tricky interaction with iterators. Most iterators don't quite behave the same way. What you need to do is the following

``` py
with closing(requests.get('https://httpbin.org/stream/5', stream=True)) as r:
    iterator = r.iter_lines()
    print 'SKIP: ', next(iterator)
    for n, l in enumerate(iterator, start=1):
        print 'PROCESS: {}'.format(n), l
```

[...]

PR #2477, opened 2015-03-06
Make Response.iter_lines() reentrant safe

plaes:
Calling r.iter_lines() multiple times doesn't lose pending data between
calls to this method.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -541,6 +541,9 @@ class Response(object):
         self._content = False
         self._content_consumed = False
 
+        # Internal buffer of partially read lines for self.iter_lines()
+        self._pending = None
+
         #: Integer Code of responded HTTP Status, e.g. 404 or 200.
         self.status_code = None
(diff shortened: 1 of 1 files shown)


plaes:
This is fix for issue #2476. So either this, or documentation should mention that Response.iter_lines() is not reentrant safe.

Lukasa:
Eh, fine by me. @sigmavirus24?

sigmavirus24:
I feel likewise "Eh". This is just going to be broken if there aren't any tests for it. It also breaks the thread-safety of a response I think. (I would hope someone's not consuming a response from multiple threads, but this makes the iterator less threadsafe.)

plaes:
Yup, this approach is broken. I'll see whether I can find a way to fix it.

plaes:
I'm closing this for now.. and submitted documentation about this issue and testcases in another pull requests.

PR #2478, opened 2015-03-07
Document Response.iter_lines() deficiencies

plaes:


Lukasa:
@plaes This is a great PR. One small note however, we don't use `unittest`, we use `pytest`. This is why your test suite failed: `unittest.expectedFailure` doesn't exist in Python 2.6. Instead, use [`pytest.mark.xfail`](http://pytest.org/latest/skipping.html#mark-a-test-function-as-expected-to-fail).

plaes:
Oops.. I push forced the fix.

Lukasa:
Hurrah! Thanks so much for this @plaes!

Lukasa:
:cake: :star:

sigmavirus24:
Thanks @plaes!

PR #2484, opened 2015-03-11
Sanitise verb string. Now we can use unicode strings in http methods.

pawlyk:
There is error (`UnicodeDecodeError: 'ascii' codec can't decode...`) when use unicode string as http verb in instance Request.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -229,7 +229,7 @@ class Request(RequestHooksMixin):
         for (k, v) in list(hooks.items()):
             self.register_hook(event=k, hook=v)
 
-        self.method = method
+        self.method = to_native_string(method)
         self.url = url
         self.headers = headers
         self.files = files
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -681,6 +681,9 @@ def to_native_string(string, encoding='ascii'):
     """
     out = None
 
+    if not string:
+        return string
+
     if isinstance(string, builtin_str):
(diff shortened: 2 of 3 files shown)


Lukasa:
Interesting.

Why do you want to do this? RFC 7230 is pretty unambiguous about what's allowed in a HTTP METHOD: namely, `token`. The ABNF defines `token` as:

```
 token          = 1*tchar

 tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                / DIGIT / ALPHA
                ; any VCHAR, except delimiters

 DIGIT          =  %x30-39
                        ; 0-9

 ALPHA          =  %x41-5A / %x61-7A   ; A-Z / a-z

 VCHAR          =  %x21-7E
                        ; visible (printing) characters
```

That would seem to guarantee that any valid method is necessarily representable in ASCII. So what's the rationale for allowing non-ASCII methods?

pawlyk:
I agree with you about RFC 7230. But documentation does't have mention that http verbs must be in ASCII, there are only list of allowed verbs. Maybe better clarify this part.

Lukasa:
That seems like a reasonable decision to me. I'll happily accept any pull request that updates the documentation (probably by updating the docstrings in `requests.request`, `requests.models.Request`, and `session.request`).

PR #2485, opened 2015-03-12
Import urllib3's Retry location from the right place

sigmavirus24:
Importing from urllib3's top-level location causes the namespace to be
urllib3.util.retry.Retry instead of
requests.packages.urllib3.util.retry.Retry. Without this fix, an using
requests with an un-vendored version of urllib3 will break when urllib3's
retry handling kicks in.

Closes shazow/urllib3#567

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -11,10 +11,10 @@ and maintain connections.
 import socket
 
 from .models import Response
-from .packages.urllib3 import Retry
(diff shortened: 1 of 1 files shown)


sigmavirus24:
@Lukasa feel free to merge this so we can release 2.5.5 tomorrow.

Lukasa:
:cake:

PR #2493, opened 2015-03-14
made the quickstart more reader friendly

### LINKED ISSUE #2418: Data argument introduced too early in Quickstart guide
I'm following the quick start guide in:
http://docs.python-requests.org/en/latest/user/quickstart/
I got confused when reading the section "Custom Headers" because the example code features the data argument, which hasn't been introduced in the guide yet (it is the next section "More complicated POST requests" that describes the data argument).
My suggestion is to swap the order of those two sections, i.e. have "More complicated POST requests" before "Custom Headers", or modify the example code in "Custom Headers" (remove data argument).

yasoob:
Closes #2418

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -186,7 +186,7 @@ For example, we didn't specify our content-type in the previous example::
 
     >>> import json
     >>> url = 'https://api.github.com/some/endpoint'
-    >>> headers = {'user-agent': 'Mozilla/5.0 (Windows NT 6.2; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/40.0.2214.115 Safari/537.36'}
+    >>> headers = {'user-agent': 'my-app/0.0.1'}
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Can we choose a better header to set? Or a better value? I can't see how a long user-agent string will be reader friendly. Some people will now start asking why all that information is necessary and what it means and why it's important. Why not use `headers={'Connection': 'close'}`?

Lukasa:
Or just shorten the user-agent. `User-Agent: my-app/0.0.1` should do just fine.

yasoob:
Made the string shorter.

Lukasa:
Hurrah! :cake:

yasoob:
Proud of my first contribution to requests :+1:  Here's a :cake: for everyone :smile:

Lukasa:
@yasoob Feel free to open a pull request adding yourself to AUTHORS.rst!

PR #2494, opened 2015-03-15
"A CVE identifier"

Ayrx:
I believe this is the more correct grammar?


Lukasa:
I believe it is too, however a CVE has now been assigned for this, so I'm going to update the changelog to refer to it instead. Thanks for spotting it though!

PR #2504, opened 2015-03-18
fix resolve redirect to pass all original arguments

pvanderlinden:
Fix for https://github.com/kennethreitz/requests/issues/2503
It is still a bit inconsistent but as mentioned in the issue, but it otherwise will break the current public api. 

This makes it possible for non urllib3 adapters to accept extra arguments.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -90,7 +90,7 @@ def merge_hooks(request_hooks, session_hooks, dict_class=OrderedDict):
 
 class SessionRedirectMixin(object):
     def resolve_redirects(self, resp, req, stream=False, timeout=None,
-                          verify=True, cert=None, proxies=None):
+                          verify=True, cert=None, proxies=None, **adapter_kwargs):
         """Receives a Response. Returns a generator of Responses."""
 
         i = 0
@@ -193,6 +193,7 @@ class SessionRedirectMixin(object):
                 cert=cert,
                 proxies=proxies,
                 allow_redirects=False,
+                **adapter_kwargs
             )
 
             extract_cookies_to_jar(self.cookies, prepared_request, resp.raw)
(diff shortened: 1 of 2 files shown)


Lukasa:
I think I'd like the `kwargs` parameter in `resolve_redirects` to have a different name: `adapter_kwargs`, maybe?

pvanderlinden:
I have changed the parameter.

Lukasa:
LGTM. The one thing I'm thinking about is whether we want to allow for forward compatibility by actually storing these kwargs in a dictionary for resolve_redirects, but I haven't decided yet. I'll let @sigmavirus24 make a call here.

requests/sessions.py
```diff
@@ -560,10 +561,6 @@ def send(self, request, **kwargs):
         # Set up variables needed for resolve_redirects and dispatching of hooks
         allow_redirects = kwargs.pop('allow_redirects', True)
         stream = kwargs.get('stream')
```
sigmavirus24: Does stream need to be extracted here?

sigmavirus24:
> The one thing I'm thinking about is whether we want to allow for forward compatibility by actually storing these kwargs in a dictionary for resolve_redirects

Could you expand on that @Lukasa? I'm not quite certain what you mean.

pvanderlinden:
Any thoughts? I would like to get this merged as it is preventing me from adding a feature to my adapter (making it open source soon).

Lukasa:
Sorry, what I meant was I'm wondering whether we should, rather than using `**kwargs` the whole way through, save the initial `**kwargs` in a dictionary which is then passed _as a dictionary_. That way, if we ever feel the need to add `**kwargs` to any of these functions later we could.

I don't think it's a big deal though.

sigmavirus24:
I'm okay with this if you are @Lukasa.

Lukasa:
Sure, why not. =)

pvanderlinden:
Thanks!

PR #2506, opened 2015-03-18
Expand on chunked handling.

Lukasa:
Once we merge in the new urllib3, this will update the docs to reflect the new best way to handle chunked transfer encoding.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -278,7 +278,8 @@ For chunked encoded responses, it's best to iterate over the data using
 :meth:`Response.iter_content() <requests.models.Response.iter_content>`. In
 an ideal situation you'll have set ``stream=True`` on the request, in which
 case you can iterate chunk-by-chunk by calling ``iter_content`` with a chunk
-size parameter of ``None``.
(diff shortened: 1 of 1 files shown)


TomasTomecek:
looks fine to me

how about specifying the size of a chunk? will it be possible?

Lukasa:
Yup, passing the chunk size parameter to `iter_content` will do it. What you _can't_ do is change that size from chunk to chunk.

TomasTomecek:
You are right! Should have checked the code.

The reason I asked because I was slightly confused by this sentence. I would probably update it to:

``` diff
-iterate chunk-by-chunk by calling ``iter_content`` with a chunk
-size parameter of ``None``.
+iterate chunk-by-chunk by calling ``iter_content`` (it is ``None``
+by default, but you can set chunk size yourself by providing an ``int``).
```

But you're the native speaker here. So I'm assuming that's just me not getting it.

Lukasa:
Agree that it needs rewording, though that wording isn't quite right (it's not `None` by default, it's `1`).

sigmavirus24:
pong

Lukasa:
Always be updating PRs.

PR #2513, opened 2015-03-23
Minor Patch TypeError thrown

Montycarlo:
Python2.7, str(err) was throwing a TypeError as it doesn't override the 'toString' method __str__.

Wrapped offending block in a quick try block, catching only TypeError exceptions.


Lukasa:
Thanks! :cake:

sigmavirus24:
@Montycarlo please see #2514. This should have been opened at [urllib3](/shazow/urllib3) and the change needs to be much better since this will introduce a breaking change to urllib3.

Lukasa:
Sorry @Montycarlo! I should have caught that earlier. =(

ropwareJB:
Sorry for wasting your time! Keep up the awesome work :)

PR #2514, opened 2015-03-23
Revert "Minor Patch TypeError thrown"

sigmavirus24:
Reverts kennethreitz/requests#2513

The change was made to urllib3 which is vendored and to which pull requests should not be accepted.


sigmavirus24:
Further, that breaks the API of urllib3 by raising an exception that is not a urllib3 exception and would introduce a breaking change to requests.

Lukasa:
Yeah, this is straight up my bad.

PR #2523, opened 2015-04-02
HTTPDigestAuth - Making it thread-safe

exvito:
The existing code counts the number of 401 responses in the num_401_calls
authenticator attribute. This is in place so as to ensure the necessary auth
header is sent, while avoiding infinite 401 loops (issue #547).

This commit makes num_401_calls an instance of threading.local() (previously
an integer), using num_401_calls.value as the counter.

It ensures that concurrent authentication requests get each their own counter
and behave as expected (otherwise every other concurrent request would have
its authentication fail).

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -65,25 +65,25 @@ class HTTPDigestAuth(AuthBase):
         self.username = username
         self.password = password
         # Keep state in per-thread local storage
-        self.tl = threading.local()
+        self._thread_local = threading.local()
 
     def init_per_thread_state(self):
         # Ensure state is initialized just once per-thread
-        if not hasattr(self.tl, 'init'):
-            self.tl.init = True
-            self.tl.last_nonce = ''
-            self.tl.nonce_count = 0
-            self.tl.chal = {}
-            self.tl.pos = None
-            self.tl.num_401_calls = None
+        if not hasattr(self._thread_local, 'init'):
+            self._thread_local.init = True
+            self._thread_local.last_nonce = ''
+            self._thread_local.nonce_count = 0
+            self._thread_local.chal = {}
+            self._thread_local.pos = None
+            self._thread_local.num_401_calls = None
 
     def build_digest_header(self, method, url):
 
-        realm = self.tl.chal['realm']
-        nonce = self.tl.chal['nonce']
-        qop = self.tl.chal.get('qop')
-        algorithm = self.tl.chal.get('algorithm')
-        opaque = self.tl.chal.get('opaque')
+        realm = self._thread_local.chal['realm']
+        nonce = self._thread_local.chal['nonce']
+        qop = self._thread_local.chal.get('qop')
+        algorithm = self._thread_local.chal.get('algorithm')
+        opaque = self._thread_local.chal.get('opaque')
 
         if algorithm is None:
             _algorithm = 'MD5'
@@ -121,12 +121,12 @@ class HTTPDigestAuth(AuthBase):
         HA1 = hash_utf8(A1)
         HA2 = hash_utf8(A2)
 
-        if nonce == self.tl.last_nonce:
-            self.tl.nonce_count += 1
+        if nonce == self._thread_local.last_nonce:
+            self._thread_local.nonce_count += 1
         else:
-            self.tl.nonce_count = 1
-        ncvalue = '%08x' % self.tl.nonce_count
-        s = str(self.tl.nonce_count).encode('utf-8')
+            self._thread_local.nonce_count = 1
+        ncvalue = '%08x' % self._thread_local.nonce_count
+        s = str(self._thread_local.nonce_count).encode('utf-8')
         s += nonce.encode('utf-8')
         s += time.ctime().encode('utf-8')
         s += os.urandom(8)
@@ -146,7 +146,7 @@ class HTTPDigestAuth(AuthBase):
             # XXX handle auth-int.
             return None
 
-        self.tl.last_nonce = nonce
+        self._thread_local.last_nonce = nonce
 
         # XXX should the partial digests be encoded too?
         base = 'username="%s", realm="%s", nonce="%s", uri="%s", ' \
@@ -165,23 +165,22 @@ class HTTPDigestAuth(AuthBase):
     def handle_redirect(self, r, **kwargs):
         """Reset num_401_calls counter on redirects."""
         if r.is_redirect:
-            self.tl.num_401_calls = 1
+            self._thread_local.num_401_calls = 1
 
     def handle_401(self, r, **kwargs):
         """Takes the given response and tries digest-auth, if needed."""
 
-        if self.tl.pos is not None:
+        if self._thread_local.pos is not None:
             # Rewind the file position indicator of the body to where
             # it was to resend the request.
-            r.request.body.seek(self.tl.pos)
-        num_401_calls = self.tl.num_401_calls
+            r.request.body.seek(self._thread_local.pos)
         s_auth = r.headers.get('www-authenticate', '')
 
-        if 'digest' in s_auth.lower() and num_401_calls < 2:
+        if 'digest' in s_auth.lower() and self._thread_local.num_401_calls < 2:
 
-            self.tl.num_401_calls += 1
+            self._thread_local.num_401_calls += 1
             pat = re.compile(r'digest ', flags=re.IGNORECASE)
-            self.tl.chal = parse_dict_header(pat.sub('', s_auth, count=1))
+            self._thread_local.chal = parse_dict_header(pat.sub('', s_auth, count=1))
 
             # Consume content and release the original connection
             # to allow our new request to reuse the same one.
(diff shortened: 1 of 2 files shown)


requests/auth.py
```diff
@@ -166,12 +167,12 @@ def handle_401(self, r, **kwargs):
             # Rewind the file position indicator of the body to where
             # it was to resend the request.
             r.request.body.seek(self.pos)
-        num_401_calls = getattr(self, 'num_401_calls', 1)
+        num_401_calls = getattr(self.num_401_calls, 'value', 1)
```
Lukasa: Why do we still have the `getattr` here?

exvito: I just adapted the existing code and as such left the `getattr` as it was. After inspection it doesn't seem needed, however: no fallback to default 1 will ever occur because the (now thread-local) value is assigned in `__call__`, when the authentication process is about to "begin".

Additionally, it shouldn't be needed in the code before this commit: `self.num_401_calls` was being assigned in `__init__`.

Would you like me to update the PR with an extra commit?

kennethreitz: @exvito that would be wonderful! notify us when you do.

exvito:
@Lukasa , @kennethreitz - Per your suggestion, I just updated the PR with commit e65360d.
Thanks for your input.

tardyp:
I think the code is racy in nearly all the areas, not just the 401 counter.
- build_digest_header is updating a number of internal variables including nonce lastnonce, etc.
- if there are several unauthenticated request at the same time, you will get several challenges, but stored in the same variables for subsequent request.

I think a more stable approach would be to store the whole state in a per thread variable. This has the drawback of doing one auth per thread, instead of trying to reuse the chalenge data from other threads.

exvito:
@tardyp : You are correct. This needs further improvement, just as you point out there are other attributes currently shared accross threads that will lead to auth failures in certain conditions.

I'll extend the per-thread approach of num_401_calls to the remaining state-tracking attributes and update the PR for review.

Regarding the drawback you refer to I think it comes as a reasonable compromise:
- This approach won't hurt the currently working single-thread case.
- It guarantees correct authentication in a multi-threaded case (which requests claims to support).
- Simple and easy to understand code change.

tardyp:
@exvito : @vincentxb and I are on the same team, and have story on our sprint to resolve this issue. I was working on it this morning, and am at the point of creating a unit test.

exvito:
Updated with commit e8d9bc5. Highlights:
- All state now in thread local storage.
- Factored out state initialization to `init_per_thread_state()` which must be called from `__init__()` for the regular, single-threaded case and, eventually, from `__call__()` in the cases where threads that did not create the auth handler invoke it.
- Maybe the `self.tl` name can be better: I opted for short, project leaders may prefer `self.thread_local` or some variation.
- Maybe the test for thread local state initialization in `__call__()` can take a different approach.

Here's a contribution. Thanks for your feedback.
PS: @tardyp does that mean you'll work from this PR or you'll take a different approach and suggest I drop this PR?

Lukasa:
This looks entirely reasonable to me. It would be ideal to have some tests for this if at all possible.

tardyp:
@exvito, we worked on the same approach at the same time. 

I opted for the name thd for the local variable, and did the same ini_per_thread call. I think it is not necessery in **init** however, only in _call_

 :+1:  

I have a unit test on a separate commit:

https://github.com/tardyp/requests/commit/e15ed2c46823183b73250503fc44146f979eb891

Please cherry-pick it!

exvito:
> I think it is not necessery in `__init__` however, only in `__call__`.

You are correct. Additionally, after taking a peek at your code, I opted for your approach of ensuring per-thread state is initialized once and only once based on the `self.tl.init` attribute.

Updating with two commits:
- Your cherry picked test case (confirmed to work nicely, thanks).
- A simpler approach at initializing per-thread state.

test_requests.py
```diff
@@ -411,6 +416,21 @@ def test_DIGESTAUTH_QUOTES_QOP_VALUE(self):
         r = requests.get(url, auth=auth)
         assert '"auth"' in r.request.headers['Authorization']
 
+    def test_DIGESTAUTH_THREADED(self):
```
Lukasa, line 419: I hate this style of function naming. I appreciate there are other tests in here that do it, but let's not use the fact that some other tests slipped past me to make a mistake here. Can we return this to being `snake_case` please?

Lukasa, line 419: Separately, while this test does work, it's both overly aggressive and doesn't necessarily reliably fail. Is there a way of us testing this without requiring us to make 100 HTTP requests to do it?

Lukasa:
I still like this idea, but I have some notes in the diff. =)

test_requests.py
exvito, line 419: I understand where @tardyp is comming from: all of the digest auth related tests are named this way... Would you recommend the name change given that?

Regarding the test reliability, both towards failure or success: this looks like an inherently complex matter. @tardyp any suggestion?

Lukasa, line 419: With a bit of work using the `threading` module, particularly `threading.Event()` it should be possible to organise a test that exactly hits this case, without hitting the network at all.

tardyp, line 419: I was indeed agressive, just to make sure. I can't think another way beyond the agressive bruteforce.

testing race conditions is always tricky, will probably require white box testing, and synchronisation test hooks inside the production code. I am not so found of these technics

exvito, line 419: I was looking for inspiration in other tests that might be in-place to ensure correct thread-safe behaviour but couldn't find any. Am I missing something?

Lukasa, line 419: No, currently there are no tests to confirm that requests is thread-safe. Such testing is really tricky without explicit language support. However, this lack of testing is why I believe that currently requests is not _truly_ thread-safe.

Again, think carefully about whether you can use `threading.Event` to coordinate two threads in such a way that you can deterministically break them (e.g. by having them get their nonces confused).

requests/auth.py
```diff
@@ -63,19 +64,26 @@ class HTTPDigestAuth(AuthBase):
     def __init__(self, username, password):
         self.username = username
         self.password = password
-        self.last_nonce = ''
-        self.nonce_count = 0
-        self.chal = {}
-        self.pos = None
-        self.num_401_calls = 1
+        # Keep state in per-thread local storage
+        self.tl = threading.local()
```
sigmavirus24: Not trying to shave any yaks here, but I'd really prefer 2 things.
1. That the thread local be internal only, e.g., prefixed with a `_`
2. That it has a better name that people can understand without having to read the entirety of the class definition (e.g, `_thread_local`)

test_requests.py
sigmavirus24, line 419: Yeah, I've been considering creating a separate repo for a much larger set of tests. They could be run each time a commit is pushed, PR is opened, and on a periodic basis (just for the sake of paranoia), but I haven't gotten around to writing them.

requests/auth.py
exvito: @sigmavirus24 - thanks for the input, I was divided between the concise vs longer, more explicit name. I've updated the PR with commit 5a69137ac871c38f4597b574e93de07d9ff9655f following your suggestion.

Also: I threw away a single utilization of a local `num_401_calls` right before the `if 'digest' in ...` in `handle_401()`.

test_requests.py
exvito, line 419: I've given some thought to (and sleep over) this - this is as far as I reached:
- Using `threading.Event` to force the coordination of threads to deterministically break them may be feasible.
- It does however imply that such code (test/instrumentation) needs to be within/inline with the code to be tested: this seems to go against the basic principle of testing - separating test code from the code being tested.
- A much more complex alternative (but separate from the code being tested) is to use/build some kind of "clever-testing-thread-runner" that forces its thread pool to behave in a controlled, deterministic fashion regardless of OS scheduling: this might need to hook into lower level runtime/debug hooks or some such... Not being aware of low-level Python runtime internals, I can't even tell if this is currently possible or if it might need extra support there.
- Last, but not least, regarding the test @tardyp shared: given that it is not guaranteed to fail, I wonder if it should be kept -- will it provide a false sense of security? or is it better than having nothing in place, given that the odds of it resulting in a "false success" are very low?
- Additionally, regarding hitting the network: by skimming through `test_requests.py` it seems lots of other tests do it - shouldn't this (admitedly important) challenge be addressed at a more general level? (it seems this can be minimized by running a local instance of httpbin and setting the `HTTPBIN_URL` environment var, which is, again, better than no choice at all)

I'd love to be wrong about the multi-thread testing: if so, please point out my mistakes.
Thanks

tardyp, line 419: I'd say if my test is failling reliabily (98%) on old unsafe implementation, and first version of this pr, it is good enough. 
When I wrote it, I verified it failed 100% on old impl ( 1 fail /1 try == 100%, uh? )
Then, there is the much bigger project of testing thread safeness of the whole library, and splitting the suite into several files. That is another much longer project

m2c

kennethreitz:
Is this code spawning a new thread?

Lukasa:
@kennethreitz The unit test for it is, but the product code doesn't. It's just to make it safe when it's called from multiple threads.

kennethreitz:
:+1:

PR #2531, opened 2015-04-05
Issue 2062: Updated custom header precedence info with .netrc example

benjaminran:
I revised the custom header documentation to include .netrc authentication like Ian recommended.


Lukasa:
\o/ :sparkles:

PR #2534, opened 2015-04-06
Copy a PreparedRequest's CookieJar reliably

### LINKED ISSUE #2527: AttributeError- LWPCookieJar instance has no attribute 'copy'
Hi,
In version 2.x:
<b>[test-case]</b>
```python
import requests
from cookielib import LWPCookieJar
cookiejar = LWPCookieJar()
req = requests.Request('GET', 'http://httpbin.org/get', cookies=cookiejar)
r = req.prepare()
r.copy()
```
<b>[traceback]</b>
...
AttributeError: LWPCookieJar instance has no attribute 'copy'

<b>[comments]</b>
It seems that not all instances of cookiejar have the copy method, but in the copy method of PreparedRequest, it call the copy method of cookies without any checks.

sigmavirus24:
When a PreparedRequests's cookie jar is not a RequestsCookieJar instance, it
will not have a "copy" method. By adding _copy_cookie_jar we can reliably copy
cookie jars so that we have an actual copy instead of the same instance on
different prepared requests.

This also updates the RequestsCookieJar.update logic to create copies of
cookies from the other jar.

Closes #2527

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -6,6 +6,7 @@ Compatibility code to be able to use `cookielib.CookieJar` with requests.
 requests.utils imports from here, so be careful with imports.
 """
 
+import copy
 import time
 import collections
 from .compat import cookielib, urlparse, urlunparse, Morsel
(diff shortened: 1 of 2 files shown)


Lukasa:
LGTM. :cake:

PR #2535, opened 2015-04-06
Fix build version number in metadata

### LINKED ISSUE #2530: __build__ out of sync? Still at 0x020503, looks like 2.5.3 rather than 2.6.0
Could it be that
```
__build__ = 0x020503
```
at https://github.com/kennethreitz/requests/blob/master/requests/__init__.py#L46
went out of sync? It looks like 2.5.3 rather than 2.6.0.

Best, Sebastian

sigmavirus24:
Closes #2530

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -43,7 +43,7 @@ is at <http://python-requests.org>.
 
 __title__ = 'requests'
 __version__ = '2.6.0'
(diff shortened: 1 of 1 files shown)


sigmavirus24:
_/me forgot to make this last night_ Sorry @Lukasa

Issue #2532, opened 2015-04-06
"PreparedRequest.prepare_cookies" can only be called once (and "succeeds" silently if called again)


smiley:
I'm writing an adapter that interfaces with a CDN, and it requires that I "sign" my requests with a cookie, but it doesn't send the cookie back in `Set-Cookie`.

I manage to send it on the first request, but then the site behind the CDN *does* send a `Set-Cookie` header (this is expected), and `prepare_cookies` called by the `Session` works its magic. The problem is, due to how `cookielib` is written, if a `Cookie` header was already composed, it won't be re-composed: (Even if cookies have changed between calls)

```python
# cookielib.py:CookieJar.add_cookie_header (starts at 1312)
# ...
attrs = self._cookie_attrs(cookies)
if attrs:
    if not request.has_header("Cookie"):
        request.add_unredirected_header(
          "Cookie", "; ".join(attrs))
#...
```

This causes a problem when any request past the first one on the same `Session` object renders my adapter worthless, as authentication fails and attempts to properly add cookies fail silently:
```python
def _prepare_request(self, request):
    """
    :type request: requests.PreparedRequest
    """
    request.prepare_cookies({self.COOKIE_NAME: self._get_authorization_cookie(request.url)})
```

`request._cookies` is modified properly, but `request.headers['Cookie']` is still its former self.

If fixing this isn't possible (it *is* a bundled library after all), maybe `requests` could delete the header before calling `CookieJar.add_cookie_header`?

sigmavirus24:
> If fixing this isn't possible (it is a bundled library after all), maybe requests could delete the header before calling CookieJar.add_cookie_header?

I assume this is happening because of [this line](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L511). And that's why you quoted code from `cookielib` a standard library. You want us to somehow patch the standard library locally?

Here's an alternative fix: Why aren't you constructing a new request, and preparing it with the new cookie header instead of trying to reuse an old prepared request?

smiley:
No, I'm suggesting that you _can_ report the issue to the people developing `cookielib` (to get this fixed or documented in future Python releases), and do one of these things:
1. Reject any future calls to `prepare_cookies` (have a local boolean for marking this was already called?)
2. Merge an existing `self._cookies` value with the new one you're creating, clear `self.headers['Cookie']` and call the correct prepare function in `cookielib`, forcing it to re-generate the header
[...]

PR #2536, opened 2015-04-06
Document that "PreparedRequest.prepare_cookies" can only be called once

smiley:
I've done the documentation changes suggested in #2532.

I also documented a way to work around this in the docstring, but I'm not sure if it'd be good to do so. (As it may lead to hacked-together solutions) Then again, other solutions like creating a brand new `Request` might lead to worse problems. (e.g.: loss of data)

@sigmavirus24, what do you think?

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -501,7 +501,15 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             self.prepare_content_length(self.body)
 
     def prepare_cookies(self, cookies):
-        """Prepares the given HTTP cookie data."""
+        """Prepares the given HTTP cookie data.
+
+        This function eventually generates a ``Cookie`` header from the
+        given cookies using cookielib. Due to cookielib's design, the header
+        will not be regenerated if it already exists, meaning this function
+        can only be called once for the life of the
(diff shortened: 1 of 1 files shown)


requests/models.py
```diff
@@ -501,7 +501,15 @@ def prepare_auth(self, auth, url=''):
...
     def prepare_cookies(self, cookies):
-        """Prepares the given HTTP cookie data."""
+        """Prepares the given HTTP cookie data.
+
+        This function eventually generates a ``Cookie`` header from the
+        given cookies using cookielib. Due to cookielib's design, the header
+        will not be regenerated if it already exists, meaning this function
+        can only be called once for the life of the
+        :class:`PreparedRequest <PreparedRequest>` object. Any subsequent calls
+        to ``prepare_cookies`` will have no actual effect, unless the "Cookie"
```
sigmavirus24, line 511: No actual effect on the cookie header right? It updated the _cookies attribute though if I remember your description properly.

smiley, line 511: Yep, that's why I wrote "actual effect" -- it changes `self._cookies` alright, but the request sent at the end is no different than the one sent if you skip the extra call.

Issue #2540, opened 2015-04-08
MisssingSchema error message does not handle unicode input correctly


ma2rten:
using requests 2.6.0, Python 2.7.6

```
import requests
requests.get(u"hällo world")
```

```
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-3-bf1281984b3a> in <module>()
----> 1 requests.get(url)

/Library/Python/2.7/site-packages/requests-2.6.0-py2.7.egg/requests/api.pyc in get(url, **kwargs)
     66 
     67     kwargs.setdefault('allow_redirects', True)
---> 68     return request('get', url, **kwargs)
     69 
     70 

/Library/Python/2.7/site-packages/requests-2.6.0-py2.7.egg/requests/api.pyc in request(method, url, **kwargs)
     48 
     49     session = sessions.Session()
---> 50     response = session.request(method=method, url=url, **kwargs)
     51     # By explicitly closing the session, we avoid leaving sockets open which
     52     # can trigger a ResourceWarning in some cases, and look like a memory leak

/Library/Python/2.7/site-packages/requests-2.6.0-py2.7.egg/requests/sessions.pyc in request(self, method, url, params, data, headers, cookies, files, auth, timeout, allow_redirects, proxies, hooks, stream, verify, cert, json)
    448             hooks = hooks,
    449         )
--> 450         prep = self.prepare_request(req)
    451 
    452         proxies = proxies or {}

/Library/Python/2.7/site-packages/requests-2.6.0-py2.7.egg/requests/sessions.pyc in prepare_request(self, request)
    379             auth=merge_setting(auth, self.auth),
    380             cookies=merged_cookies,
--> 381             hooks=merge_hooks(request.hooks, self.hooks),
    382         )
    383         return p

/Library/Python/2.7/site-packages/requests-2.6.0-py2.7.egg/requests/models.pyc in prepare(self, method, url, headers, files, data, params, auth, cookies, hooks, json)
    302 
    303         self.prepare_method(method)
--> 304         self.prepare_url(url, params)
    305         self.prepare_headers(headers)
    306         self.prepare_cookies(cookies)

/Library/Python/2.7/site-packages/requests-2.6.0-py2.7.egg/requests/models.pyc in prepare_url(self, url, params)
    359         if not scheme:
    360             raise MissingSchema("Invalid URL {0!r}: No schema supplied. "
[...]

PR #2547, opened 2015-04-10
Fix MisssingSchema error message does not handle unicode input correctly #2540

sh1buy:


Lukasa:
I think this won't work on Python 3 with a binary URL, though I need to test that.

sh1buy:
You rigth. This add 'b' to url. 

```
requests.exceptions.MissingSchema: Invalid URL b'example.com': 
No schema supplied. Perhaps you meant http://b'example.com'?
```

sh1buy:
Whether it is necessary to display  URL completely in error message?
Will be enough only hostname?
For example, instead of the:

```
requests.exceptions.MissingSchema: Invalid URL 'example.com/page?query=blabla&foo=23': 
No schema supplied. Perhaps you meant 'http://example.com/page?query=blabla&foo=23'?
```

Output this:

```
requests.exceptions.MissingSchema: Invalid URL 'example.com/...': 
No schema supplied. Perhaps you meant 'http://example.com/...'?
```

Then we can solve this bug with applying IDNA to the hostname.

Lukasa:
The better thing to do might be to call `to_native_str`, from our utility package.

sh1buy:
This work in both Python's versions.

Lukasa:
Can you wrap the line at 79 characters?

sigmavirus24:
Thanks @sh1buy! :sparkles: :cake: :sparkles:

PR #2548, opened 2015-04-11
Add a timeout to select(), otherwise we can block forever on a bad socket

natecode:
All the other select.select() calls in this module use a timeout. Without this bugfix, a server that fails to complete the handshake will block forever. We've seen this in production.


Lukasa:
Can you open this pull request on urllib3? All your code changes are there.

natecode:
Please re-open this issue. I just checked `urllib3` and it has this same fix in it so `requests` needs to update your code from urllib3.

natecode:
It appears great minds think alike https://github.com/shazow/urllib3/commit/14d09fc8fd9df697e534e8e10cbe6f913289c748

Lukasa:
Indeed. However, there's no need to re-open the issue, we'll bring a new urllib3 version before our next release.

PR #2551, opened 2015-04-13
EditorConfig

treyhunner:
Previously proposed and declined in #893.

I am opening this because I think it is time to reassess whether EditorConfig is appropriate for this project.

I did some GitHub diving and found the following Python projects which use EditorConfig:

- [flask-wtf](https://github.com/lepture/flask-wtf)
- [cookiecutter-django](https://github.com/pydanny/cookiecutter-django)

I also found the following Python-powered websites that use EditorConfig:

- [MDN](https://github.com/mozilla/kuma)
- [gratipay](https://github.com/gratipay/gratipay.com)
- [mozilla.org](https://github.com/mozilla/bedrock)

This file is a human-readable auto-enforced whitespace style guide for all files (*not just Python*).

For:
- **users of editors with built-in support** (e.g. PyCharm, Coda 2, SourceLair) this file will ensure the code style is enforced automatically.
- **users with a plugin installed** (available for most major editors) this file will also auto-enforce the code style
- **all other users** this simply serves as a white space style guide.


Lukasa:
Thanks for this!

I do not believe that EditorConfig is appropriate at this time. In general, I resent the relentless expansion of dotfiles that various tools require people to add to the project root. I appreciate that these tools are close to many people's hearts, and that they provide value for plenty of people, but it's extremely difficult to justify any tool that is not actually considered 'enforced' by some form of automated tooling. The risk is that if we accept this it becomes difficult to reject it for other tools, and we rapidly end up with an expansion of dotfiles that have no love.

For this reason, I'm afraid I don't think I can accept this. =( Sorry!

PR #2570, opened 2015-04-25
Add new release process to docs

### LINKED ISSUE #2569: [Proposal] New Release Management Process
In v2.6.1 we pulled in a new copy of urllib3. Typically this doesn't negatively affect us but in this case we've seen a bunch of new issues from the chunked encoding refactor in urllib3. Because that kind of stuff *should* not break in a `.x` release I'm proposing we follow this process when preparing releases (and that we document it for the future):

For an `v*.*.x` release, e.g., `v2.6.1`, `v2.6.2`, `v2.6.3`, etc. we *only* include changes to requests itself. This means anything that does not live in `requests/packages/{urllib3,chardet}/*`.

For a `v*.x.0` release, e.g., `v2.6.0`, `v2.7.0`, etc. we include changes to requests *and* update the vendored libraries. The same goes for a `vx.0.0` release, but of course those have other implications as well.

sigmavirus24:
Closes #2569

### DIFF
--- a/docs/community/release-process.rst
+++ b/docs/community/release-process.rst
@@ -1,0 +1,51 @@
+Release Process and Rules
+=========================
+
+.. versionadded:: v2.6.2
+
+Starting with the version to be released after ``v2.6.2``, the following rules 
+will govern and describe how the Requests core team produces a new release.
+
+Major Releases
+--------------
+
+A major release will include breaking changes. When it is versioned, it will 
+be versioned as ``vX.0.0``. For example, if the previous release was 
+``v10.2.9`` the next version will be ``v11.0.0``.
+
+Breaking changes are changes that break backwards compatibility with prior 
+versions. If the project were to change the ``text`` attribute on a 
+``Response`` object to a method, that would only happen in a Major release.
+
+Major releases may also include miscellaneous bug fixes and upgrades to 
+vendored packages.
+
+Minor Releases
+--------------
+
+A minor release will not include breaking changes but may include 
+miscellaneous bug fixes and upgrades to vendored packages. If the previous 
+version of Requests released was ``v10.2.9`` a minor release would be 
+versioned as ``v10.3.0``.
+
+Minor releases will be backwards compatible with releases that have the same 
+major version number. In other words, all versions that would start with 
+``v10.`` should be compatible with each other.
+
+Hotfix Releases
+---------------
+
+A hotfix release will only include bug fixes that were missed when the project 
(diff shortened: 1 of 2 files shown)


docs/community/release-process.rst
```diff
@@ -0,0 +1,51 @@
...
+A major release will include breaking changes. When it is versioned, it will 
+be versioned as ``vX.0.0``. For example, if the previous release was 
+``v10.2.9`` the next version will be ``v11.0.0``.
+
+Breaking changes are changes that break backwards compatibility with prior 
+versions. If the project were to change the ``text`` attribute on a 
+``Response`` object to a method, that would only happen in a Major release.
+
+Major releases may also include miscellaneous bug fixes and upgrades to 
+vendored packages.
```
kevinburke: maybe a note about "we don't do this often" / maybe once a year or once every 2 years.

docs/community/release-process.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Minor Releases
+--------------
+
+A minor release will not include breaking changes but may include 
+miscellaneous bug fixes and upgrades to vendored packages. If the previous 
+version of Requests released was ``v10.2.9`` a minor release would be
```
Lukasa: This example is potentially somewhat ambiguous, because Python used to operate under the rule that after vX.Y.9 came vX.Y+1.0, which is what this _could_ look like.

Lukasa:
Looks good, some small notes.

docs/community/release-process.rst
sigmavirus24: These are all hypothetical version numbers. I can lower `9` to `7` in all examples if that makes you more comfortable with this.

docs/community/release-process.rst
sigmavirus24: What about "We do our best to maintain backwards compatibility and we will only create a major release when it is absolutely necessary."?

docs/community/release-process.rst
Lukasa: It'll remove the ambiguity, certainly.

docs/community/release-process.rst
sigmavirus24: Added something along these lines to the latest commit. Thanks for the suggestion @kevinburke

docs/community/release-process.rst
sigmavirus24: Done

docs/community/release-process.rst
kevinburke: +1

Issue #2581, opened 2015-05-01
The docstring for `get` in `api.py` should be more clear.


mhavard999:
This is the docstring for `get` from `api.py`:
```python
    """Sends a GET request.

    :param url: URL for the new :class:`Request` object.
    :param \*\*kwargs: Optional arguments that ``request`` takes.
    :return: :class:`Response <Response>` object
    :rtype: requests.Response
    """
```
So when someone does `help(requests.get)` they won't see anything about the optional kwarg `params` used to attach query string parameters to the request.  If they're extra clever, they'll understand what "Optional arguments that \`\`request\`\` takes" mean and do `help(requests.request)`, which then explains the `params` kwarg parameter.  Note that the docstring of `requests.post` mentions `data` and `json`, even though those params are explained in the docstring of `requests.requests`.  Since query string parameters are as relevant to GET requests as body and json parameters are to POST requests, should the `requests.get` be changed to:
```python
def get(url, params=None, **kwargs):
    """Sends a GET request.

    :param url: URL for the new :class:`Request` object.
    :param params: (optional) Dictionary or bytes to be sent in the query string for the :class:`Request`.
    :param \*\*kwargs: Optional arguments that ``request`` takes.
    :return: :class:`Response <Response>` object
    :rtype: requests.Response
    """

    kwargs.setdefault('allow_redirects', True)
    return request('get', url, params=params, **kwargs)
```

Current version for reference:
```python
def get(url, **kwargs):
    """Sends a GET request.

    :param url: URL for the new :class:`Request` object.
    :param \*\*kwargs: Optional arguments that ``request`` takes.
    :return: :class:`Response <Response>` object
    :rtype: requests.Response
    """

    kwargs.setdefault('allow_redirects', True)
    return request('get', url, **kwargs)
```

Lukasa:
Closed by #2582.

PR #2582, opened 2015-05-01
Make docstring of api.get more clear

mhavard999:
I opened issue #2581, and since it is such a small change, I just went ahead and opened the pull request immediately.

This makes the docstring for `requests.get` clearer so when someone does `help(requests.get)` they can see that `params` is an optional kwarg that can be used to attach query string parameters to a GET request.  I think this change makes it symmetric with `requests.post` which specifically mentions `data` and `json` for body and json parameters.  As I mentioned in #2581, query string parameters are as relevant to GET requests as json and body parameters are to POST requests, so this change makes a lot of sense in that respect.

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -55,17 +55,18 @@ def request(method, url, **kwargs):
     return response
 
 
-def get(url, **kwargs):
+def get(url, params=None, **kwargs):
     """Sends a GET request.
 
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this @mhavard999, it looks great! :cake:

PR #2586, opened 2015-05-03
Cut 2.7.0 this weekend

Lukasa:
Ok, I believe we've left enough time for the chunked stuff to bake, and no further issues seem to have come out of the woodwork. At this point, I think we should cut 2.7.0 and begin the switchover to our new release process.

@shazow are you open to tagging a new urllib3 release, and if so what commit would you like us to use? If you don't really care, by default we'd use shazow/urllib3@74073791a3429d9b9f375563954f57f2181599dc.

Things to do:

- [x] Update urllib3.
- [x] Confirm that the #2455 problem sites still work.
- [x] Update changelog.
- [ ] Tag.
- [ ] Push release.
- [ ] Have a drink.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,17 @@
 Release History
 ---------------
 
+2.7.0 (2015-05-03)
+++++++++++++++++++
+
+This is the first release that follows our new release process. For more, see
+[our documentation](http://docs.python-requests.org/en/latest/community/release-process/).
+
+**Bugfixes**
+
+- Updated urllib3 to 1.10.4, resolving several bugs involving chunked transfer
+  encoding and response framing.
+
 2.6.2 (2015-04-23)
 ++++++++++++++++++
 
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -42,8 +42,8 @@ is at <http://python-requests.org>.
 """
 
 __title__ = 'requests'
-__version__ = '2.6.2'
-__build__ = 0x020602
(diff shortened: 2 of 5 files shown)


Lukasa:
In my ideal world we'd update the `certifi` bundle as well, but I'm now strongly averse to trying to do too much with these releases. @sigmavirus24?

sigmavirus24:
I'm not strongly opposed to it. I'd like confirmation that the problem sites in #2455 at least work. Thoughts?

Lukasa:
Agreed. Adding it to the list.

Lukasa:
Alright, the chunked keep-alive problem does appear to be resolved after synchronizing urllib3.

t-8ch:
Hm, this downgraded the embedded version string of urllib3.
I think it is confusing to have a quite arbitrary version in the bundled library.

Thoughts:
- Wait for a @shazow to cut a new release and update the version in urllib3 itself, then pull that.
- Rewrite the version string to be actual the commit id.

shazow:
Tagged and pushed urllib3 v1.10.4.

Lukasa:
And I updated the commit to resolve that problem. The only delta between the previous commit and the new one is the version identifier.

Lukasa:
@sigmavirus24 Any objections to me merging this PR (manually without a merge commit, because we don't need it) and pushing the release?

sigmavirus24:
0 (zero) objections here.

Lukasa:
Done.

PR #2587, opened 2015-05-03
Remove the __bool__ and __nonzero__ response methods

sigmavirus24:
Many people expect to be able to say:

    response = make_request(url)

    if response:
        body = response.content

Where the first part should test for whether or not response is None.
Instead, the __bool__ and __nonzero__ methods return response.ok, so if
the response is actually a 4xx or 5xx response, then the user would
expect to get the body of the response.

By removing these methods, we restore the functionality that most users
expect.

Closes #2002 

----

**Note** this is development for 3.0.0

### DIFF
--- a/3.0-HISTORY.rst
+++ b/3.0-HISTORY.rst
@@ -1,0 +1,11 @@
+3.0.0 (2015-xx-xx)
+++++++++++++++++++
+
+- Remove the ``__bool__`` and ``__nonzero__`` methods from a ``Response``
+  object.
+
+  This has been a planned feature for over a year. The behaviour is surprising
+  to most people and breaks most of the assumptions that people have about
+  Response objects. This resolves issue `#2002`_
+
+.. _#2002: https://github.com/kennethreitz/requests/issues/2002
--- a/requests/models.py
+++ b/requests/models.py
@@ -619,14 +619,6 @@ class Response(object):
     def __repr__(self):
         return '<Response [%s]>' % (self.status_code)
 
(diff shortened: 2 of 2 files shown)


Lukasa:
For the 3.0.0 stuff, I think we should try to be adding changelog updates for this: ideally, fairly thorough ones.

Lukasa:
Otherwise, this looks totally reasonable. =)

sigmavirus24:
Since we've discussed keeping this branch around for a while and the likelihood of having another release (at least), I made a different file to store the 3.0 history to avoid messy merge conflicts. Let me know if you just want it in `HISTORY.rst` instead

3.0-HISTORY.rst
```diff
@@ -0,0 +1,11 @@
+3.0.0 (2015-xx-xx)
+++++++++++++++++++
+
+- Remove the ``__bool__`` and ``__nonzero__`` methods from a ``Response``
+  object.
+
+  This has been a planned feature for over a year. The behaviour is exciting
```
Lukasa: s/exciting/surprising/?

Lukasa:
I think you accidentally a word, but this is otherwise perfect.

3.0-HISTORY.rst
sigmavirus24: That's what I get for writing this late at night while talking to a friend about their exciting news.

sigmavirus24:
The word is unaccidentallyied

PR #2588, opened 2015-05-04
use reST link syntax

gutworth:


HISTORY.rst
```diff
@@ -7,7 +7,8 @@ Release History
 ++++++++++++++++++
 
 This is the first release that follows our new release process. For more, see
-[our documentation](http://docs.python-requests.org/en/latest/community/release-process/).
+`our documentation
```
sigmavirus24, line 10: tsk tsk tsk @Lukasa

sigmavirus24:
Thanks for catching this @gutworth! :sparkles: :cake: :sparkles:

HISTORY.rst
Lukasa, line 10: Markup is tricky. =(

PR #2597, opened 2015-05-12
do not blindly catch all AttributeErrors

t-8ch:
see shazow/urllib3#618


sigmavirus24:
So, perhaps this is a stupid question, by why is there this condition in the code in the first place?

Lukasa:
It's to allow you to return file-like objects that aren't urllib3's HTTPResponse class. This is super common in tests, for example.

This LGTM.

Issue #2599, opened 2015-05-13
Auth handler should be calling close, not release_conn


sigmavirus24:
On [this line](https://github.com/kennethreitz/requests/blob/956ec8321a76f0b5a9f5bcb591cf6e77d773f819/requests/auth.py#L182) we should be using the response's "close" method. We shouldn't be explicitly calling a method on the "raw" attribute.

Lukasa:
Resolved in #2601.

sahutd:
Should I add a test to ensure close gets called?
Would using a mock object be the right way?

PR #2601, opened 2015-05-14
Auth handler calls close instead of raw.close_conn

sahutd:
Issue #2599


Lukasa:
\o/

Lukasa:
thanks for this! :cake: :sparkles:

PR #2603, opened 2015-05-14
Change "Staring new HTTP/S connection..." log message to DEBUG level

jefferyaronhalt:
Of course there are ways to work around this on a per-execution basis (programmatically or manually), but I figured I'd address what I've heard from various developers as "the only annoying thing about python requests".


sigmavirus24:
This is a change to urllib3 and so belongs on [urllib3](/shazow/urllib3)

goodboybeau:
Ah, thanks...

PR #2606, opened 2015-05-19
Don’t break IPv6 address format

Aeyoun:
No idea what this does to older supported Python versions. In Python 3.4.2, at least, you can’t strip away the brackets otherwise no IPv6 address will resolve.


Lukasa:
Thanks for this!

Two notes. First, this is a change in [urllib3](https://github.com/shazow/urllib3), please open this pull request on that repository. Secondly, given the comment above I don't think we can unilaterally remove this logic. Likely we'll have to detect what versions are affected.

Issue #2593, opened 2015-05-09
contextlib.closing(response) is not enough when using sessions and stream = True


Beluki:
The requests documentation suggests using the following idiom to partially read response bodies (e.g. when stream = True):

```python
from contextlib import closing

with closing(requests.get('http://httpbin.org/get', stream=True)) as r:
    # Do things with the response here.
```

I thought this would be applicable to sessions, but it seems it's not enough.
The following program reliably raises an exception (BadStatusLine) even while it uses contextlib.closing, and the only way to stop that from happening is to consume the whole response content:

```python
import contextlib
import random
import requests

urls = [
    'http://reddit.com',
    'http://news.ycombinator.com',
    'http://github.com',
    'http://bitbucket.org',
]

try:
    session = requests.Session()
    
    while True:
        try:
            with contextlib.closing(session.get(random.choice(urls), stream = True)) as r:
                # uncomment the line below and the errors dissapear:
                # r.content
                pass

        except Exception as ex:
            print(str(ex))

except KeyboardInterrupt:
    pass
```

Is this the expected behavior or actually a bug? (to be fair the documentation doesn't actually say that the idiom is also applicable to requests coming from sessions).

Lukasa:
I think this is expected, but it could arguably be worked around.

The problem is that urllib3 doesn't know whether we consumed the response, so it doesn't know that the socket has unread data in it. We should be able to support that case, I think.

sigmavirus24:
Wait. I have a question.

In your first example, you have a comment "Do things with the response here". If you put that in your while loop with a "pass" statement, do you see the same behaviour? If so, this issue seems a bit misleading as it is currently presented.

Beluki:
Hi.

> In your first example, you have a comment "Do things with the response here". If you put that in your while loop with a "pass" statement, do you see the same behaviour?

The errors disappear if:

``` python
# I just use requests.get instead of sessions:
with contextlib.closing(requests.get(random.choice(urls), stream = True)) as r:
    # no need to consume the whole content, no errors:
    pass
```

``` python
# I use sessions but I consume the whole response body:
with contextlib.closing(session.get(random.choice(urls), stream = True)) as r:
    r.content
```

In other words, the problem is that `closing(requests.get(...), stream = True)` is always safe regardless of the content, but `closing(session.get(...), stream = True)` is not, you must consume the response content.

Beluki:
Just so that everything is as clear as possible:

[...]

PR #2610, opened 2015-05-24
fix contextlib.closing bug for sessions where content is not consumed…

colindickson:
… (issue #2593)


sigmavirus24:
Why are you making 25 requests to three websites that we don't use anywhere else in the test suite?

colindickson:
mainly as an attempt to reproduce the bug demonstrated in the original issue.   i suppose i could simplify it a great deal

sigmavirus24:
Please don't create a new pull request. If you continue to push commits to this branch they'll just show up here.

sigmavirus24:
Also, let me be clear, your fix here is 100% correct.

colindickson:
test should be better now :)

Lukasa:
LGTM. :cake:

PR #2631, opened 2015-06-05
Handle empty chunks

neosab:
Empty chunk in request body could prematurely signal end of chunked
transmission. As a result, the terminating zero-size chunk sent by
'requests' can be interpretted as bad request by the recepient. I
have used the same logic used by httplib to handle such cases.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -388,7 +388,10 @@ class HTTPAdapter(BaseAdapter):
                     low_conn.endheaders()
 
                     for i in request.body:
-                        low_conn.send(hex(len(i))[2:].encode('utf-8'))
+                        chunk_size = len(i)
+                        if chunk_size == 0:
+                            continue
+                        low_conn.send(hex(chunk_size)[2:].encode('utf-8'))
                         low_conn.send(b'\r\n')
                         low_conn.send(i)
                         low_conn.send(b'\r\n')


neosab:
Resubmitted https://github.com/kennethreitz/requests/pull/2388 on proposed/3.0.0 branch.

requests/adapters.py
```diff
@@ -388,7 +388,10 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
                     low_conn.endheaders()
 
                     for i in request.body:
-                        low_conn.send(hex(len(i))[2:].encode('utf-8'))
+                        chunk_size = len(i)
+                        if chunk_size == 0:
+                            break
```
Lukasa: Do you mean `break` here, or `continue`?

neosab: break - making it consistent with the existing behavior where the connection would have anyway terminated upon sending this zero size chunk.

Lukasa: If we're trying to be consistent with existing behaviour, I think we're setting the bar too low. With the PR open against 3.0.0 we can design a sensible behaviour, rather than a compatible one, and I think that the sensible behaviour is to do our best to consume the entire iterator.

sigmavirus24: So if someone gives us a generator like:

``` py
(chunk for chunk in b'foo', b'bar', b'', b'biz')
```

I think they'll expect it all to be sent. So I agree with @Lukasa. That said, we definitely need to also document this to death. I am willing to bet someone at sometime will want that to be a break for some bizarre reason. =P

neosab: Agreed. I changed this to ignore empty chunks. Let me know if this looks good.

sigmavirus24:
:+1: LGTM

Lukasa:
LGTM as well. =) We should note this change somewhere before we merge it and lose it.

sigmavirus24:
Changes are noted in https://github.com/kennethreitz/requests/pull/2646. Merging. :sparkles: :cake: :sparkles:

sigmavirus24:
Thanks @neosab

PR #2640, opened 2015-06-13
Avoid double releasing chunked upload connections

### LINKED ISSUE #2636: Same HTTPConnection object returned to pool twice
Not totally sure how to reproduce this yet but I am definitely getting a case where I am getting the "Connection pool is full" message because the same connection object is being returned to the pool twice  (and not surprisingly the second time the pool says that it is full).

I am using gevent (yes I realize that is probably crucial), but even so ...

The first time the connection gets returned is in requests.adapters:

410                 else:
411                     # All is well, return the connection to the pool.
412                     conn._put_conn(low_conn)

The second time the connection gets returned is in requests.packages.urllib3.response:

285         finally:
286             if self._original_response and self._original_response.isclosed():
287                 self.release_conn()

I'd love to  be able to supply a simple example, but the code path is a little to convoluted for me to work out quickly so I'm looking for hints/tips here if you have any.

Thanks,
Giles

Lukasa:
Resolves #2636

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -407,9 +407,6 @@ class HTTPAdapter(BaseAdapter):
                     # Then, reraise so that we can handle the actual exception.
                     low_conn.close()
                     raise
-                else:
-                    # All is well, return the connection to the pool.
-                    conn._put_conn(low_conn)
 
         except (ProtocolError, socket.error) as err:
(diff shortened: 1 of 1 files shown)


sigmavirus24:
:sparkles: :cake: :sparkles:

PR #2641, opened 2015-06-15
make api.py support keepalive

duanhongyi:
Use an object pool, making api.py support for keepalive

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -12,6 +12,24 @@ This module implements the Requests API.
 """
 
 from . import sessions
+from . import pool
+
+
+SESSION_RECYCLE_COUNT = 3000
+
+
+def _session_creater():
+    session = sessions.Session()
+    session._recycle_count = 0
+    return session
+
+
+def _session_after_remove(session):
+    session.close()
+
+
+SESSION_POOL = pool.ObjectPool(_session_creater,
+        after_remove=_session_after_remove)
 
 
 def request(method, url, **kwargs):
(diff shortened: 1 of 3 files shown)


Lukasa:
Thanks for this @duanhongyi!

However, I'm :-1: on this idea. My primary objection is that this adds implicit global state. Generally speaking I don't think libraries should maintain any internal state at all except what is truly necessary for their function. Wherever possible we should provide 'state objects' to users that the user has control over, ensuring that users own the lifetimes of objects: we already do this.

I think having a pool of sessions in the background is a bad idea. It'll lead to bug reports from users who aren't expecting it and it'll lead to complaints from users who don't like the memory profile it causes. Most importantly, it makes it _very very difficult_ to obtain reproducible behaviour out of the top-level API, because what exactly happens on a given request is actually dependent on the entire lifetime of the program up until that point. `requests.get()` may or may not work depending on whether there are cookies present in the particular `Session` you're using. It may or may not work depending on whether the remote server supports keepalive connections (a surprising number don't handle it well).

I'll let @sigmavirus24 express an opinion as well, but I'm sorry, I doubt we'll merge this.

duanhongyi:
@Lukasa 

I agree with you, but the global shared socket connection, which is common in programming languages such as Java, is also practical. and perhaps adapters.py is more appropriate for doing this.

Lukasa:
Any user who wants it can easily achieve a very similar effect:

```
from requests import session as requests
```

This gives the implicit shared pool that everyone wants. Bear in mind the session API is a strict superset of the `api.py` API, so it should be a simple replacement.

sigmavirus24:
I agree with @Lukasa. This will also likely negatively affect things like grequests which build on the api module's design. I also don't think we want this in `requests/adapters.py`.

PR #2650, opened 2015-06-23
Display content as part of HTTP error messages

kuldeeprishi:


Lukasa:
Unlike #2648 I think this is a not a good idea.

The content of a message is potentially extremely large or unbounded. It should not appear in tracebacks. Additionally, if you've set `stream=True` because you're expecting a large response in any form then the body may not have been downloaded yet, in which case we will actually block behind obtaining the request body before we throw the exception. Generally I'd say an exception we _know_ we want to throw should not block behind I/O before being thrown.

If you want to log this information out yourself you can do: catch the exception from `raise_for_status` and then consume the content yourself. =)

kuldeeprishi:
Ok Thanks for the feedback :)

PR #2656, opened 2015-06-29
Allow get_netrc_auth to raise parse/permission errors to caller

### LINKED ISSUE #2654: utils.get_netrc_auth silently fails when netrc exists but fails to parse
My .netrc contains a line for the github auth, [like this](https://gist.github.com/wikimatze/9790374).

It turns out that `netrc.netrc()` doesn't like that:

```
>>> from netrc import netrc
>>> netrc()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/netrc.py", line 35, in __init__
    self._parse(file, fp, default_netrc)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/netrc.py", line 117, in _parse
    file, lexer.lineno)
netrc.NetrcParseError: bad follower token 'protocol' (/Users/david/.netrc, line 9)
```

`get_netrc_auth` catches the `NetrcParseError` [but just ignores it](https://github.com/kennethreitz/requests/blob/master/requests/utils.py#L106).

At least having it emit a warning would have saved some hair-pulling.

dpursehouse:
If the netrc file exists but cannot be parsed or read, get_netrc_auth
silently fails.

Add a new argument `ignore_errors` which when set to False will cause
any parse/permission errors to be raised to the caller.  The default
value is True, which means the default behavior is unchanged.

Fixes #2654

Change-Id: I7436aaaf593178673ab84fd9e7ab4bcb0e3fe75e

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -67,7 +67,7 @@ def super_len(o):
         return len(o.getvalue())
 
 
-def get_netrc_auth(url):
+def get_netrc_auth(url, ignore_errors=True):
     """Returns the Requests tuple auth for a given url from netrc."""
 
     try:
@@ -106,7 +106,8 @@ def get_netrc_auth(url):
         except (NetrcParseError, IOError):
             # If there was a parsing error or a permissions issue reading the file,
             # we'll just skip netrc auth
-            pass
+            if not ignore_errors:
+                raise
 
     # AppEngine hackiness.
     except (ImportError, AttributeError):
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -67,7 +67,7 @@ def super_len(o):
         return len(o.getvalue())
 
 
-def get_netrc_auth(url, ignore_errors=True):
+def get_netrc_auth(url, raise_errors=False):
     """Returns the Requests tuple auth for a given url from netrc."""
 
     try:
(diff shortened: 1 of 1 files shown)


dpursehouse:
With a badly formed netrc file, `netrc.netrc` fails:

```
>>> from netrc import netrc
>>> netrc()
  File "/usr/lib/python2.7/netrc.py", line 32, in __init__
    self._parse(file, fp)
  File "/usr/lib/python2.7/netrc.py", line 94, in _parse
    file, lexer.lineno)
netrc.NetrcParseError: bad follower token 'fssdfs' (/home/user/.netrc, line 23)
```

Default behavior of `requests.utils.get_netrc_auth` is to silently fail:

```
>>> from requests.utils import get_netrc_auth
>>> get_netrc_auth("www.example.com")
```

With this patch, passing `ignore_errors=False` allows the parse error to be raised:

```
>>> get_netrc_auth("www.example.com", ignore_errors=False)
  File "requests/utils.py", line 101, in get_netrc_auth
    _netrc = netrc(netrc_path).authenticators(host)
  File "/usr/lib/python2.7/netrc.py", line 32, in __init__
    self._parse(file, fp)
  File "/usr/lib/python2.7/netrc.py", line 94, in _parse
    file, lexer.lineno)
netrc.NetrcParseError: bad follower token 'fssdfs' (/home/user/.netrc, line 23)
>>> 
```

sigmavirus24:
This is simple enough. Personally I'd rather have something like

``` py
from requests import utils

utils.get_netrc_auth('www.example.com', raise_errors=True)
```

I've found people tend to become more confused when passing `False` as an override instead of `True`.

dpursehouse:
@sigmavirus24 thanks for the feedback.  Added a new commit to change the logic as suggested.

sigmavirus24:
Cool. LGTM. I'll let @Lukasa weigh in.

Thanks @dpursehouse

Lukasa:
This LGTM.

PR #2661, opened 2015-07-03
Only pass useful timeouts to _get_conn

Lukasa:
So `_get_conn` doesn't use a urllib3 `Timeout` object, it only takes a float. It also only does stuff if `block` is `True`. Given that by default we use `block=False`, it won't do anything, so let's not use it. However, if someone changes that with a global constant they may want to be able to change this too, so let's make it a named constant.


shazow:
:custard:

sigmavirus24:
LGTM

PR #2666, opened 2015-07-13
Adds extra tests for CaseInsensitiveDict

gvangool:
This adds extra tstes for `requests.structures.CaseInsensitiveDict`

- Test for NotImplemented in `__eq__`
- Adds test for `copy()`
- Adds test for `__repr__()`

### DIFF
--- a/test_requests.py
+++ b/test_requests.py
@@ -1224,6 +1224,7 @@ class TestCaseInsensitiveDict(unittest.TestCase):
         del othercid['spam']
         assert cid != othercid
         assert cid == {'spam': 'blueval', 'eggs': 'redval'}
+        assert cid.__eq__(object()) == NotImplemented
 
     def test_setdefault(self):
         cid = CaseInsensitiveDict({'Spam': 'blueval'})
@@ -1260,6 +1261,20 @@ class TestCaseInsensitiveDict(unittest.TestCase):
         assert frozenset(i[0] for i in cid.items()) == keyset
         assert frozenset(cid.keys()) == keyset
         assert frozenset(cid) == keyset
+
+    def test_copy(self):
+        cid = CaseInsensitiveDict({
+            'Accept': 'application/json',
+            'user-Agent': 'requests',
+        })
+        assert cid == cid.copy()
+
+    def test_repr(self):
+        cid = CaseInsensitiveDict({
+            'Accept': 'application/json',
+            'user-Agent': 'requests',
+        })
+        assert repr(cid) == "{'Accept': 'application/json', 'user-Agent': 'requests'}"
 
 
 class UtilsTestCase(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


test_requests.py
```diff
@@ -1224,6 +1224,7 @@ def test_equality(self):
         del othercid['spam']
         assert cid != othercid
         assert cid == {'spam': 'blueval', 'eggs': 'redval'}
+        assert cid.__eq__(object()) == NotImplemented
```
sigmavirus24: Shouldn't this be

``` py
with pytest.raises(NotImplementedError):
    cid == object()
```

We shouldn't be directly invoking dunder methods anywhere in the code base, not even tests.

test_requests.py
```diff
@@ -1261,6 +1262,20 @@ def test_preserve_last_key_case(self):
...
         assert frozenset(cid) == keyset
 
+    def test_copy(self):
+        cid = CaseInsensitiveDict({
+            'Accept': 'application/json',
+            'user-Agent': 'requests',
+        })
+        assert cid == cid.copy()
+
+    def test_repr(self):
```
sigmavirus24, line 1275: I, for one, do not agree with testing reprs. @Lukasa thoughts?

test_requests.py
gvangool: Except that it's not raised, but returned. So not sure how else to test this.

sigmavirus24: If you have to access private methods to test something, you shouldn't be testing it.

I thought that returning `NotImplemented` triggered a `NotImplementedError` but that doesn't happen in all cases it seems.

``` py
cid == object()
# => False
```

But I don't think we strictly need this assertion anyway. You're testing an implementation detail, not a public part of the API (which is all we should ever test).

test_requests.py
Lukasa, line 1275: I only ever test reprs when I think they're an important part of the API. In this case I don't really mind.

test_requests.py
gvangool: @sigmavirus24 I've changed it in 47e71d1774745543b743bdf570d9d8e7151355e6

sigmavirus24:
@Lukasa how does this look to you? Looks fine to me. I don't mind a couple extra tests personally.

test_requests.py
```diff
@@ -1261,6 +1262,20 @@ def test_preserve_last_key_case(self):
         assert frozenset(cid.keys()) == keyset
         assert frozenset(cid) == keyset
 
+    def test_copy(self):
+        cid = CaseInsensitiveDict({
+            'Accept': 'application/json',
+            'user-Agent': 'requests',
+        })
+        assert cid == cid.copy()
```
Lukasa: Should this test also assert that the copy does not lead to two objects that are literally the same? e.g., should we also add `assert cid is not cid.copy()`?

Lukasa:
I have one note, but I'm happy to add in extra tests.

test_requests.py
gvangool: I believe I fixed this in 3b325f1bc5df476c4c4e93c5a742d81e2c5bd950

gvangool:
@Lukasa I believe I've addressed your concern for that :)

Lukasa:
Yeah, that looks good to me. \o/

Issue #2668, opened 2015-07-15
CaseInsensitiveDict.__repr__ is unordered


jayvdb:
CaseInsensitiveDict.__repr__ uses dict, and therefore the results are unordered.

However TestCaseInsensitiveDict.test_repr attempts to compare the __repr__ with an ordered representation.  This fails on Python 3.

jayvdb:
The alternative is to order the `__repr__`, but by what order..?

PR #2669, opened 2015-07-15
Test CaseInsensitiveDict.__repr__ unordered

jayvdb:
InsensitiveDict.__repr__ does not order the members, which on
Python 3 causes TestCaseInsensitiveDict.test_repr to fail.

Fixes issue #2668


sigmavirus24:
We should just remove this test. As I said when it was added, there's no need to have it.

Lukasa:
I'm happy to remove this test rather than mess about with it.

PR #2670, opened 2015-07-15
Re-enable Travis-CI builds

jayvdb:


jayvdb:
Ideally this isnt merged until #2668 is fixed.

sigmavirus24:
We will not be turning on Travis CI. ci.kennethreitz.org just needs a proper kick. /cc @kennethreitz

jayvdb:
The main repo doesnt need to enable Travis. This helps other developers testing on their forks.

Lukasa:
The main repo should be testing pull requests, which will help other developers on their forks. =)

jayvdb:
That only helps developers when they submit a pull request.  Some developers like to check their work against all configurations in the build matrix before doing a pull request.
If travis was enabled for pull requests, https://github.com/kennethreitz/requests/pull/2666 would not have been merged as travis would have commented on the pull request that it was broken.
If http://ci.kennethreitz.org/ is the preferred approach, it should be submitting comments to pull requests, or people with merge rights should be checking it manually before merging.

Lukasa:
> If http://ci.kennethreitz.org/ is the preferred approach, it should be submitting comments to pull requests, or people with merge rights should be checking it manually before merging.

No-one is denying that. ci.kennethreitz.org is currently broken. It should be functioning correctly. Currently it's the preferred approach, but obviously we need to get it working again.

As to developers on their forks, we encourage developers to open pull requests early if they want to do testing. Again, all of this assumes our process is functioning correctly: right now it is not. This is an organisational issue within the project, and re-adding travis will not fix it, especially because Travis is _very_ flaky when running our tests and regularly fails when it shouldn't.

PR #2672, opened 2015-07-15
Fix quickstart "Custom Headers" example intro

petedmarsh:
Previously this section prefaced an example with:

    For example, we didn't specify our content-type

But, the actual example set a custom user-agent header on the request. This
changes it to say "user-agent" instead which matches the given example.


petedmarsh:
Alternatively the example could be changed to actually set the content-type header instead of the user-agent header - either way I think they should be consistent :)

Lukasa:
Thanks @petedmarsh, I missed this when I changed the example from content-type to user-agent!

Issue #2675, opened 2015-07-17
Using header with a float value in a request results in error on Python 3.


sunu:
I have an example below. 

This seems to be a bug in `http.client` as it doesn't convert floats to string representation. But as a workaround we can may be explicitly convert any floats in the header to string in [`urlopen`](https://github.com/kennethreitz/requests/blob/master/requests/packages/urllib3/connectionpool.py#L421)?

```
$ ipython                
Python 3.4.0 (default, Apr 11 2014, 13:05:11) 
Type "copyright", "credits" or "license" for more information.

IPython 3.2.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import requests

In [2]: requests.__version__
Out[2]: '2.7.0'

In [3]: url = "http://sunu.in"

In [4]: headers = {'x-wait': 0.2}

In [5]: requests.get(url, headers=headers)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-951b2e0c9a38> in <module>()
----> 1 requests.get(url, headers=headers)

/home/sunu/gsoc/splash/env3/lib/python3.4/site-packages/requests/api.py in get(url, params, **kwargs)
     67 
     68     kwargs.setdefault('allow_redirects', True)
---> 69     return request('get', url, params=params, **kwargs)
     70 
     71 

/home/sunu/gsoc/splash/env3/lib/python3.4/site-packages/requests/api.py in request(method, url, **kwargs)
     48 
     49     session = sessions.Session()
---> 50     response = session.request(method=method, url=url, **kwargs)
     51     # By explicitly closing the session, we avoid leaving sockets open which
     52     # can trigger a ResourceWarning in some cases, and look like a memory leak

/home/sunu/gsoc/splash/env3/lib/python3.4/site-packages/requests/sessions.py in request(self, method, url, params, data, headers, cookies, files, auth, timeout, allow_redirects, proxies, hooks, stream, verify, cert, json)
    463         }
    464         send_kwargs.update(settings)
--> 465         resp = self.send(prep, **send_kwargs)
    466 
    467         return resp

/home/sunu/gsoc/splash/env3/lib/python3.4/site-packages/requests/sessions.py in send(self, request, **kwargs)
    571 
    572         # Send the request
--> 573         r = adapter.send(request, **kwargs)
[...]

PR #2676, opened 2015-07-17
Convert non-string header values to their string representation.

sunu:
Fixes https://github.com/kennethreitz/requests/issues/2675

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -25,7 +25,7 @@ from .exceptions import (
     HTTPError, MissingSchema, InvalidURL, ChunkedEncodingError,
     ContentDecodingError, ConnectionError, StreamConsumedError)
 from .utils import (
-    guess_filename, get_auth_from_url, requote_uri,
+    convert_to_string, guess_filename, get_auth_from_url, requote_uri,
     stream_decode_response_unicode, to_key_val_list, parse_header_links,
     iter_slices, guess_json_utf, super_len, to_native_string)
 from .compat import (
@@ -399,7 +399,10 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         """Prepares the given HTTP headers."""
 
         if headers:
-            self.headers = CaseInsensitiveDict((to_native_string(name), value) for name, value in headers.items())
+            self.headers = CaseInsensitiveDict((
+                                to_native_string(name),
+                                convert_to_string(value)
+                            ) for name, value in headers.items())
         else:
             self.headers = CaseInsensitiveDict()
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -693,6 +693,17 @@ def to_native_string(string, encoding='ascii'):
     return out
 
 
+def convert_to_string(obj):
+    """
+    Given a non-string object (neither unicode, nor bytes), returns a string
+    representation of that object.
+    """
+
+    if not isinstance(obj, basestring):
+        obj = str(obj)
+    return obj
+
+
 def urldefragauth(url):
     """
     Given a url remove the fragment and the authentication part
--- a/test_requests.py
+++ b/test_requests.py
@@ -1079,6 +1079,13 @@ class RequestsTestCase(unittest.TestCase):
         next(r.iter_lines())
         assert len(list(r.iter_lines())) == 3
 
+    def test_request_with_float_header(self):
+        """Headers with float value should not result in error."""
+        header = {'wait': 2.5}
+        r = requests.get(httpbin('headers'), headers=header)
+        assert r.status_code == 200
+        assert str(header['wait']) in r.text
+
 
 class TestContentEncodingDetection(unittest.TestCase):
 


Lukasa:
The change you've made is in urllib3. @shazow may want this change, in which case we should accept it over at the main urllib3 repository. If he does not want it, we should instead move it into the requests code.

shazow:
-1 on having this in urllib3's urlopen, that's far too low-level to mangle user input. Requests is probably the more appropriate layer for this.

Alternatively, it could live in `HTTPHeaderDict` once we get it usable as input as well as output, but that's blocked on https://github.com/shazow/urllib3/pull/633.

Lukasa:
That's pretty much what I thought.

In that case, I think the right place for this is `requests.models.PreparedRequest.prepare_headers()`.

sunu:
Thanks for the review @Lukasa and @shazow :) I've updated the PR now. 
https://github.com/kennethreitz/requests/blob/master/test_requests.py#L1275 fails for me on Python 3 because the keys are not in order, but looking at the implementation the keys are not expected to be in order?

test_requests.py
```diff
@@ -1079,6 +1079,12 @@ def test_response_iter_lines_reentrant(self):
         next(r.iter_lines())
         assert len(list(r.iter_lines())) == 3
 
+    def test_request_with_float_header(self):
+        """Headers with float value should not result in error."""
+        header = {'wait': 2.5}
+        r = requests.get(httpbin('get'), headers=header)
+        assert r.status_code == 200
```
Lukasa, line 1086: Can you assert the header got reflected back correctly?

requests/models.py
```diff
@@ -400,6 +400,11 @@ def prepare_headers(self, headers):
 
         if headers:
             self.headers = CaseInsensitiveDict((to_native_string(name), value) for name, value in headers.items())
+            # Explicitly convert any non-string header value to its string
+            # representation.
+            for key, val in self.headers.items():
+                if not isinstance(val, basestring):
+                    self.headers[key] = str(val)
```
Lukasa: This is the pretty close to the right place, but not really right. It'd be better if we could do it in the generator comprehension above.

Basically, wrap the `isinstance` block up in a function and put it in `requests.utils`. Probably call it something like `convert_to_string`. Then just call it on each value in the generator comprehension on L402.

Lukasa:
Yup, it's a known issue. =)

requests/models.py
sunu: I was thinking of doing this and then went into lazy mode. :D Let me fix it up real quick.

sunu:
@Lukasa I've updated the PR.

requests/models.py
```diff
@@ -399,7 +399,10 @@ def prepare_headers(self, headers):
         """Prepares the given HTTP headers."""
 
         if headers:
-            self.headers = CaseInsensitiveDict((to_native_string(name), value) for name, value in headers.items())
+            self.headers = CaseInsensitiveDict((
+                                to_native_string(name),
+                                convert_to_string(value)
+                            ) for name, value in headers.items())
```
Lukasa, line 405: This is really weird at this point, but can we do this instead?

``` python
self.headers = CaseInsensitiveDict(
    (to_native_string(name), convert_to_string(value))
    for name, value in headers.items()
)
```

This just reads a little bit better to me.

requests/utils.py
```diff
@@ -693,6 +693,17 @@ def to_native_string(string, encoding='ascii'):
     return out
 
 
+def convert_to_string(obj):
+    """
+    Given a non-string object (neither unicode, nor bytes), returns a string
+    representation of that object.
+    """
+
```
Lukasa, line 701: Can you remove this blank line please?

requests/utils.py
```diff
@@ -693,6 +693,17 @@ def to_native_string(string, encoding='ascii'):
     return out
 
 
+def convert_to_string(obj):
+    """
+    Given a non-string object (neither unicode, nor bytes), returns a string
+    representation of that object.
+    """
```
Lukasa, line 700: This docstring isn't quite right. It should be:

```
Given an object, converts it to a string. If the object is already an instance of `basestring`,
this is a no-op. Otherwise, uses the string representation of that object.
```

test_requests.py
```diff
@@ -1079,6 +1079,13 @@ def test_response_iter_lines_reentrant(self):
         next(r.iter_lines())
         assert len(list(r.iter_lines())) == 3
 
+    def test_request_with_float_header(self):
+        """Headers with float value should not result in error."""
+        header = {'wait': 2.5}
+        r = requests.get(httpbin('headers'), headers=header)
+        assert r.status_code == 200
+        assert str(header['wait']) in r.text
```
Lukasa, line 1087: Heh, not quite enough.

httpbin returns JSON. For this web request, it'd return something like this:

```
{
  "args": {}, 
  "headers": {
    "Accept": "*/*", 
    "Accept-Encoding": "gzip, deflate", 
    "Connection": "keep-alive", 
    "Host": "http2bin.org", 
    "User-Agent": "python-requests/2.7.0 CPython/2.7.10 Darwin/14.4.0", 
    "Via": "1.1 http2bin.org", 
    "Wait": "2.5"
  }, 
  "origin": "81.129.190.37", 
  "url": "http://http2bin.org/get"
}
```

So we'd like to assert that the value of the header was right. The assertion is: `assert r.json()['headers']['Wait'] == '2.5'`. (Or something close to it, the capitalisation might be wrong here.)

sigmavirus24:
See discussion on #2675 before continuing work here. I'm not sure we want to encourage this.

sunu:
Closing based on discussion on #2675 
Alas, it couldn't be my first contribution to requests. May be some time in the future :)

PR #2680, opened 2015-07-18
Fix docs for passing a list of values for a query string

peplin:
The special `[]` notation at the end of the field name is not necessary to get a field to appear with multiple values in the query string.


Lukasa:
Yeah, this is _mostly_ true. I'm going to merge this as-is @peplin, but I'd really love it if you'd follow up with another pull request that adds a paragraph about _why_ you might want to add the `[]` (namely, because many servers expect that notation).

Thanks so much for the contribution! :cake: :sparkles: :cake:

PR #2692, opened 2015-07-28
Dealing with removal of Authorization header on redirect and diff domain

casahome2000:
for services that require Authorization in the header, all redirect calls to a different domain would result in a 403-Forbidden.  Authorization should be preserved just as Authentication is.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -210,15 +210,6 @@ class SessionRedirectMixin(object):
         headers = prepared_request.headers
         url = prepared_request.url
 
-        if 'Authorization' in headers:
-            # If we get redirected to a new host, we should strip out any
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Sorry, we won't be accepting this. This is specifically written to fix a [vulnerability](http://www.cvedetails.com/cve/CVE-2015-2296/) that was reported to us recently. Without knowing your specific use-case I can't advise you on how to replicate the original behaviour, but if you post specifics to [our Q&A area](https://stackoverflow.com/questions/tagged/python-requests) you'll get an excellent answer.

casahome2000:
Thanks for looking into this request as well as providing the reference as to why it's not a good idea.  Appreciate the time.

PR #2696, opened 2015-07-29
Use debug instead of info to log new connection

luca3m:


sigmavirus24:
This belongs on urllib3, not here.

PR #2706, opened 2015-08-04
Fix merge setting for not preserving original order of dict parameters

ak1r4:
Fix a bug introduced by https://github.com/kennethreitz/requests/pull/1921

Bug: the ordered dictionary (the default `dict_class`) is accidentally converted back to normal dictionary.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -63,11 +63,9 @@ def merge_setting(request_setting, session_setting, dict_class=OrderedDict):
     merged_setting.update(to_key_val_list(request_setting))
 
     # Remove keys that are set to None.
-    for (k, v) in request_setting.items():
+    for (k, v) in merged_setting.items():
         if v is None:
             del merged_setting[k]
-
-    merged_setting = dict((k, v) for (k, v) in merged_setting.items() if v is not None)
 
     return merged_setting
 
--- a/test_requests.py
+++ b/test_requests.py
@@ -125,6 +125,13 @@ class RequestsTestCase(unittest.TestCase):
             "http://example.com/path?key=value#fragment", params={"a": "b"}).prepare()
         assert request.url == "http://example.com/path?key=value&a=b#fragment"
 
+    def test_params_original_order_is_preserved_by_default(self):
+        param_ordered_dict = collections.OrderedDict((('z', 1), ('a', 1), ('k', 1), ('d', 1)))
+        session = requests.Session()
+        request = requests.Request('GET', 'http://example.com/', params=param_ordered_dict)
+        prep = session.prepare_request(request)
+        assert prep.url == 'http://example.com/?z=1&a=1&k=1&d=1'
+
     def test_mixed_case_scheme_acceptable(self):
         s = requests.Session()
         s.proxies = getproxies()


sigmavirus24:
Can you explain how this is a bug? In what context do you expect order to preserved but it isn't?

Lukasa:
I think the bug here is that we don't use the chosen dict subclass, which means that an ordered dict might be lost. I think I'm tentatively +1 on this change.

sigmavirus24:
Based on the _test_ I'm guessing this affects `params`, `files`, `data`, and `headers`. In that case I'm +1 too, but I'd still like confirmation rather than accepting a fix without a bug description.

ak1r4:
@sigmavirus24 @Lukasa This bug was found when we generate signature based on the order of parameters (signature would be a hash of the url, for example), while on the server side, the signature is validated using the same process based on the query string passed to the server. In this case, the order of the parameters were changed by requests library underneath.

On the other hand, the approach #1921 took is not recommended either, because the order of keys even in normal dictionaries depends on when the insertion and deletion happens, so it is better not to reconstruct the dictionary, do a `del` instead. See the example below:

``` python
>>> d = {'b': 1, 'a': 1}
>>> d.update({'d': 1, 'z': 1, 'f': 1})
>>> d.items()
[('a', 1), ('b', 1), ('z', 1), ('f', 1), ('d', 1)]
>>> d1 = dict(d)
>>> d1.items()
[('a', 1), ('d', 1), ('b', 1), ('z', 1), ('f', 1)]
>>> d1 == d
True
>>> d1.items() == d.items()
False
```

ak1r4:
@sigmavirus24 @Lukasa Any thoughts on this PR?

Lukasa:
I'm :+1:: @sigmavirus24?

sigmavirus24:
:+1: from me too

Lukasa:
Alright, let's try to land this in 2.8.0. @sigmavirus24 I'll go ahead and merge it to the 2.8.0 branch and update the changelog appropriately.

Lukasa:
Ok, merged into the 2.8.0 branch in 5d7392f140ed476cdc99f4178e425ce2fbb05884.

kennethreitz:
:sparkles: :cake: :sparkles:

mjpieters:
This introduced a bug on Python 3; you are deleting keys while iterating over the dictionary items so Python 3 will throw a `RuntimeError: dictionary changed size during iteration` exception. Easily reproduced by using `requests.get('http://httpbin.org/get', headers={'foo': None})`:

``` python
>>> requests.get('http://httpbin.org/get', headers={'foo': None})
  File "/Users/mj/Development/venvs/stackoverflow-3.4/requests/requests/api.py", line 69, in get
  File "/Users/mj/Development/venvs/stackoverflow-3.4/requests/requests/api.py", line 50, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/Users/mj/Development/venvs/stackoverflow-3.4/requests/requests/sessions.py", line 455, in request
    prep = self.prepare_request(req)
  File "/Users/mj/Development/venvs/stackoverflow-3.4/requests/requests/sessions.py", line 382, in prepare_request
    headers=merge_setting(request.headers, self.headers, dict_class=CaseInsensitiveDict),
  File "/Users/mj/Development/venvs/stackoverflow-3.4/requests/requests/sessions.py", line 66, in merge_setting
    for (k, v) in merged_setting.items():
  File "/Users/mj/Development/Library/buildout.python/parts/opt/lib/python3.4/_collections_abc.py", line 503, in __iter__
    for key in self._mapping:
  File "/Users/mj/Development/venvs/stackoverflow-3.4/requests/requests/structures.py", line 60, in <genexpr>
    return (casedkey for casedkey, mappedvalue in self._store.values())
RuntimeError: dictionary changed size during iteration
```

mjpieters:
A work-around would be to collect the keys first:

``` python
none_keys = [k for k, v in merged_setting.items() if v is None]
for key in none_keys:
    del merged_setting[key]
```

mjpieters:
Fix available in #2737; ironically the test suite already caught this but I guess it wasn't yet run on 3.x? I'm ignoring the `test_expires_valid_str` failure I'm seeing, just so you know. :-P

PR #2712, opened 2015-08-07
Document use of sessions as context managers

### LINKED ISSUE #2580: Documentation for context manager?
The documentation makes only a single reference to the fact that `requests` provides some sort of context manager, and it's in the release notes from version 0.6 in 2011:

> New persistient sessions object and context manager

Would it be possible to get examples of how to implement these context managers? Specifically, I'm looking to replace the following code:

```python
    with contextlib.closing(urllib2.urlopen(self.url)) as dl:
```

I can't find any place that documents how to use an equivalent `requests` context manager, so I'll have to dig through the code to find out. It would be nice to get a bit more visibility into the `requests` API...

lukasgraf:
This adds some documentation on how to use sessions as **context managers** (fixes #2580). I also added a brief mention of **connection pooling** and its performance benefits.

I still got one question though: Is instantiating [`Session`](https://github.com/kennethreitz/requests/blob/9b067db19e20226dcb3aa407605d30942d085050/requests/sessions.py#L267) directly the proper API to document? I noticed that there's also a [`session()` factory function](https://github.com/kennethreitz/requests/blob/9b067db19e20226dcb3aa407605d30942d085050/requests/sessions.py#L674-L677) that explicitly mentions context management in its docstring. But I then saw this in the [API changes section](http://www.python-requests.org/en/latest/api/?highlight=backwards%20compatibility#api-changes):

> The `Session` API has changed. Sessions objects no longer take parameters. `Session` is also now capitalized, but it can still be instantiated with a lowercase `session` for backwards compatibility.

So I therefore used `Session()` in the examples, and I'm assuming that factory function is only there for backwards compatibility. Should I update its docstring with something like `Alias for backwards compatibility`?

Let me know how you'd like it documented, and I'll update my PR accordingly.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -12,7 +12,10 @@ Session Objects
 
 The Session object allows you to persist certain parameters across
 requests. It also persists cookies across all requests made from the
-Session instance.
+Session instance, and will use ``urllib3``'s `connection pooling`_. So if
+you're making several requests to the same host, the underlying TCP
+connection will be reused, which can result in a significant performance
+increase (see `HTTP persistent connection`_).
 
 A Session object has all the methods of the main Requests API.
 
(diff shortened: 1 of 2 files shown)


kennethreitz:
This is great, thanks!

:sparkles: :cake: :sparkles:

PR #2713, opened 2015-08-07
Docs: Clarify that method-level parameters are not persisted in sessions.

### LINKED ISSUE #2488: Need documentation about "Cookies set on individual Requests through a `Session` are not persisted to the session"
It took me a while to figure it out. Thought it might be a bug initially, I then came across #1791 & #1728.

Since this feature by design is quite confusing. I feel it deserves some more attention apart from simply mention it in the changelog (0b68037).

Can we have it as a tip/hint some where in the documentation (or quick start)? And suggest the proper way to prepare the persistent cookies.

lukasgraf:
This adds a paragraph to the **session docs** that clarifies the fact that **method-level parameters are not persisted across requests**, even when a session is being used (fixes #2488).

As an example I used cookies, and included a pointer to the [Cookie utility functions](http://docs.python-requests.org/en/latest/api/#cookies). In order to be able to link to that section I added some section labels in `docs/api.rst` (prefixed with `api-`, because otherwise there would be label collisions).

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -81,6 +81,8 @@ Status Code Lookup
     >>> requests.codes['\o/']
     200
 
+.. _api-cookies:
+
 Cookies
 ~~~~~~~
 
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -45,6 +45,24 @@ Any dictionaries that you pass to a request method will be merged with the
 session-level values that are set. The method-level parameters override session
 parameters.
 
+Note, however, that method-level parameters will *not* be persisted across
+requests, even if using a session. This example will only send the cookies
+with the first request, but not the second::
+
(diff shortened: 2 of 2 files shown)


sigmavirus24:
Eh, I'd rather not add a bunch of labels to one file just to be able to use _1_ in the docs.

lukasgraf:
@sigmavirus24 got it, I'll remove all the ones except the one for the cookies section. Should I stick with the `api-` prefix or drop that as well?

sigmavirus24:
The api- prefix makes sense. Thanks.

lukasgraf:
@sigmavirus24 updated

sigmavirus24:
Thanks @lukasgraf! :sparkles: :cake: :sparkles:

PR #2728, opened 2015-08-15
Docs: Fix links to `timeouts` section by using :ref:

### LINKED ISSUE #2698: 404 for timeouts advanced documentation links
Three files [in this search](https://github.com/kennethreitz/requests/search?l=python&q=<user/advanced.html#timeouts>&type=Code&utf8=✓) has links to `<user/advanced.html#timeouts>` which works good for documentation generated on the local machine.
But on the http://docs.python-requests.org/ there is no ".html" in pages urls so 404 Page not found is returned.

I think docs.python-requests.org will be ok if you change links to `<user/advanced/#timeouts>` but local machine documentation will be broken.

lukasgraf:
This fixes three links (from docstrings) to the [<`timeouts`>](http://docs.python-requests.org/en/latest/user/advanced/#timeouts) section by using `:ref:` instead of linking to `.html` files (which are only available when building docs locally, but not in the published docs on RTD).

Fixes #2698

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -327,8 +327,8 @@ class HTTPAdapter(BaseAdapter):
         :param request: The :class:`PreparedRequest <PreparedRequest>` being sent.
         :param stream: (optional) Whether to stream the request content.
         :param timeout: (optional) How long to wait for the server to send
-            data before giving up, as a float, or a (`connect timeout, read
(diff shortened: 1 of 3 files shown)


Lukasa:
LGTM, thanks! :cake: :sparkles: :cake:

PR #2736, opened 2015-08-20
quick fix to make session cookies work with mocked HTTP responses

FodT:
I ran into an issue working with Responses/HTTMock where session cookies were not being set. Tracked it down to here, looked like a quick (possibly dirty?) fix. Doesn't break any tests!

Throwing it up here for wider consideration. Comments welcome.

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -120,14 +120,16 @@ def extract_cookies_to_jar(jar, request, response):
     :param response: urllib3.HTTPResponse object
     """
     if not (hasattr(response, '_original_response') and
-            response._original_response):
-        return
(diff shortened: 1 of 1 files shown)


Lukasa:
Generally speaking, we don't really believe it's the role of this library to fix abstraction errors in Responses or HTTMock. Our goal is to function with the standard library, which this code does. In rare cases we allow for really obvious differences (e.g. `StringIO`), but we are _extremely_ reluctant to add more special case code to account for libraries that _pretend_ to be other libraries. This is particularly true when it comes to attribute access, given that Responses or HTTMock could easily just provide that attribute in their mocked response.

I recommend opening this issue with the respective projects. =) Sorry!

PR #2737, opened 2015-08-21
Avoid altering dictionary size during iteration

mjpieters:
Fixes `RuntimeError` on Python 3 introduced in 53ea23128e375ef84511fd2b22b740eca15d0854


Lukasa:
Yup, this looks right to me. Thanks @mjpieters!

PR #2740, opened 2015-08-24
params of proxy_headers are inconsistent with its docstrings

qingyunha:


Lukasa:
Thanks @qingyunha! :cake:

PR #2741, opened 2015-08-24
Implement per-host proxies

### LINKED ISSUE #2722: Proxies that depend on the URL
Currently the proxies dictionary supports only schemes as keys.  In networks with multiple proxies, this may not be sufficient for picking the correct proxy.  I propose that we allow scheme+hostname, hostname, and scheme (in that order) as keys, searched in that order.  Changing this line, https://github.com/kennethreitz/requests/blob/2440b6f089f4c53e96cd2ece80e92774b72ed243/requests/adapters.py#L242, could search first for the scheme+hostname, then the hostname, then the scheme.

I'm not sure of all the ramifications of this in other parts of the codebase, but this would greatly help when we have a more complicated proxy setup than one just relying on schemes.

jasongrout:
This change allows the proxy dict to be have entries of the form
`{'<scheme>://<hostname>': '<proxy>'}`.  Host-specific proxies will be used in
preference to the scheme-specific proxies (i.e., proxy dict entries with keys
like 'http' or 'https').

Fixes #2722

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -483,7 +483,10 @@ To use HTTP Basic Auth with your proxy, use the `http://user:password@host/` syn
     }
 
 To give a proxy for a specific scheme and host, use the
-`scheme://host` form for the key::
+`scheme://hostname` form for the key.  This will match for
+any request to the given scheme and exact hostname.
+
+::
 
     proxies = {
       "http://10.20.1.128": "http://10.10.1.10:5323",
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -459,17 +459,11 @@ If you need to use a proxy, you can configure individual requests with the
     import requests
 
     proxies = {
-      "http://10.20.1.128": "http://10.10.1.10:5323",
-      "https://10.20.1.140": "http://10.10.1.10:6431"
       "http": "http://10.10.1.10:3128",
       "https": "http://10.10.1.10:1080",
     }
 
     requests.get("http://example.org", proxies=proxies)
-
-A request's ``<scheme>://<hostname>`` is looked up in the proxies
-dictionary first, to provide host-specific proxies.  If no
-host-specific proxy is found, the request's scheme is looked up.
 
 You can also configure proxies by setting the environment variables
 ``HTTP_PROXY`` and ``HTTPS_PROXY``.
@@ -488,8 +482,15 @@ To use HTTP Basic Auth with your proxy, use the `http://user:password@host/` syn
         "http": "http://user:pass@10.10.1.10:3128/",
     }
 
-Note that proxy URLs must include the scheme.
+To give a proxy for a specific scheme and host, use the
+`scheme://host` form for the key::
 
+    proxies = {
+      "http://10.20.1.128": "http://10.10.1.10:5323",
+    }
+
+Note that proxy URLs must include the scheme.
+    
 .. _compliance:
 
 Compliance
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -459,11 +459,17 @@ If you need to use a proxy, you can configure individual requests with the
     import requests
 
     proxies = {
+      "http://10.20.1.128": "http://10.10.1.10:5323",
+      "https://10.20.1.140": "http://10.10.1.10:6431"
       "http": "http://10.10.1.10:3128",
       "https": "http://10.10.1.10:1080",
     }
 
     requests.get("http://example.org", proxies=proxies)
+
+A request's ``<scheme>://<hostname>`` is looked up in the proxies
+dictionary first, to provide host-specific proxies.  If no
+host-specific proxy is found, the request's scheme is looked up.
 
 You can also configure proxies by setting the environment variables
 ``HTTP_PROXY`` and ``HTTPS_PROXY``.
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -239,7 +239,8 @@ class HTTPAdapter(BaseAdapter):
         :param proxies: (optional) A Requests-style dictionary of proxies used on this request.
         """
         proxies = proxies or {}
-        proxy = proxies.get(urlparse(url.lower()).scheme)
+        u = urlparse(url.lower())
+        proxy = proxies.get(u.scheme+'://'+u.hostname, proxies.get(u.scheme))
 
         if proxy:
             proxy = prepend_scheme_if_needed(proxy, 'http')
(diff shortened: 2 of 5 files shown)


Lukasa:
This looks reasonable enough to me.

Thoughts @sigmavirus24?

sigmavirus24:
Nope. This thoroughly breaks backwards compatibility. This changes the behaviour to **only** behave for per-host which means anyone passing `proxies={'http': ..., 'https': ...}` will see everything fail as soon as they upgrade.

We should support both, not one or the other.

sigmavirus24:
Closing/re-opening to trigger a new CI build

sigmavirus24:
Also there aren't any tests for this

Lukasa:
This does support both.

It does a lookup for `u.scheme+'://'+u.hostname`, and if that does not exist falls back to the get for `u.scheme`. Or have I missed something?

jasongrout:
@Lukasa - that's what I intended, and what I'm sure the code does.  Fall back to the scheme lookup if the scheme+hostname lookup fails.  Would you rather I make it 5 lines of code to make that clearer?

jasongrout:
@sigmavirus24 - this is my first contribution to requests.  How does testing work in this package?

jasongrout:
I added a note to the documentation about this.

sigmavirus24:
> It does a lookup for u.scheme+'://'+u.hostname, and if that does not exist falls back to the get for u.scheme. Or have I missed something?

Nesting it like that made me miss it without having had enough coffee. It should be done separately to be very explicit.

As for tests, they go in the `test_requests.py` file.

jasongrout:
@sigmavirus24, I didn't see any tests in test_requests.py for the current proxies argument to request methods.  I saw a few places where it is set from the urllib getproxies function, presumably to do some actual network calls.  I also saw two tests for the no_proxy environment variable.  Is there a test for the proxies argument to a request that I missed?  I'm happy to write one, just wanted to make sure I didn't miss it (and I was looking for a model for how to write a test that is consistent with your framework).

Lukasa:
Testing proxies code is tricky, normally, because of our tendency to do 'live' testing, and because this function is very side-effect-y. The best way to do it might be to pull this 'find the correct proxy url' logic into its own function in `util.py` and then unit test that specifically to verify the logic. I think that would keep everyone happy. =)

jasongrout:
@Lukasa, I've done that now.  I'm not very clear on how this plays with the environment proxies, especially in rebuild_proxies() in the sessions code (https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L228).  Any help from someone that already knows the call flow and how the proxies are changed at each step is appreciated.  I think I don't need to do anything, but I'm not absolutely sure.

requests/adapters.py
```diff
@@ -272,13 +272,10 @@ def request_url(self, request, proxies):
...
-        :param proxies: A dictionary of schemes to proxy URLs.
+        :param proxies: A dictionary of schemes or schemes and hosts to proxy URLs.
         """
-        proxies = proxies or {}
-        scheme = urlparse(request.url).scheme
-        proxy = proxies.get(scheme)
-
-        if proxy and scheme != 'https':
+        proxy = select_proxy(request.url, proxies)
+        if proxy and not request.url.lower().startswith('https'):
```
Lukasa: This is a regressing, breaking change. It's possible that `request.url` may begin with `https` but not actually be a HTTPS request. Please keep the urlparse logic in place. =)

Lukasa:
One note here, otherwise fine.

The change you've made should be ok, but we do need to verify it. I can't do that right now, but I'll hopefully get to it Monday.

requests/adapters.py
jasongrout: I used that because it is also used at https://github.com/jasongrout/requests/blob/per-host-proxy/requests/adapters.py#L171.  I wondered about the correctness.  Sounds like we should fix it there too?

jasongrout: FYI, it looks like there is an order of magnitude difference on my machine between using startswith and using urlparse, from about 300 ns for startswith, to 3 microseconds for urlparse.

requests/adapters.py
```diff
@@ -272,12 +272,10 @@ def request_url(self, request, proxies):
...
-        :param proxies: A dictionary of schemes to proxy URLs.
+        :param proxies: A dictionary of schemes or schemes and hosts to proxy URLs.
         """
-        proxies = proxies or {}
-        scheme = urlparse(request.url).scheme
-        proxy = proxies.get(scheme)
-
+        proxy = select_proxy(request.url, proxies)
+        scheme = urlparse(request.url.lower()).scheme
         if proxy and scheme != 'https':
```
jasongrout, line 279: How come we don't do `if proxy and scheme == 'http':`, like the documentation seems to indicate?

requests/adapters.py
Lukasa: That performance cost is fine: the virtue of correctness here is higher than the virtue of speed. ;)

jasongrout:
@Lukasa - a friendly ping about reviewing this.  Thanks again for helping this along!

docs/user/advanced.rst
```diff
@@ -482,8 +482,15 @@ To use HTTP Basic Auth with your proxy, use the `http://user:password@host/` syn
...
-Note that proxy URLs must include the scheme.
+To give a proxy for a specific scheme and host, use the
+`scheme://host` form for the key::
 
+    proxies = {
+      "http://10.20.1.128": "http://10.10.1.10:5323",
+    }
+
+Note that proxy URLs must include the scheme.
+
```
Lukasa, line 496: Can we comment here that this is _not_ a longest-prefix-match, but it must match _exactly_? E.g. If you set `proxies={'http://10.21.1.128': 'some_url'}`, you will not get that proxy for `http://10.21.1.128:8080`. I feel like this could otherwise trip people up.

Lukasa:
One small docs note. Otherwise, a live test appears to validate that this works. =)

docs/user/advanced.rst
jasongrout, line 496: actually, it would work, since it's only scheme://hostname that is checked, not the port.  I'll still add a note.

jasongrout:
I changed the docs to emphasize that the per-host proxy matches only on the scheme and exact hostname.  Thanks!

Lukasa:
Ok, I'm officially happy with this. @sigmavirus24, can I get you to take a quick look?

sigmavirus24:
Looks good to me. Thanks for your hard work and diligence @jasongrout

Lukasa:
Hurrah! @sigmavirus24 do we want to merge this before 2.8.0 or hold off until the release after?

sigmavirus24:
May as well pull it into 2.8.0. Seems like @jasongrout needs this sooner rather than later.

Lukasa:
Cool!

jasongrout:
Thanks!

jasongrout:
@Lukasa - just curious, is there a planned release date for 2.8.0?

Lukasa:
There is not. I'm hoping to get to it this week, maybe tomorrow, but currently 2.8.0's release is a best-effort kind of thing.

PR #2742, opened 2015-08-24
Include file-like object in data arg description

everett-toews:
If `requests.api.request()` uses `requests.sessions.Session.request()` to get it's work done then the data arg description should match between the two.


sigmavirus24:
Thanks @everett-toews ! :sparkles: :cake: :sparkles:

Issue #1558, opened 2013-08-27
Make instances of PreparedRequest class pickable


piotr-dobrogost:
See [How can I save a python HTTP requests from python-requets so I can execute it later?](http://stackoverflow.com/q/18462749/95735) question on Stack Overflow.

sigmavirus24:
Is that the only remaining class that isn't Pickleable? If so, I'll work on it for 2.0.

Lukasa:
Looks like it might be. I'd love this in 2.0. =)

ssbarnea:
:+1: If we make a patch for the current version, would you accept it? -- I want to improve the python-jira library without having to use a non-released version of requests.

Lukasa:
@ssbarnea To the best of my knowledge we have no plans to release any further 1.2.X point releases. The next planned release is 2.0.0. I'm potentially open to you making a patch against master, but I'd want to check with @kennethreitz first.

kennethreitz:
Go for it

```
--
```

Kenneth Reitz

Lukasa:
Sweet, feel free to patch against master then. =)

Lukasa:
Ok, looking at this again we don't currently make any of the trio of `Request`, `PreparedRequest` and `Response` pickleable. `Response` is potentially difficult to pickle (file descriptors, underlying urllib3 response object isn't pickleable), but the other two are the things worth making pickleable anyway. We should shoot for both of them I think.

sigmavirus24:
I'm not (quite frankly) convinced that a `Request` object should be entirely pickleable. It is an ephemeral object that we use to construct the `PreparedRequest`. `PreparedRequest` objects do not store a reference to it, so I'm 90% sure it does not need to be pickleable. We should probably make it pickleable anyway though it is a seriously minor introduction of code and it will cover us for the future.

I'm wondering if @shazow would be opposed to making the `urllib3` response object pickleable. If he were to approve, then I think it would make sense to also make the Response object pickleable.

Lukasa:
The biggest issue with pickling `PreparedRequest`s is pickling hooks. We could be super-clever and try to pickle the function objects but I'm not convinced that's a great plan. Alternative suggestions?

shazow:
Hrm, yea not sure if it makes a lot of sense. Especially since the Response object has a lot of state within it, state which is network-dependent too (is the socket still open? has the body been read? etc).

What would it take to make urllib3's Response object pickleable?

ssbarnea:
[...]

PR #2757, opened 2015-09-05
Add PreparedRequest pickling tests

keyan:
Trying to address #1558 by first creating a failing test, but I'm not sure when `PreparedRequest`s fail to be pickleable. Pulling early to see if anyone has suggestions on how to re-create the issue, also maybe these tests are useful on their own as well.

### DIFF
--- a/test_requests.py
+++ b/test_requests.py
@@ -802,6 +802,43 @@ class RequestsTestCase(unittest.TestCase):
         assert r.request.url == pr.request.url
         assert r.request.headers == pr.request.headers
 
+    def test_prepared_request_is_pickleable(self):
+        p = requests.Request('GET', httpbin('get')).prepare()
+
+        # Verify we can pickle the request.
+        assert pickle.dumps(p)
+        r = pickle.dumps(p)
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks @keyan! From the issue:

> On a prepared request, the only thing I think can hurt us is the case when the body attribute is a file object. Is there a specific hook that we're concerned about not being pickleable?

So the concerns seem to be body file objects and hooks.

sigmavirus24:
@keyan do you have intentions to update this with the tests we were looking to add?

keyan:
Hey @sigmavirus24, I could use some advice on how to add a file object in the test environment. I just added a commit with a failing test containing a hook. Not sure how to address the actual issue though.

Lukasa:
@keyan It is often easiest to just create a file handle to the file under test: `open(__file__)`.

kennethreitz:
Closing due to inactivity, (now) merge conflict, and failing tests. These look very much needed though, so please open another PR when ready to proceed!

PR #2758, opened 2015-09-05
Fix RFC Quotation

mhils:
Minor nitpick: The quote in http://www.python-requests.org/en/latest/user/quickstart/#response-headers is currently not formatted properly. This PR fixes this.


Lukasa:
Fantastic @mhils! Thanks! :cake: :sparkles: :cake:

Issue #2756, opened 2015-09-04
Json supersedes data in prepare_body


xEtherealx:
When not a stream, json supersedes data in prepare_body:
https://github.com/kennethreitz/requests/blob/f5dacf84468ab7e0631cc61a3f1431a32e3e143c/requests/models.py#L446

This conflicts with the docstring, which indicates that json is only used when data is not specified:
https://github.com/kennethreitz/requests/blob/f5dacf84468ab7e0631cc61a3f1431a32e3e143c/requests/models.py#L195

Lukasa:
Thanks for this! Yes, that's well spotted: we're definitely in conflict with the documented behaviour.

Lukasa:
I may try to find someone who wants to get started in requests to see if they want to fix this: it's a perfect fix for someone relatively new to OSS.

taion:
BTW, `files` also supersedes `json` - the docs could use a small update there.

Lukasa:
This is resolved in 2.8.0. Thanks all!

PR #2761, opened 2015-09-07
Fix for Json supersedes data in prepare_body #2756

ravipra:
Now json will be attached to the body only if data is not present. 
(I hope the fix is not more complicated than this.)


ravipra:
Tests are failing. Will figure out how to run the tests, fix it and will resubmit. Apologies for the trouble!

PR #2763, opened 2015-09-08
Fixed issue #2756

ravipra:
Now 'json' parameter will be used to prepare body only if the 'data'
parameter is not present

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -414,7 +414,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         content_type = None
         length = None
 
-        if json is not None:
+        if not data and json is not None:
             content_type = 'application/json'
             body = complexjson.dumps(json)
 
@@ -443,7 +443,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             if files:
                 (body, content_type) = self._encode_files(files, data)
             else:
-                if data and json is None:
+                if data:
                     body = self._encode_params(data)
                     if isinstance(data, basestring) or hasattr(data, 'read'):
                         content_type = None
--- a/requests/models.py
+++ b/requests/models.py
@@ -414,7 +414,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         content_type = None
         length = None
 
-        if not data and json is not None:
+        if data == {} and json is not None:
             content_type = 'application/json'
             body = complexjson.dumps(json)
 
--- a/test_requests.py
+++ b/test_requests.py
@@ -1062,6 +1062,13 @@ class RequestsTestCase(unittest.TestCase):
         assert 'application/json' in r.request.headers['Content-Type']
         assert {'life': 42} == r.json()['json']
 
+    def test_json_param_post_should_not_override_data_param(self):
+        r = requests.Request(method='POST', url='http://httpbin.org/post',
+                             data={'stuff'.encode('utf-8'): 'elixr'},
+                             json={'music'.encode('utf-8'): 'flute'})
+        prep = r.prepare()
+        assert 'stuff=elixr' == prep.body
+
     def test_response_iter_lines(self):
         r = requests.get(httpbin('stream/4'), stream=True)
         assert r.status_code == 200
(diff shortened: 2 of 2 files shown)


requests/models.py
```diff
@@ -414,7 +414,7 @@ def prepare_body(self, data, files, json=None):
         content_type = None
         length = None
 
-        if json is not None:
+        if not data and json is not None:
```
Lukasa: This should really be `if data is None`.

Lukasa:
One minor comment, otherwise I'm fine. =)

ravipra:
With `if data is None` I get the following test failure. Please advice.

```
|======================================================================
ERROR: test_json_param_post_content_type_works (__main__.RequestsTestCase)
----------------------------------------------------------------------
   File "test_requests.py", line 1062, in 
test_json_param_post_content_type_works
     assert 'application/json' in r.request.headers['Content-Type']
   File "/home/ravi/oss/requests-temp/requests/structures.py", line 54, 
in __getitem__
     return self._store[key.lower()][1]
KeyError: 'content-type'

----------------------------------------------------------------------
Ran 148 tests in 85.057s

FAILED (errors=1)
```

Lukasa:
Ugh, you're right, our code sets data to the empty dict. Nevermind, I rescind that feedback. This LGTM, @sigmavirus24?

requests/models.py
sigmavirus24: Since we apparently can't use `if data is None` can we compare to an empty dictionary? Explicit is so much better than implicit.

test_requests.py
```diff
@@ -1062,6 +1062,13 @@ def test_json_param_post_content_type_works(self):
         assert 'application/json' in r.request.headers['Content-Type']
         assert {'life': 42} == r.json()['json']
 
+    def test_json_param_post_should_not_override_data_param(self):
+        r = requests.Request(method='POST', url='http://httpbin.org/post',
+                             data={'stuff'.encode('utf-8'): 'elixr'},
```
sigmavirus24: Why are we calling `str.encode`?

ravipra: Copy-paste from the other testcase! :) Will remove.

requests/models.py
ravipra: Sure, will change.

kennethreitz:
:+1:

requests/models.py
```diff
@@ -414,7 +414,7 @@ def prepare_body(self, data, files, json=None):
         content_type = None
         length = None
 
-        if json is not None:
+        if data == {} and json is not None:
```
kennethreitz, line 417: I am skeptical of this line.

ueg1990, line 417: Skeptical of the comparison? If yes, following is a suggestion

``` python
if not data and json is not None:
    ....
```

sigmavirus24, line 417: @kennethreitz why are you skeptical? We should be explicit in what we're expecting to see if only to benefit future "us". `not data` is what was there and caused some minor confusion for @Lukasa when he initially reviewed this.

sumitbinnani:
The documentation also needs to be updated.
The `files` also supersedes `json` as already pointed out by @taion in comments of #2756

Lukasa:
@sumitbinnani I'd happily accept a pull request that adds only that documentation change.

sigmavirus24:
This looks okay to me. How about you @Lukasa? We should probably not have this pending after having merged the documentation for next week.

Lukasa:
I'm happy for us to merge this in for 2.8.0.

requests/models.py
gyzpunk, line 417: This line introduce an other issue, when the request is created with the models.Request object.
In this case, if we instanciate the Request object only with `json` argument (no `data`), then `data` is initialized with an empty list (not dict).

Then when the preparation process arrive to the previous comparison, json is never taken into account (`[] != {}`)

I can propose this : 

``` python
if not data and json:
    [...]
```

Lukasa, line 417: @gyzpunk Did you try with 2.8.1?

kennethreitz, line 417: `if not data and json is not None` is what I would have written, yes :)

PR #2764, opened 2015-09-09
minor refactoring to function

suhussai:
I was just looking through the code base and noticed this small "improvement" I could make. Is the focus more on readability or is this kind of refactoring acceptable as well? I understand that this format does decrease readability a little bit, so I didn't look into it too much, thought I'd get feedback first.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -265,14 +265,7 @@ def dict_from_cookiejar(cj):
 
     :param cj: CookieJar object to extract cookies from.
     """
-
-    cookie_dict = {}
-
-    for cookie in cj:
-        cookie_dict[cookie.name] = cookie.value
-
-    return cookie_dict
-
+    return {cookie.name:cookie.value for cookie in cj}
 
 def add_dict_to_cookiejar(cj, cookie_dict):
     """Returns a CookieJar from a key/value dictionary.


sigmavirus24:
Dictionary comprehensions were introduced in Python 2.7. Prior to that they are invalid syntax, e.g., for Python 2.5 and 2.6, both of which are supported versions for ~~pep8~~ requests. 

Thanks for contributing though @suhussai!

Perhaps you could look through the bugs and find one that doesn't have a linked PR to try to work on.

Cheers,
Ian!

**Edit** I said pep8 but I meant requests. I clearly can't keep my projects straight anymore. :(

suhussai:
Your right, I noticed that in the tests. Even though dictionary comprehension isn't supported directly in python2.6, this format works according to http://stackoverflow.com/questions/1747817/python-create-a-dictionary-with-list-comprehension:

```
dict((cookie.name, cookie.value) for cookie in cj)
```

I installed python2.6/2.5 and ran an example program using that format, and it worked as expected. Do we want to include this other format, which is supported by at least python 2.5 and greater?

sigmavirus24:
So `dict()` is actually bizarrely slow. That said, being a style issue I don't think we're concerned enough to change it.

suhussai:
Ah, I didn't know that. Thanks for the information.

PR #2768, opened 2015-09-10
Replace hardcoded class with dunder class attribute

ueg1990:
Small suggestion as this was something I heard in Raymond Hettinger's talk at PyCon 2015

### DIFF
--- a/requests/structures.py
+++ b/requests/structures.py
@@ -72,7 +72,7 @@ class CaseInsensitiveDict(collections.MutableMapping):
 
     def __eq__(self, other):
         if isinstance(other, collections.Mapping):
-            other = CaseInsensitiveDict(other)
+            other = self.__class__(other)
         else:
             return NotImplemented
         # Compare insensitively
@@ -80,7 +80,7 @@ class CaseInsensitiveDict(collections.MutableMapping):
 
     # Copy is required
     def copy(self):
-        return CaseInsensitiveDict(self._store.values())
+        return self.__class__(self._store.values())
 
     def __repr__(self):
         return str(dict(self.items()))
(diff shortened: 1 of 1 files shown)


ueg1990:
Btw if this is accepted, i can do it for the other modules/python files...in that way i can also read some of the requests code :stuck_out_tongue:

requests/structures.py
```diff
@@ -90,7 +90,7 @@ class LookupDict(dict):
 
     def __init__(self, name=None):
         self.name = name
-        super(LookupDict, self).__init__()
+        super(self.__class__, self).__init__()
```
jmoldow, line 93: This is not correct. The class name must always be hard-coded when using `super()`.

Lukasa:
I certainly want the change in `super` removed: as pointed out in the diff comment by @jmoldow, that isn't valid. Otherwise, I suspect it makes the class somewhat more subclass-friendly, which I don't object to.

ueg1990:
i am not sure why this is considered invalid. can you elaborate?

jmoldow:
There is a reason that `super()` requires two arguments. It's because the first argument can't be deduced from `self`, because the first argument isn't (in general) `self.__class__`.

If you have a method-resolution order C -> B -> A -> object, `C.__init__` is going to call `super(C, self).__init__`, which will call `B.__init__`, which will call `super(B, self).__init__`, which will call `A.__init__`. The only call where the first argument is `self.__class__` is the first one, but you don't know in advance what will even be the first one.

The first argument needs to be the hard-coded class name, in order for super to figure out what the super object should refer to.

sigmavirus24:
Actually `__class__` and `self.__class__` are a whole different level of magic that will thoroughly break this for any subclass. Sorry @ueg1990 this is not something I will allow in requests.

ueg1990:
lol i do not have any objection to that. learnt something new in python :)
onto the next pull request

PR #2770, opened 2015-09-11
Use context manager for Session

ueg1990:
Since a context manager is defined for session, i thought it would be better (and more pythonic) to use it. All the tests passed. Let me know what you guys think

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -46,12 +46,8 @@ def request(method, url, **kwargs):
       <Response [200]>
     """
 
-    session = sessions.Session()
-    response = session.request(method=method, url=url, **kwargs)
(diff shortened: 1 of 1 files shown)


Lukasa:
I'm afraid we actually don't want this.

The use-case is extremely uncommon, but it's possible that if an exception is hit during request processing that a user may want to keep hold of state from the connection pool or urllib3 layer. Using the context manager potentially prevents that by forcing the Session closed.

Basically, remember that a context manager is not the same as this:

``` python
obj = something()
obj.action()
obj.close()
```

It's the same as this:

``` python
obj = something()

try:
    obj.action()
finally:
    obj.close()
```

PR #2772, opened 2015-09-11
Move CA Certificates documentation next to SSL Cert Verification

jimbrowne:
When reading the SSL Cert Verification documentation and working on directing requests to use a system's local CA store I never noticed the CA Certificates documentation as it is at the tail end of the Advanced page.  I suggest it would be better to have the CA Certificates documentation immediately follow the documentation regarding SSL Cert Verification.

No content change, just a relocation of existing verbiage.


Lukasa:
Thanks @jimbrowne! :cake: :sparkles: :cake:

PR #2774, opened 2015-09-13
Some pythonic suggestions for requests/hooks.py

ueg1990:


kennethreitz:
these look pretty good to me! :sparkles: :cake: :sparkles:

PR #2775, opened 2015-09-13
Removed unused module from hooks.py

ueg1990:


Lukasa:
Thanks @ueg1990! :cake:

sigmavirus24:
You know someone's going to complain that they were using `hooks.defaultdict` now, right? ;)

PR #2782, opened 2015-09-21
Does't escaped ASCII characters when post json data by default

tao12345666333:


sigmavirus24:
@tao12345666333 could you explain this? Why is this necessary? Did you encounter a bug? We don't just accept code thrown over the wall without an explanation.

tao12345666333:
@sigmavirus24 Thanks for your reply.

When I use requests to post json data, the json data will be escaped by default. 
My code is here:

```
#!/usr/bin/env python
# coding=utf-8
import requests


def push_message():
    body = {
        'touser': '@all',
        'msgtype': 'text',
        'agentid': 0,
        'text': {
            'content': '测试'
        },
        'safe': 0
    }

    res = requests.post(
        url='http://mytest_url',
        json=body,
        headers={
            'Content-Type': 'application/json',
            'encoding': 'utf-8'
        }
    )

    print res.content

if __name__ == '__main__':
    push_message()
```

The api server can't parse data correctly, `if include \\uxxxx will create fail!` (The api server is a third party)

Because of requests's default json dumps, all Chinese character will be escaped `\\uxxx`, so I want to change default json.dumps() by it's `ensure_ascii` options.

tao12345666333:
Hi， @sigmavirus24  I have modified the code before, does not change the original behavior, but to provide a keyword argument. Using this because escape does not an action of `prepare_body()` function.
Therefore, don't use explicit arguments.

kennethreitz:
@tao12345666333 in this case, you should simply encode the json data yourself and pass it through the `data` parameter. Thanks for sharing, though! I see where you're coming from.

tao12345666333:
@kennethreitz Thanks! I will try again. :-)

PR #2786, opened 2015-09-23
adapters: Fix bug causing ingored read_timeout for chunked requests

swistakm:
Whenever chunked request was made (using generators) read timeout value was ignored because socket timeout was not set.

This can be a problem when system default timeout is very long because blocks process waiting for response for a period of time that cannot be controlled.

Changes:

* `HTTPAdapter.send()` now sets read timeout on `low_conn.sock` socket when request body is chunked.
* socket.error, socket.timeout are explicitely translated to urllib3 exceptions using `conn._raise_timeout`
  as expected with non-chunked request bodies to be later translated to `requests` exceptions with existing code.
* test added for timeouts on chunked-message body requests

Additional notes:

* due to convention to use httbin in requests tests and no posibility
 to tests POST timeouts with `delay/:n` endpoint (method not allowed)
 test for this bug uses GET with chunked message-body. This is OK
 because RFC-2616 does not explicitely say that message-body is
 forbidden in GET requests. And also it works

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -392,7 +392,16 @@ class HTTPAdapter(BaseAdapter):
                         low_conn.send(b'\r\n')
                     low_conn.send(b'0\r\n\r\n')
 
-                    r = low_conn.getresponse()
+                    # Control timeout manually because we do not use
+                    # 'urlopen' but internal connection object.
+                    if hasattr(low_conn, 'sock'):
+                        low_conn.sock.settimeout(timeout.read_timeout)
+
+                    try:
+                        r = low_conn.getresponse()
+                    except (socket.error, socket.timeout) as err:
+                        conn._raise_timeout(err, url, timeout.read_timeout)
+
                     resp = HTTPResponse.from_httplib(
                         r,
                         pool=conn,
--- a/test_requests.py
+++ b/test_requests.py
@@ -1508,6 +1508,19 @@ class TestTimeout:
         except ReadTimeout:
             pass
 
+    def test_read_timeout_chunked(self):
+        def chunked():
+            yield 'foo'
+            yield 'bar'
+
+        try:
+            # note: don't worry, according to HTTP/1.1 spec you actually can
+            #       issue a GET request with a message body
+            requests.get(httpbin('delay/10'), data=chunked(), timeout=(None, .1))
+            assert False, "The recv() request should time out."
+        except ReadTimeout:
+            pass
+
     def test_connect_timeout(self):
         try:
             requests.get(TARPIT, timeout=(0.1, None))
--- a/test_requests.py
+++ b/test_requests.py
@@ -1510,8 +1510,8 @@ class TestTimeout:
 
     def test_read_timeout_chunked(self):
         def chunked():
-            yield 'foo'
-            yield 'bar'
+            yield 'foo'.encode()
+            yield 'bar'.encode()
 
         try:
             # note: don't worry, according to HTTP/1.1 spec you actually can


sigmavirus24:
Paging @shazow to make sure we don't step on any toes here with urllib3 by accident.

shazow:
Looks painful but don't think urllib3 would break this.

requests/adapters.py
```diff
@@ -392,7 +392,16 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
...
-                    r = low_conn.getresponse()
+                    # Control timeout manually because we do not use
+                    # 'urlopen' but internal connection object.
+                    if hasattr(low_conn, 'sock'):
+                        low_conn.sock.settimeout(timeout.read_timeout)
+
+                    try:
+                        r = low_conn.getresponse()
+                    except (socket.error, socket.timeout) as err:
+                        conn._raise_timeout(err, url, timeout.read_timeout)
```
sigmavirus24, line 403: I'm not quite a fan of using a private method here. I'm otherwise, okay with the rest of this change.

swistakm, line 403: Yeah me too. I have considered few options and I think this is the lesser evil. Only other option (that is reasonable) is to duplicate behaviour of `_raise_timeout` i.e. handling different compatibility quirks (like SSL timeouts, EAGAIN in py2, see: [code](https://github.com/kennethreitz/requests/blob/master/requests/packages/urllib3/connectionpool.py#L300-L315)). I think using this particular "private" method introduces less of technology dept.

To be honest I think the only true and reasonable solution would be to fix the obvious deficiency of `urllib3` package, because once we did:

```
low_conn = conn._get_conn(timeout=DEFAULT_POOL_TIMEOUT)
```

there is no other way to do our work without calling `HTTPConnectionPool` internals. In fact the code section that starts from line 372 is like a big inlined excerpt from some non existing  `HTTPConnectionPool` method that should be somewhere there.

I think this should be eventually handled in `urllib3` instead of here, but with existing dependencies policy in **requests** it is impossible to achieve that in single PR/patch.

Lukasa, line 403: @shazow Would you be interested in taking a patch for sending chunked uploads in urllib3?

shazow, line 403: Absolutely!

Lukasa, line 403: Ok then, I recommend we do this "properly". Let's add support for chunked uploads to urllib3, and then requests can switch to using it.

swistakm:
This went how I expected. It was closed but never fixed. If it won't be fixed this way but is planned to be done in urllib3 can we just reopen it as an issue so it will be left as a reminder that such issue exists?

swistakm:
Also I found another reason why it may be worth fixing. Some servers can send a response before whole request is transmitted. I know HTTP was designed as a synchronous protocol but it looks like a lot of software implements it asynchronously.

One of notable examples is Jetty. In case of issues like unhandled read timeouts on its side it sends response before even consuming a request. I noticed that various services using this server may behave in unexpected ways if they send error response but client is still writing its response (this is another issue by the way). Currently I experience an issue where requests lock indefinitely on `sendto` sys call and never quit due to lack of timeout when talking to Jetty server using chunked transfer encoding.

I think that instead of saying that there is better way to do it we could merge this as a working fix to a critical issue (never-ending blocking calls) even if not elegant and only then think about rectifying this in urllib3.

Lukasa:
@swistakm We have merged chunked upload support into urllib3. We have not updated to the version of urllib3 that contains that fix, but when we do we'll also remove our own support for timeouts.

However, requests should _never_ lock on sending a socket: that suggests the connection is still up but Jetty is not consuming any data from the socket, leading to the TCP stream being forced open.

PR #2792, opened 2015-10-01
Update test dependencies

Lukasa:
Brings our test dependencies up to date.


Lukasa:
I think there's no good reason not to do this so I'm merging immediately.

PR #2800, opened 2015-10-02
Update Quickstart documentation of Response.json()

### LINKED ISSUE #2799: Quickstart note about json rendering is misleading
In the Quickstart document, the [JSON Response Content](http://docs.python-requests.org/en/v2.0-0/user/quickstart/#json-response-content) section says:

> In case the JSON decoding fails, `r.json` raises an exception. For example, if the response gets a 401 (Unauthorized), attempting `r.json` raises `ValueError: No JSON object could be decoded`

This is a bit misleading, especially that the [`json()`](https://github.com/kennethreitz/requests/blob/v2.7.0/requests/models.py#L798) method makes no attempt to validate that. It is accurate to say that failure to decode JSON will result in an error being raised. However, receiving 401 does not necessarily mean that JSON decoding will fail. Some APIs might return a JSON object in the response body (e.g. `{ 'error': 'Unauthorized' }`), in which case `r.json()` will work fine.

hosamaly:
Fixes #2799

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -145,8 +145,14 @@ There's also a builtin JSON decoder, in case you're dealing with JSON data::
     [{u'repository': {u'open_issues': 0, u'url': 'https://github.com/...
 
 In case the JSON decoding fails, ``r.json`` raises an exception. For example, if
-the response gets a 401 (Unauthorized), attempting ``r.json`` raises ``ValueError:
-No JSON object could be decoded``
+the response gets a 204 (No Content), or if the response contains invalid JSON,
+attempting ``r.json`` raises ``ValueError: No JSON object could be decoded``.
+
+It should be noted that the success of the call to ``r.json`` does **not**
+indicate the success of the response. Some servers may return a JSON object in a
+failed response (e.g. error details with HTTP 500). Such JSON will be decoded
+and returned. To check that a request is successful, use ``r.ok == True`` or
+``r.raise_for_status()``.
 
 
 Raw Response Content
--- a/requests/models.py
+++ b/requests/models.py
@@ -616,11 +616,8 @@ class Response(object):
 
     @property
     def ok(self):
-        try:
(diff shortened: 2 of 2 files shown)


Lukasa:
This is really good! One note: can we remove the reference to `response.ok`? It's at risk of deprecation, so I'd rather not add new references to it.

hosamaly:
Thanks. I beg to differ on `response.ok`. I find it useful and advise people to use it, although I'd prefer to change its implementation to avoid raising and catching exceptions. It could be as simple as `return self.status_code < 400`. I could update the pull request if you wish.

Lukasa:
The problem with `ok` is that it's not entirely clear what exactly it means: it's ambiguous and weird. Changing it would be better, yeah. =)

hosamaly:
I've updated the implementation of `ok` and added a docstring to it.

Lukasa:
Sorry, that was me being really unclear. I didn't want you to change `ok` itself, I wanted you to change the reference to it. =( Sorry.

hosamaly:
Feel free to edit the description to fit your needs. I still think using `ok` is clearer for new users (to whom the Quickstart guide is targeted), IMHO.

Lukasa:
To be clear, I'm not going to accept a change to the implementation of `ok`, especially one that changes the contract. If you'd rather not make further changes to this, that's fine, I can merge it and make the changes myself.

hosamaly:
My change to the implementation of `ok` certainly did not change the contract; it still behaves exactly the same. And certainly feel free to accept it or not; it isn't really important.

I'm just not comfortable changing the reference to `ok` in the Quickstart guide, so I'd rather you do it yourself if you really think it's better without it.

sigmavirus24:
@Lukasa would you like to revert these changes, or should I?

Lukasa:
Thanks so much @hosamaly!

hosamaly:
You're welcome @Lukasa . Thanks for accepting it.

Issue #2785, opened 2015-09-23
Information leakage in default user agent


asieira:
I've noticed that requests adds a user-agent header by default that looks like the following: `python-requests/2.7.0 CPython/2.7.9 Linux/3.14.44-32.39.amzn1.x86_64`.

Including the CPython and OS kernel version are an unnecessary information leakage that could have security implications, as per the [OWASP documentation](https://www.owasp.org/index.php/Information_Leakage).

I would like to suggest the default is changed to follow the 'bot' convention described [here](https://en.wikipedia.org/wiki/User_agent#Format_for_automated_agents_.28bots.29) with the requests version and a link to the requests documentation: `python-requests/<version> (http://http://www.python-requests.org/)`.

sigmavirus24:
Eh, I'm 50/50 on this. I agree the default user-agent isn't ideal. I don't agree that this is a large problem. I'd like to hear @Lukasa's opinion on this but he's away on vacation (as I should be).

asieira:
Agreed that this is not a critical issue, but it is a security problem nonetheless.

Thank you very much for the awesome project, enjoy your vacation and let me know if there's anything I can do to help. :)

sigmavirus24:
So, I've ruminated a bit on this now that I've had a chance and I'm not quite convinced it's a security problem. 
1. this is easily overridden by the user
2. your attackers without knowledge of your use of requests, can not be certain that the information there is entirely accurate or if it has been faked (just like other user-agents can just as easily be used)
3. attackers in this case are not very clear
   - An attacker could be the server your intentionally talking to (e.g., example.com) but at that point, there are potential other attacks that they could carry out even without (for example) knowledge of your Python version or kernel information
   - An attacker could be someone who is actively performing a MITM attack on a connection, in that case you have much larger problems than the attacker knowing your Python version or kernel information

In short, I think we need a much better threat model here before we change the existing behaviour. I'm fine documenting that for some use-cases the default user-agent is perhaps not ideal, and they can reference the existing documentation that shows how to set a UA.

asieira:
My responses:
1. the fact that there is a workaround (setting another UA value) doesn't make the default any less secure;
2. since this is the default, I have to imagine the vast majority of requests-based software will be leaking valid data through this mechanism, so the odds are in favor of the attackers;
3. let me try to help you understand the potential threats here:
[...]

PR #2803, opened 2015-10-05
Update default user-agent (#2785)

asieira:
* Updated default user agent so that it includes only the requests
version;
* Added PyCharm / IDEA project directory to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ requests.egg-info/
 *.swp
 *.egg
 env/
+.idea/
 
 .workon
 
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,6 @@ requests.egg-info/
 *.swp
 *.egg
 env/
-.idea/
 
 .workon
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -552,33 +552,7 @@ def select_proxy(url, proxies):
 
 def default_user_agent(name="python-requests"):
     """Return a string representing the default user agent."""
-    _implementation = platform.python_implementation()
-
-    if _implementation == 'CPython':
-        _implementation_version = platform.python_version()
-    elif _implementation == 'PyPy':
-        _implementation_version = '%s.%s.%s' % (sys.pypy_version_info.major,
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -15,6 +15,7 @@ requests.egg-info/
 *.swp
 *.egg
 env/
+.idea/
```
sigmavirus24: Please remove this. Pull requests should be singly purposed and not include unrelated changes.

Lukasa:
If we can clean this up relatively swiftly it'll make it into 2.8.0: otherwise, it'll fall out until 2.8.1.

.gitignore
asieira: It seemed like the right thing to do since there was no CONTRIBUTING document explaining what the rules were and this seemed like the most straightforward way to avoid getting the project files into the code both for me and future contributors who happen to use PyCharm / IDEA. 

You recommend I open a separate issue just for this one-line change to .gitignore, then? If that's the right thing to do, I'll do it.

sigmavirus24:
I wouldn't be comfortable including this in 2.8.1. 2.9.0 definitely, but not 2.8.1

Lukasa:
@asieira No, I think I recommend adding it to your local `.git/info/exclude` file. Text-editor specific config really shouldn't go into the project repository, the world just gets really complex really fast that way.

.gitignore
sigmavirus24: No, I'm recommending you remove this line and use a [Global gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

asieira:
Ok, will do that.

Lukasa:
LGTM. @sigmavirus24?

sigmavirus24:
@Lukasa will you handle the release notes for this or should I do it later tonight?

Lukasa:
@sigmavirus24 I'll do it.

@asieira Thanks so much for this change! :sparkles: :cake: :sparkles:

asieira:
:bow: my pleasure, thank you all again for the awesome package. :)

PR #2805, opened 2015-10-05
allows user paths for ca_cert verify path

a-tal:
currently you get the following error:

```SSLError("bad ca_certs: '~/example.crt'", Error([('system library', 'fopen', 'No such file or directory'), ('BIO routines', 'BIO_new_file', 'no such file'), ('x509 certificate routines', 'X509_load_cert_crl_file', 'system lib')],))```

### DIFF
--- a/requests/packages/urllib3/connection.py
+++ b/requests/packages/urllib3/connection.py
@@ -1,4 +1,5 @@
 import datetime
+import os
 import sys
 import socket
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this! However, this is a change in the [urllib3](https://github.com/shazow/urllib3) library, which we vendor in. Can you please open this pull request over there? Thanks!

PR #2808, opened 2015-10-07
Adding Entrust G2 CA Cert

hombremuchacho:
Adding Entrust G2 CA Cert to cacerts.pem


sigmavirus24:
I could be mistaken but I'm rather certain that we don't accept additions to the certificate bundle directly. They have to go through [certifi](/certifi/python-certifi) first. @Lukasa am I right?

Lukasa:
We do not accept changes to the cert bundle in requests in this manner: certainly not one certificate at a time. 

Would you like to open an issue that discusses the actual problem you have encountered?

hombremuchacho:
@Lukasa @sigmavirus24 Thanks for clarifying how this works. I'll open an issue. We use the requests library as part of our project and are missing a certain certificate -- this certificate is present in certifi; however, it hasn't been added to requests. https://github.com/certifi/python-certifi/blob/master/certifi/cacert.pem#L5026

aehlke:
To clarify, Certifi was not a drop-in solution for us as it was missing at least one cert that's bundled with Requests.

PR #2814, opened 2015-10-08
Use general null check for JSON

### LINKED ISSUE #2813: Requests json argument is ignored in 2.8.0
In version `2.7.0` I was using `Requests` constructor with `json` parameter specified, like:
```
json_data = [ ... ];
r = requests.Request('POST', url, json=json_data)
```

And in the `2.8.0` version this code does not work, because I do not specify `data` parameter. Inside constructor we get `self.data` assigned as `[] if data is None else data`, so we finally get `self.data` equal to `[]`.

And because of this commit: 7a0cd16c1a5be16f3df395ee12ed11d0ddde7837 we get `json` parameter ignored, because now we have to explicitly specify `data={}` when calling constructor, like:
```
json_data = [ ... ];
r = requests.Request('POST', url, json=json_data, data={})
```

I think that this issue definitely breaks documented API.

Lukasa:
Resolves #2813.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -414,7 +414,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         content_type = None
         length = None
 
-        if data == {} and json is not None:
+        if not data and json is not None:
             content_type = 'application/json'
             body = complexjson.dumps(json)
 


Lukasa:
I think we're going to start needing some tests that our set of data-y parameters interact correctly together when using different parts of the API.

sigmavirus24:
Or we need to stop duplicating logic everywhere that will break logic in different places.

Lukasa:
Or both?

sigmavirus24:
Yes. I commented in the wrong place. I greatly dislike the reliance on False-y values for tests. That should have been on #2813

PR #2825, opened 2015-10-14
Add warnings about text/binary mode files.

### LINKED ISSUE #2725: Requests.put blocking on Windows (Tika-Python)
Over in the [Apache Tika Python Port](http://github.com/chrismattmann/tika-python/) I'm noticing in [tika-python#44](http://github.com/chrismattmann/tika-python/issues/44) and in [tika-python#58](http://github.com/chrismattmann/tika-python/issues/58) some odd behavior with requests on Python 2.7.9. For whatever reason, when using a file handle and putting it with requests.put it blocks and blocks until finally it gets (correctly) a BadStatusLine back after a timeout. Anyone else seen this?

Lukasa:
Resolves #2725.

@sigmavirus24 I'm wondering if we should add a warning to `super_len` for whenever we decide on the length of the file using `os.fstat`, if the file has not been opened in binary mode. It should be easy enough to do. Thoughts?

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -315,6 +315,15 @@ file-like object for your body::
     with open('massive-body', 'rb') as f:
         requests.post('http://some.url/streamed', data=f)
 
+.. warning:: It is strongly recommended that you open files in `binary mode`_.
+             This is because Requests may attempt to provide the
+             ``Content-Length`` header for you, and if it does this value will
+             be set to the number of *bytes* in the file. Errors may occur if
+             you open the file in *text mode*.
+
+.. _binary mode: https://docs.python.org/2/tutorial/inputoutput.html#reading-and-writing-files
+
+
 .. _chunk-encoding:
 
 Chunk-Encoded Requests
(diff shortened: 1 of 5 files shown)


sigmavirus24:
@Lukasa do we want to add that warning in `super_len` and then make it a hard exception in 3.0.0?

Lukasa:
@sigmavirus24 Yeah, probably.

chrismattmann:
&hearts; thanks brosephs @sigmavirus24 @Lukasa

sigmavirus24:
@Lukasa do you want me to send that separately or do you want to do that as part of this PR?

Lukasa:
Ah, I forgot about that. Let's do it here.

Lukasa:
Ok, I added a warning that should fire once the first time this happens (no need to spam the crap out of people with it). It's strictly a subclass of `DeprecationWarning`, which I think is ok too.

sigmavirus24:
Excellent work @Lukasa . Thanks!

PR #2827, opened 2015-10-15
Fixed usage example in sessions.py

daniel-mueller:
The usage example for sessions.py's Session class is not correct in my opinion. The get method in fact returns a `Response <Response>` object, the notation "200" may be misleading the user into thinking only the http statuscode will be returned. Furthermore the output you get on console if typing the example in is different from the example and similar to my change.

The two usage examples in models.py are also similar to my change.

Thank you for considering.


Lukasa:
Thanks for this @daniel-mueller!

PR #2829, opened 2015-10-15
Use buffering for HTTP responses on chunked requests

dblia:
For non-chunked requests, the request is performed using the 'urlopen'
method that underlying uses buffering for the HTTP responses (for Python
2.7+ versions). For chunked requests though, the request is made via a
different code path and so the 'getresponse' method is called without
using buffering. So, the response is consumed per single byte via a
recv() call on the underlying socket.

This patch, fixes that issue to mimic the non-chucked request behavior.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -394,7 +394,15 @@ class HTTPAdapter(BaseAdapter):
                         low_conn.send(b'\r\n')
                     low_conn.send(b'0\r\n\r\n')
 
-                    r = low_conn.getresponse()
+                    # Receive the response from the server
+                    try:
+                        # For Python 2.7+ versions, use buffering of HTTP
+                        # responses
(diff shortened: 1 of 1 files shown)


Lukasa:
This seems like a sensible change to me, I'm :+1:. @sigmavirus24?

sigmavirus24:
What's the benefit?

Lukasa:
Without `buffering`, `httplib` will issue a lot of 1-byte reads in order to satisfy the constraints of the `read` call. That's...not good.

sigmavirus24:
Sorry, I was reading the commit message/PR message backwards. LGTM :+1:

Lukasa:
Thanks for this @dblia! :cake: :sparkles: :cake:

dblia:
Thanks for the immediate response @Lukasa and @sigmavirus24!

PR #2836, opened 2015-10-19
Use session.verify

causton81:
session.verify is currently ignored. Didn't have time for a unit test, but this is one solution.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -455,6 +455,9 @@ class Session(SessionRedirectMixin):
 
         proxies = proxies or {}
 
+        if verify is None:
+            verify = self.verify
+
         settings = self.merge_environment_settings(
             prep.url, proxies, stream, verify, cert
         )
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,5 @@
+mock
+contextlib2
 py==1.4.30
 pytest==2.8.1
 pytest-cov==2.1.0
--- a/test_requests.py
+++ b/test_requests.py
@@ -11,7 +11,9 @@ import unittest
 import collections
 import contextlib
 
+from contextlib2 import ExitStack
 import io
+import mock
 import requests
 import pytest
 from requests.adapters import HTTPAdapter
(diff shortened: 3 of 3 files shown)


Lukasa:
@causton81 What evidence do you have that `Session.verify` is ignored? This works fine for me:

``` python
>>> import requests
>>> s = requests.Session()
>>> s.verify = False
>>> s.get('https://kennethreitz.org/')
/Users/cory/.pyenv/versions/2.7.10/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py:789: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/Users/cory/.pyenv/versions/2.7.10/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py:789: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
<Response [200]>
```

causton81:
Sorry for the lack of details. The problem is when either `REQUESTS_CA_BUNDLE` or `CURL_CA_BUNDLE` is set in the environment and session.trust_env==True, then session.verify is ignored.  I added a unit test to demonstrate. Just for completeness I missed this for awhile because I was testing with Python 2.7.3 which ignores cert issues. I noticed it when testing with Python 2.7.9 and env variable REQUESTS_CA_BUNDLE.

Lukasa:
Ah, yes, I see the problem now.

We have a similar issue with proxy settings, where proxies set in the environment can override proxies set on the session. I'm beginning to wonder if we can fix two problems in one go here, by taking the `trust_env` block of `merge_environment_settings` and moving it below the `merge_setting` lines in that function.

@causton81 This is definitely a real bug, but we need a bit of time to work out what the fix is and what branch it'll go on to (while this is strictly a bug fix, it's been in the product long enough that it's also an API compatibility break and needs to be dealt with cautiously).

@sigmavirus24 What do you think about my proposed rearrangement of `merge_environment_settings`?

sigmavirus24:
@Lukasa the rearrangement is something we had discussed for 3.0.0 because it's fundamentally backwards incompatible. I'm in favor of it though because I think the order of precendence should be (in order of least to most important):
- Environment
- Session settings
- Per-call settings

(For those features to which this applies)

Lukasa:
Ok, let's get this done for 3.0.0.

Lukasa:
Ok, I'm closing this in favour of #2839.

PR #2839, opened 2015-10-21
Make sure we build environment settings properly.

Lukasa:
Resolves #2836.

This actually requires a slightly tricky re-ordering here: the proxies work needs to happen separately to make sure keys set to `None` DTRT.


sigmavirus24:
This looks like :cake: to me!

kennethreitz:
Rebase. Also appears as though it should have been merged six months ago.

Lukasa:
Rebase complete.

PR #2843, opened 2015-10-23
Timeout used by `resolve_redirects` generator can now depend on URL

tristan0x:
`timeout` parameter given to the `Session::resolve_redirects` member method
can be a `callable` object taking the URL in parameter.

It allows you to specify different timeout values for URLs encountered
by the generator.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -181,13 +181,19 @@ class SessionRedirectMixin(object):
             proxies = self.rebuild_proxies(prepared_request, proxies)
             self.rebuild_auth(prepared_request, resp)
 
+            # Ask callable object the proper timeout according to url.
+            if callable(timeout):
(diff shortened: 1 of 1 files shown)


Lukasa:
How is it expected that you'd use this? By setting `allow_redirects=False` and calling `resolve_redirects` manually?

tristan0x:
Yes exactly!

Lukasa:
Ok. =)

So, generally speaking, we don't allow the parameters to `resolve_redirects` to be different to those that you can pass to `requests.*`: it just leads to too much confusion. In this case, it should be safe enough to step through `resolve_redirects` one URL at a time and check the `Location` header as you go, using the information there to determine what timeout you'd like to apply.

Given that we are in feature freeze, I don't think this meets the bar for a merge into the library. Sorry! =(

Issue #2844, opened 2015-10-23
_encode_params return bytes that break urlencode


gagoman:
Simple example:

```
import requests


r = requests.get('http://python.org', params=b'test=foo')


print(r.status_code)
```

Trace:

```
Traceback (most recent call last):
... skip ...
    url = requote_uri(urlunparse([scheme, netloc, path, None, query, fragment]))
  File "C:\Python35\lib\urllib\parse.py", line 383, in urlunparse
    _coerce_args(*components))
  File "C:\Python35\lib\urllib\parse.py", line 111, in _coerce_args
    raise TypeError("Cannot mix str and non-str arguments")
TypeError: Cannot mix str and non-str arguments
```

I know, that this one is a bad use case, but API allows us to do so.

Should we do something?

P.S. Python 3.4/3.5 on win 8.1 x64

Lukasa:
I think what we need to do here is to change `encode_params` to make sure that it calls `to_native_string` on its argument before returning it. This seems like a bug we can fix pretty swiftly, and is well suited to a new contributor, so I'll mark this as `Contributor Friendly`.

untitaker:
This fix seems completely broken. It prevents me from uploading binary data through PUT.

sigmavirus24:
@untitaker please open a new issue.

untitaker:
See #2030

touilleMan:
Could you release a quick 2.9.1 with this fix plz ?

I was waiting the 2.9.0 for the REQUEST_CA_BUNDLE env var but cannot use it for the moment due to this bug ;-(

Lukasa:
@touilleMan We're planning to release 2.9.1 on Monday.

touilleMan:
Awesome, you guys rock ! ^^

yami:
Looking forward to the new release. This bug breaks everything, since we use Chinese...

PR #2845, opened 2015-10-23
Fix issue #2844

gagoman:
Closes https://github.com/kennethreitz/requests/issues/2844

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -81,7 +81,7 @@ class RequestEncodingMixin(object):
         """
 
         if isinstance(data, (str, bytes)):
-            return data
+            return to_native_string(data)
         elif hasattr(data, 'read'):
             return data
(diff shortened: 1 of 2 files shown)


Lukasa:
\o/ Looks good to me! :cake: :sparkles:

requests/models.py
```diff
@@ -81,7 +81,7 @@ def _encode_params(data):
         """
 
         if isinstance(data, (str, bytes)):
-            return data
+            return to_native_string(data)
         elif hasattr(data, 'read'):
             return data
```
akhomchenko, line 86: I'm not sure if we need to wrap this.

sigmavirus24, line 86: You don't. It's a file-like object. Calling `to_native_string` on it is absolutely the wrong thing. That's why @Lukasa merged this. :)

PR #2859, opened 2015-11-05
Move test suite offline using pytest-httpbin

### LINKED ISSUE #2184: Port test suite to pytest-httpbin
https://github.com/kevin1024/pytest-httpbin

Lukasa:
Resolves #2184, and should allow us to move back to Travis for our testing.

### DIFF
--- a/test_requests.py
+++ b/test_requests.py
@@ -1136,7 +1136,7 @@ class TestRequests(object):
         with contextlib.closing(s.get(httpbin('stream/4'), stream=True)) as response:
             pass
 
-        assert not response._content_consumed is False
+        assert response._content_consumed is False
         assert response.raw.closed
 
     @pytest.mark.xfail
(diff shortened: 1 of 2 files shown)


Lukasa:
There _ought_ to be a lot of improvements to doing this. Tests should be a lot closer to reproducible, and the runtime should improve greatly. This is still not really how I want this test suite to look, but it's a _lot_ closer.

test_requests.py
```diff
@@ -431,22 +448,7 @@ def test_DIGESTAUTH_QUOTES_QOP_VALUE(self):
         r = requests.get(url, auth=auth)
         assert '"auth"' in r.request.headers['Authorization']
 
-    def test_DIGESTAUTH_THREADED(self):
```
sigmavirus24, line 434: Could we add a `# TODO(...): Add threaded HTTPDigestAuth test`?

sigmavirus24:
One request, then I'll press the big green button. :)

Alternatively, we can add a bug to track the work of adding that test.

Lukasa:
I think it's better to add a bug than a note in the code: we'll just lose the note. ;)

sigmavirus24:
Thanks @Lukasa!

Issue #2215, opened 2014-09-10
Using st_size on FIFO files or stdin will cause super_len to return 0 as the length


cecton:
Hello,

### Use Case
I run a process using ```subprocess``` and give the stdout as data of a POST request. Since it is a file-descriptor, it has a fileno, and ```super_len()``` "stat" the file descriptor and use the st_size to determine the length.

The problem is: if the st_size returned is 0, it is possibly due to other reasons, in this case: the length is not determined (despite st_size is 0). But ```requests``` send a header specifying a ```Content-Length``` of 0 whilst it is actually not determined.

### Work Around
Wrap the stream in a generator, so the method ```fileno()``` cannot be accessed anymore.

```python
request.post(url, data=(x for x in p.stdout))
```

### Another Use Case
```
[14:05:31] ~ > mkfifo /tmp/a
[14:05:35] ~ > stat /tmp/a
  File: ‘/tmp/a’
  Size: 0               Blocks: 0          IO Block: 4096   fifo
Device: 1eh/30d Inode: 514910      Links: 1
Access: (0644/prw-r--r--)  Uid: ( 1001/ openerp)   Gid: (  100/   users)
Access: 2014-09-10 14:05:35.415835558 +0200
Modify: 2014-09-10 14:05:35.415835558 +0200
Change: 2014-09-10 14:05:35.415835558 +0200
 Birth: -
```
The size of a named pipe is 0.

### Fix
According to ```man 2 stat```:

> For most files under the /proc directory, stat() does not return the file size in the st_size field; instead the field is returned with the value 0.

I guess ```super_len()``` should return st_size if st_size if bigger than 0, otherwise ```None```.

Lukasa:
Thanks for reporting this issue!

This is an interesting case. Here are my thoughts. First, on the positive side, the effect of this change is that we'd switch to chunked encoding for the upload. This is good, that's clearly the correct behaviour when passed a named pipe.

On the downside, how should we react to a user passing us a file of zero length to upload? Sending content-length 0 has always been fairly clear. Sending a chunked upload of zero length is a bit weird, but not the end of the world, and it works fine.

In practice I'm +0.5 on the proposed change. I think we gain more than we lose.

cecton:
Maybe we should check if the user has specified `Transfer-Encoding` in the header, because, according to the RFC, it conflicts with Content-Length and therefore there is no use to check the file size.

My code will look like this instead:

``` python
request.post(url, data=p.stdout,
             headers={
[...]

PR #2861, opened 2015-11-05
Fix breakage introduced by 8f591682

Lukasa:
Turns out we broke chunked uploads in 8f591682e6a901baf0cc8a0393b5930252f0318a. I discovered this while investigating #2215.

While we were here, I added a test that would at least have caught errors as egregious as this one.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -398,10 +398,10 @@ class HTTPAdapter(BaseAdapter):
                     try:
                         # For Python 2.7+ versions, use buffering of HTTP
                         # responses
-                        r = conn.getresponse(buffering=True)
+                        r = low_conn.getresponse(buffering=True)
                     except TypeError:
                         # For compatibility with Python 2.6 versions and back
-                        r = conn.getresponse()
+                        r = low_conn.getresponse()
 
                     resp = HTTPResponse.from_httplib(
                         r,
--- a/test_requests.py
+++ b/test_requests.py
@@ -1133,6 +1133,13 @@ class TestRequests(object):
         next(r.iter_lines())
         assert len(list(r.iter_lines())) == 3
 
+    def test_chunked_upload(self, httpbin):
+        """Can safely send generators."""
+        data = (c for c in 'abc')
+        r = requests.post(httpbin('post'), data=data, stream=True)
+        assert r.status_code == 200
+        assert r.request.headers['Transfer-Encoding'] == 'chunked'
+
 
 class TestContentEncodingDetection(unittest.TestCase):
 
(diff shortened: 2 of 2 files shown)


Lukasa:
To note: this breakage did not make it into any shipped version of requests, so we don't need to schedule a patch release for this.

sigmavirus24:
Why is the build failing?

Lukasa:
Good question. The test behaves fine on my
Mac, but it's a tricky one because chunked encoding and WSGI don't necessarily play nice.

Lukasa:
Hmm, so this failure is intermittent on my machine.

Lukasa:
(my machine == Linux box I found)

Lukasa:
I wonder if this is actually a problem with the WSGI server being used by pytest-httpbin. Specifically, I think it might be related to the speed of the server.

Lukasa:
Specifically, I think what's happening here is that the chunked-encoding header is being stripped, and so the server is sending the 200 right away and then killing the connection. This works _sometimes_ because the socket sends happen before the connection is closed, but sometimes the server closes the connection before we can get there.

This is pretty annoying, because it means it's likely very hard to find a good test-case here.

Lukasa:
More importantly, it means we can't really have tests _at all_ for chunked transfer encoding.

Lukasa:
I honestly don't know what to do about this. To get a test that confirms we don't break this we need either to change the WSGI server used by pytest-httpbin (the only options are ones that are _really_ heavyweight and probably not suitable) or start writing new tests that don't use a httpbin at all.

The second is plenty do-able, it just requires some py.test test fixtures and some socket work. But it makes me a bit nervous given @kennethreitz's attitude to changing our test layout in the past: I'd want to move all the tests to a `test/` subdirectory so that I can add the necessary `conf.py` file, and also start writing new tests that explicitly use sockets rather than a web server.

@kennethreitz, do you have any objection to me adding some more thorough testing around here? The reality is that our chunked upload logic is _entirely_ untested, and has been since it was originally written. That is a recipe for accidental breakage, and it makes me really nervous.

Lukasa:
Ok, so I pushed a new commit that doesn't contain the test. This should pass, and the fix _is_ right, but I don't want us to lose track of the need for better testing here so I'm going to open a separate issue for that.

Lukasa:
@sigmavirus24 This should be good to merge.

sigmavirus24:
@Lukasa I'm :+1: but I think we should work on fixing our test suite. If Kenneth comes along and decimates it again, then we at least had test coverage for some period of time there.

PR #2869, opened 2015-11-10
unnest deep loop

vienno:
Flat is better than nested. Instead of nesting `if` statements, check for the opposite early in the loop and `continue` if the opposite is the case.

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -145,9 +145,11 @@ def remove_cookie_by_name(cookiejar, name, domain=None, path=None):
     for cookie in cookiejar:
         if cookie.name != name:
             continue
-        if domain is None or domain == cookie.domain:
-            if path is None or path == cookie.path:
-                clearables.append((cookie.domain, cookie.path, cookie.name))
+        if domain and domain != cookie.domain:
+            continue
+        if path and path != cookie.path:
+            continue
+        clearables.append((cookie.domain, cookie.path, cookie.name))
 
     for domain, path, name in clearables:
         cookiejar.clear(domain, path, name)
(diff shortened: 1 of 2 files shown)


Lukasa:
Good idea @vienno!

If we're going to un-nest this, let's un-nest it the whole way. Want to break out the next nested couplet as well and use the same logic?

miroli:
How about now @Lukasa ? I rewrote the third `if` statement as well, for consistency.

requests/cookies.py
```diff
@@ -143,10 +143,13 @@ def remove_cookie_by_name(cookiejar, name, domain=None, path=None):
     """
     clearables = []
     for cookie in cookiejar:
-        if cookie.name == name:
-            if domain is None or domain == cookie.domain:
-                if path is None or path == cookie.path:
-                    clearables.append((cookie.domain, cookie.path, cookie.name))
+        if cookie.name != name:
+            continue
+        if domain and domain != cookie.domain:
```
Lukasa: You changed this slightly. It needs to be `if domain is not None`.

requests/cookies.py
```diff
@@ -143,10 +143,13 @@ def remove_cookie_by_name(cookiejar, name, domain=None, path=None):
...
+            continue
+        if path and path != cookie.path:
```
Lukasa: Similarly, this needs to be `if path is not None`.

Lukasa:
Thanks, much better! I have one small note, inline on the diff. =)

miroli:
Of course! It's been updated.

Lukasa:
Looks good to me @vienno! Would you like to add yourself to AUTHORS.rst?

miroli:
Sure!

Lukasa:
Thanks so much @vienno! :sparkles: :cake: :sparkles:

PR #2880, opened 2015-11-12
cache Response.text

j0hnsmith:
`Response.text` is a pure function, seems pointless doing the same conversions in repeated calls. 

Cached
```
In [4]: %timeit any(word in resp.text for word in words)
The slowest run took 6.20 times longer than the fastest. This could mean that an intermediate result is being cached 
100000 loops, best of 3: 17.4 µs per loop
```

Not cached
```
In [5]: %timeit any(word in resp.text_old for word in words)                                                                            
10000 loops, best of 3: 64.4 µs per loop
```

`len(words) == 3` in these examples, in a real world use case it could be much longer.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -753,31 +753,33 @@ class Response(object):
         non-HTTP knowledge to make a better guess at the encoding, you should
         set ``r.encoding`` appropriately before accessing this property.
         """
+        if not hasattr(self, '_text'):
+            if not self.content:
+                self._text = str('')
 
-        # Try charset from content-type
-        content = None
-        encoding = self.encoding
+            else:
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this! However, `Response.text` is _not_ a pure function, and cannot be safely cached. This is actually explicitly called out [in the docs](http://docs.python-requests.org/en/latest/user/quickstart/#response-content):

> If you change the encoding, Requests will use the new value of `r.encoding` whenever you call `r.text`.

This is intentional and desirable. For this reason, we want to prevent text from being cached. 

Sorry we can't accept your contribution! =(

PR #2882, opened 2015-11-14
Please add a Session.close()

fzzo:
Is there a reason why the .close() has yet to be implemented, the equivalent of the .close() from the original requests library? To ensure the session is closed, without leaving any open/dead/hanging/zombie sockets, would a Session.close() be the way to avoid having socket nightmares on a multi-threaded crawler utilizing requesocks.Session()?

Trying to do:

```
sesh = requesocks.Session()

# do stuff

sesh.close()
```

From requests:

```
     def close(self):
        """Closes all adapters and as such the session"""
        for v in self.adapters.values():
            v.close()
```

Thanks!


Lukasa:
Uh...this is the repository for the original requests library. We _do_ have a close method.

PR #2891, opened 2015-11-19
Specify minimum pyOpenSSL version

lorenzph:
urllib3 requires "set_tlsext_host_name" which was only added in
pyOpenSSL 0.13. As some distributions (e.g. Ubuntu 12.04) still ship an
older version enforce the correct minimum version during installation.

I will submit a similar pull requests to the urllib3 project.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,6 +69,6 @@ setup(
         'Programming Language :: Python :: 3.5',
     ),
     extras_require={
-        'security': ['pyOpenSSL', 'ndg-httpsclient', 'pyasn1'],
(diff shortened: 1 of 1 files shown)


Lukasa:
It's mind-boggling to me that anyone is shipping a PyOpenSSL that old. Amusingly, the reason we need set_tlsext_host_name is to enable SNI, which is literally the point of using PyOpenSSL with requests, so any earlier version of PyOpenSSL wouldn't do what we need it to do anyway. What a world we live in. Thanks for spotting this!

sigmavirus24:
_reserves comments on Long Term Support versions of distributions harming user security in the interested of perceived stability_

Issue #2866, opened 2015-11-07
Better testing for chunked uploads.


Lukasa:
We have a whole code branch that does chunked uploads that is not and has never been tested. That's a problem, because sometimes it breaks and we don't notice (#2861). I'd like to add more testing for chunked uploads.

However, we can't use httpbin to do this testing. This is because the WSGI spec staunchly refuses to let applications see chunked transfer encoding at the app layer. For this reason, to test chunked transfer encoding requires a new style of testing, one that doesn't involve running a WSGI server but instead running something that can see the bytes *as they hit the wire*. Basically, something like the socket level tests of urllib3.

BraulioVM:
One possibility would be adapting urllib3's solution to our case. This new server we'd make doesn't need to support the whole HTTP protocol, because we just need it for the chunked transfer encoding thing, right?

Lukasa:
I think we wouldn't actually use a proper server at all, just a pair of sockets.

kgstan:
Г

вторник, 24 ноября 2015 г. пользователь Cory Benfield написал:

> I think we wouldn't actually use a proper server at all, just a pair of
> sockets.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/kennethreitz/requests/issues/2866#issuecomment-159189923>
> . Тштххшгшшззжзхмц
>


-- 
С уважением,
Рыскулов Чингиз

WWW.WEEKENDS.KG <http://www.weekends.kg/>
email: chingizkg@gmail.com <weekends.kg@gmail.com>
тел: +996 777 808888
Skype: k <http://weekends.kg/>gstan

BraulioVM:
Right, ok. 

Both #2860 and this issue come from the same problem: pytest-httpbin's server doesn't seem to support the whole spectrum of our testing needs properly (not saying it is bad software, it's great). We could solve both issues at the same time just by making our socket server able to listen to listen to _parallel_ requests. The problem is that it would be necessary to hardcode the response headers in the threaded's auth case, which is far from optimal.

Lukasa:
Even that is overdesigning it. All we need is to be able to receive a request and send a response. Parallel requests aren't a concern, just one is totally fine. We need a listening socket that can send a pre-canned 200 response when a request is received then read all the data. That's literally all we need to do.

BraulioVM:
[...]

PR #2896, opened 2015-11-23
[WIP] Set 'Transfer-Encoding: chunked' if data is a file with length 0

### LINKED ISSUE #2215: Using st_size on FIFO files or stdin will cause super_len to return 0 as the length

BraulioVM:
Fixes #2215 

This is the first time I contribute to the project so I am not sure at all I put that code where it belongs. I know I have to add tests for this, but wanted to show you this beforehand.

Did I do this right?

Thank you

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -434,7 +434,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             if files:
                 raise NotImplementedError('Streamed bodies and files are mutually exclusive.')
 
-            if length is not None:
+            if length:
                 self.headers['Content-Length'] = builtin_str(length)
             else:
                 self.headers['Transfer-Encoding'] = 'chunked'
@@ -446,8 +446,6 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
                 if data:
                     body = self._encode_params(data)
                     if isinstance(data, basestring) or hasattr(data, 'read'):
-                        if hasattr(data, 'fileno') and length == 0:
-                            self.headers['Transfer-Encoding'] = 'chunked'
                         content_type = None
                     else:
                         content_type = 'application/x-www-form-urlencoded'
--- a/requests/models.py
+++ b/requests/models.py
@@ -446,6 +446,8 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
                 if data:
                     body = self._encode_params(data)
                     if isinstance(data, basestring) or hasattr(data, 'read'):
+                        if hasattr(data, 'fileno') and length == 0:
+                            self.headers['Transfer-Encoding'] = 'chunked'
                         content_type = None
                     else:
                         content_type = 'application/x-www-form-urlencoded'


Lukasa:
@BraulioVM This is a good first pass! I have a draft of this fix sitting around on my laptop, so when I get home I'll show you how I approached it and let you take the ideas you like best.

BraulioVM:
@Lukasa Thanks!

I have just seen #2866, which won't allow me to implement the tests I want, so I might as well think of a solution for that issue.

Lukasa:
So I think this change is not quite right. What we want is a simple change around line 437. Right now we test `length` for whether it is `None`: we should just broaden that to say `if length`. That way, we'll fall through to chunked encoding for zero-length files.

BraulioVM:
I see. What I did didn't make any sense because file objects are _streams_. However, this change won't just affect files, it will affect any stream with length 0. Things will keep working, but I thought we just wanted to set 'Transfer-Encoding: chunked' if `data` was a file object with `super_len(data) == 0`.

Lukasa:
@BraulioVM That's actually intentional: sending a chunked zero-length stream is fine, there won't be a problem there. That way, the code is clearer. =)

So, this change looks good to me, but I'll let @sigmavirus24 review it.

BraulioVM:
Is there anything else I should do regarding this PR?

Lukasa:
@BraulioVM Nope, just waiting for @sigmavirus24 to get enough time to swing by and take a look at it. =)

sigmavirus24:
Woops! Sorry. I lost track of this with recent goings on IRL. This looks great to me! Thanks @BraulioVM

PR #2897, opened 2015-11-24
Test socket server

### LINKED ISSUE #2866: Better testing for chunked uploads.

BraulioVM:
Fixes #2866.

### What is this?
This is an implementation of a basic socket server that is able to run concurrently with the main thread. This might be useful for testing our 'Transfer-Encoding: chunked' functionality as pytest-httpbin isn't able to handle this kind of encoding.

### How does it work?
You provide the Server constructor with a handler that will be called once the socket server is listening. This handler will be able to accept clients and send and receive any info it wants to. The main thread will be able to query the server by creating a normal socket. 

### Syntax proposed
I think that using a Context-Manager is a great way of making it easy to use the Server. 
````python
with Server(handler) as host, port:
    sock = socket.socket()
    sock.connect((host, port))
   ...
````
### Doubts
* urllib3 tries to use a IPv6 socket and falls back to IPv4 when it can't. Should I try doing the same thing?
* I made a new directory for dummyserver, but don't know whether that was the right thing to do.
* I am not sure whether I'm testing my code properly. Code is being tested in very different ways in ````test_requests.py````

### Roadmap
* Handle exceptions in Server.
* Add features such as ````basic_response_handler```` in urllib3's implementation.
* Implement more tests
* Implement 'Transfer-Encoding: chunked' tests using the dummy server.

Is this a good idea? Or should I stop doing this right now?

### CODE REVIEW

File: dummyserver/server.py [hunk 1]
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/python
+
+import threading, socket
+
+
+
+"""
+Dummy server using for unit testing
```
sigmavirus24: If this is the docstring for the Server class, why isn't it in-line?

sigmavirus24:
Can we perhaps not call it dummyserver? Maybe something more accurate, e.g., testserver?

Lukasa:
Calling it a testserver feels like a good idea. As to the approach, this is definitely going in the right general direction, at least as far as I'm concerned. Obviously you need to replace the test that's currently there with one that actually tests requests itself, but that's fine.

We also need to be on the lookout for flaky tests here. Threaded tests can get tricky.

BraulioVM:
Ok, I'll call it testserver.

The test I added is there just for testing the server. It might be weird to add tests for a testing utility but I think it is necessary as it could be broken in many no trivial ways. If you don't like having requests tests and testserver tests in the same file (which actually looks bad) we could do the following things:
- Having a separate file for testing testserver
- Extracting testserver as a separate package and developing it with its tests independently from requests
- Not testing it at all

BraulioVM:
I think I broke the CI server with tests that hang forever. I don't know whether I should do anything to fix that

Lukasa:
Yup, that seems right. They don't consistently hang, but they do sometimes hang. Are you able to run the tests locally @BraulioVM?

Lukasa:
PS @sigmavirus24 I don't seem to be able to stop the builds: can you do it?

sigmavirus24:
I cannot stop the tests.

Lukasa:
Well...that's not ideal.

BraulioVM:
@Lukasa I ran the tests locally on python2.7 and know why they hang. I have run the tests for the last version both on 2.7 and 3.4, and they are fine. Is there any easy way to replicate the build environment of the CI server? 

I am sorry for breaking it, I never thought there wasn't going to be a timeout or something like that.

BraulioVM:
Ok, I already have a failing test for chunked uploads (in python3, the 'Transfer-Encoding' header is not being set and python2.7 doesn't even let me use a generator as `data`)

BraulioVM:
Do you like the overall direction of this PR? Do you think I have gone too far? As I am new with this project I'm not sure whether this is the way to go

Lukasa:
I think in general it's ok, but I need to spend some time playing with it tomorrow.

Lukasa:
@BraulioVM So, that test doesn't fail for me, it passes, but more importantly it does run!

Having played with this quickly on my local machine, this seems generally good: the test runs, the structure of the function mostly seems to be good, so I'm happy to have this as the basis of a set of more specific tests. I would like us to add some more helpers, but we don't need to do that right now: we can definitely tackle it later.

Well done @BraulioVM! Did you have more work you wanted to do before you proposed this for merging?

BraulioVM:
Thank you @Lukasa! I am happy this is finally going to be useful. 

I don't mind adding those helpers you talk about, but as I am not totally sure about how this server will be used I don't know what helpers should be implemented. I guess some logic for parsing the chunked content must be implemented in order to fully test the library, however, I don't think that logic should be built into TestServer.

There's no more work I was planning on doing. I was hoping you would tell some other things you wanted this PR to have (such as those helpers).

Lukasa:
So, some helpers that might be useful:
- something that lets you easily return a complete response from a HTTP/1.1 string.
- something that returns 200 OK but saves the request off to be examined

Ideally we'd want these with tests that use them, to confirm they work.

BraulioVM:
Ok! Sure, everything I implement will be tested. 

Just for clarification, when you say "return a complete response from a HTTP/1.1 string" you mean the _string_ contains the http headers and the body of the response and not just its body, right? Or is it just the body?

Lukasa:
That's correct, headers and body. =)

BraulioVM:
Ok, done!

I added an attribute to the TestServer class called `handler_results`. It is a list which stores the return values of the request handler. This allows the server instance returned by  `text_response_server` to store the requests by returning their content.

I have added tests for those two features as well.

BraulioVM:
Anything else I should do in order to get this merged?

Lukasa:
Not right now @BraulioVM, we're pretty quiet over the holiday period so we probably won't find time to focus on reviewing this until the new year.

BraulioVM:
Ok, that's perfect!

Happy holydays!

BraulioVM:
I have just finished my exams and have time to work on this again if it is heading the right direction. I have read https://github.com/kennethreitz/requests/issues/2866 and believe you want something similar to this, but you may prefer a different implementation anyway.

Should I keep working on this or look for another issue to work on?

Lukasa:
I believe this is generally going in the right direction, but now that @kennethreitz is more active he may want to see if this is something he wants. @sigmavirus24 and I are definitely :+1: on this for improving our testing, but it's up to Kenneth.

BraulioVM:
Ok, I'll wait to see what Kenneth thinks of this

kennethreitz:
I have no opinion :)

Lukasa:
Ok, I'll give this a review tomorrow and see where we are.

test_requests.py
```diff
@@ -1177,6 +1181,24 @@ def test_precedence(self):
         encodings = requests.utils.get_encodings_from_content(content)
         assert encodings == ['HTML5', 'HTML4', 'XML']
 
+    def test_chunked_upload(self):
+        """ can safely send generators """
```
Lukasa: Can you remove the excess whitespace here?

testserver/server.py
```diff
@@ -0,0 +1,106 @@
...
+            handler_result = self.handler(sock)
+
+            self.handler_results.append(handler_result)
+        
+        if self.wait_to_close_event:
+            self.wait_to_close_event.wait()
+        
+    def __enter__(self):
+       self.start()
+       self.ready_event.wait()
```
Lukasa: Having waits without timeouts is really problematic, because it runs the risk of grinding the entire test suite to a halt. Better to have all the waits have a finite timeout (5 seconds is nice).

testserver/server.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/python
+
+import threading 
+import socket
+import select
+
+
+def consume_socket_content(sock, chunks=65536, timeout=0.5):
+    content = ""
+    more_to_read = select.select([sock], [], [], timeout)[0]
```
Lukasa: Do we gain anything from having `select` called here?

testserver/server.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/python
+
+import threading 
+import socket
+import select
+
+
+def consume_socket_content(sock, chunks=65536, timeout=0.5):
```
Lukasa: There's no need to make `chunks` an argument here, we don't use it anywhere. Just set it in the method body.

testserver/server.py
```diff
@@ -0,0 +1,106 @@
...
+    content = ""
+    more_to_read = select.select([sock], [], [], timeout)[0]
+
+    while more_to_read:
+        new_content = sock.recv(chunks).decode("utf-8")
+
+        if len(new_content) == 0:
```
Lukasa: Idiomatically, this should be `if not new_content:`

testserver/server.py
```diff
@@ -0,0 +1,106 @@
...
+            more_to_read = False # empty recv means the socket disconnected
+
+        else:
+            content += new_content 
+            # stop reading if no new data is received for a while 
+            more_to_read = select.select([sock], [], [], timeout)[0]
```
Lukasa: I'm nervous about this `select` based approach to draining sockets. It runs the risk of being a bit brittle: slow test machines may run into problems, and fast test machines may be delayed up to half a second in a test.

I presume that the reason we're doing this is to avoid problems with needing the test suite to understand HTTP, which I can understand, but we need to be wary of this in the future and consider coming up with a different approach.

Lukasa: While we're here, we may want to refactor this whole block. Rather than the `if/else`, let's try:

``` python
while more_to_read:
    new_content = sock.recv(chunks)
    if not new_content:
         break
    content += new_content
    more_to_read = select.select([sock], [], [], timeout)[0]
```

testserver/server.py
```diff
@@ -0,0 +1,106 @@
...
+import socket
+import select
+
+
+def consume_socket_content(sock, chunks=65536, timeout=0.5):
+    content = ""
+    more_to_read = select.select([sock], [], [], timeout)[0]
+
+    while more_to_read:
+        new_content = sock.recv(chunks).decode("utf-8")
```
Lukasa: Why are we decoding the content here?

testserver/server.py
```diff
@@ -0,0 +1,106 @@
...
+ 
+        self.wait_to_close_event = wait_to_close_event
+        self.ready_event = threading.Event()
+        self.stop_event = threading.Event()
+
+    @classmethod
+    def text_response_server(cls, text, request_timeout=0.5, **kwargs):
+        def text_response_handler(sock):
+            request_content = consume_socket_content(sock, timeout=request_timeout)
+            sock.send(text.encode())
```
Lukasa: A blank encode statement is almost never going to be what you want. You should pick an encoding. =)

testserver/server.py
```diff
@@ -0,0 +1,106 @@
...
+        return server
+
+    @classmethod
+    def basic_response_server(cls, **kwargs):
+        server = cls.text_response_server(
+            "HTTP/1.1 200 OK\r\n" +
+            "Content-Length: 0\r\n\r\n", **kwargs
+        )
+
+        return server
```
Lukasa: This is a minor style nit, but you don't need the blank line above `return` unless the thing you're returning is not the thing you were just working on. As all the methods in this class are pretty simple and really only affect one object, you can usually remove the empty line above `return`.

testserver/server.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def _handle_requests_and_close_server(self, server_sock):
+        for _ in range(self.requests_to_handle):
+            sock = server_sock.accept()[0]
+            handler_result = self.handler(sock)
+
+            self.handler_results.append(handler_result)
+        
+        if self.wait_to_close_event:
+            self.wait_to_close_event.wait()
```
Lukasa: It's a little bit confusing, when reading the `run` method, that there's a `wait_to_close_event` that isn't obviously waited on before calling `socket.close()`. It might be better to hoist this up into `run`.

That will also let us deal with the slightly confusing name of this function, which claims that it closes the server when it does not. =)

test_requests.py
```diff
@@ -1177,6 +1181,24 @@ def test_precedence(self):
...
+        block_server = threading.Event()
+        server = Server.basic_response_server(wait_to_close_event=block_server)
+        data = (i for i in [b'a', b'b', b'c']) 
+
+        with server as (host, port):
+            url = 'http://{}:{}/'.format(host, port)
+            r = requests.post(url, data=data, stream=True)
+            block_server.set() # release server block
```
Lukasa: Do we need this event? I think the server won't get closed until it has handled the request, so presumably we don't need to gate it.

test_requests.py
```diff
@@ -1177,6 +1181,24 @@ def test_precedence(self):
...
+
+            assert r.status_code == 200
+            assert r.request.headers['Transfer-Encoding'] == 'chunked'
```
Lukasa: We should bring this out of the context manager.

Lukasa:
Ok @BraulioVM I've left some notes in the diff. =)

Unfortunately, while you've been away we moved the tests to a `test` subdirectory, so you'll need to move this change onto that directory structure. Sorry!

test_requests.py
BraulioVM: We could have problems if the data upload takes more than 0.5 seconds (`consume_socket_content` has a 0.5 timeout). I don't think that happens very often, but wanted the tests to be correct without assuming that the generator data is going to be sent without breaking any timeouts.

BraulioVM:
Ok, I have had several doubts while merging with master:
- I have put `TestTestServer` test cases inside `test_util.py` because that's the file I've found  best for it. Should I change it? I had to import requests in order to test the server, but I believe you did not import it on purpose.
- Is there any better place for `test_chunked_upload`?

tests/test_requests.py
```diff
@@ -1452,3 +1451,8 @@ def test_vendor_aliases():
 
     with pytest.raises(ImportError):
         from requests.packages import webbrowser
+
+
+
+if __name__ == '__main__':
+    unittest.main()
```
Lukasa: Don't add this to the bottom of the test file: these tests don't run properly in unittest, they need to be executed in py.test.

tests/test_utils.py
```diff
@@ -229,3 +234,124 @@ def test_select_proxies(url, expected):
...
+        data = "bananadine"
+
+        with server as address:
+            sock = socket.socket() 
+            sock.connect(address)
+            time.sleep(1.5)
+            sock.send(data.encode())
+            sock.close()
+            
+        assert server.handler_results[0] == data
```
Lukasa: I think it'd be best to move these to a test file of their own.

tests/testserver/server.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/python
```
Lukasa: This isn't an executable script, so you don't need the shebang here.

tests/testserver/server.py
```diff
@@ -0,0 +1,107 @@
...
+        content += new_content 
+        # stop reading if no new data is received for a while 
+        more_to_read = select.select([sock], [], [], timeout)[0] 
+
+    return content
+
+
+
+class Server(threading.Thread):
+    """ Dummy server using for unit testing """
```
Lukasa: We still don't need the extra whitespace here. =)

tests/testserver/server.py
```diff
@@ -0,0 +1,107 @@
...
+    WAIT_EVENT_TIMEOUT = 5
+
+    def __init__(self, handler, host='localhost', port=0, requests_to_handle=1, wait_to_close_event=None):
+        threading.Thread.__init__(self)
```
Lukasa: It's generally preferred to use `super(Server, self).__init__()` here.

tests/testserver/server.py
```diff
@@ -0,0 +1,107 @@
...
+    @classmethod
+    def text_response_server(cls, text, request_timeout=0.5, **kwargs):
+        def text_response_handler(sock):
+            request_content = consume_socket_content(sock, timeout=request_timeout)
+            sock.send(text.encode('utf-8'))
+
+            return request_content
+
+
+        server = Server(text_response_handler, **kwargs)
```
Lukasa: This is a `classmethod`, so you should change this line to be `server = cls(text_response_handler, **kwargs)`. That way, if anyone decides to subclass this object, they'll get the right behaviour.

tests/testserver/server.py
```diff
@@ -0,0 +1,107 @@
...
+        return server
+
+    @classmethod
+    def basic_response_server(cls, **kwargs):
+        server = cls.text_response_server(
+            "HTTP/1.1 200 OK\r\n" +
+            "Content-Length: 0\r\n\r\n", **kwargs
+        )
+        return server
```
Lukasa: You can save yourself a line here: just do `return cls.text_response_server()`.

tests/testserver/server.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+        server = Server(text_response_handler, **kwargs)
+        return server
+
+    @classmethod
+    def basic_response_server(cls, **kwargs):
+        server = cls.text_response_server(
+            "HTTP/1.1 200 OK\r\n" +
+            "Content-Length: 0\r\n\r\n", **kwargs
```
Lukasa: Can you bring the `**kwargs` onto their own line? It makes it a bit clearer that it's there.

tests/testserver/server.py
```diff
@@ -0,0 +1,107 @@
...
+    def run(self):
+        try:
+            sock = self._create_socket_and_bind()
+            # in case self.port = 0
+            self.port = sock.getsockname()[1]
+            self.ready_event.set()
+            self._handle_requests(sock)
+
+            if self.wait_to_close_event:
+                self.wait_to_close_event.wait(self.WAIT_EVENT_TIMEOUT)
```
Lukasa, line 70: I'm a little nervous about this: should this be inside the `finally` block? I'm concerned that if an error gets hit we'll close the connection earlier than expected, and that will lead to misleading concerns elsewhere. But I'm not _certain_ of that.

Lukasa:
Ok, we're making some good progress here @BraulioVM! We'll have a few more rounds of code review while we get this stuff settled, and right now our Jenkins runs are hanging (I'll investigate why).

Did you lose `test_chunked_upload` entirely? It doesn't seem to be in the diff at all: commit f17ef753d2c1f4db0d7f5aec51261da1db20d611 just removed it, without moving it _to_ anywhere.

tests/test_requests.py
BraulioVM: Ok that was not on purpose, I must have pasted it without noticing it was there. Fixed

tests/testserver/server.py
BraulioVM, line 70: I'm not sure. If an exception is thrown then there's something that the test implementer needs to take care of (he may have written an incorrect request handler, for example). I believe that it is a good thing that the server gets closed as soon as possible if an exception is found, otherwise the client sockets will be sending messages to a server that is behaving in unexpected ways and that could hide bugs.

BraulioVM:
I'm sorry for needing this much reviewing. I hope that I can get the code right from the beginning with my next PR.

Lukasa:
@BraulioVM I can't stress this enough: you don't need to apologise for code review. Code review is a healthy part of the development process, and it doesn't matter how good you are: you _will_ get code review, and it's a good thing. You should see some of my patches on Twisted, which have had more than 100 code-review comments on them. ;)

tests/test_requests.py
```diff
@@ -1452,3 +1453,17 @@ def test_vendor_aliases():
...
+    server = Server.basic_response_server(wait_to_close_event=block_server)
+    data = (i for i in [b'a', b'b', b'c']) 
+
+    with server as (host, port):
+        url = 'http://{0}:{1}/'.format(host, port)
+        r = requests.post(url, data=data, stream=True)
+        block_server.set() # release server block
+
+    assert r.status_code == 200
+    assert r.request.headers['Transfer-Encoding'] == 'chunked'
```
Lukasa: So, I'm wondering if we should add a new test file for this kind of test. Urllib3 has a `test_socketlevel.py` file, we might want to consider having a `test_lowlevel.py` with this as the first test inside it.

tests/test_requests.py
```diff
@@ -1452,3 +1453,17 @@ def test_vendor_aliases():
 
     with pytest.raises(ImportError):
         from requests.packages import webbrowser
+
+def test_chunked_upload():
+    """can safely send generators"""
+    block_server = threading.Event()
```
Lukasa: This is a really tiny note, but I think that it would be better if we called this `close_server`. It just feels a bit clearer to me.

tests/test_testserver.py
```diff
@@ -0,0 +1,128 @@
...
+        answer = b"yeah, success"
+        def handler(sock):
+            text = sock.recv(1000)
+            assert text == question 
+            sock.send(answer)
+        
+        with Server(handler) as (host, port):
+            sock = socket.socket()
+            sock.connect((host, port))
+            sock.send(question)
```
Lukasa: So here's a useful thing to know: `sock.send()` may not send all the data you pass to it. It has a return value that is how many bytes of the data to sent actually get sent.

You may find it helpful to replace this `send` with `sendall`, and probably all instances of it in this pull request. =)

tests/test_testserver.py
```diff
@@ -0,0 +1,128 @@
...
+    def test_server_closes(self):
+        with Server.basic_response_server() as (host, port):
+            sock = socket.socket()
+            sock.connect((host, port))
+
+            sock.close()
+
+        with pytest.raises(socket.error):
+            new_sock = socket.socket()
+            new_sock.connect((host, port))
```
Lukasa, line 37: What's this test doing?

Lukasa, line 37: Oh, wait, I see. Ok, it might be useful to add doc strings to these tests that say what specific function they're testing.

tests/test_testserver.py
```diff
@@ -0,0 +1,128 @@
...
+            "HTTP/1.1 200 OK\r\n" + 
+            "Content-Length: 6\r\n" +
+            "\r\nroflol"
+        )
+
+        with server as (host, port):
+            r = requests.get('http://{0}:{1}'.format(host, port))
+
+            assert r.status_code == 200
+            assert r.text == 'roflol'
```
Lukasa: `text` is always going to be Unicode, so let's change this to `u'roflol'` just to confirm that this always works.

tests/test_testserver.py
```diff
@@ -0,0 +1,128 @@
...
+            assert r.headers['Content-Length'] == '6' 
+            
+    def test_basic_response(self):
+        with Server.basic_response_server() as (host, port):
+            r = requests.get('http://{0}:{1}'.format(host, port))
+            assert r.status_code == 200
+            assert r.text == ''
```
Lukasa: Same here, let's make this the empty Unicode string `u''`.

tests/test_utils.py
```diff
@@ -151,7 +153,6 @@ def test_precedence(self):
         '''.strip()
         assert get_encodings_from_content(content) == ['HTML5', 'HTML4', 'XML']
 
-
```
Lukasa, line 154: We don't need any of the diff in this file. =)

tests/testserver/server.py
```diff
@@ -0,0 +1,101 @@
+import threading 
+import socket
+import select
+
+def consume_socket_content(sock, timeout=0.5):
```
Lukasa, line 6: Can we put two lines between each function and class at the top level?

tests/testserver/server.py
Lukasa, line 70: That sounds reasonable to me, I'm prepared to accept that. =) We'll revisit if it turns out to be a problem.

BraulioVM:
Ok, done. I am ready for the next round!

Lukasa:
Hmm, the Jenkins builds for this PR are all messed up. Lemme see if I can coerce Jenkins into starting a new build.

Lukasa:
Ok, so Jenkins is now being helpful. The specific problem is import paths on Python 3, which is a result of the way the tests are invoked. I suspect you can fix that by changing the import path of test server, so that instead of doing `from testserver.server import Server`, you do `from .testserver.server import Server`.

Want to give that a try?

BraulioVM:
I went with `from tests.testserver.server import Server`. Is that ok? Or is the other option considered to be better in this particular case?

BraulioVM:
I'll review tonight all the style issues my PR has already had and look for them in other parts of my code in order to fix them. Anything else I should be doing? 

Glad to see the tests are going just right in jenkins.

Lukasa:
@BraulioVM I think I'm happy with this as-is at this point, but I'm also a bit too close to it now to be objective. I'd like to ask @sigmavirus24 to do a review as well now, if he has time: just to make sure the code is as good as it can be. Otherwise, I think this is very close to being ready to merge.

BraulioVM:
Ok, I'll wait then!

kennethreitz:
Let's get this merged or closed soon — it's been open since late November.

BraulioVM:
If you think there is something wrong with this PR tell me and I will fix it

Lukasa:
@sigmavirus24, what are the odds that you'll have time in the next 7 days to do a review?

tests/testserver/server.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _handle_requests(self, server_sock):
+        for _ in range(self.requests_to_handle):
+            sock = server_sock.accept()[0]
+            handler_result = self.handler(sock)
+
+            self.handler_results.append(handler_result)
+        
+    def __enter__(self):
+       self.start()
```
sigmavirus24: The indentation here is not consistent.

sigmavirus24:
One nit that we could probably fix up ourselves if we want to. Otherwise :+1: from me.

BraulioVM:
Fixed!

kennethreitz:
✨ 🍰 ✨

PR #2917, opened 2015-12-05
requests/auth: Handle an empty 'qop' attribute in an Authenticate challenge

### LINKED ISSUE #2916: HTTPDigestAuth: Empty qop header in server challenge results in Authorization: None
This is a bit of a weird one, and I'm fairly certain the server is misbehaving slightly, but since I have less control over that:

When issuing a request with the following:
```
        response = requests.post(url,
            auth=HTTPDigestAuth('admin', 'foobar'))
```

The server returns back a challenge with the following ```WWW-Authenticate``` header:

```
reply: 'HTTP/1.1 401 Unauthorized\r\n'
header: Server: nginx
header: Date: Sat, 05 Dec 2015 02:08:43 GMT
header: Content-Type: text/html; charset=utf-8
header: Content-Length: 188
header: Connection: keep-alive
header: WWW-Authenticate: Digest realm="foobar_api_auth", qop="", nonce="a12059eaaad0b86ece8f62f04cbafed6", algorithm="MD5", stale="false"
```
Note the ```qop``` header with an empty value. That doesn't appear to be valid per RFC 2617:

>       qop-value         = "auth" | "auth-int" | token

That being said, RFC 2617 does specify that unrecognised values MUST be ignored. Instead, it looks like requests drops the entire authorization header:

```

        if qop is None:
            respdig = KD(HA1, "%s:%s" % (nonce, HA2))
        elif qop == 'auth' or 'auth' in qop.split(','):
            noncebit = "%s:%s:%s:%s:%s" % (
                nonce, ncvalue, cnonce, 'auth', HA2
                )
            respdig = KD(HA1, noncebit)
        else:
            # XXX handle auth-int.
            return None
```

This results in the following being sent back to the challenge:

```
send: 'POST /json HTTP/1.1\r\nHost: foo.com\r\nContent-Length: 94\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nUser-Agent: python-requests/2.8.1\r\nConnection: keep-alive\r\nContent-Type: application/json\r\nAuthorization: None\r\n\r\n{"data": "more}'
```

I'd be more than happy to work on a PR for this issue, but would like to confirm that a PR that updates the handling of ```qop``` to be tolerant to unrecognized tokens/characters would be acceptable. Arguably, the server should be omitting the ```qop``` header completely, but RFC compliance in requests would probably make it more tolerant to misbehaving servers like this one.

Thanks!

matt-jordan:
Some malfunctioning HTTP servers may return a qop directive with no token, as
opposed to correctly omitting the qop directive completely. For example:

```
header: WWW-Authenticate: Digest realm="foobar_api_auth", qop="",
        nonce="a12059eaaad0b86ece8f62f04cbafed6", algorithm="MD5",
        stale="false"
```

Prior to this patch, requests would respond with a 'None' Authorization header.
While the server is certainly incorrect, this patch updates requests to be
more tolerant to this kind of shenaniganry. If we receive an empty string for
the value of the qop attribute, or some tokens that are not recognized (that
is, not 'auth' or 'auth-int'), we instead treat that as if the qop attribute
was simply not provided.

Closes #2916

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -136,7 +136,7 @@ class HTTPDigestAuth(AuthBase):
         if _algorithm == 'MD5-SESS':
             HA1 = hash_utf8('%s:%s:%s' % (HA1, nonce, cnonce))
 
-        if qop is None:
+        if not qop:
             respdig = KD(HA1, "%s:%s" % (nonce, HA2))
         elif qop == 'auth' or 'auth' in qop.split(','):
             noncebit = "%s:%s:%s:%s:%s" % (


Lukasa:
Thanks for this!

I think this change is more aggressive than we need actually. Is there any reason you didn't take the simpler path and just change the conditional to look for falsy values of `qop`?

matt-jordan:
Just got a bit trigger happy on trying to handle unrecognized values in the `qop` attribute. A much simpler PR is now up. :smile:

Lukasa:
\o/

Given that `None` is falsy, I think we can simplify this a little bit further to simply check `if not qop`.

matt-jordan:
Good point. Simpler patch incoming.

sigmavirus24:
I pulled this down and ran the tests locally. All seems well on this end.

PR #2919, opened 2015-12-07
[DRAFT] WIP in finding a solution to data encoding check

ArcTanSusan:
Use existing method `to_native_string()` as a check.

This is a first attempt to resolve https://github.com/kennethreitz/requests/issues/2638, so marking this PR as [DRAFT]. I'd appreciate feedback on this.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -444,6 +444,12 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
                 (body, content_type) = self._encode_files(files, data)
             else:
                 if data:
+                    # Check if data is encoded
+                    if isinstance(data, list) and isinstance(data[0], tuple):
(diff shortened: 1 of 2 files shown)


Lukasa:
Yeah, so how we approach this is basically dependent on whether we think it's acceptable for `to_native_string()` to go in `encode_params`. I personally think that's the best place for it, because it allows us to avoid duplicating the large amount of knowledge `encode_params` has about how this variable is structured. However, IIRC @sigmavirus24 originally objected to that idea so I'd like to hear from him.

sigmavirus24:
This should be its own function that we can test thoroughly because I think there's a lot of room to do this wrong.

If we have similar checking in a lot of places, make it one function that we run often and that we test well. Our `_encode_*` methods are too large as it is and barely tested at all

PR #2923, opened 2015-12-09
Add hint to :param verify.

bsamek:
It is not clear that :param verify defaults to True. The way the verify
portion of the docstring is written it looks like it defaults to False, and
you have to pass in True if you'd like the SSL cert to be verified, but the
opposite is the case.


Lukasa:
Thanks @bsamek! :sparkles: :cake: :sparkles:

PR #2926, opened 2015-12-11
Refactor default params

bsamek:
Specified the default argument for params that have a default in the docstring
so that the default is easier to see from the code. Modified the docstring in
api.py to match the docstring in sessions.py.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -162,3 +162,4 @@ Patches and Suggestions
 - Smiley Barry (`@smiley <https://github.com/smiley>`_)
 - Shagun Sodhani (`@shagunsodhani <https://github.com/shagunsodhani>`_)
 - Robin Linderborg (`@vienno <https://github.com/vienno>`_)
+- Brian Samek(`@bsamek <https://github.com/bsamek`_)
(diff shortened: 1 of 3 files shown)


bsamek:
I also noticed that there may be a problem in test_requests.py. Lines 524-5 reference the httpbin_ca_bundle fixture, but there is no fixture by that name. If I am reading this correctly, this means that the test is passing in None instead of a cert path. Does this make sense?

sigmavirus24:
Hey @bsamek! Thanks for the PR.

I don't think we want to change the default parameters to the `request` method. We want to be able to discern between something that's a session setting and a request setting which we cannot do if we change that signature. We use [`merge_setting`](https://github.com/bsamek/requests/blob/master/requests/sessions.py#L49..L53) to make that determination and I don't think we want to break this.

Please revert that as the docstring changes _are_ desirable.

bsamek:
Reverted the changes to the parameters. Kept the docstring changes.

Should I submit an issue regarding my earlier questions about the test?

sigmavirus24:
There absolutely is such a fixture: https://github.com/kevin1024/pytest-httpbin/blob/master/pytest_httpbin/plugin.py#L40

sigmavirus24:
Thanks again @bsamek !

bsamek:
Ohhhhhhhhh I see. Thanks @sigmavirus24 !

sigmavirus24:
Anytime @bsamek :) Thanks for the PR!

PR #2927, opened 2015-12-15
Prepare 2.9.0 release.

Lukasa:
This should contain the set of changes required for us to push 2.9.0 out the door.


Lukasa:
@sigmavirus24 Can you confirm for me that you're happy with these release notes?

HISTORY.rst
```diff
@@ -10,6 +10,28 @@ dev (XXXX)
...
+  send the content length for the number of bytes we will actually read, rather
+  than the total size of the file, allowing partial file uploads.
+- When uploading file-like objects, if they are empty or have no obvious
+  content length we set ``Transfer-Encoding: chunked`` rather than
+  ``Content-Length: 0``.
+- We correctly receive the response in buffered mode when uploading chunked
+  bodies.
+- We now handle being passed a query string as a bytestring on Python 3, by
+  decoding it as UTF-8.
+- Sessions are now closed in a timely manner when using the function API,
```
sigmavirus24: I would prefer

```
Sessions are now closed in all cases (exceptional and not) when using the functional API rather than leaking and waiting for the garbage collector to clean them up.
```

requests/packages/urllib3/connection.py
```diff
@@ -6,18 +8,13 @@
 from .packages import six
 
 try:  # Python 3
-    from http.client import HTTPConnection as _HTTPConnection, HTTPException
+    from http.client import HTTPConnection as _HTTPConnection
```
sigmavirus24, line 11: Ugh, I need to remember to go clean all of these imports up in urllib3 because some of the super shitty "future" modules that "backport" Python 3 module names to Python 2 will break things for us here and for urllib3 in general.

There's nothing wrong with 2.9.0, this is just something that I know will bite us.

sigmavirus24, line 11: Specifically I'm talking about:
- https://github.com/timothycrosley/pies/issues/49
- https://github.com/PythonCharmers/python-future

They both cause an incredible amount of breakage for this pattern of Python 2/3 compat (that doesn't happen when using `six.moves`).

sigmavirus24:
You also need to bump `__version__` and `__build__`. :)

Lukasa:
@sigmavirus24 I intentionally did not do that, so that we have a single commit entitled `v2.9.0`, rather than a merge commit that hits it. =D

sigmavirus24:
Got it. Well the rest of this looks excellent to me. Feel free to do that and ping me for the merge. :)

PR #2928, opened 2015-12-15
make setup.py upload happy

heni:
distutils brokes when we use tuple based meta values
canonic way is define classifiers as list not a tuple

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
     install_requires=requires,
     license='Apache 2.0',
     zip_safe=False,
-    classifiers=(
+    classifiers=[
         'Development Status :: 5 - Production/Stable',
         'Intended Audience :: Developers',
         'Natural Language :: English',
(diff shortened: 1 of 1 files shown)


Lukasa:
@heni We've been using a tuple here since [literally the start of the project](https://github.com/kennethreitz/requests/commit/0477018761c67152cdcc0b83d56f27e701e65b9e), which _used_ distutils. I do not believe this can possibly be the problem you're experiencing.

heni:
I want to redistribute this package into private repository of our project and with tuple for classifiers I have broken next chain of commands
  pip3 install --download=. --no-binary=:all: requests
  tar xf requests-2.9.0.tar.gz
  cd requests-2.9.0
  python3 setup.py bdist_wheel upload -r private-repo

Lukasa:
Distutils does not work with wheel, you have to use setuptools.

sigmavirus24:
Also `setup.py upload` most likely is _not_ performing certificate validation. Use `twine` instead.

PR #2936, opened 2015-12-18
Handle bytes and unicode URLs for netloc

Lukasa:
Spotted this on my local machine running tests on Python 3. Annoyingly, we didn't see this on the CI rig because the Jenkins box doesn't have a netrc file, so it never gets that far in the function.

However, tests preparing bytes URLs fail here on Python 3 because the splitter is the wrong type. This change ensures that we use the appropriate type.

Feels like this is safe for 2.9.1.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -116,7 +116,10 @@ def get_netrc_auth(url, raise_errors=False):
         ri = urlparse(url)
 
         # Strip port numbers from netloc
-        host = ri.netloc.split(':')[0]
+        if isinstance(url, str):
+            host = ri.netloc.split(u':')[0]
+        else:
+            host = ri.netloc.split(b':')[0]
 
         try:
             _netrc = netrc(netrc_path).authenticators(host)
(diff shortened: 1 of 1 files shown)


requests/utils.py
```diff
@@ -116,7 +116,10 @@ def get_netrc_auth(url, raise_errors=False):
         ri = urlparse(url)
 
         # Strip port numbers from netloc
-        host = ri.netloc.split(':')[0]
+        if isinstance(url, str):
```
sigmavirus24, line 121: Suggestion:

``` py
splitstr = b':'
if isinstance(url, str):
    splitstr = splitstr.decode('utf-8')
host = r.netloc.split(splitstr)[0]
```

My reasoning is that while we don't test against Python 3.2, pip still supports 3.2 and so we need to support it as best as we can as well. `u''` is not supported on 3.2 and that's why the above might be a bit friendlier.

Lukasa, line 121: Done. =)

sigmavirus24:
Woot! :+1: 

Thanks @Lukasa (also YAY OUR CI IS WORKING AGAIN)

PR #2937, opened 2015-12-18
Release 2.9.1

Lukasa:
This release is scheduled for Monday. I'm aiming to set this up now so that it's an easy release to push on Monday.

@shazow, what are the odds that we can get a urllib3 patch release 1.13.1 containing the fix for shazow/urllib3#761 by Monday? I'd owe you lots of :custard: if we could get it! :heart:

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,15 @@
 Release History
 ---------------
 
+2.9.1 (2015-12-21)
+++++++++++++++++++
+
+**Bugfixes**
+
+- Resolve regression introduced in 2.9.0 that made it impossible to send binary
+  strings as bodies in Python 3.
+- Fixed errors when calculating cookie expiration dates in certain locales.
+
 2.9.0 (2015-12-15)
 ++++++++++++++++++
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -12,6 +12,10 @@ Release History
   strings as bodies in Python 3.
 - Fixed errors when calculating cookie expiration dates in certain locales.
 
+**Miscellaneous**
+
+- Updated bundled urllib3 to 1.13.1.
+
 2.9.0 (2015-12-15)
 ++++++++++++++++++
 
@@ -42,6 +46,10 @@ Release History
   contains no token, by treating it the same as if no ``qop`` directive was
   provided at all.
 - Minor performance improvements when removing specific cookies by name.
+
+**Miscellaneous**
+
+- Updated urllib3 to 1.13.
 
 2.8.1 (2015-10-13)
 ++++++++++++++++++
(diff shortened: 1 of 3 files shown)


Lukasa:
Note: please don't merge this until we're ready to release on Monday! We may land other patches between now and then, and we'll want to make sure the changelog is updated!

For those who are interested: the reason we're enforcing a delay between 2.9.0 and 2.9.1 is to try to land as many regression fixes as possible in a single release, rather than rapidly releasing several versions as issues become apparent to us. Now that requests is a widely used project it becomes important to be as reliable as possible in our release management to avoid making people's lives too hard.

sigmavirus24:
@Lukasa are you thinking about making an exception to [our release document](https://github.com/kennethreitz/requests/blob/master/docs/community/release-process.rst#hotfix-releases) in this case?

Lukasa:
@sigmavirus24 Yeah, good question, but I think I am, given that the urllib3 release would itself be a patch release. Any concerns there? @ralphbean @eriol?

shazow:
@Lukasa Could do. Could try to sneak in SOCKS support too, if you wanna.

Also I noticed some of your changelist includes an update note to the urllib3 version, some don't. Dunno if that's intentional.

Lukasa:
@shazow I'd rather leave SOCKS out for now: makes 1.13.1 very clearly a simple bugfix release.

Yeah, the missing changelist notes are in error, we should try to fix it.

shazow:
Sounds good, I'll push out v1.13.1 before monday (maybe today).

sigmavirus24:
@Lukasa yeah, if it's just a couple small bug-fixes, that's fine by me. I think we've been bitten in the past by urllib3 having a lot more than just bug fixes when we pulled it in.

ralphbean:
:+1: here.  I just finished packaging 2.9.0.  Bumping to 2.9.1/1.13.1 shouldn't be an issue.  Thanks!

Lukasa:
@ralphbean Frankly I don't recommend releasing 2.9.0, it had enough nasty bugs that going straight to 2.9.1 is going to be the better approach.

eriol:
:+1: from me. I will skip 2.9.0 since 2.9.1 is going to be released soon, thanks!

shazow:
urllib3 v1.13.1 is live.

ralphbean:
> Frankly I don't recommend releasing 2.9.0, it had enough nasty bugs that going straight to 2.9.1 is going to be the better approach.

Acknowledged.  I attached a comment to let people know on the 1.9.0 [Fedora updates](https://bodhi.fedoraproject.org/updates/python-requests).

Lukasa:
\o/ Thanks @shazow!

sigmavirus24:
@Lukasa is this still a WIP? I believe you wanted to cut this today

Lukasa:
Heh, was just waiting for you. No longer a WIP, should be good to go.

sigmavirus24:
:sparkles: :cake: :sparkles: 

:sparkler: :shipit:

PR #2948, opened 2015-12-28
advanced: use "client.*" to designate client certificate

vincentbernat:
Using "server.crt" is confusing as one may try to put the server certificate while this is really the client certificate that should be put here. Instead, use "client.cert", "client.key" and "client.pem".


Lukasa:
Good spot, thanks! :sparkles: :cake: :sparkles:

PR #2952, opened 2015-12-29
Discard 0-byte chunks from request body iterators

tipabu:
Previously, an empty chunk in a request body iterator would be
faithfully translated to a 0-length chunk on the wire, which the server
would interpret as the end of the transfer. That is, a request like

    def data():
        yield 'foo'
        yield ''
        yield 'bar'
    requests.put(url, data=data())

... would cause the server to receive a chunk for 'foo', receive an
empty chunk, and close the connection before receiving 'bar'.

Further, if `Connection: keep-alive` is set, the server will attempt (and
fail) to parse the next chunk, serve a 400, and close the connection.

Now, the empty chunk will be ignored, and the server should receive all
of the actual bytes emitted from the iterable.


sigmavirus24:
This is a breaking change that cannot be merged for 2.x but has been fixed for 3.0.0 in #2631 by @neosab.

Issue #2562, opened 2015-04-23
Merge with requesocks


ikus060:
Any plan to merge modification made to requests to support SOCKS proxy ?

See https://github.com/dvska/requesocks

sigmavirus24:
It doesn't seem likely that we'll merge with requesocks as that hasn't been updated in over 2 years. Also that author never propsed a merge on their own.

ikus060:
May I propose to add this feature to requests ?

Lukasa:
@ikus060 Many people have proposed it and even begun work: see #2425, #2156, #1982, shazow/urllib3#486, shazow/urllib3#284, shazow/urllib3#68, and shazow/urllib3#44.

You are welcome to attempt to do the work. The initial implementation should be over in urllib3: I recommend that you start work there.

If you were simply _requesting_ the feature, the core team does not believe it's an important enough priority to justify our development time. In practice, the absence of SOCKS support does not appear to be limiting the success of the library. For that reason, unless someone does the work for the implementation themselves, I do not believe we'll be implementing it ourselves.

Issue #2425, opened 2015-01-25
SOCKS proxy support


jorenham:
I would like to request support for Socket Secure protocol. There exists a fork (https://github.com/dvska/requesocks) with this capability, but it is limited (python 2 only). 

Currently the ```AssertionError: Not supported proxy scheme socks5``` is thrown when attempting to connect to a socks5 proxy.

A quick Google search (https://www.google.nl/search?q=socks+proxy+list) illustrates the popularity of the protocol. Even Tor and I2P use it.

sigmavirus24:
https://github.com/kennethreitz/requests/issues?q=is%3Aissue+socks+is%3Aclosed would have been a good search before opening this issue as well.

sigmavirus24:
https://github.com/kennethreitz/requests/pull/478 was the only effort anyone made towards including this support in requests over 2 years ago. As of now, https://github.com/shazow/urllib3/pull/486 is still in progress. Until that is finished there is nothing we can do about supporting socks proxies.

loretoparisi:
If I do 

``` python
#proxy
        # SOCKS5 proxy for HTTP/HTTPS
        proxies = {
            'http' : "socks5://myproxy:9191",
            'https' : "socks5://myproxy:9191"
        }

        #headers
        headers = {

        }

        url='http://icanhazip.com/'
        res = requests.get(url, headers=headers, proxies=proxies)
```

I get `AssertionError: Not supported proxy scheme socks5`

AlJohri:
@loretoparisi make sure you have the latest version of `requests` and also install `PySocks`. here is an example of using Tor (via a socks5 proxy) with requests: https://github.com/AlJohri/python-tor-examples

itJunky:
How can i open .onion domains with this code?
Now i getting error:

```
requests.exceptions.ConnectionError: HTTPConnectionPool(host='zqktlwi4fecvo6ri.onion', port=80): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fd35ba09b10>: Failed to establish a new connection: [Errno -2] Name or service not known',))
```

AlJohri:
You need to use the `socks5h` scheme ([stackoverflow](https://stackoverflow.com/questions/42971622/fetching-a-onion-domain-with-requests)):

https://github.com/requests/requests/blob/e3f89bf23c53b98593e4248054661472aacac820/requests/packages/urllib3/contrib/socks.py#L155-L160

[...]

PR #2953, opened 2015-12-30
SOCKS Proxy Support

Lukasa:
This pull request *finally* adds SOCKS proxy support to the Requests module, pretty much four years *to the day* since we were first asked (in #324 for those who want to follow it).

This pull request relies on a pull request I have opened against urllib3, so we can't land it until urllib3 1.14.1 is released containing that code. However, once that's done this PR should be ready to rock.

I'm setting this up for the 2.10.0 milestone as I see no reason to wait for 3.0.0 for SOCKS support.

Resolves #324, #723, #1147, #1982, #2156, #2425, and #2562.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -303,17 +303,16 @@ class HTTPAdapter(BaseAdapter):
         """
         proxy = select_proxy(request.url, proxies)
         scheme = urlparse(request.url).scheme
-        proxy_scheme = ''
-        if proxy:
-            proxy_scheme = urlparse(proxy).scheme
 
-        if proxy and proxy_scheme.lower().startswith('socks'):
-            # Socks proxies behave like the proxy isn't there at all.
-            url = request.path_url
-        elif proxy and scheme != 'https':
+        is_proxied_http_request = (proxy and scheme != 'https')
+        using_socks_proxy = False
+        if proxy:
+            proxy_scheme = urlparse(proxy).scheme.lower()
+            using_socks_proxy = proxy_scheme.startswith('socks')
+
+        url = request.path_url
+        if is_proxied_http_request and not using_socks_proxy:
             url = urldefragauth(request.url)
-        else:
-            url = request.path_url
 
         return url
 
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -302,7 +302,8 @@ class HTTPAdapter(BaseAdapter):
         """
         proxy = select_proxy(request.url, proxies)
         scheme = urlparse(request.url).scheme
-        proxy_scheme = urlparse(proxy).scheme
+        if proxy:
+            proxy_scheme = urlparse(proxy).scheme
 
         if proxy and proxy_scheme.lower().startswith('socks'):
             # Socks proxies behave like the proxy isn't there at all.
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -19,7 +19,7 @@ from .packages.urllib3.util.retry import Retry
 from .compat import urlparse, basestring
 from .utils import (DEFAULT_CA_BUNDLE_PATH, get_encoding_from_headers,
                     prepend_scheme_if_needed, get_auth_from_url, urldefragauth,
-                    select_proxy)
+                    select_proxy, to_native_string)
 from .structures import CaseInsensitiveDict
 from .packages.urllib3.exceptions import ClosedPoolError
 from .packages.urllib3.exceptions import ConnectTimeoutError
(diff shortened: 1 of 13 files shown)


jezdez:
:pray:

requests/adapters.py
```diff
@@ -150,14 +156,26 @@ def proxy_manager_for(self, proxy, **proxy_kwargs):
         :returns: ProxyManager
         """
         if not proxy in self.proxy_manager:
```
sigmavirus24: While we're here, let's reformat this a bit:

``` py
if proxy in self.proxy_manager:
    manager = self.proxy_manager[proxy]
elif proxy.lower().startswith('socks'):
    username, password = get_auth_from_url(proxy)
    manager = self.proxy_manager[proxy] = SOCKSProxyManager(
        # ...
    )
else:
    proxy_headers = self.proxy_headers(proxy)
    manager = self.proxy_manager[proxy] = proxy_from_url(
        # ...
    )
return manager
```

I feel like this will be easier to read

requests/adapters.py
```diff
@@ -284,7 +302,13 @@ def request_url(self, request, proxies):
         """
         proxy = select_proxy(request.url, proxies)
         scheme = urlparse(request.url).scheme
-        if proxy and scheme != 'https':
+        if proxy:
+            proxy_scheme = urlparse(proxy).scheme
+
+        if proxy and proxy_scheme.lower().startswith('socks'):
```
sigmavirus24: You conditionally create `proxy_scheme` but unconditionally reference it here unless my coffee hasn't kicked in. Shouldn't this block look like:

``` py
proxy_scheme = ''
if proxy:
    proxy_scheme = urlparse(proxy).scheme

if proxy and proxy_scheme.lower().startswith('socks'):
    url = request.path_url
# ...
```

Lukasa: I'm being a little tricksy here, because `and` is short-circuiting in Python. This means that I actually only reference `proxy_scheme` here if `proxy` is truthy, which is the condition under which we populate it in the first place.

However, it's definitely a bit unclear, so I'm happy to change it.

requests/adapters.py
Lukasa: Done

requests/adapters.py
Lukasa: Done.

requests/adapters.py
```diff
@@ -284,7 +303,14 @@ def request_url(self, request, proxies):
         """
         proxy = select_proxy(request.url, proxies)
         scheme = urlparse(request.url).scheme
-        if proxy and scheme != 'https':
+        proxy_scheme = ''
+        if proxy:
+            proxy_scheme = urlparse(proxy).scheme
+
+        if proxy and proxy_scheme.lower().startswith('socks'):
```
sigmavirus24: I see how you were being tricksy now. I wonder if we can be better about this.

We essentially have the same code path (`url = request.path_url`) if we're using a 'socks' proxy or if we're making an HTTPS request. What if we do

``` py
proxied_http_request = proxy and scheme != 'https'
using_socks_proxy = False
if proxy:
    using_socks_proxy == urlparse(proxy).scheme.lower().startswith('socks')

url = request.path_url
if proxied_http_request and not using_socks_proxy:
    url = urldefragauth(request.url)
```

I puzzled through my logic with 

| proxy | http_scheme && !socks_proxy_scheme | https_scheme | socks_proxy_scheme |
| --- | --- | --- | --- |
| truthy | urldefragauth(request.url) | request.path_url | request.path_url |
| falsey | request.path_url | request.path_url | request.path_url |

where `http_scheme` is where `(scheme != 'https') is True` and `https_scheme` is where `(scheme != 'https') is False`. Hopefully the table makes sense.

Lukasa: Ok, I'm going with that. I degolfed this a little bit to aid clarity, but not much.

sigmavirus24:
:+1: on the code and :+1: on the milestone.

sigmavirus24:
@Lukasa what do you think about having a `socks` extra for requests that installs `PySocks` and whatever else?

Lukasa:
Done. =)

kennethreitz:
Does PySocks require any compilation during installation? We should consider vendoring it if we want to consider this an actual feature of Requests.

Lukasa:
PySocks requires no compilation. However, we'll have to rewrite import paths in urllib3 if you want to vendor PySocks.

kennethreitz:
Well, if it's going to be used by a very small subset of users (likely the case), it's likely fine to leave it as an optional dependency.

We could pack it in without modifying urllib3's code, technically:

(`packages/__init__.py`)

``` python
try:
    from . import socks
    sys.modules['socks'] = socks
except ImportError:
    try:
        import socks
    except ImportError:
        pass
```

This is obviously not the best idea, but it does function properly.

kennethreitz:
Actually, simply adding the following line to `requests/packages/__init__.py` makes it function properly (assuming `socks.py` and `sockshandler.py` are present):

```python
from .urllib3 import contrib
```

sigmavirus24:
Mucking with sys.modules in the past has caused us enough headaches that I'd rather not go too far down that path again

kennethreitz:
I don't disagree :)

loretoparisi:
@Lukasa is this was merged or not? Btw, how to configure the socks proxy?

Lukasa:
This has not been merged.

When it is released, you can configure the socks proxy as you would configure other proxies: `proxies={'http': 'socks5://localhost:8888/'}`, for example.

loretoparisi:
@Lukasa so it should work like

``` python
#proxy
        # SOCKS5 proxy for HTTP/HTTPS
        proxies = {
            'http' : "socks5://myproxy:9191",
            'https' : "socks5://myproxy:9191"
        }

        #headers
        headers = {

        }

        url='http://icanhazip.com/'
        res = requests.get(url, headers=headers, proxies=proxies)
```

Lukasa:
Correct. Of course, that won't work unless you use a patched version of requests that contains both this patch _and_ an updated urllib3, _and_ you install the appropriate optional dependencies.

loretoparisi:
@Lukasa uhm, so supposed to use on google appengine, I have to wait the `devappserver2` python sdk update...

Lukasa:
On GAE I suspect using a SOCKS proxy is going to take a while. I should also note that to use a SOCKS proxy on GAE will require that you enable socket support, which as I understand it is substantially more expensive than simply using the standard httplib replacement.

loretoparisi:
btw, In the meanwhile I'm doing

``` shell
>>> import requesocks
>>> import requests
>>> session = requesocks.session()
>>> session.proxies = {
...     'http' : "socks5://myproxy.net:9050",
...     'https' : "socks5://smyproxy.net:9050"
... }
>>> url='http://icanhazip.com/'
>>> resp = requests.get(url)
>>> resp
<Response [200]>
>>> resp.text
u'89.97.90.230\n'
>>> resp.text.strip()
u'89.97.90.230'
>>> resp = session.get(url)
>>> resp.text.strip()
u'171.25.193.78'
>>> 
```

that in most envs will work ( a part again appengine that lacks of `pwd` support 😡

kennethreitz:
Is this done? Let's merge it if so.

Lukasa:
We need an updated urllib3. I think @shazow is expecting to release this week.

shazow:
Released!

kennethreitz:
Sweet, let's do it.

kennethreitz:
We'll include it in the next release, which can therefore happen this week (or even today).

Lukasa:
@kennethreitz I'm away from my laptop, so if you're in a merging mood you can land the urllib3 update and then merge this.

Please note that it makes everyone's life easier if you merge in urllib3 directly from the brand new tag (1.15). Just a suggestion!

darvid:
The anticipation is killing me.

dequis:
What's [WIP] about this? Previous comments seem to suggest it's done, just waiting for the release.

(Also, referencing https://github.com/shazow/urllib3/pull/762 and #478)

kennethreitz:
:sparkles: :cake: :sparkles:

darvid:
DJ Khaled

kennethreitz:
Released, announced.

PR #2960, opened 2016-01-05
Rename {Missing,Invalid}Schema to *Scheme

whit537:
As @merwok pointed out at https://github.com/kennethreitz/requests/pull/512#commitcomment-1296187, schemes are what they're called, not schemas. I got tripped up by this at https://github.com/CenterForOpenScience/osf.io/pull/4745#issuecomment-169038417, where "InvalidSchema" made me think I was hitting a database error, when in fact it was a configuration error in the format of a URL.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -253,7 +253,7 @@ API Changes
       }
 
       # In requests 1.x, this was legal, in requests 2.x,
-      #  this raises requests.exceptions.MissingSchema
+      #  this raises requests.exceptions.MissingScheme
       requests.get("http://example.org", proxies=proxies)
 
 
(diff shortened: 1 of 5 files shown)


Lukasa:
Thanks @whit537. Unfortunately, this is a breaking change so we can't accept it until 3.0.0: can you adjust this PR to point to the proposed/3.0.0 branch instead please?

chadwhitacre:
Sure thing. Closing this PR ...

chadwhitacre:
It occurs to me that it'd be possible to implement this in a backwards-compatible way, but since that would involve more than simply renaming the error classes (I guess you'd make an `InvalidScheme` class that subclasses `InvalidSchema`) it would take more time and might deserve a test or two. I can't commit to that work so I'm leaving this PR closed.

PR #2961, opened 2016-01-05
Rename {Missing,Invalid}Schema to *Scheme

whit537:
As @merwok pointed out at https://github.com/kennethreitz/requests/pull/512#commitcomment-1296187, schemes are what they're called, not schemas. I got tripped up by this at https://github.com/CenterForOpenScience/osf.io/pull/4745#issuecomment-169038417, where "InvalidSchema" made me think I was hitting a database error, when in fact it was a configuration error in the format of a URL.

This PR is a redo of #2960 against `proposed/3.0.0` instead of `master`.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -251,7 +251,7 @@ API Changes
       }
 
       # In requests 1.x, this was legal, in requests 2.x,
-      #  this raises requests.exceptions.MissingSchema
+      #  this raises requests.exceptions.MissingScheme
       requests.get("http://example.org", proxies=proxies)
 
 
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -71,12 +71,12 @@ class TooManyRedirects(RequestException):
     """Too many redirects."""
 
 
-class MissingSchema(RequestException, ValueError):
-    """The URL schema (e.g. http or https) is missing."""
+class MissingScheme(RequestException, ValueError):
+    """The URL scheme (e.g. http or https) is missing."""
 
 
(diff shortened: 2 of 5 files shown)


Lukasa:
\o/ Thanks @whit537! :sparkles: :cake: :sparkles:

chadwhitacre:
Huzzah, thanks! :-)

@Lukasa++

sigmavirus24:
@Lukasa please add release notes for this. I think it may be a good idea to have some shim for this as people may be catching `InvalidSchema` or `MissingSchema` exceptions and they won't be able to catch those anymore in 3.0

Lukasa:
I'm away from the laptop for a bit, but I will do.

chadwhitacre:
> I think it may be a good idea to have some shim for this as people may be catching `InvalidSchema` or `MissingSchema` exceptions and they won't be able to catch those anymore in 3.0

If you add a shim then you're basically backwards-compatible, right? (That's what I was getting at with https://github.com/kennethreitz/requests/pull/2960#issuecomment-169070702.)

sigmavirus24:
@whit537 yeah, it isn't _necessary_ for 3.0 though and there's no way to tell people to move to those new messages unless we forcibly break things for them though.

What we could do is add `InvalidScheme` and `MissingScheme` as sub-classes of their now defunct relatives so that people can start (with 2.10.0) using those exceptions so they're future proofed for 3.0.0. I'm not sure if @Lukasa would agree that there is value in that though.

Lukasa:
I'm open to that plan.

PR #2970, opened 2016-01-14
Rebase to 2.9.1 changes

katakumpo:
I just rebased the proposed 3.0.0 to current master near 2.9.1 to further help with socks support there also :)


hwms:
As long as it pushes the 3.0 to current upgrades (and works of course) i am happy with it :)

PR #2972, opened 2016-01-20
Return request & response with TooManyRedirects

munro:
I'm using requests to fetch data from websites that I don't maintain, and sometimes websites have broken cyclic redirects, but still return useful responses.  So I would like to grab the last cut off response, similar to the browser.

The problem is that `TooManyRedirects` doesn't throw the ending response state for me to inspect, so this PR adds that.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -110,7 +110,11 @@ class SessionRedirectMixin(object):
                 resp.raw.read(decode_content=False)
 
             if i >= self.max_redirects:
-                raise TooManyRedirects('Exceeded %s redirects.' % self.max_redirects)
+                raise TooManyRedirects(
+                    'Exceeded %s redirects.' % self.max_redirects,
+                    request=prepared_request,
+                    response=resp
+                )
 
             # Release the connection back into the pool.
             resp.close()
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -110,11 +110,7 @@ class SessionRedirectMixin(object):
                 resp.raw.read(decode_content=False)
 
             if i >= self.max_redirects:
-                raise TooManyRedirects(
-                    'Exceeded %s redirects.' % self.max_redirects,
-                    request=prepared_request,
-                    response=resp
-                )
+                raise TooManyRedirects('Exceeded %s redirects.' % self.max_redirects, response=resp)
 
             # Release the connection back into the pool.
             resp.close()
--- a/test_requests.py
+++ b/test_requests.py
@@ -24,7 +24,7 @@ from requests.cookies import cookiejar_from_dict, morsel_to_cookie
 from requests.exceptions import (ConnectionError, ConnectTimeout,
                                  InvalidSchema, InvalidURL, MissingSchema,
                                  ReadTimeout, Timeout, RetryError, TooManyRedirects)
-from requests.models import PreparedRequest, DEFAULT_REDIRECT_LIMIT
+from requests.models import PreparedRequest
 from requests.structures import CaseInsensitiveDict
 from requests.sessions import SessionRedirectMixin
 from requests.models import urlencode
@@ -192,10 +192,11 @@ class TestRequests(object):
         try:
             requests.get(httpbin('redirect', '50'))
         except TooManyRedirects as e:
-            assert e.request is not None
-            assert len(e.response.history) == DEFAULT_REDIRECT_LIMIT
+            assert '/relative-redirect/20' in e.request.url
+            assert e.request.url == e.response.url
+            assert len(e.response.history) == 30
         else:
-            assert False
+            pytest.fail()
 
     def test_HTTP_302_TOO_MANY_REDIRECTS_WITH_PARAMS(self, httpbin):
         s = requests.session()
@@ -203,10 +204,11 @@ class TestRequests(object):
         try:
             s.get(httpbin('redirect', '50'))
         except TooManyRedirects as e:
-            assert e.request is not None
+            assert '/relative-redirect/45' in e.request.url
+            assert e.request.url == e.response.url
             assert len(e.response.history) == 5
         else:
-            assert False
+            pytest.fail()
 
     # def test_HTTP_302_ALLOW_REDIRECT_POST(self):
     #     r = requests.post(httpbin('status', '302'), data={'some': 'data'})
--- a/test_requests.py
+++ b/test_requests.py
@@ -191,7 +191,7 @@ class TestRequests(object):
     def test_HTTP_302_TOO_MANY_REDIRECTS(self, httpbin):
         try:
             requests.get(httpbin('redirect', '50'))
-        except TooManyRedirects, e:
+        except TooManyRedirects as e:
             assert e.request is not None
             assert len(e.response.history) == DEFAULT_REDIRECT_LIMIT
         else:
@@ -202,7 +202,7 @@ class TestRequests(object):
         s.max_redirects = 5
         try:
             s.get(httpbin('redirect', '50'))
-        except TooManyRedirects, e:
+        except TooManyRedirects as e:
             assert e.request is not None
             assert len(e.response.history) == 5
         else:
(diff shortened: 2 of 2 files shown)


sigmavirus24:
@Lukasa I'm +0.5 on this. The reason I'm not a full +1 is that we might start leaking sockets this way.

Lukasa:
@sigmavirus24 We shouldn't leak sockets here: the redirect logic consumes all the data from the connection, which should return it back to the connection pool, and that happens before we raise `TooManyRedirects`.

test_requests.py
```diff
@@ -188,6 +188,26 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
         assert r.history[0].status_code == 302
         assert r.history[0].is_redirect
 
+    def test_HTTP_302_TOO_MANY_REDIRECTS(self, httpbin):
+        try:
+            requests.get(httpbin('redirect', '50'))
+        except TooManyRedirects as e:
+            assert e.request is not None
```
Lukasa: Can we expand on this assertion a bit? For example, check some properties on the request, and confirm it's ok? Similarly, check the response?

test_requests.py
```diff
@@ -188,6 +188,26 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
...
+            assert len(e.response.history) == DEFAULT_REDIRECT_LIMIT
+        else:
+            assert False
```
Lukasa: Rather than using `assert False`, let's use [`pytest.fail`](https://pytest.org/latest/example/simple.html#writing-well-integrated-assertion-helpers) so that we can have a convenient message telling us what went wrong.

Lukasa:
Got a few small notes, but this looks really good! :cake:

test_requests.py
```diff
@@ -188,6 +188,28 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
         assert r.history[0].status_code == 302
         assert r.history[0].is_redirect
 
+    def test_HTTP_302_TOO_MANY_REDIRECTS(self, httpbin):
+        try:
+            requests.get(httpbin('redirect', '50'))
+        except TooManyRedirects as e:
+            assert '/relative-redirect/20' in e.request.url
+            assert e.request.url == e.response.url
+            assert len(e.response.history) == 30
```
munro, line 198: I removed the `DEFAULT_REDIRECT_LIMIT` reference, so if it's changed this test will break.  There's probably people relying on it, so this will raise a discussion.

requests/sessions.py
```diff
@@ -110,7 +110,7 @@ def resolve_redirects(self, resp, req, stream=False, timeout=None,
                 resp.raw.read(decode_content=False)
 
             if i >= self.max_redirects:
-                raise TooManyRedirects('Exceeded %s redirects.' % self.max_redirects)
+                raise TooManyRedirects('Exceeded %s redirects.' % self.max_redirects, response=resp)
```
munro, line 113: Switched to just setting the response.  `RequestException` does `self.request = self.response.request`, so I figured that would be more consistent.

test_requests.py
```diff
@@ -188,6 +188,28 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
...
+        else:
+            pytest.fail()
```
munro: I didn't leave a message, because it seemed more consistent with the rest of the tests to just let the reason be implied.

munro:
> Got a few small notes, but this looks really good! :cake:

Yay!  Thanks a lot for the fast response! @sigmavirus24 @Lukasa

test_requests.py
sigmavirus24: Let's say "Expected redirect to raise TooManyRedirects but it did not" please?

test_requests.py
```diff
@@ -188,6 +188,28 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
         assert r.history[0].status_code == 302
         assert r.history[0].is_redirect
 
+    def test_HTTP_302_TOO_MANY_REDIRECTS(self, httpbin):
+        try:
+            requests.get(httpbin('redirect', '50'))
+        except TooManyRedirects as e:
+            assert '/relative-redirect/20' in e.request.url
```
sigmavirus24: I think this is a bit too specific. If we change what we use to test (or httpbin changes its paths... which it did a few months ago to introduce `/relative-redirect/:n` then this test will start failing.

test_requests.py
```diff
@@ -188,6 +188,28 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
...
+            assert e.request.url == e.response.url
```
sigmavirus24: I don't think this assertion helps anything.

test_requests.py
```diff
@@ -188,6 +188,28 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
...
+        else:
+            pytest.fail()
+
+    def test_HTTP_302_TOO_MANY_REDIRECTS_WITH_PARAMS(self, httpbin):
+        s = requests.session()
+        s.max_redirects = 5
+        try:
+            s.get(httpbin('redirect', '50'))
+        except TooManyRedirects as e:
+            assert '/relative-redirect/45' in e.request.url
```
sigmavirus24: Same concerns as above.

test_requests.py
```diff
@@ -188,6 +188,28 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
...
+            assert e.request.url == e.response.url
+            assert len(e.response.history) == 5
+        else:
+            pytest.fail()
```
sigmavirus24: This message might be slightly different, e.g., "Expected custom max number of redirects to be respected but was not."

test_requests.py
munro: I think it could help catch off by one logic where the previous redirect request is returned instead of the one used to get the last request.  I'll swap it for a more direct comparison instead.

sigmavirus24: On January 20, 2016 1:14:50 PM CST, Ryan Munro notifications@github.com wrote:

> > @@ -188,6 +188,28 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self,
> > httpbin):
> >          assert r.history[0].status_code == 302
> >          assert r.history[0].is_redirect
> > -    def test_HTTP_302_TOO_MANY_REDIRECTS(self, httpbin):
> > -        try:
> > -            requests.get(httpbin('redirect', '50'))
> > -        except TooManyRedirects as e:
> > -            assert '/relative-redirect/20' in e.request.url
> > -            assert e.request.url == e.response.url
> 
> I think it could help catch off by one logic where the previous
> redirect request is returned instead of the one used to get the last
> request.  I'll swap it for a more direct comparison instead.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/kennethreitz/requests/pull/2972/files#r50303029

## What about e.request is e.response.request? I still don't think this is necessary but that's a more clear check. Could also use a comment.

test_requests.py
```diff
@@ -188,6 +188,30 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
         assert r.history[0].status_code == 302
         assert r.history[0].is_redirect
 
+    def test_HTTP_302_TOO_MANY_REDIRECTS(self, httpbin):
+        try:
+            requests.get(httpbin('relative-redirect', '50'))
+        except TooManyRedirects as e:
+            url = httpbin('relative-redirect', '20')
```
munro, line 195: Checked some other tests, and found they were just using `httpbin` to generate URLs for assertion.  I think it's good that if the `httpbin` API changes, that this test will break, as opposed to silently failing.  Also it provides further testing that the redirection logic is working.

sigmavirus24:
LGTM. @Lukasa thoughts?

Lukasa:
LGTM.

PR #2973, opened 2016-01-20
Stop falling back to distutils in setup.py

sigmavirus24:
distutils has been deprecated for years now and pip is refusing to install
projects using it. As such, *some* users are reporting problems installing
requests. Let's avoid this entirely by not falling back to distutils at all.


Lukasa:
Still, we just don't need this fallback now.

PR #2976, opened 2016-01-22
Fix api doc of debug logging for Python 3

shoma:
`httplib` was renamed in Python 3.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -184,11 +184,13 @@ API Changes
       import requests
       import logging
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This then breaks for users on Python 2. Can you simply add comments explaining how to do this on Python 3 instead of removing the recipe for people using Python 2?

shoma:
@sigmavirus24 

Ok, I'll update.

Lukasa:

PR #2984, opened 2016-01-29
Remove int type on HTTPAdapter's max_retries argument doc

rbcarson:
The argument description says `max_retries` can be an int, or a `urllib3.util.Retry` object. However, the type declared in the doc restricts it to int only. Passing a Retry object causes a warning in PyCharm code inspection.

The behaviour described for `max_retries` is accurate: HTTPAdapter passes `max_retries` to `Retry.from_int()`, which just returns the input if `max_retries` is a Retry object.


kennethreitz:
Thanks!

PR #2991, opened 2016-02-02
[WIP] Test suite refactoring

Stranger6667:
Hello, I'd like to propose some refactorings for current test suite. First of all it is more about testing convenience, to make development process easier.

Current changes:
- Changed test running to setup.py. It allows to use some tuning on dependencies during test runs
- Updated classifiers list. All tests run fine on Python 2.6 and PyPy. And almost all on Jython. I think it will not be a problem to deal with 2 failing tests, but will see.
- Added config file for coverage. But I'm not sure about omitting "*/packages/*".
- Added tox.ini to run tests against different Python versions locally.
- Removed `unittest` from tests. It is not required for `py.test`. Also I've removed `_multiprocess_can_split_` because it is only related to `nose`.
- Added some of py.test parametrization.
- Replaced `assert False` with `pytest.fail`.
- Fixed typo and added some minor improvements.

Some thoughts and questions:
- Coverage report shows, that code [here](https://github.com/kennethreitz/requests/compare/master...Stranger6667:tests-refactoring?expand=1#diff-56c2d754173a4a158ce8f445834c8fe8R616) is not executed. So, is there problem in hooks execution?
- Split test module to some logically separated files. Current file is kind of huge. May be it will be better to have test suite in multiply files?
- Python 3.2. Basically all tests pass on Python 3.2 / PyPy3 (based on Python 3.2). But `httpbin` dependency is not compatible with this Python version. But `requests` seems to be compatible. Any plans to keep compatibility with Python 3.2? May be rewrite some tests for Python 3.2 or something?
- Jython. Most things are OK on Jython. Only 2 tests are failing - `TestRequests.test_pyopenssl_redirect` and `TestMorselToCookieExpires.test_expires_valid_str`. I don't know detail yet, but I hope, that it won't be a problem to fix it. Is it reasonable to run test suite against Jython? 
- Travis CI. I saw some CI-related lines in `Makefile`, but I don't see any info about in docs. Can you help me here? Also I can add some Travis CI config to run all configurations on it (including Jython)
- Public coverage report. May I add some `codecov/coveralls` integration? Will it be helpful?

It is not a complete lists and I'll be glad to hear what do you think about it. Also I'll be happy to write more tests here :)
Thanks!

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[run]
+omit = */packages/*
+
+[report]
+show_missing = true
+precision = 2
+exclude_lines = pytest.fail
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ nosetests.xml
 junit-report.xml
 pylint.txt
 toy.py
-tox.ini
 violations.pyflakes.txt
 cover/
 build/
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -162,4 +162,5 @@ Patches and Suggestions
 - Smiley Barry (`@smiley <https://github.com/smiley>`_)
 - Shagun Sodhani (`@shagunsodhani <https://github.com/shagunsodhani>`_)
 - Robin Linderborg (`@vienno <https://github.com/vienno>`_)
-- Brian Samek(`@bsamek <https://github.com/bsamek`_)
+- Brian Samek (`@bsamek <https://github.com/bsamek>`_)
+- Dmitry Dygalo (`@Stranger6667 <https://github.com/Stranger6667>`_)
--- a/Makefile
+++ b/Makefile
@@ -6,10 +6,10 @@ init:
 test:
 	# This runs all of the tests. To run an individual test, run py.test with
 	# the -k flag, like "py.test -k test_path_is_not_double_encoded"
-	py.test test_requests.py
+	python setup.py test -a "test_requests.py"
 
 coverage:
-	py.test --verbose --cov-report term --cov=requests test_requests.py
+	python setup.py test -a "--verbose --cov-report term --cov=requests test_requests.py"
 
 ci: init
 	py.test --junitxml=junit.xml
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,36 @@
 #!/usr/bin/env python
 
 import os
+import platform
 import re
 import sys
 
 from codecs import open
 
 from setuptools import setup
+from setuptools.command.test import test as TestCommand
+
+
+JYTHON = platform.system() == 'Java'
+
+
+class PyTest(TestCommand):
+    user_options = [('pytest-args=', 'a', "Arguments to pass into py.test")]
+
+    def initialize_options(self):
+        TestCommand.initialize_options(self)
+        self.pytest_args = []
+
+    def finalize_options(self):
+        TestCommand.finalize_options(self)
+        self.test_args = []
+        self.test_suite = True
+
+    def run_tests(self):
+        import pytest
+
+        errno = pytest.main(self.pytest_args)
+        sys.exit(errno)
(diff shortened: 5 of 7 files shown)


Lukasa:
Thanks for this! I'll handle answering some questions, but I know that @kennethreitz has got opinions about some of the changes you're making here so I'll let him weigh in to do the actual review work.

To your first question, I suspect that test is just poorly written.

To Python 3.2 support, requests explicitly considers Python 3.2 a best-effort platform. There's no reason to continue to support it formally, because Python 3.2 is no longer supported by the PSF: the last release was in 2014. The only reason we appear to be compatible with Python 3.2 is because pip needed to support it, and even pip is planning to drop support.

To Jython: we'd like to work on Jython, but it's moderately painful for us to get a CI suite for it. It's not unreasonable to run the tests against Jython though.

Re: Travis. We moved away from Travis because our test suite used to run against the real httpbin instance, and Travis' network access was sufficiently flaky that our tests failed a lot. We've therefore been using Jenkins for a while. However, now that our tests are entirely offline I'm interested in moving back to Travis. Of course, that's @kennethreitz's call.

For coverage, I see no reason to use anything but coverage.py, but again, this is @kennethreitz's area.

kennethreitz:
- The rewrite of `test_requests.py` looks great to me. I originally wanted our test suite to be fully compatible with unittest's test runner (and it originally was), but over time this slowly became less and less the case. 
- Jython is a very rarely used Python implementation. If someone fixes a Jython-specific bug in a way that doesn't complicate the Requests codebase, I'm happy for us to accept the patch. But, our level of support should (and does) stop there. 
- I don't think having `tox.ini` is very useful, as we won't be using it in any official capacity. Use of `tox.ini` requires that every specified Python installation be available in order to be used effectively, which is quite difficult to do properly on most systems. This project shipped `tox.ini` long ago, but it was removed for a number of reasons. Tox is a wonderful local testing tool, and anyone is free to use it with Requests — that doesn't mean we need to provide the file, though. It adds (yet) another file to the root of the repository, which I would like to keep as clean as possible.
- I have little opinion regarding coverage, except to echo @Lukasa's `coverage.py` statement. If we can get rid of the `.coveragerc` file, that would be great. Again, I'd like to keep those files to a minimum.

kennethreitz:
Regarding Travis: we used to utilize Travis, but I quickly grew frustrated with it reporting false negatives for our test suite on a regular basis. I was also frustrated by them randomly removing a few older versions of Python one day, as well. I was also frustrated by all the email notifications it sent me (many regarding forks of kennethreitz/requests). I'm also not a fan of its UI. :)

That was a long time ago, and I'm sure it's improved quite a bit since then. 

Our current CI integration with my Jenkins server seems to be suiting our needs perfectly fine. I personally quite enjoy having a private Jenkins server. However, I wouldn't be against considering a switch back to Travis if there was an obvious benefit to myself, @Lukasa, or @sigmavirus24. I'm not aware of any. Are there any?

kennethreitz:
Merged. Made some changes. 
- Removed `.coveragerc` (this may be useful to add later).
- Removed `tox.ini`, see above.
- Changed the recommended way to run tests (`$ make tests`) back to using `$ py.test test_requests.py` (instead of `$ python setup.py test`). Why? Because this method outputs many lines of noise before running the suite, and creates/requires a `requests.egg-info` directory to be present in my development environment, which I do not need nor want. 
- Removed references to Jython in `setup.py`.

Hope none of that sounded critical! I just wanted to document my perspective behind the decisions. Thanks so much for this patch! The improvements to the test suite are great. 

:sparkles: :cake: :sparkles:

Stranger6667:
Thank you for the review and merge! I'm happy to contribute :)
I still have some questions about tests, but I'll create new issues for that.

PR #2992, opened 2016-02-03
Use TTLCache to cache proxy_bypass.

lazywei:
As mentioned in https://github.com/kennethreitz/requests/issues/2988, we might leverage the LRU cache to speed up the `proxy_bypass`, which might cause unexpected delay.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -12,6 +12,7 @@ that are also useful for external consumption.
 import cgi
 import codecs
 import collections
+import functools
 import io
 import os
 import platform
@@ -19,6 +20,7 @@ import re
 import sys
 import socket
 import struct
+import time
 import warnings
 
 from . import __version__
@@ -548,7 +550,8 @@ def should_bypass_proxies(url):
     # exceptions we've seen, though: this call failing in other ways can reveal
     # legitimate problems.
     try:
-        bypass = proxy_bypass(netloc)
+        invalid_indicator = int(time.time() / 60)
+        bypass = _cached_proxy_bypass(netloc, invalid_indicator)
     except (TypeError, socket.gaierror):
         bypass = False
 
@@ -556,6 +559,11 @@ def should_bypass_proxies(url):
         return True
 
     return False
+
+@functools.lru_cache(maxsize=64)
+def _cached_proxy_bypass(netloc, _):
+    """Use LRU to cache the proxy_bypass."""
+    return proxy_bypass(netloc)
 
 def get_environ_proxies(url):
     """Return a dict of environment proxies."""
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -549,6 +549,13 @@ def should_bypass_proxies(url):
     # of Python 2.6, so allow this call to fail. Only catch the specific
     # exceptions we've seen, though: this call failing in other ways can reveal
     # legitimate problems.
+    #
+    # Additionally, the proxy_bypass function might cause unexpected delay when
+    # it called socket.gethostbyaddr internally (refer to #2988). Thus we cache
+    # the function call. Also, to invalidate the cache from time to time (as we
+    # don't know when to expire the cache exactly), we use an invalid_indicator
+    # here, which will be different for every 60 seconds, to expire the cache
+    # even with the same netloc.
     try:
         invalid_indicator = int(time.time() / 60)
         bypass = _cached_proxy_bypass(netloc, invalid_indicator)
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -12,7 +12,6 @@ that are also useful for external consumption.
 import cgi
 import codecs
 import collections
-import functools
 import io
 import os
 import platform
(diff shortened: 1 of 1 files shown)


requests/utils.py
```diff
@@ -548,7 +550,8 @@ def should_bypass_proxies(url):
     # exceptions we've seen, though: this call failing in other ways can reveal
     # legitimate problems.
     try:
-        bypass = proxy_bypass(netloc)
+        invalid_indicator = int(time.time() / 60)
+        bypass = _cached_proxy_bypass(netloc, invalid_indicator)
```
Lukasa: So I think with this approach we _definitely_ need to comment what we're doing here, to make sure that people coming down the line later understand what's going on.

I'd also love it if we tested this, though I'm not immediately sure how best to do that.

Lukasa:
The bigger problem here is that `functools.lru_cache` is not available in Python 2. It should be fairly trivial to lift the code into a backport module and vendor it though.

Lukasa:
For general code review I'm not best placed to do it because I suggested the approach to begin with, so let's tag in @sigmavirus24 and @kennethreitz as well.

requests/utils.py
sigmavirus24: Why is the calculation for invalid_indicator inside a try block catching an exception that is completely unrelated?

lazywei:
OK. I'll add the doc latter. As for the backport for LRU cache, do you think it should reside in `compat.py` or in a separate module?

Lukasa:
I think it should go under `packages`.

requests/utils.py
lazywei: You're right, I should move it outside of the try block to make the exception catching simple.
However, since we wrap the `proxy_bypass` by `_cached_proxy_bypass`, do you think we should put the try block here? Or should we put it inside the wrapper function?

schlamar:
> The bigger problem here is that functools.lru_cache is not available in Python 2. It should be fairly trivial to lift the code into a backport module and vendor it though.

IMO you should invalidate single items in the cache based on the timestamp added, so you would need a custom cache function anyway...

lazywei:
@Lukasa @schlamar I agree I'd need to make a custom cache function. Do you think it is appropriate to make it a standalone package and vendor it under `packages`?

Lukasa:
I do. =)

lazywei:
I just found there is a 3rd party package that can do the cache for us: https://github.com/tkem/cachetools
Specifically, we can use TTL cache, instead of making a timestamp invalid_indicator ourself.
How do you think?

Lukasa:
Requests doesn't like bringing in third-party modules as dependencies if we can avoid it.

lazywei:
I thought if I make a standalone package it would also be considered third party, right? Then would it make any difference for using another caching tool?
Thanks

Lukasa:
> I thought if I make a standalone package it would also be considered third party, right? Then would it make any difference for using another caching tool?

Sorry, I'm not sure that I understand what you're proposing here.

schlamar:
I think a simple TTL cache can be implemented in 50-80 LOC or even less so there is no need to vendor a new standalone package nor another third-party module. I would just put it into `requests.utils`.

lazywei:
sure, no problem. will do it in utils.
Marc Schlaich notifications@github.com於 2016年3月21日 週一，下午7:24寫道：

kennethreitz:
This pull request has been open for over two months now. Let's do what we can to get it merged or closed :)

lazywei:
To remember which functions have been cached, I'd need to store it somewhere.
Which approach, using a scope (under utils.py) variable or using a Class's instance variable, would you prefer?

lazywei:
@kennethreitz @Lukasa Sorry for taking so long. I implemented a naive TTLCache for caching the `proxy_bypass` function call. Please have a look at it and let me know your thoughts.
If there is no big problem with the implementation, I'll add some documentation and testing if necessary later.

Thanks.

requests/utils.py
```diff
@@ -38,6 +39,92 @@
...
+        return _ListNode(key=None, val=None, expired_at=None)
+
+
+class TTLCache(object):
+    def __init__(self, ttl=3600, maxitems=64):
+        self.ttl = ttl
+        self.key_to_node = {}
+        self.maxitems = maxitems
+
+        self.list_head = _ListNode.dummy_node()
```
schlamar, line 78: I'm not sure if the complexity of this linked list is worth its benefits. I'd rather prefer a linear search for the oldest entry when `maxitems` is reached. But please wait for other opinions before you change anything. 

Or use at least a `collections.deque` instead of a linked list, where you can `popleft` and `append` (right) with O(1), too.

requests/utils.py
```diff
@@ -541,14 +628,23 @@ def should_bypass_proxies(url):
...
+    # Additionally, the proxy_bypass function might cause unexpected delay when
+    # it called socket.gethostbyaddr internally (refer to #2988). Thus we cache
+    # the function call. Also, to invalidate the cache from time to time (as we
+    # don't know when to expire the cache exactly), we use the TTLCache
+    # mechanism here.
     try:
-        bypass = proxy_bypass(netloc)
+        bypass = proxy_bypass_cacher.get(netloc)
+        if bypass is None:
+            bypass = proxy_bypass_cacher.set(netloc, proxy_bypass(netloc))
```
schlamar, line 647: `set` doesn't return the value so this won't work.

```
bypass = proxy_bypass(netloc)
proxy_bypass_cacher.set(netloc, bypass)
```

requests/utils.py
```diff
@@ -719,3 +815,6 @@ def urldefragauth(url):
     netloc = netloc.rsplit('@', 1)[-1]
 
     return urlunparse((scheme, netloc, path, params, query, ''))
+
+def ttl_cache(func, ttl, maxsize=64):
```
schlamar, line 819: This is not used?!

schlamar:
I reviewed it. A second opinion would be nice.

requests/utils.py
```diff
@@ -38,6 +39,92 @@
 DEFAULT_CA_BUNDLE_PATH = certs.where()
 
 
+class _ListNode(object):
```
sigmavirus24, line 42: This also doesn't belong in requests/utils.py as far as I'm concerned. We have a separate file for internal things like this, I think structs.py

kennethreitz:

PR #2993, opened 2016-02-03
Changes to support sending a complete response object when 407 errors…

angelcaido19:
… is found while doing the CONNECT to proxy. Added a way to stablish a connection using connect over an already connected socket

Trying to connect to a "https" web page using a proxy uses a "HTTP CONNECT" verb to tell the proxy that we want to connect to a https destination.
With "basic authentication" the "CONNECT" is sent with the appropriate headers so the proxy can authenticate the request and do the request to the https web server. (That headers are setup by the requests library)
With "NTLM" we use a plugin (requests-ntlm) that is set as a hook and it is called on every request response. For example: with HTTP, the requests library sends a "GET" to a web page, as a response we get a HTTP-407 that is handled by the plugin hook to do the NTLM negotiation and authentication
In a HTTPS scenario, the first connection to the proxy is being done with the "CONNECT". At that time, requests does not set any authentication header (because it is handled by a plugin) and we get an 407 authentication Error.

This Patch tries to avoid raising an exception on "Proxy Authentication" errors and it sends back a complete http response so it can be handled by the requests-ntlm plugin and reuse an already connected and authenticated socket to do the connection

### DIFF
--- a/requests/packages/urllib3/connection.py
+++ b/requests/packages/urllib3/connection.py
@@ -10,9 +10,11 @@ from .packages import six
 try:  # Python 3
     from http.client import HTTPConnection as _HTTPConnection
     from http.client import HTTPException  # noqa: unused in this module
+    from http.client import PROXY_AUTHENTICATION_REQUIRED
 except ImportError:
     from httplib import HTTPConnection as _HTTPConnection
     from httplib import HTTPException  # noqa: unused in this module
+    from httplib import PROXY_AUTHENTICATION_REQUIRED
 
 try:  # Compiled with SSL?
     import ssl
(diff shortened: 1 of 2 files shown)


Lukasa:
@angelcaido19 Sorry, but this includes changes in urllib3. You'll need to propose this patch over there first, I'm afraid. =(

angelcaido19:
@Lukasa Hey thanks. Help me with this: Is this "https://github.com/shazow/urllib3" the oficial urllib3 repo?
Thanks again!!

Lukasa:
Ah, sorry I didn't link it: yeah it is!

PR #3003, opened 2016-02-09
Added pprint() function to Response and PreparedRequest objects for human-reading goodness

7uice:
This feature is aimed at making debugging/following HTTP traffic easier on a developer. Many times in the past, I would send Requests' traffic through Burp proxy in order to have a visual representation of my HTTP traffic. With this pull request, a developer can simply call pprint() on their HTTP requests to see a Burp-like view of the traffic.

Example usage:
```
import requests

r = requests.get("http://www.praetorian.com")
r.request.pprint()
r.pprint(body=False)
```

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -524,6 +524,22 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         for event in hooks:
             self.register_hook(event, hooks[event])
 
+    def pprint(self, body=True):
+        """Prints the HTTP traffic of `PreparedRequest` object in 
+        a human-readable format
+
+        :param body: If True, prints the body of the response
+
+        """
+
+        s = ""
+        s += "%s %s HTTP/1.1" % (self.method, self.path_url)
+        for header,value in self.headers.items():
+            s += "\n%s: %s" % (header, value)
+        if self.body is not None and body:
+            s += "\n%s\n" % self.body
+        print(s)
+
 
 class Response(object):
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This already exists (to a degree) in the requests-toolbelt: https://toolbelt.readthedocs.org/en/latest/dumputils.html

mtj42:
My major problem with the dump utility is that it is not as granular as pprint(), which gives a developer the ability to dump a single request or a response without the body attached (good for when the response is a large static page). The dump util returns a lot more data than is usually needed while monitoring Requests traffic.

Lukasa:
@7uice Presumably we could add flags to `dump` to resolve that problem though?

Generally speaking I'm in favour of keeping this kind of code out of requests: it becomes one more thing we have to maintain.

gitpraetorianlabs:
@Lukasa It looks like this could be accomplished using utils.dump by adding a `print_body=True` flag and wrapping the following lines in a conditional statement:

`line 78: bytearr.extend(prefix + _coerce_to_bytes(request.body))`

`line 106: bytearr.extend(response.content)`

This would allow a user to print Request & Response while not flooding their screen with data returned in the response. If it's preferred to keep this out of Requests, I can move the code over to the utils.dump code.

Lukasa:
I think that would be better for now. =) At some time that may make its way over here some time in the future. =)

mtj42:
Sounds good :) Closing this for now and moving the feature into utils.dump.

PR #3005, opened 2016-02-11
Use time.monotonic() for tracking elapsed time where available

scop:


sigmavirus24:
Hey @scop I was wondering if you could provide detail about why this is desirable and why we should change the existing behaviour as well as a description of when/where monotonic will be available and where/when it won't.

scop:
Most of the requested details should be answered by the `time.monotonic` docs: https://docs.python.org/3/library/time.html#time.monotonic

In a nutshell, this change makes `r.elapsed` unaffected by system clock changes where `time.monotonic` is available. There are no changes to behavior on setups where it isn't. It was introduced in Python 3.3. In 3.3 and 3.4 it wasn't available on all platforms, but in 3.5 it is.

sigmavirus24:
Have you run into issues with the elapsed attribute that this would help with?

scop:
No, but I have elsewhere with similar things, just thought it'd be pretty much a no-brainer to fix in requests as well.

requests/sessions.py
```diff
@@ -10,7 +10,11 @@
 """
 import os
 from collections import Mapping
-from datetime import datetime
+from datetime import timedelta
+try:
+    from time import monotonic
+except:
```
kennethreitz, line 16: `except ImportError`

kennethreitz, line 16: This code should be moved into https://github.com/kennethreitz/requests/blob/master/requests/compat.py

kennethreitz:
I don't think this is anything we need to be concerned about, and it adds unnecessary complexity to one of our simplest (and least used) features. I understand the thought process behind this PR, and if we were all using 3.5, then I'd probably accept it (with a few improvements).

PR #3014, opened 2016-02-14
PreparedRequest.dump()

kennethreitz:
```
>>> r = requests.get('http://httpbin.org/ip')

>>> print r.request.dump()
REQUESTS/2.9.1 GET http://httpbin.org/ip
Connection: keep-alive
Accept-Encoding: gzip, deflate
Accept: */*
User-Agent: python-requests/2.9.1

### CODE REVIEW

**Comment by kennethreitz** [comment]
I think `render` is a good name if it was reporting `HTTP/1.1` in the strings. Another name may be more appropriate now.

**Comment by kennethreitz** [comment]
I find it entertaining how this showcases just how much we completely disregard header order.

**Comment by kennethreitz** [comment]
Thinking of removing Response's render.

**Comment by Lukasa** [comment]
Maybe `dump()`?

**Comment by kennethreitz** [comment]
@Lukasa hmmm, that works.

**Comment by EmilStenstrom** [comment]
This looks exactly like what I was hoping for.

A couple of comments (feel free to disregard and ship anyway, I'm already happy):
1. I'm not sure about the name **render()**. The method feels like a variant of `__str__` but with all the gory details. I expect it to be used mainly for debugging, so maybe just debug() instead?
2. I really would like both of them (Response.render() and Request.render()). For debugging purposes I often need both the sent request and the response. I then take both of them and send them in an e-mail to the one with the server, so they can debug why my requests are not working right. Without a way to render responses I still would have to have my own utility for that.
3. Not that fond of the "REQUESTS/2.9.1" string. Maybe we could do PreparedRequest<[string_here]> as a way to mimic `__repr__`?

**Comment by kennethreitz** [comment]
@EmilStenstrom `render` has been renamed to `dump`.

**Comment by EmilStenstrom** [comment]
dump() works for me!

File: requests/models.py [hunk 1]
```diff
@@ -524,6 +525,20 @@ def prepare_hooks(self, hooks):
...
+        useful for debugging."""
+
+        r = ('REQUESTS/{version} {method} {url}\n{headers}'.format(
+            version=__version__,
+            method=self.method,
+            url=self.url,
+            headers='\n'.join('{}: {}'.format(k, v) for k, v in self.headers.items()),
+        ))
+        if body and self.body:
+            r += '\n\n{body}'.format(body=self.body)
```
sigmavirus24: What if body is a file-like object? This won't render the way you expect.

Lukasa: The same is true for generators and other weird iterators.

sigmavirus24: That is to say, it doesn't make sense to handle this differently from how the toolbelt already handles this. We puzzled out how to do this properly over there months ago and the code should be liftable to avoid having to reinvent the wheel and catch the same problems.

sigmavirus24:
> Not that fond of the "REQUESTS/2.9.1" string. Maybe we could do PreparedRequest<[string_here]> as a way to mimic **repr**?

I agree with @EmilStenstrom. I don't like that much either. It's confusing and may make some people think we're doing some weird protocol instead of HTTP.

requests/models.py
kennethreitz: Currently, they're represented by their repr, which seems like a perfectly fine solution to me. What behavior does the toolbelt exhibit in that scenario?

EmilStenstrom:
Concrete suggestion of syntax for dump():

``` python
>>> import requests
>>> r = requests.get('http://httpbin.org/ip')
>>> r.request
<PreparedRequest [GET]>
>>> r.request.dump()
<PreparedRequest [
    GET http://httpbin.org/ip
    Connection: keep-alive
    Accept-Encoding: gzip, deflate
    Accept: */*
    User-Agent: python-requests/2.9.1
]>
```

EmilStenstrom:
Any rationale for removing Response.dump()? I find it really useful for debugging a remote service in the same way that Request.dump() is useful.

kennethreitz:
@EmilStenstrom hmm, that's not a bad way to present it! I like it. 

For the record, I'm still not 100% on including this. I'm still tossing the idea back and forth in my mind. This syntax definitely helps, though.

This syntax also helps support the inclusion of `Response.dump`. It just didn't make sense with the other syntax.

kennethreitz:
@EmilStenstrom updated the syntax. 

```
<PreparedRequest [
    POST http://httpbin.org/post
    User-Agent: python-requests/2.9.1
    Connection: keep-alive
    Accept-Encoding: gzip, deflate
    Accept: */*
    Content-Length: 3447
    Content-Type: multipart/form-data; boundary=bf96155646964bf082b7a30531aa7c7e
]>
```

kennethreitz:
Using colors to display warning about non string-like content bodies now. I like where this is going. 

screen shot 2016-02-16 at 1 12 22 am

Things to do:
- [x] disable color output when not a tty
- [x] potentially color the rest of the output, not sure how i feel about that.

kennethreitz:
Ooooh, it's so pretty (and cross-platform).

screen shot 2016-02-16 at 1 51 13 am

I dig this.

EmilStenstrom:
I'm very happy with how this turned out, and by also including Response.dump() all three of my concerns have been delt with. It's ready to ship in my mind.

kennethreitz:
@EmilStenstrom as soon as I'm done perfecting `PreparedRequest`, I'll move on to `Response` :)

kennethreitz:
Two methods are now included, `dump()`, and `dumps()`. 

_dumps_ is for people that want to grab a string representation of the dump and use it programatically. _dump_ prints the dump directly to the screen, in colorful glory. Colors can optionally be disabled by passing `colored=False`.

kennethreitz:
Reponse is done! 

screen shot 2016-02-16 at 3 13 46 am

I think this is ready to ship.

kennethreitz:
@EmilStenstrom Response is now included (see above). :cake:

Thanks a lot for your help with this, you've been instrumental ;)

kennethreitz:
@sigmavirus24 @Lukasa last call for comments. Will merge/release tomorrow.

Lukasa:
I have a preference to remove the functionality to dump colours, rather than vendor in two more modules, but otherwise this is fine by me. =)

sigmavirus24:
Yeah. So, I agree with @Lukasa whole-heartedly. I don't think we should be vendoring libraries just for colorizing this output. This feels like a vast step outside the domain of this library - HTTP. That kind of behaviour belongs elsewhere. We also don't even remotely attempt to reflect whether or not the user is attempting to use proxies, which I think will be far more confusing for those users when they expect to see that output here.

To be clear:
- I think we're violating the law of constraints of limiting this library to HTTP
- I think we're doing our users a grave disservice by having this functionality here in such a naive implementation
- I think we're adding more (vendored) dependencies than we should ever rightfully need

kennethreitz:
If I hear the word _vendored_ from either one of you again, you're fired :)

But in all seriousness, thoughts on colorization noted. I quite like it, and am not 100% on it yet myself (but am leaning towards yes). 

Now, the purpose of these methods are to provide a convenient way to print out the state of a `PreparedRequest` and a `Response` when building applications by hand (I typically resort to printing `PreparedRequest.__dict__`, which is terrible. Proxies are not involved at this point, and no user should expect any proxy information to be present at this point. The output clearly shows that we are representing the `PreparedRequest` object, and they can look that up in the documentation if they get confused.

sigmavirus24:
> Proxies are not involved at this point, and no user should expect any proxy information to be present at this point. 

I disagree. The usage you showed above (reproduced below for clarity)

``` py
>>> r = requests.get(url)
>>> r.request.dump()
...
```

Is a request that has been sent. Requests has no way to distinguish that, realistically, but the user knows it has been sent and will expect that information. I know it's hard for us to easily distinguish between a PreparedRequest that someone has made using the PreparedRequest flow and one that was already sent, but users tend to not care if we can determine that or not. Introducing this here will open us up to a host of bug reports about not including various pieces of information in the dump.

> If I hear the word vendored from either one of you again, you're fired :)

Yeah, I'm just thinking of two groups of people:
- Developers who closely follow requests development and are constantly looking to pick fights with us over different choices (vendoring, certifi vs system bundles, etc.)
- Downstream redistributors

The former I don't particularly care about because they generally include people whose opinion I already deprioritize (for various reasons). The latter I'm not worried too much about in this case because colorama is almost certainly already packaged in their repositories. What I care about in that case is when someone does `{package-manager-install} python-requests` and suddenly sees a few new dependencies that look completely unrelated to requests. I think the inclusion of those libraries should be a large warning sign.

And now I'll shut up ;)

kennethreitz:
> Is a request that has been sent. Requests has no way to distinguish that, realistically, but the user knows it has been sent and will expect that information. I know it's hard for us to easily distinguish between a PreparedRequest that someone has made using the PreparedRequest flow and one that was already sent, but users tend to not care if we can determine that or not. Introducing this here will open us up to a host of bug reports about not including various pieces of information in the dump.

That information would be expected to be present in the headers, right? If so, then anyone manually printing `r.request.headers` today would be equally confused. I don't see that to be the case. Perhaps this is just surfacing a bug that `r.request` should include that information (I don't necessarily think that it should).

sigmavirus24:
@kennethreitz the proxy information is tricky. We have a hacky way of incorporating it into the information in the toolbelt, but I don't think this implementation would have the same ability to determine if a proxy is in use.

kennethreitz:
I have no concern for proxy representation here, but if you're concerned for user confusion, we can make a special note in the documentation. `PreparedRequests` and `Response` have no information or knowledge of proxies, and the `dump()` of their contents has no need to represent such information.

sigmavirus24:
If we're going to document as notes everything that might confuse a person with these implementations, those notes will probably outweigh the documentation of the feature itself.

kennethreitz:
I see nothing here that would be confusing to anyone.

EmilStenstrom:
The release notes will be soooo nice with those colorful outputs!

kennethreitz:
Methods are now simply `PreparedRequest.pretty()` and `Response.pretty()`.

sigmavirus24:
Method names aside, I'm still -1 on this. I think we're destroying the line between an HTTP library and an HTTP library with a lot of whizbang features. We've rejected other feature requests in the past because we're only an HTTP library and I think this will open the door for those to be re-evaluated on a case-by-case basis.

StewPoll:
I'm just gonna jump in here with this. I've not contributed to this in the past, but from what I can see this would make an easy way for developers like myself who use Requests to log the requests they make and the response they get. While I don't care for the colourisation of the text when you dump it, it does add something for any CLI libraries that use Requests. (Edge cases I know) 

It is a whizbang feature, but it is a useful whizbang that would save other developers having to develop their own way of getting this information to log it if needed.

The colourisation and vendoring of another package shouldn't be a reason to not include the feature entirely. Although the fact that colouring requires another package would be a good reason to not include the colouring. Maybe just have it as an optional feature?

kennethreitz:
@TetraEtc thanks for your thoughts!

I think the colors are enjoyable and make the information easier to parse. Why do you not care for them? Just preference?

StewPoll:
If you're viewing it using a CLI the colours would definitely help, in the same way that syntax highlighting helps in your IDE.

If I were to use this I'd mainly use the string representation (what you were calling dumps, not sure what you're calling it now with pretty) and saving that to a log file on the off chance that I needed to debug something.

kennethreitz:
@TetraEtc for that, there is `pretty(colors=False)`. Also, colors are automatically disabled if the code is not running in an interactive console.

sigmavirus24:
> I'm just gonna jump in here with this. I've not contributed to this in
> the past, but from what I can see this would make an easy way for
> developers like myself who use Requests to log the requests they make
> and the response they get. While I don't care for the colourisation of
> the text when you dump it, it does add something for any CLI libraries
> that use Requests. (Edge cases I know) 
> 
> It is a whizbang feature, but it is a useful whizbang that would save
> other developers having to develop their own way of getting this
> information to log it if needed.

This is already the case. There is code you can easily utilize to do this, albeit in a more accurate and potentially more verbose way.

StewPoll:
> TetraEtc for that, there is pretty(colors=False). Also, colors are automatically disabled if the code is not running in an interactive console.

That's why I don't care that much either way. It's not something I'd use much. I care about the entire feature, not the colouring specifically

StewPoll:
> This is already the case. There is code you can easily utilize to do this, albeit in a more accurate and potentially more verbose way.

It's entirely possible I'm simply just not aware of how to do this, but this appears to be an incredibly easy way of doing this when needed.

requests/models.py
```diff
@@ -849,3 +917,62 @@ def close(self):
...
+            key = colored.green(key).color_str
+            value = colored.cyan(value).color_str
+
+            # Add the formatted header to our collection.
+            header = '{0}: {1}'.format(key, value)
+            dump_.append(header)
+
+        # Prepare body for dump.
+        if body:
+            # Add an extra newline; gotta keep 'em seperated!
```
alexwlchan, line 955: Nitpick: "seperated" ~> "separated"

sigmavirus24:
> It's entirely possible I'm simply just not aware of how to do this, but this appears to be an incredibly easy way of doing this when needed.

https://toolbelt.readthedocs.org/en/latest/dumputils.html

requests/models.py
kennethreitz, line 955: Thanks!

sigmavirus24:
Also worth noting that we're re-implementing part of https://github.com/jkbrzt/httpie

kennethreitz:
Random thoughts:
- it's so pretty!
- i have absolutely no problem including this, in any way shape or form
- however, requests is an intensely massive project, so "playing it safe" is probably best
- this annoys me, and i'm not sure how I feel about it
- therefore, I may do it anyway
- therefore, I may not
- my cat loves boxes

No further :+1:s or :-1:s needed at this point, but random anecdotes are more than welcome!

ghost:
My cat loves leather chairs. Doesn't like boxes at all.

EmilStenstrom:
When thinking about this, know that there are two different features included here: 
1. Dumping a request or response as a string
2. Making said string pretty.
I think accepting 1 should be considered playing it safe.

kennethreitz:
@EmilStenstrom absolutely, that'll happen no matter what. :)

(unless I change my mind, for some reason)

kennethreitz:
I don't actually think including the pretty formatting matters in any way, shape, or form, to be completely honest. Meaning, I see absolutely no reason not to. I think it's a nice feature, and I would be very pleasantly surprised to find that level of polish in my favorite HTTP library. 

That being said, I'm taking extra time to be considerate of @Lukasa and @sigmavirus24's objections (regardless of final result).

kennethreitz:
My cat also loves hiding inside my recliner.

sigmavirus24:
> I would be very pleasantly surprised to find that level of polish in my favorite HTTP library.

And I'd be very confused why an HTTP library is doing anything with colorized output. `¯\_(ツ)_/¯` I think we have different definitions of the word "polish" w/r/t what a library should do.

> I'm taking extra time to be considerate of @Lukasa and @sigmavirus24's objections (regardless of final result). 

Sounds like you're going to merge this regardless of our objections? If so, I'm not sure why you would take extra time.

alexwlchan:
I like the idea of pretty-printing request bodies, but I’m uneasy about colorised output, at least in the core library.

A few disconnected thoughts:
- This seems like something that might be useful when I’m in an interactive session (colorisation and all). But most of the time I’m not in an interactive session; requests is part of a larger application and is just one piece of the whole. It churns away in the background; nobody is inspecting the raw HTTP requests.
  
  Pretty-printing is less useful there. I’m not going to log the details of every HTTP header request.
- I tend to agree with others that this pretty-printing, especially colorisation, feels a bit odd in a library for handling HTTP requests.
  
  Which is a shame, because it seems like something I’d find useful on occasion – but it doesn’t feel like a core function.
- Might it be worth pushing into requests_toolbelt instead (at least the colorisation aspect)? Seem like that might be a better home for this.
- With my corporate hat on, vendoring in an extra module will probably cause hassle in environments where you need legal sign-off for new third-party software (especially since colorama has a different license to requests). I would rather avoid that hassle – especially if it’s for a feature that’s orthogonal to core HTTP, and which will never actually be used!
- My cat loves pawing at cardboard flaps, especially the ones you get on Amazon book packaging.

kennethreitz:
@sigmavirus24 if I was sure, I would have merged it a week ago :)

kennethreitz:
e.g. I am currently leaning towards inclusion, while swaying back and forth. Simply using this thread to think out loud.

kennethreitz:
My cat enjoys long naps on flannel blankets on couches.

kennethreitz:
My 6-year-old goldfish is currently living in a small man-made pond at my ex-girlfriend's house. 

I regret that decision :P

sigmavirus24:
I avoid adopting pets because open source projects need just as much love and are just as alive (or so I try to convince myself). ;)

Lukasa:
So, while we're taking feedback, here's my 2¢:

The philosophy of this project since about v1.0, at least as I understand it, has been to aggressively resist scope creep. We have rejected innumerably more feature requests than we've accepted, saying "no" to almost all feature requests that failed to meet one of three criteria:
1. It's extremely difficult or impossible to implement the feature from outside the library.
2. It's a convenience feature that would be used by a substantial majority of our users.
3. There is some subtlety in "correctness" of the feature that many users would miss, and that we have an opportunity to get right for them.

As far as I can see this feature fits into none of those categories.

To the first point, it is not difficult to implement this feature from outside Requests: it is exactly as hard there as it is here. The objects being used here are public and are regularly exposed to users: there's nothing that users would ordinarily be unable to find or struggle to reach.

To the second point, this feature would not be used by a majority of our users. The colorizing feature in particular suffers here because it only works when printing directly to a terminal and having a user observe those responses. That is a vanishingly small use-case of requests compared to automated use, and while it's important that those users have a good experience, I don't see value in implementing something solely for them unless it meets criteria 1 or 3. The dumping feature, even ignoring colorizing, is also something that is unlikely to be used by a substantial majority of our users, particularly because it is thoroughly ill-suited to logfiles. This is for two reasons: first, the dump contains newlines which make parsing logfiles extremely unpleasant; and second, the dump is really quite verbose and contains a lot of extraneous information that will serve only to bloat log files.

To the third point, the only subtlety in correctness here is not dealt with by this code, it is _encouraged_ by this code, and that is the fact that the representation dumped here is not the representation as received from or sent to the network. It is _related_ to those representations, but in no way conforms to them. Users who do not grasp this subtlety are, in my opinion, likely to be confused by the representation used here. The example used in the original post is great for this:

```
REQUESTS/2.9.1 GET http://httpbin.org/ip
Connection: keep-alive
Accept-Encoding: gzip, deflate
Accept: */*
User-Agent: python-requests/2.9.1
```

In this example the `Host` header is missing entirely, and the request URI is not what is actually sent on the network (that would be `/ip`, unless a proxy is in use in which case it _is_ what would be sent on the network, unless we're making a HTTPS requests and then there's a whole separate web request that we aren't showing at all here). I should note also that the REQUESTS/2.9.1 block is not in the same place as the HTTP/1.1 block would be in a request, though I don't know if that's intentional or not. The `Response` suffers even more from this because the response headers have been thoroughly transformed by the time they get to requests and frequently look almost nothing like they did when they were received from the network.

All of that goes to say that I don't believe we'd accept this feature if it came from outside the project. I think it represents a maintenance burden for the future, because once we implement this we'll need to support it, and I don't believe that it pays for that burden in utility. And I haven't even begun to mention that to implement this feature we are adding two implicit dependencies that will now require updates in order to fix bugs that affect this rendering and that will create substantial work downstream of us to unbundle.

Generally speaking we have been fairly aggressive about getting features _out_ of this library that are not required. I think this represents a step in the other direction, and I don't believe that we gain enough by doing it to justify it.

Of course, @kennethreitz, this is still your project and so my opinion is only that. =) If you merge this, I'll help maintain it.

kennethreitz:
@Lukasa failurel! The above post stated:

> No further :+1:s or :-1:s needed at this point, but random anecdotes are more than welcome!

My cat won't leave me alone today.

jwg4:
It seems to me that '2. It's a convenience feature that would be used by a substantial majority of our users.' is a no-brainer. Every Requests user who makes mistakes would like to see this output in a debugger.

sigmavirus24:
Thanks for playing @jwg4 but the instructions in this thread clearly state that only random anecdotes are welcome at this point.

kennethreitz:
You have a cool avatar @jwg4.

piotr-dobrogost:
https://en.wikipedia.org/wiki/Sierpinski_triangle

justanr:
I'd like to see this included, right now I'm logging requests with a rather naive:

``` python
# inside of session
def send(self, request, ...):
    self._logger(self._level, "Sending request: {}".format(pformat(request.__dict__))
    resp = super().send(request, ...)
    self._logger(self._level, "Got back: {}".format(pformat(resp.__dict__))
    return resp
```

As for the cat tax, my cat enjoys chasing my dog off her bed.

kennethreitz:
Closing, for now. I still think this is useful, and the branch still exists. But, something just doesn't feel right.

PR #3020, opened 2016-02-17
Fix #3017: Whitepace characters surrounding a URL should be ignored

### LINKED ISSUE #3017: Whitepace characters surrounding a URL should be ignored
It might be beneficial to unite the behavior with other tools (including browsers) and start ignoring whitespaces surrounding a URL. See the examples:

```
$ curl -s "https://github.com/ " -o /dev/null -w "%{http_code}"; echo
200

$ python -c 'import urllib2; print(urllib2.urlopen("https://github.com/ ").getcode())'
200

$ python -c 'import requests; print(requests.get("https://github.com/ ").status_code)'
404
```

Also, W3C [defines](https://www.w3.org/TR/2012/WD-html5-20121025/links.html#attr-hyperlink-href) the `a` tag's `href` attribute as...

> The href attribute on a and area elements must have a value that is a valid URL potentially surrounded by spaces.

Where _URL potentially surrounded by spaces_ is [defined](https://www.w3.org/TR/2012/WD-html5-20121025/urls.html#valid-url-potentially-surrounded-by-spaces) as...

> A string is a valid non-empty URL potentially surrounded by spaces if, after stripping leading and trailing whitespace from it, it is a valid non-empty URL.

I know this is not a killer argument for the requests library but it would be beneficial if it behaved the same way as the other approaches. It would also make HTML processing with requests (and e.g. beautifulsoup4) easier.

geckon:
My attempt to fix [the issue I reported](https://github.com/kennethreitz/requests/issues/3017). Feel free to comment my changes, especially if you don't agree with them.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -164,3 +164,4 @@ Patches and Suggestions
 - Robin Linderborg (`@vienno <https://github.com/vienno>`_)
 - Brian Samek (`@bsamek <https://github.com/bsamek>`_)
 - Dmitry Dygalo (`@Stranger6667 <https://github.com/Stranger6667>`_)
+- Tomáš Heger (`@geckon <https://github.com/geckon>`_)
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,13 @@
 Release History
 ---------------
 
+2.9.2 (???)
++++++++++++
+
+**Bugfixes**
+
+- Don't use redirect_cache if allow_redirects=False
+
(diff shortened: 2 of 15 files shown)


sigmavirus24:
@geckon please rebase this against that branch instead. If you need help, let me know and I can give you some instructions that should work in this case. Thanks for handling this! :cake:

geckon:
Hopefully I did it right the third time: https://github.com/kennethreitz/requests/pull/3021
I believe this PR can be closed, right?

Issue #3022, opened 2016-02-18
exceptions missing from __init__.py


jtpereyda:
It is traditional to include the public API in `__init__.py`.

Consider this test code:
```python
import requests
import requests.exceptions

a = requests.exceptions.ConnectTimeout
```

This will run properly, but PyCharm raises a warning, saying, "Cannot find reference 'exceptions' in `__init__.py`".

This could be fixed by changing an `__init__.py` line from:

    from . import utils

to

    from . import requests, utils

While this is technically a [bug ](https://youtrack.jetbrains.com/issue/PY-18550) in PyCharm, requests' __init__.py already imports utils, apparently for this very purpose. So it would be consistent to import exceptions as well.

sigmavirus24:
This isn't _technically_ a bug in PyCharm. It is a bug that they should be fixing. If we did something like this for each editor that does nonsensical things like this, I'm not sure we'd have a very coherent project.

jtpereyda:
@kennethreitz Resolved this by changing the docs (see PR #3023).

Looks like http://docs.python-requests.org/en/master/api/#exceptions is still missing ConnectTimeout and ReadTimeout. Should they be added to `__init__.py`?

kennethreitz:
@jtpereyda oddly, those two exceptions are being _raised_ during the sphinx build. I haven't dug into it yet. 

But.... anyone can investigate! ;)

kmad1729:
Is this still a reproducible bug? Both (python2 and python3) sphinx builds are clean:

```
(p2_venv) requests $ make docs
cd docs && make html
make[1]: Entering directory '/home/kmaduri/Documents/programming/python/requests/docs'
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v1.3.5
loading pickled environment... failed: unsupported pickle protocol: 4
loading intersphinx inventory from http://urllib3.readthedocs.org/en/latest/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 18 source files that are out of date
updating environment: 18 added, 0 changed, 0 removed
reading sources... [100%] user/quickstart                                     
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] user/quickstart                                      
generating indices... genindex py-modindex
highlighting module code... [100%] requests.auth                              
writing additional pages... search
copying static files... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
[...]

PR #3023, opened 2016-02-18
Add exceptions into __init__.py

jtpereyda:
Addresses Issue #3022.

Also added the JetBrains IDE settings folder, `.idea/`, to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ requests.egg-info/
 *.swp
 *.egg
 env/
+.idea/
 
 .workon
 
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -55,6 +55,7 @@ try:
 except ImportError:
     pass
 
+from . import exceptions
 from . import utils
 from .models import Request, Response, PreparedRequest
 from .api import request, get, head, post, patch, put, delete, options


Lukasa:
I'm going to leave this up to @kennethreitz. As you stated in #3022, this is totally unnecessary, but neither is it a maintenance burden, so I have no objection to it. Thanks for your contribution! :cake:

kennethreitz:
We already have all the exceptions available in the public API:

``` python
from .exceptions import (
    RequestException, Timeout, URLRequired,
    TooManyRedirects, HTTPError, ConnectionError,
    FileModeWarning,
)
```

sigmavirus24:
I'm with @kennethreitz this isn't necessary. Because of the line @kennethreitz is citing, users can already just do

``` py
import requests

try:
    r = requests.get(url)
except requests.ConnectionError:
    pass

# or

try:
   requests.get(url)
except requests.exceptions.ConnectionError:
   pass
```

Both work perfectly fine. Beyond that, we don't need a gitignore line for PyCharm/idea projects because that should already be in your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

jtpereyda:
@kennethreitz Thanks, I was wondering about that line. Two questions:
1. That import makes them available as `requests.RequestException`. It's a subtle difference, but [this doc page](http://docs.python-requests.org/en/master/api/#exceptions) instructs users to include them as `requests.exceptions.RequestException`. Which one is the preferred or supported method?
   - If we want to include both, we could specify both in `__init__.py`.
   - If `requests.RequestException` is preferred (which I like), perhaps we should update the docs.
2. It looks like `ConnectTimeout` is missing from the list. Should it be added?

jtpereyda:
@sigmavirus24 Thanks for the global gitignore tip! I hadn't heard about that.

sigmavirus24:
@jtpereyda it's a subtle but unimportant difference. Try this in an interactive session:

``` py
>>> import requests
>>> try:
...     requests.get('http://localhost:8080')
... except requests.ConnectionError:
...     pass
...
>>> try:
...     requests.get('http://localhost:8080')
... except requests.exceptions.ConnectionError:
...     pass
...
>>>
```

Both already work perfectly fine. You can use either one or both interchangeably. If `ConnectTimeout` is missing, that should be added to the list of exceptions imported. But like I said, as it is now, we do not need to change anything for `requests.exceptions.RequestException` to work. It will simply work.

kennethreitz:
@jtpereyda the "recommended" way is, in fact, `requests.RequestException`, as `requests.exceptions` is really just an internal namespacing.

That section of the docs are auto-generated from the codebase. I'd love to fix that section if we can. 

However, at this time, there's nothing broken about the docs or the code. We're just discussing room for possible usability improvement (which is the main goal of this library).

kennethreitz:
@jtpereyda fixed the docs! http://docs.python-requests.org/en/master/api/#exceptions

jtpereyda:
@kennethreitz Thanks!

PR #3024, opened 2016-02-19
Added unit tests for utils module

Stranger6667:
More unit tests for `utils` module. Also there is small cleanup for this module - removed unused imports & variable.
Lines coverage increased from 46% to 75%, but I'd like to ask about some functions, which are not used in `requests` at all (e.g `dict_to_sequence`, `from_key_val_list`, etc). Is there some plan for them? I wanted to write unit tests for it, but I'm not sure, may be they are staged to be deleted and it is not reasonable to cover them with unit tests.

BR


Lukasa:
Some of the utils module functions are left in place to avoid breaking users that may be using them directly, even though they're not used in requests any more. There's certainly no plan to remove them prior to 3.0.0.

kennethreitz:
I'm down for removing ones that _don't belong there_ for 3.0.

Issue #3028, opened 2016-03-03
"in" operation on resp.cookies should not raise CookieConflictError


trilogysci:
Hosts which have the same cookie define for multiple domains should not raise 
"CookieConflictError: There are multiple cookies with name,"  when performing "in" 
 It should short cut and return True immediately when found.

Lukasa:
So the problem here is that the default implementation of `collections.Mapping.__contains__` calls in to `__getitem__`, which for the `RequestsCookieJar` requires that there be no duplicate. In this instance, we probably get the most bang for our buck by just overriding `__contains__` to call the parent implementation, and then catch `CookieConflictError` and return `True` in that instance.

I'm marking this as contributor friendly because the fix is fairly easy, and it'd be a good one for someone to pick up as a first open source change.

davidsoncasey:
I've been using requests for a while and thinking I'd contribute, so I might take a shot at this.

kennethreitz:
@davidsoncasey Awesome!

jamalmoir:
I'd like to give this one a go if that's OK.

Cheers.

PR #3032, opened 2016-03-05
Changes for #3028

davidsoncasey:
I've overwritten the `__contains__` method of `RequestsCookieJar` as discussed in the issue. It looks to be working as expected and tests pass, but this is my first contribution to this project so I could be missing something.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -767,7 +767,7 @@ class TestRequests:
         items = jar.items()
         assert len(items) == 2
 
-    def test_cookie_duplicte_names_raises_cookie_conflict_error(self):
+    def test_cookie_duplicate_names_raises_cookie_conflict_error(self):
         key = 'some_cookie'
         value = 'some_value'
         path = 'some_path'
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -764,6 +764,8 @@ class TestRequests:
         jar.set(key, value, domain=domain1)
         jar.set(key, value, domain=domain2)
         assert key in jar
+        items = jar.items()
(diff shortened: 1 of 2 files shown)


tests/test_requests.py
```diff
@@ -754,6 +754,30 @@ def test_cookie_as_dict_items(self):
...
+        value = 'some_value'
+        domain1 = 'test1.com'
+        domain2 = 'test2.com'
+
+        jar = requests.cookies.RequestsCookieJar()
+        jar.set(key, value, domain=domain1)
+        jar.set(key, value, domain=domain2)
+        assert key in jar
+        items = jar.items()
+        assert len(items) == 2
```
Lukasa, line 768: Can this test also validate that we get a CookieConflictError here too?

tests/test_requests.py
```diff
@@ -754,6 +754,30 @@ def test_cookie_as_dict_items(self):
...
+
+    def test_cookie_duplicte_names_raises_cookie_conflict_error(self):
```
Lukasa: Typo of duplicate.

Lukasa:
Thanks for this @davidsoncasey, this pretty much looks right! I left a few small notes inline for you to take a look at, but one those are addressed we can merge this!

davidsoncasey:
@Lukasa great, thanks for taking a look! I'll have a chance later today to make those edits.

davidsoncasey:
@Lukasa I added a couple fixes, take a look and let me know if there's anything else!

Lukasa:
\o/ Thanks so much @davidsoncasey! :sparkles: :cake: :sparkles:

PR #3034, opened 2016-03-06
Support non-ASCII Reason Phrase as per RFC2616

erydo:
RFC2616 specifies that header values and reason phrases are encoded in in ISO-8859-1 or RFC2047 (MIME for text). RFC7230, which obsoletes RC2616, recommends treating non-ASCII header values as 'opaque data'.

In the case of the Reason Phrase, however, we should try to decode it as described due to its possible/encouraged localization and due to its purpose for display to the user; treating it as 'opaque data' is not useful and prevents its intended use.

Note that several production servers, including Apache Tomcat, send responses conforming to RFC2616. Servers that respond only in US-ASCII are unaffected by this change. This *only* regresses (by introducing possible encoding errors) for servers that send responses conforming to neither RFC, which I believe are either uncommon or non-existent.

Fixes kennethreitz/requests#3009

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -19,7 +19,7 @@ from .packages.urllib3.util.retry import Retry
 from .compat import urlparse, basestring
 from .utils import (DEFAULT_CA_BUNDLE_PATH, get_encoding_from_headers,
                     prepend_scheme_if_needed, get_auth_from_url, urldefragauth,
-                    select_proxy)
+                    select_proxy, get_rfc2047_text_as_unicode)
 from .structures import CaseInsensitiveDict
 from .packages.urllib3.exceptions import ClosedPoolError
 from .packages.urllib3.exceptions import ConnectTimeoutError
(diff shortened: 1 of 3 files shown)


Lukasa:
I'm sorry, but this reasoning is still not accurate, and so this code is still not right.

Firstly, RFC 2616 is not relevant anymore except inasmuch as it provides guidance to what older implementations may do.

Secondly, this argument is wrong:

> In the case of the Reason Phrase, however, we should try to decode it as described due to its possible localization and its purpose for display to the user; treating it as 'opaque data' is not useful and prevents its intended use.

RFC 7230 is very clear on this:

> The reason-phrase element exists for the sole purpose of providing a textual description associated with the numeric status code, mostly out of deference to earlier Internet application protocols that were more frequently used with interactive text clients.  A client SHOULD ignore the reason-phrase content.

That is to say, we are strongly advised by RFC 7230 to ignore the reason phrase _entirely_ rather than to do any special handling of it.

Now, I do believe that we should aim to fix the original problem, but the original problem is _entirely_ that the default encode/decode cycle of Python was busted. For that reason, we should restrict the scope of our fix to the `raise_for_status` method (any anywhere else where we try to do computation on the reason phrase), at which point we should use `requests.utils.to_native_string` with the `ISO-8859-1` encoding. That does the minimal translation to avoid anything exploding when doing the string interpolation, but otherwise leaves the reason phrase unmolested.

sigmavirus24:
@erydo expressed on the original issue that they were going hands off. That implies to me that they're done working on this if we're not going to accept their changes. If I'm wrong, I'll be happy to reopen this if they wish to follow our recommendations for the appropriate fix. Otherwise, this will sit open for a while and I'd rather that not happen.

PR #3041, opened 2016-03-09
Fix autofunction Reference

bsandrow:
`requests.codes` is a class (`LookupDict`), not a function.


kennethreitz:
:sparkles: :cake: :sparkles:

PR #3042, opened 2016-03-09
Fix api.rst References

bsandrow:
`api.rst` references `requests.ConnectTimeout` and `requests.ReadTimeout`, but they aren't imported into the top-level of the package.


kennethreitz:
:sparkles: :cake: :sparkles:

PR #3043, opened 2016-03-09
Verify Certificate never False 

tusharmakkar08:
Since `and` condition is there, `verify is not True` will be always false.

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -144,11 +144,10 @@ def remove_cookie_by_name(cookiejar, name, domain=None, path=None):
     """
     clearables = []
     for cookie in cookiejar:
-        if cookie.name != name:
-            continue
-        if domain is not None and domain != cookie.domain:
-            continue
-        if path is not None and path != cookie.path:
+        bool_cookie = (cookie.name != name)
+        bool_domain = (domain is not None and domain != cookie.domain)
+        bool_path = (path is not None and path != cookie.path)
+        if bool_cookie or bool_domain or bool_path:
             continue
         clearables.append((cookie.domain, cookie.path, cookie.name))
 
@@ -242,19 +241,11 @@ class RequestsCookieJar(cookielib.CookieJar, collections.MutableMapping):
 
     def list_domains(self):
         """Utility method to list all the domains in the jar."""
-        domains = []
-        for cookie in iter(self):
-            if cookie.domain not in domains:
-                domains.append(cookie.domain)
-        return domains
+        return list(set([cookie.domain for cookie in iter(self)]))
 
     def list_paths(self):
         """Utility method to list all the paths in the jar."""
-        paths = []
-        for cookie in iter(self):
-            if cookie.path not in paths:
-                paths.append(cookie.path)
-        return paths
+        return list(set([cookie.path for cookie in iter(self)]))
 
     def multiple_domains(self):
         """Returns True if there are multiple domains in the jar.
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -164,3 +164,4 @@ Patches and Suggestions
 - Robin Linderborg (`@vienno <https://github.com/vienno>`_)
 - Brian Samek (`@bsamek <https://github.com/bsamek>`_)
 - Dmitry Dygalo (`@Stranger6667 <https://github.com/Stranger6667>`_)
+- Tushar Makkar <tusharmakkar08@gmail.com>(`@tusharmakkar08 <https://github.com/tusharmakkar08>`_)
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -164,4 +164,4 @@ Patches and Suggestions
 - Robin Linderborg (`@vienno <https://github.com/vienno>`_)
 - Brian Samek (`@bsamek <https://github.com/bsamek>`_)
 - Dmitry Dygalo (`@Stranger6667 <https://github.com/Stranger6667>`_)
-- Tushar Makkar <tusharmakkar08@gmail.com>(`@tusharmakkar08 <https://github.com/tusharmakkar08>`_)
+- Tushar Makkar <tusharmakkar08@gmail.com> (`@tusharmakkar08 <https://github.com/tusharmakkar08>`_)
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -197,7 +197,10 @@ class HTTPAdapter(BaseAdapter):
             conn.ca_cert_dir = None
 
         if cert:
-            conn.cert_file, conn.key_file = (cert[0], cert[1]) if not isinstance(cert, basestring) else (cert, None)
+            if not isinstance(cert, basestring):
+                conn.cert_file, conn.key_file = cert[0], cert[1]
+            else:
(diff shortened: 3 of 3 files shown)


sigmavirus24:
Hey @tusharmakkar08,

Thanks for sending this pull request. Could you explain what bug you were seeing that made you change this? What instigated this change? It's not clear that this exhibits the same behaviour it did before, especially since tests have failed pretty consistently across all versions of Python.

tusharmakkar08:
@sigmavirus24 : In line 174 in `adapters.py` `if url.lower().startswith('https') and verify`  is there and then `if not verify` is there. Since `and` condition is there, `if not verify` won't be `True`. The other thing is more pythonic way of if/else i.e in line 200

sigmavirus24:
@tusharmakkar08 your commit no longer affects that region and simply turns an easy to read if/else statement into a ternary which is not "pythonic" nor is it easier to read.

I'll ask again: What bug are you seeing that prompted this (if there was any bug)?

tusharmakkar08:
There was no bug per se. I was trying to remove lines for increase in readability.

Lukasa:
Fewer lines does not lead to an increase in readability. In this instance, the ternary conditional is particularly unclear when combined with tuple unpacking assignment as well as creating an intermediate tuple in many cases. 

In this instance, I suggest that the only clear readability improvement would come from using tuple unpacking in the first branch of the conditional. I don't think any change beyond that is warranted.

sigmavirus24:
We're not interested in random refactorings that change logic for no apparent reason. Sorry @tusharmakkar08

tusharmakkar08:
@Lukasa : I have done couple of more changes and few of them are [anti-patterns](https://www.quantifiedcode.com/app/issue_class/62b0f5b7b69e4a2498568b32bfa30991) as suggested by http://docs.quantifiedcode.com/python-code-patterns/ . This isn't random refactoring, we are removing anti-patterns from code @sigmavirus24. 

Thanks.

kennethreitz:
@tusharmakkar08 while your contributions are appreciated, I want to give you feedback — all of the changes present in this pull request actually _reduce_ readability, significantly.

requests/cookies.py
```diff
@@ -242,19 +241,11 @@ def items(self):
 
     def list_domains(self):
         """Utility method to list all the domains in the jar."""
-        domains = []
-        for cookie in iter(self):
-            if cookie.domain not in domains:
-                domains.append(cookie.domain)
-        return domains
+        return list(set([cookie.domain for cookie in iter(self)]))
```
tusharmakkar08, line 244: How is this reducing readability @kennethreitz ? Can you please elaborate? This is a classic example of "Complex is better than complicated" according to Zen of Python"

requests/cookies.py
```diff
@@ -144,11 +144,10 @@ def remove_cookie_by_name(cookiejar, name, domain=None, path=None):
     """
     clearables = []
     for cookie in cookiejar:
-        if cookie.name != name:
-            continue
-        if domain is not None and domain != cookie.domain:
-            continue
-        if path is not None and path != cookie.path:
+        bool_cookie = (cookie.name != name)
```
tusharmakkar08, line 147: @kennethreitz : I read about this anti-pattern [here](https://www.quantifiedcode.com/app/issue_class/62b0f5b7b69e4a2498568b32bfa30991) and also according to [this](https://www.youtube.com/watch?v=8bZh5LMaSmE&feature=youtu.be) video by Sandi Metz, anyday this would be preferred over the current one.

tusharmakkar08:
@sigmavirus24 @Lukasa @kennethreitz: Since requests is being mentioned over [here](http://docs.python-guide.org/en/latest/writing/reading/) among the best python codes, I believe these kind of antipatterns shouldn't exist over here.

requests/cookies.py
Lukasa, line 147: Your first link has nothing to do with this pattern. Your first link objects to having a _nested_ if conditional that represents the entire body of the outer _if_ conditional. This is not that. This has consecutive if conditionals _at the same level_ that each continue the loop. Sandi _also_ complains about nested conditionals, which again is not something we have here. Sandi's complaints are about differing levels of abstractions, which are not true here. We are operating at one

This is widely considered a _very good_ pattern for handling complex conditional logic in loops because it splits each check out to its own line that can be considered by itself.

And, to be clear, the variable names you're using here actively reduce the clarity of this block. `bool_cookie` is a very problematic variable name because it makes the conditional at the bottom really unclear. _All_ of these boolean metavariables are to do with cookies! The question is, is this cookie the one we're looking for.

There is a problem here with assuming that because there are multiple `if` conditionals in this code it is unclear. This code is about as clear as it gets. Each `if` conditional follows the exact same pattern except for the first, which does not allow `name` to be `None`. All the rest say "if this parameter was passed to us, check it matches the cookie we're looking at". There is nothing complex or complicated about this code.

The _best_ way to clean this code up would be to add a comment simply explaining that. It'd be maybe two lines:

``` python
# Loop over the cookies in the jar, to find any that match the filter
# criteria passed by the user (same name, domain, and path)`.
```

tusharmakkar08, line 147: @Lukasa Thanks for clearing things up. I agree though it's not `nested` if's but wouldn't it be good if multiple `if's` can be reduced to a single `if` statement. Sandi _mentioned_ to have kind of `linear` structure when overviewed which is again accomplished by this change. I agree with the issue in variables naming and I apologize for the wrong naming. I will update the variable names and add the comment as mentioned by you. 
What is your feedback on the changes I mentioned in the `list_domains` function?

Lukasa, line 147: > Sandi mentioned to have kind of linear structure when overviewed which is again accomplished by this change.

Sandi talked about heirarchy. She would have been fine with the code as is: there's one level of indentation. Indeed, in Ruby there'd have been none, because you can do each of these as a one-line conditional with a `continue` at the end.

> but wouldn't it be good if multiple if's can be reduced to a single if statement.

No. Unequivocally not. When it comes to code structure, it is rarely the case that one approach is unequivocally better than the other: it depends entirely on the details. For example, here's another way to get all this code into one if statement:

``` python
if (name == cookie.name) and (domain is None or domain == cookie.domain) and (path is None or path == cookie.path):
    clearables.append((cookie.domain, cookie.path, cookie.name))
```

But I think we can agree that this conditional is not better than what we're doing right now. If having multiple conditionals is clearer than having one, there's nothing wrong with having multiple conditionals. And in this case, we get a bit of a speedup from it too (avoid doing checks we don't need to do by bailing on them early).

requests/cookies.py
kennethreitz, line 244: This is one of the better changes present.

sigmavirus24:
@tusharmakkar08 please keep in mind that your continuous arguments are emailing roughly 841 people. Since it doesn't seem as if you will respectfully let this conversation end and it serves to benefit none of those 841 people (because it has little value to the project as a whole and is likely _not_ why they are subscribed) I am ending the conversation forcefully by locking it.

kennethreitz:
Antipatterns are an antipattern. :)

PR #3048, opened 2016-03-13
Added unit tests for structures module

Stranger6667:


tests/test_structures.py
```diff
@@ -0,0 +1,72 @@
...
+    def dict_instance(self):
+        instance = CaseInsensitiveDict()
+        instance['Accept'] = 'application/json'
+        return instance
+
+    def test_list(self, dict_instance):
+        assert list(dict_instance) == ['Accept']
+
+    def test_getitem(self, dict_instance):
+        assert dict_instance['aCCept'] == 'application/json'
```
Lukasa: For this kind of test it's really valuable to have quite a few examples with wildly different cases. Something like:

``` python
@pytest.mark.parametrize('key', ['aCCept', 'Accept', 'accept', 'ACCEPT', 'aCcEpT'])
def test_getitem(self, dict_instance, key):
    assert dict_instance[key] == 'application/json'
```

```
```

tests/test_structures.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def test_delitem(self, dict_instance):
+        del dict_instance['aCcept']
+        assert 'aCCept' not in dict_instance
+        assert 'Accept' not in dict_instance
```
Lukasa: Similarly, this can be parametrised.

Lukasa:
@sigmavirus24 @kennethreitz Can one of you two do a separate code review of this? @Stranger6667 is here with me at PyCon SK writing some of these tests. I'm happy with these, but it'd be good if one of you two gave the ok/not-ok.

requests/structures.py
```diff
@@ -85,6 +85,7 @@ def copy(self):
     def __repr__(self):
         return str(dict(self.items()))
 
+
```
sigmavirus24: Again, the changes in this file are not unit tests, they're strictly cosmetic

tests/test_structures.py
```diff
@@ -0,0 +1,75 @@
+# coding: utf-8
+import pytest
+
+from requests.structures import CaseInsensitiveDict, LookupDict
+
+
+class TestCaseInsensitiveDict:
+
+    @pytest.fixture
+    def dict_instance(self):
```
sigmavirus24: This doesn't use the `self` parameter and as such shouldn't be defined on the class. There should also be a docstring here explaining what dict_instance does for a quick glance.

In fact nothing in this class uses `self` and as such each should be its own function.

tests/test_structures.py
```diff
@@ -0,0 +1,75 @@
...
+            ({'AccePT': 'application/json'}, True),
+            ({}, False),
+            (None, False)
+        )
+    )
+    def test_equality(self, dict_instance, other, result):
+        assert (dict_instance == other) is result
+
+
+class TestLookupDict:
```
sigmavirus24, line 53: Same feedback about this class. Use more descriptive function names and make these all independent functions please.

Alternatively, don't use fixtures passed into each method, but instead have a "set-up" fixture that runs for each method and defines the `instance` on `self`.

That would justify fairly bland test names grouped inside classes.

Stranger6667:
Hello @Lukasa @sigmavirus24 !
I've updated this PR :)

tests/test_structures.py
```diff
@@ -0,0 +1,79 @@
...
+
+class TestCaseInsensitiveDict:
+
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        """
+        CaseInsensitiveDict instance with "Accept" header.
+        """
+        self.case_insensitive_dict = CaseInsensitiveDict()
+        self.case_insensitive_dict['Accept'] = 'application/json'
```
Lukasa, line 15: Ah, sorry @Stranger6667, but I think @sigmavirus24 wanted [something more like this](https://pytest.org/latest/xunit_setup.html#method-and-function-level-setup-teardown).

tests/test_structures.py
```diff
@@ -0,0 +1,79 @@
...
+
+class TestLookupDict:
+
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        """
+        LookupDict instance with "bad_gateway" attribute.
+        """
+        self.lookup_dict = LookupDict('test')
+        self.lookup_dict.bad_gateway = 502
```
Lukasa, line 61: Same here.

Lukasa:
Awesome, we're very close now @Stranger6667. One small note. =)

kennethreitz:
Looks pretty solid!

PR #3049, opened 2016-03-13
Added unit tests for hooks module

Stranger6667:


Lukasa:
This also seems reasonable to me, but @Stranger6667 is sat right next to me. ;) @sigmavirus24/@kennethreitz, mind doing an extra review for me? This is +1 from me.

ryandebruyn:
Hello folks seems somehow I got placed on this email alias By mistake.
Anyway I can be removed.

Thank you
-Ryan

Lukasa:
@ryandebruyn This is managed by GitHub, it's not an email alias. You'll need to take some action to remove yourself via GitHub.

requests/hooks.py
```diff
@@ -14,6 +14,7 @@
 """
 HOOKS = ['response']
 
+
```
sigmavirus24: I know this is pep8, but it's unnecessary and is not a unit test. ;)

tests/test_hooks.py
```diff
@@ -0,0 +1,23 @@
+# coding: utf-8
+import pytest
+
+from requests.hooks import dispatch_hook, default_hooks
```
sigmavirus24: I would rather just import `hooks` from requests instead of the two functions.

sigmavirus24:
@ryandebruyn you may be "Watching" this repository in which case only you can fix that by visiting the project page while logged in and "un-watching" it.

Stranger6667:
Hello, @sigmavirus24 @Lukasa !
I've updated this PR :)

tests/test_hooks.py
```diff
@@ -0,0 +1,22 @@
...
+    return value[1:]
+
+
+@pytest.mark.parametrize(
+    'hooks_list, result', (
+        (hook, 'ata'),
+        ([hook, lambda x: None, hook], 'ta'),
+    )
+)
+def test_hooks(hooks_list, result):
```
Lukasa, line 17: So, this test is now pretty opaque: looking at the parametrized results don't make it enormously clear what it's supposed to be testing. A docstring or comment that explains how this test should function would be good.

Lukasa:
I have one small note. =)

tests/test_hooks.py
kennethreitz, line 17: +1

kennethreitz:
@Stranger6667 I don't like having tons of pull requests open, so I'm merging this. If you wouldn't mind, it would be appreciated if you opened another PR that addresses the notes @Lukasa left on your test.

PR #3051, opened 2016-03-15
Fix 'Transfer-Encoding: chunked' bug when upload an empty file

monhz:
Though we have the `stdout` (which is also a `fileno` type with length=0) to care about, when we post or put a normal empty file, we still can not force to set the `Transfer-Encoding` to `chunked`.

I think it is clearly meaning that "I don't want to transfer the file in chunked type", when someone set the `Content-length=0` in headers. `requests` does too much in this case, and it will bring out some problem when requesting some servers which not support chunked type. The `Transfer-Encoding` parameter should be user's response to add into the headers.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -439,7 +439,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
 
             if length:
                 self.headers['Content-Length'] = builtin_str(length)
-            else:
+            elif 'Content-Length' not in self.headers:
                 self.headers['Transfer-Encoding'] = 'chunked'
         else:
             # Multi-part file uploads.


Lukasa:
I am :-1: on this change.

The reality is that I don't care about servers that don't support chunked transfer encoding. [RFC 7230 Section 4.1](https://tools.ietf.org/html/rfc7230#section-4.1) says:

> A recipient MUST be able to parse and decode the chunked transfer coding.

And this is hardly a new requirement. RFC 2616, the original HTTP/1.1 specification that was published _nearly 15 years ago_ (in 1999), says in [Section 3.6.1](https://tools.ietf.org/html/rfc7230#section-4.1) that:

> All HTTP/1.1 applications MUST be able to receive and decode the "chunked" transfer-coding

This means that any server that cannot handle chunked transfer encoding is either a) a HTTP/1.0-only server, which is incompatible with requests (a HTTP/1.1 library); or b) deliberately non-compliant with the now _fifteen year old_ HTTP/1.1 specification. Requests has _no duty_ to bow to the whims of server developers that chose only to support more than 15 year old protocols or to ignore specifications when they do so, or to operators of websites that are stuck in the 1990s.

And even if that were not true, your reasoning about chunked transfer encoding:

> The Transfer-Encoding parameter should be user's response to add into the headers.

is not now and has never been how requests operates. The requests team has quite consistently told users not to add their own content-length or transfer-encoding headers to the headers dictionary, because requests will ignore them.

The upshot is that sending an empty chunked body is far and away the nicest way to deal with this situation.

PR #3053, opened 2016-03-16
remove simplejson import, native python json import only

digitaldavenyc:
Pull Request derived from the discussion on issue https://github.com/kennethreitz/requests/issues/3052

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -5,7 +5,7 @@ pythoncompat
 """
 
 from .packages import chardet
-import json
+
 import sys
 
 # -------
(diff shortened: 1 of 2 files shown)


Lukasa:
This seems reasonable to me, and it's certainly the smallest diff that could achieve the job. I do wonder if we shouldn't just remove this from `compat` now though, and instead change the imports in the modules that use it.

digitaldavenyc:
@Lukasa That's a good point. There is actually only file that is importing json, [models.py](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L34) So the diff would be about the same. Updated the PR

sigmavirus24:
This should target proposed/3.0.0 not master.

Issue #3052, opened 2016-03-15
Why default to simplejson?


digitaldavenyc:
I've recently run into an issue where another library installed simple-json and because the requests library defaults to it if available, it caused all of our json requests to fail due to a decoding problem.

I'm unclear why requests even defaults to simple-json anymore. I'd be happy to contribute to a PR to make the json library for requests more controllable but wanted to submit an issue first. Or perhaps there is another way I'm unaware of that would allow more control over which json library requests will use.

Lukasa:
Thanks for this report! Please see issue #2516 for the last time this was discussed.

Lukasa:
So to be clear, I am now as I was then open to removing simplejson as an option entirely, but @kennethreitz wanted it there. =)

digitaldavenyc:
I'm not saying it should be removed ... but perhaps there is a better way to control the import. It's possible for libraries to conflict with each other. Could there be a way to specify the json library to requests?

kennethreitz:
Historically, I was under the impression that a freshly-compiled `simplejson` has better performance characteristics than the `json` module that is built into Python 2.6.

kennethreitz:
I believe the "speed-ups" weren't enabled in the `json` module until 2.7.

Recalling from memory here, so I may not be correct.

Lukasa:
Controlling imports is _extremely_ fraught because it relies on delaying the import until some later time in the execution. I am pretty averse to doing that because it causes bugs to appear at weird places and can generally make life weirdly tricky. If we're going to keep it we should keep it, if we're going to remove it we should remove it.

kennethreitz:
The last three comments here state our stance nicely https://github.com/kennethreitz/requests/pull/2516#issuecomment-89005463

digitaldavenyc:
So what would you suggest to do when two libraries conflict with each other? Monkey patch the requests library to control which json import is used?

kennethreitz:
Conflict with eachother? I don't understand. If simplejson doesn't function properly on your system, then you probably find out why, or uninstall it.

kennethreitz:
I'm not against removing if it's really a problem for some users, as I don't think it would impact anyone in any measurable way, but it's currently one of those nice final touches that makes using Requests a pleasure — an extreme level of thoughtfulness.

Five years ago, people cared a lot more about the speed of their encoding/decoding serialization libraries than they seem to today. This was best practice. I don't think it's harmful either.

kennethreitz:
[...]

PR #3056, opened 2016-03-16
Import native Python JSON only

digitaldavenyc:
Pull Request derived from the discussion on issue #3052

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -21,12 +21,6 @@ is_py2 = (_ver[0] == 2)
 #: Python 3.x?
 is_py3 = (_ver[0] == 3)
 
-try:
-    import simplejson as json
-except (ImportError, SyntaxError):
-    # simplejson does not support Python 3.2, it throws a SyntaxError
-    # because of u'...' Unicode literals.
-    import json
 
 # ---------
 # Specifics
--- a/requests/models.py
+++ b/requests/models.py
@@ -32,7 +32,7 @@ from .utils import (
 from .compat import (
     cookielib, urlunparse, urlsplit, urlencode, str, bytes, StringIO,
     is_py2, chardet, builtin_str, basestring)
-from .compat import json as complexjson
+import json as complexjson
 from .status_codes import codes
 
 #: The set of HTTP status codes that indicate an automatically


Lukasa:
LGTM.

sigmavirus24:
:sparkles: :cake: :sparkles:

kennethreitz:
Really, no longer a need for `json` to be in `compat` at all anymore, let alone be called `complexjson`. 

But, that would require touching a lot more code.

digitaldavenyc:
@kennethreitz I actually was going to rename complexjson to json but there are json variable references all over the models.py file and decided it was probably best not to refactor it. If you still think it's a good idea I'm game to change it.

digitaldavenyc:
Perhaps there a different name we could use for the import so it would require less touching of code. `jsonp`?

kennethreitz:
Ah, prob best to `import json as complexjson` in `models.py` then (instead of `compat.py`).

It's a clever name, and I enjoy cleverness.

digitaldavenyc:
Ha! Agreed, complexjson is a good alt name. Best to either change it to json or leave it as is.

kennethreitz:
PR looks perfect.

sigmavirus24:
The reason we need `complexjson` in `models.py` is because we accept `json` as a method/function parameter and for those cases we can't also reference `json` the module without a name collision. Leaving it as such makes life a bit simpler.

That said, I'm open to changing it to one of the following
- `loljson`
- `whyjson`
- `hailjson`
- `zomgjson`
- `whatevenisjson`

And probably quite a few other names. ;)

kennethreitz:
- `mwson` (micro web services object notation)

kennethreitz:
- `perhapsjson`
- `itcouldbejson`
- `pleasedontsayjason`

kennethreitz:
No, it's fine as is.

sigmavirus24:
I know. I'm just kidding about alternate names. Alternatively
- `mycathatesjson`
- `oneweirdtrickforserializinganddeserializingjson`

kennethreitz:
- `magicjson`
- `sacredjson`
- `chopjsoncarrywater`
- `cosmicjson`

digitaldavenyc:
I got it ...
- `trumpjson`

You guys are gonna delete the PR now aren't you :(

kennethreitz:
TrumpJSON™: It's unbelievable, really, the best JSON you'll ever experience. Trust me, I know JSON.

digitaldavenyc:
It overwrites all your code with this way better code, trust me it's better....

sigmavirus24:
It's a huuuuuuuuuuuuuuuuuuge download. It makes serialization great again

crccheck:
anyway this can get in a 2.x release? It's been two years now and `simplejson` is even less relevant than when this was authored

nicmcd:
What is the status on this?

PR #3057, opened 2016-03-17
Clarify that SSL verification is on by default

kevinburke:
Generally if a kwarg is present it indicates that an option other than the
default is being specified. Putting `verify=True` in the first code sample
for SSL confused me, because it seemed to indicate that you had to specify
`verify=True` to get SSL verification. The opposite is true; SSL verification
is turned on by default and you have to specify `verify=False` to opt out of
SSL verification.

Updates the docs to make this more clear. Furthermore, connections to
https://kennethreitz.com currently time out instead of presenting an invalid
certificate, so I replaced this domain with https://requestb.in, which presents
the same error message as is currently there.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -192,15 +192,16 @@ applied, replace the call to :meth:`Request.prepare()
 SSL Cert Verification
 ---------------------
 
-Requests can verify SSL certificates for HTTPS requests, just like a web browser.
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks @kevinburke! :sparkles: :cake: :sparkles:

t-8ch:
What about https://self-signed.badssl.com/? (I am not affiliated, just an occasional user)
The problem is, that is validates without SNI...

PR #3059, opened 2016-03-17
Raise a ProxyError for proxy related connection issues

alexanderad:
This one aims to address https://github.com/kennethreitz/requests/issues/3050


Lukasa:
Cool, I'm happy with this. @sigmavirus24 `ProxyError` is a `ConnectionError` subclass, so I think this can go in the next minor release. Agreed?

sigmavirus24:
I agree. Do we have things we've merged that are not suitable for a minor release though?

Lukasa:
I don't think so.

PR #3060, opened 2016-03-18
Consolidate logic for changing method during redirects

benweatherman:
I only moved the code into a function, there was no actual change to the code. I added a few tests to ensure we're doing things correctly.

The real point of me doing this is to make it easier to bring back `strict_mode` functionality. For you requests youngsters in the crowd, `strict_mode` followed the spec for redirects meaning the method wouldn't change to a GET. The current code follows the browser convention of changing the method to a GET when doing a 302 redirect. However, lots of servers want you to follow the standards (the nerve!) so I'd like to override the logic. Now that the method changing logic is in `rebuild_method`, I can simply override that function instead of overriding the entire `resolve_redirects` function as suggested by kennethreitz/requests#1325

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -116,7 +116,6 @@ class SessionRedirectMixin(object):
             resp.close()
 
             url = resp.headers['location']
-            method = req.method
 
             # Handle redirection without scheme (see: RFC 1808 Section 4)
             if url.startswith('//'):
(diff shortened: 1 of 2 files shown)


Lukasa:
So, I'm totally happy with this code change: I think it's a reasonable refactoring that helps improve some clarity, _as well_ as allow @benweatherman to perform the override you want.

I'm also delighted to see some extra tests added for this. I think this change is wonderful! Pinging @kennethreitz or @sigmavirus24 for an extra round of review before merging.

kennethreitz:
:sparkles: :cake: :sparkles:

benweatherman:
Yay! Thanks y'all for getting this merged quickly. Now for the release... :joy_cat:

PR #3069, opened 2016-03-29
Update docstring and API doc to document ability to add per-file headers in multipart POST

achermes:
I recently needed this function and could not find it in the docs - luckily SO pointed me at PR #1640.

However, I feel like this really should be documented, so here is my attempt.

Note:
* I have not run any tests whatsoever over this as this is "just a docs change" - I don't know if that's ok. 
* The API docs seem a bit inaccurate in that they do not mention that ``files`` can also be a list. Similarly the docsstring in ``_encode_files`` does not discuss the case where instead of a tuple a file-obj is provided.


Lukasa:
\o/ This looks great, thanks @achermes! :sparkles: :cake: :sparkles:

PR #3080, opened 2016-04-08
Add 421 Misdirected Request.

Lukasa:
The 421 Misdirect Request status code was originally added in [RFC 7540](https://tools.ietf.org/html/rfc7540#section-11.7), and is going to be actively used in [RFC 7838](https://www.rfc-editor.org/rfc/rfc7838.txt). This adds it to our status code registry.


kennethreitz:
:sparkles: :cake: :sparkles:

PR #3082, opened 2016-04-08
Added a test to show a faulty behaviour when posting binary data for an object with no __iter__

tzickel:
An object with read / seek / tell but without \__iter__ triggers a code path in requests which causes the stream to seek to the start instead of the position it was when it was passed into the data of the request. The faulty code is in models.py in the function prepare_content_length where it seeks to the start of the stream.

The tests fail currently because this is a test to expose the bug.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -484,12 +484,9 @@ class TestRequests:
 
     def test_POSTBIN_SEEKED_OBJECT_WITH_NO_ITER(self, httpbin):
 
-        class BufferedStream(object):
+        class TestStream(object):
             def __init__(self, data):
-                if isinstance(data, buffer):
-                    self.data = data
-                else:
-                    self.data = buffer(data)
+                self.data = data.encode()
                 self.length = len(self.data)
                 self.index = 0
 
@@ -498,10 +495,10 @@ class TestRequests:
 
             def read(self, size=None):
                 if size:
-                    ret = buffer(self.data, self.index, size)
+                    ret = self.data[self.index:self.index + size]
                     self.index += size
                 else:
-                    ret = buffer(self.data, self.index)
+                    ret = self.data[self.index:]
                     self.index = self.length
                 return ret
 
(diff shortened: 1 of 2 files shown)


tests/test_requests.py
```diff
@@ -482,6 +482,51 @@ def test_POSTBIN_GET_POST_FILES(self, httpbin):
...
             requests.post(url, files=['bad file data'])
 
+    def test_POSTBIN_SEEKED_OBJECT_WITH_NO_ITER(self, httpbin):
+
+        class BufferedStream(object):
+            def __init__(self, data):
+                if isinstance(data, buffer):
+                    self.data = data
+                else:
+                    self.data = buffer(data)
```
Lukasa: This is over-general I think. We only ever pass one type here, so the if branch is unnecessary.

tests/test_requests.py
```diff
@@ -482,6 +482,51 @@ def test_POSTBIN_GET_POST_FILES(self, httpbin):
...
+            def tell(self):
+                return self.index
+
+            def seek(self, offset, where=0):
+                if where == 0:
+                    self.index = offset
+                elif where == 1:
+                    self.index += offset
+                elif where == 2:
+                    self.index = self.length + offset
```
Lukasa: Again, we only seek one kind of way here so we can avoid the over-generality I think. We can probably safely assert that `where == 0` and then just act like we're in that case.

tests/test_requests.py
```diff
@@ -482,6 +482,51 @@ def test_POSTBIN_GET_POST_FILES(self, httpbin):
...
+        test = BufferedStream('test')
+        post1 = requests.post(httpbin('post'), data=test)
+        assert post1.status_code == 200
+        assert post1.json()['data'] == 'test'
+
+        test = BufferedStream('test')
+        test.seek(2)
+        post2 = requests.post(httpbin('post'), data=test)
+        assert post2.status_code == 200
+        assert post2.json()['data'] == 'st'
```
Lukasa, line 525: Can we either add a new test or extend this one to assert the values of the content-length headers are also right? I have a suspicion we get them wrong.

Lukasa:
Ok awesome, this is a great first step! Are you interested in providing a fix as well?

tzickel:
I have made a dummy test so it works both on python 2 and 3. I hope the added fix is ok with all possible interactions (first time looking at requests code), but it might still not work with the redirect bug.

Lukasa:
I think this works well, at least in the basic case. As you correctly pointed out, there's a question about how this will work with redirects, but that's harder for us to test at this moment (though we'll get there shortly I hope).

@kennethreitz, want a final review here?

kennethreitz:
This looks like a sane approach to me.

Lukasa:
Ok @kennethreitz, you can hit the big green merge button whenever you're ready.

kennethreitz:
Gah, @sigmavirus24 beat me to it

PR #3089, opened 2016-04-13
add requests.util.parse funtions

sleshep:
help things easy

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -148,7 +148,7 @@ def guess_filename(obj):
     """Tries to guess the filename of the given object."""
     name = getattr(obj, 'name', None)
     if (name and isinstance(name, basestring) and name[0] != '<' and
-            name[-1] != '>'):
+                name[-1] != '>'):
         return os.path.basename(name)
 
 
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this!

However, I don't believe we need either of these.

`parse_rawpostdata_to_dict` is a less general version of `urlparse.parse_qs` from the standard library, so users should simply use that. `parse_rawcookies_to_dict` is a less general buggy version of `cookielib.CookieJar.add_cookie_header`. Both of these have better versions available in the standard library, so I'm afraid that we're not going to merge this: they simply add no useful function that is not available elsewhere.

Thanks for the work though!

sleshep:
@Lukasa 
But i can't use `cookielib.CookieJar.add_cookie_header` to add a raw cookie to requests
can you give me a example?
thanks!

Lukasa:
@sleshep What specifically are you trying to achieve? You appear to have a cookie header on hand and want to attach it to a request. Is that right?

sleshep:
@Lukasa 
Yes, I copy a cookie from browser want to attach it to a request.
How to do this with `cookielib.CookieJar.add_cookie_header`?

Lukasa:
Oh I'm sorry, I didn't fully understand your use case. You want to do this:

``` python
import Cookie
import requests

c = Cookie.SimpleCookie()
c.load('user=a;a=b;c=1')

s = requests.Session()
for v in c.values():
    s.cookies.set_cookie(requests.cookies.morsel_to_cookie(v))
```

Then, running requests through that session will use the cookies appropriately.

Note that this won't tightly scope them to domains unless the domain is set in the cookie directly, but you can use the standard functionality in the Cookie module to manipulate the morsels appropriately.

sleshep:
@Lukasa 
Thanks.
 but I think use `Cookie.SimpleCookie` is too complex,use this pull request to very easy to achive.

sigmavirus24:
@sleshep As a project, we don't agree. This pull request does not implement the same functionality that the standard library does. It also implements only a portion of what can be found without this pull request regardless of one's perception of ease of use.

If these functions are working for you, you should keep using them, but we are not going to accept them into requests or maintain them since there are significantly more mature and robust solutions available to us that we do not have to maintain.

sleshep:
@sigmavirus24 
Ok.thank you

sleshep:
if found a simple way to  achieve my target.

``` python
from urlparse import parse_qsl
from requests import Session
cookies='user=a;a=b;c=1'
params='user=a&a=b&c=1'
s=requests.Session()
s.cookies.update(dict(parse_qsl(cookies)))
s.post('http://someurl',data=dict(parse_qsl(params)))
```

it work well.
Thanks all of you.
@kennethreitz 
@Lukasa 
@sigmavirus24

Issue #3090, opened 2016-04-13
Session.close() doesn't work with proxies


bodgit:
I have an upstream server that supports both basic authentication and SSL client certificates for authenticating a client.

I have the following contrived requests client (with a session object) that hits the upstream server using basic auth and then switches to using SSL client certificates and repeats the request:

```python
#!/usr/bin/env python

import requests

session = requests.Session()
session.verify = './cacert.pem'
session.auth = ('bob', 'password')

response = session.get('https://1.2.3.4/')

session.auth = None
session.cert = ('./cert.pem', './key.pem')
session.close()

response = session.get('https://1.2.3.4/')
```

Because the client certificates need to be presented at the beginning of the connection I use `session.close()` to tear down any active connections before attempting the second request. If I don't do this, the session connection pooling kicks in and reuses the existing connection from the first request which won't have presented any certificates and so the second request is unauthorised.

This is all understood however I then had a need for the client to access the server through a proxy so the code is updated to just add the proxy configuration:

```python
#!/usr/bin/env python

import requests

session = requests.Session()
session.verify = False
session.auth = ('bob', 'password')

session.proxies = {
  'https': 'http://alice:secret@5.6.7.8:3128/',
}

response = session.get('https://1.2.3.4/')

session.auth = None
session.cert = ('./cert.pem', './key.pem')
session.close()

response = session.get('https://1.2.3.4/')
```

Now the second request is unauthorised; `session.close()` seems to have no effect and doesn't tear down the connections through the proxy; there are no TCP FIN packets initiated from the client which I do see when no proxy is used.

The upstream server has keepalives enabled so if I add a sleep for longer than the idle timeout then the server terminates the connection between itself and the proxy, then the proxy terminates the connection to the client and then the second request is successful.

I appreciate this is probably a corner case but any ideas here? I would expect `session.close()` to correctly tear down the connections to the proxy. I've reproduced this against version 2.9.1.

Lukasa:
Great spot, this is definitely a bug!

The bug is in `HTTPAdapter.close`: this currently clears the basic `PoolManager`, but doesn't clear any instantiated `ProxyManager` objects. That means that they inadvertently get preserved, which makes using them for this use-case untenable.

This bug is easily fixed, though: clearing out `HTTPAdapter.proxy_manager` is going to be the way to go.

In the meantime @bodgit, to work around this problem you can not just close the Session but actively mount new `HTTPAdapters`:

``` python
session.close()

[...]

PR #3091, opened 2016-04-13
Clear any pooled proxy connections

bodgit:
Based on the explanation of the bug in #3090 this PR closes any pooled connections for any active ProxyManager object associated with the session.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1188,6 +1188,17 @@ class TestRequests:
         next(r.iter_lines())
         assert len(list(r.iter_lines())) == 3
 
+    def test_session_close_proxy_clear(self, mocker):
+        proxies = {
+          'one': mocker.Mock(),
+          'two': mocker.Mock(),
+        }
+        session = requests.Session()
+        mocker.patch.dict(session.adapters['http://'].proxy_manager, proxies)
+        session.close()
+        proxies['one'].clear.assert_called_once_with()
+        proxies['two'].clear.assert_called_once_with()
+
 
 class TestCaseInsensitiveDict:
 
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -264,10 +264,12 @@ class HTTPAdapter(BaseAdapter):
     def close(self):
         """Disposes of any internal state.
 
-        Currently, this just closes the PoolManager, which closes pooled
-        connections.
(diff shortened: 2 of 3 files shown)


sigmavirus24:
This looks great to me. @bodgit could you add some tests to ensure this doesn't regress?

bodgit:
I'm wondering what/how I can test this. I had a look through `tests/test_requests.py` but I could only see one proxy-related test. Do you have any suggestions?

sigmavirus24:
Create a dictionary with values which are mocks and then assert that `clear()` was called once on each one?

Lukasa:
To be clear, my code review is +1 on this too. @sigmavirus24 has the best testing approach here, I think we just need to mock it out.

bodgit:
The suggestion makes sense, however I don't know how to do this with pytest. I'm trying to find an existing example I can use as a starting point...

bodgit:
I've managed to come up with the following:

``` python
    def test_session_close_proxy_clear(self, mocker):
        proxies = {
          'one': mocker.Mock(autospec=True),
          'two': mocker.Mock(autospec=True),
        }
        session = requests.Session()
        mocker.patch.dict(session.adapters['http://'].proxy_manager, proxies)
        session.close()
        proxies['one'].clear.assert_called_once_with()
        proxies['two'].clear.assert_called_once_with()
```

I had to pull in `pytest-mock`. It passes with my change and fails when I revert it so it looks like it does the job.

Lukasa:
I think that looks fine to me. If you want to go ahead and update this pull request (you can just push your new commits to the branch), we'll review more formally. =)

tests/test_requests.py
```diff
@@ -1188,6 +1188,17 @@ def test_response_iter_lines_reentrant(self, httpbin):
         next(r.iter_lines())
         assert len(list(r.iter_lines())) == 3
 
+    def test_session_close_proxy_clear(self, mocker):
+        proxies = {
+          'one': mocker.Mock(autospec=True),
```
sigmavirus24: Why are you using `autospec` here? You're not actually spec'ing out a ProxyManager.

bodgit:
Fixed.

sigmavirus24:
Thanks @bodgit! :tada:

PR #3096, opened 2016-04-14
Change _store of CaseInsensitiveDict to OrderedDict

### LINKED ISSUE #3038: Order of request headers should be preserved when sent to origin server
#### Summary

If I made a request with code that looks like:

```
    requests.get(
        'http://127.0.0.1:8765/',
        headers=OrderedDict([
            ('X-Header-1', 'ignoreme'),
            ('X-Header-2', 'ignoreme'),
            ('X-Header-3', 'ignoreme'),
            ('X-Header-4', 'ignoreme'),
        ])
    )
```

I expect the recipient HTTP server to receive the X headers in the same order at they were listed in the code above. However I receive them in a random non-deterministic order.

#### Environment

* Requests 2.9.1
* Python 3.4.1

#### Repro Steps

* Download files in the following gist: https://gist.github.com/davidfstr/0dca5a133fa397cadd92
* Run reflect_server.py in terminal window 1.
* Run reflect_client.py in terminal window 2.

#### Expected Results

* Terminal window 1 shows a request made with the X headers in consecutive order from 1 to 4.

#### Actual Results

* Terminal window 1 shows a request made with the X headers in random non-deterministic order.

#### Notes

The following closed older issues suggest that preservation of order is indeed important:

* https://github.com/kennethreitz/requests/issues/284, https://github.com/kennethreitz/requests/issues/179
* https://github.com/kennethreitz/requests/issues/2668

piotrjurkiewicz:
This will preserve order of request headers when passed to request method as `OrderedDict`.

Closes #3038.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -315,6 +315,26 @@ class TestRequests:
         prep = ses.prepare_request(req)
         assert 'Accept-Encoding' not in prep.headers
 
+    def test_headers_preserve_order(self, httpbin):
+        """Preserve order when headers provided as OrderedDict."""
+        ses = requests.Session()
+        ses.headers = collections.OrderedDict()
+        ses.headers['Accept-Encoding'] = 'identity'
+        ses.headers['First'] = '1'
+        ses.headers['Second'] = '2'
+        headers = collections.OrderedDict([('Third', '3'), ('Fourth', '4')])
+        headers['Fifth'] = '5'
+        headers['Second'] = '222'
+        req = requests.Request('GET', httpbin('get'), headers = headers)
+        prep = ses.prepare_request(req)
+        items = prep.headers.items()
+        assert items[0] == ('Accept-Encoding', 'identity')
+        assert items[1] == ('First', '1')
+        assert items[2] == ('Second', '222')
+        assert items[3] == ('Third', '3')
+        assert items[4] == ('Fourth', '4')
+        assert items[5] == ('Fifth', '5')
+
     @pytest.mark.parametrize('key', ('User-agent', 'user-agent'))
     def test_user_agent_transfers(self, httpbin, key):
 
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -325,9 +325,9 @@ class TestRequests:
         headers = OrderedDict([('Third', '3'), ('Fourth', '4')])
         headers['Fifth'] = '5'
         headers['Second'] = '222'
-        req = requests.Request('GET', httpbin('get'), headers = headers)
+        req = requests.Request('GET', httpbin('get'), headers=headers)
         prep = ses.prepare_request(req)
-        items = prep.headers.items()
+        items = list(prep.headers.items())
         assert items[0] == ('Accept-Encoding', 'identity')
         assert items[1] == ('First', '1')
         assert items[2] == ('Second', '222')
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -318,11 +318,11 @@ class TestRequests:
     def test_headers_preserve_order(self, httpbin):
         """Preserve order when headers provided as OrderedDict."""
         ses = requests.Session()
-        ses.headers = collections.OrderedDict()
+        ses.headers = OrderedDict()
         ses.headers['Accept-Encoding'] = 'identity'
         ses.headers['First'] = '1'
         ses.headers['Second'] = '2'
-        headers = collections.OrderedDict([('Third', '3'), ('Fourth', '4')])
+        headers = OrderedDict([('Third', '3'), ('Fourth', '4')])
         headers['Fifth'] = '5'
         headers['Second'] = '222'
         req = requests.Request('GET', httpbin('get'), headers = headers)
--- a/requests/structures.py
+++ b/requests/structures.py
@@ -10,6 +10,8 @@ Data structures that power Requests.
 
 import collections
 
+from .compat import OrderedDict
+
 
 class CaseInsensitiveDict(collections.MutableMapping):
     """
@@ -40,7 +42,7 @@ class CaseInsensitiveDict(collections.MutableMapping):
 
     """
     def __init__(self, data=None, **kwargs):
-        self._store = collections.OrderedDict()
+        self._store = OrderedDict()
         if data is None:
             data = {}
         self.update(data, **kwargs)
--- a/requests/structures.py
+++ b/requests/structures.py
@@ -40,7 +40,7 @@ class CaseInsensitiveDict(collections.MutableMapping):
 
     """
     def __init__(self, data=None, **kwargs):
-        self._store = dict()
+        self._store = collections.OrderedDict()
         if data is None:
             data = {}
         self.update(data, **kwargs)


Lukasa:
Thanks for this! I'd love a test that proves that this works: do you feel like you're up to writing one?

piotrjurkiewicz:
OK, now tests should pass on Python 2.6 as well.

kennethreitz:
@piotrjurkiewicz the tests fail on 3.3 and 3.4 now http://ci.kennethreitz.org/job/requests-pr/974/

This reminds me, I need to add Python 3.5 to the CI server. I didn't realize it was missing.

tests/test_requests.py
```diff
@@ -315,6 +315,26 @@ def test_headers_on_session_with_None_are_not_sent(self, httpbin):
         prep = ses.prepare_request(req)
         assert 'Accept-Encoding' not in prep.headers
 
+    def test_headers_preserve_order(self, httpbin):
+        """Preserve order when headers provided as OrderedDict."""
+        ses = requests.Session()
+        ses.headers = OrderedDict()
```
sigmavirus24, line 321: Why not use a `CaseInsensitiveDict` here too?

tests/test_requests.py
```diff
@@ -315,6 +315,26 @@ def test_headers_on_session_with_None_are_not_sent(self, httpbin):
...
+        ses.headers['Accept-Encoding'] = 'identity'
+        ses.headers['First'] = '1'
+        ses.headers['Second'] = '2'
+        headers = OrderedDict([('Third', '3'), ('Fourth', '4')])
+        headers['Fifth'] = '5'
+        headers['Second'] = '222'
+        req = requests.Request('GET', httpbin('get'), headers = headers)
+        prep = ses.prepare_request(req)
+        items = prep.headers.items()
```
sigmavirus24: You either need to set your expected headers as a list and then zip the items of that list together with the output from `items()` here or you need to call `list(prep.headers.items())` because an `ItemsView` is not indexable.

piotrjurkiewicz: You're right, my fault, sorry for this.

sigmavirus24:
@kennethreitz the tests here seem to have hung in Jenkins. Is there anyway to add a timeout to test runs?

tests/test_requests.py
sigmavirus24, line 321: @piotrjurkiewicz I'm still curious about your choice here.

kennethreitz:
@sigmavirus24 just added a build timeout plugin, configured for 5 minutes for PR builds. 

Don't worry, I'm getting increasingly frustrated with maintaining Jenkins at the moment.

tests/test_requests.py
kennethreitz, line 321: All tests pass! I may cut a new release tonight, and would love to include this, so would love an answer to this question@piotrjurkiewicz :)

sigmavirus24:
> Don't worry, I'm getting increasingly frustrated with maintaining Jenkins at the moment. 

I've been maintaining my own for a while. I feel your pain and I'm happy to help out more.

kennethreitz:
I mean, I've been managing Jenkins servers since the Hudson days, but this is the only one I've had to deal with in the past three years, and my annoyances with how it requires at least some level of upkeep are growing :)

I do think it's an overall less frustrating experience for this project than using Travis would be, so worth the effort. Travis has seen much improvements over the past two years, though, so at some point, this may no longer be true.

tests/test_requests.py
piotrjurkiewicz, line 321: Common practice is to pass `headers` as `dict`. I have thought that an average user would intuitively use `OrderedDict` in place of `dict`, when he want order to be maintained. `CaseInsensitiveDict` is somewhat internal to requests, so many users may not be even aware of its existence.

But I can change it to `CaseInsensitiveDict` if you want.

sigmavirus24, line 321: People might use that, or they might use a list of tuples (which holds the same benefit).

piotrjurkiewicz, line 321: > they might use a list of tuples

In fact, they can't: https://github.com/kennethreitz/requests/issues/3038#issuecomment-193699603

`Session.headers` and `Request.headers` are required to be mappings: https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L55-L60

davidfstr, line 321: > Common practice is to pass headers as dict. I have thought that an average user would intuitively use OrderedDict in place of dict, when he want order to be maintained.

+1. Definitely what I've been doing.

kennethreitz:
:sparkles: :cake: :sparkles:

palacinka:
Is this fixed in version 2.10.0 ?

```
import requests
from collections import OrderedDict



headers = OrderedDict( [
('User-Agent', 'Mozilla/5.0 (Windows NT 6.3; WOW64; rv:46.0) Gecko/20100101 Firefox/46.0'),
('Accept', 'application/json, text/javascript, */*; q=0.01'),
('Accept-Language', 'cs,en-US;q=0.7,en;q=0.3'),
('Accept-Encoding', 'gzip, deflate'),
('Connection', 'keep-alive'),
] )

rsp = requests.get('http://api.wipmania.com',headers=headers)

```

order of headers not preserved on Windows 8.1  python 2.7.9   tested with Wireshark

edit: it seems to work fine only if I use session like this

```
import requests
from collections import OrderedDict

headers = OrderedDict( [
('User-Agent', 'Mozilla/5.0 (Windows NT 6.3; WOW64; rv:46.0) Gecko/20100101 Firefox/46.0'),
('Accept', 'application/json, text/javascript, */*; q=0.01'),
('Accept-Language', 'cs,en-US;q=0.7,en;q=0.3'),
('Accept-Encoding', 'gzip, deflate'),
('Connection', 'keep-alive'),
] )

ses = requests.Session()
ses.headers = OrderedDict()
ses.get('http://website.com',headers=headers)
```

Lukasa:
The reason the header order is being overridden in your case is because of the way requests merges the two different dictionaries in the `Session` and the request kwargs.

By default, a requests `Session` already contains several keys:

```
>>> s = requests.Session()
>>> s.headers
{'Connection': 'keep-alive', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'User-Agent': 'python-requests/2.10.0'}
```

You'll note, then, that when you send headers using the `headers` kwarg, the order of the keys in the Session is preserved in priority to the order of the keys in the `headers` kwarg. This is the expected result of using the dict on the `Session` as the base into which the request dict is merged to update.

Trying to get the entirely intuitive behaviour (where the request header defines the order in preference to the `Session` order) is somewhat frustrating. Right now the code looks like this:

``` python
actual_headers = CaseInsensitiveDict(to_key_val_list(session.headers))
actual_headers.update(to_key_val_list(request.headers))
```

We'd need to change the code to

``` python
actual_headers = CaseInsensitiveDict(to_key_val_list(request.headers))
actual_headers.update(to_key_val_list(session.headers))
actual_headers.update(to_key_val_list(request.headers))
```

This would lead to the exact same result as we currently have but would prioritise the _order_ of the request dict rather than the `Session` dict. The cost is that we do substantially extra computation in order to achieve this relatively minor effect.

I am open to making this change, but it does rather feel like using a sledgehammer to crack a nut. @kennethreitz @sigmavirus24?

sigmavirus24:
@Lukasa I think this is something we need to document instead of work around. I'm not sure "fixing" that particular behaviour wouldn't introduce some other subtle bug.

kennethreitz:
I don't disagree with @sigmavirus24. While I do feel like this pattern should "just work" as requested, I feel like it's extremely uncommon for someone to want/need this, and we should not bend over backwards to accomplish this.

Lukasa:
Ok, so I think we're in agreement: we'd rather document this than try to change the logic.

PR #3103, opened 2016-04-18
Implement #2149

gastlygem:
Though #2149 is marked as closed, I had to do this because 1) I can't just give up requests and 2) I can't change the non-standard URLs I'm using. So I did a little hack. Now it goes like this:

    >>> from urllib.parse import quote
    >>> requests.models.DEFAULT_QUOTE_VIA = quote
    >>> requests.models.DEFAULT_SAFE = '/?$=%'
    >>> r = requests.get('http://reddit.com/', params={'$foo': 'bar baz'})
    >>> r.url
    'https://www.reddit.com/?$foo=bar%20baz'

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,20 +7,8 @@ Requests: HTTP for Humans
 .. image:: https://img.shields.io/pypi/dm/requests.svg
         :target: https://pypi.python.org/pypi/requests
 
-**Warning:** This is a GMO version of the original requests library. Although it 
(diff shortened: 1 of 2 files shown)


Lukasa:
I'm -1 on this: we have an appropriate hook to do this already (the PreparedRequest), and I'm disinclined to want to add configurability here, _particularly_ in the form of global singleton state: this is a recipe for really tricky to debug issues down the road.

weiwei:
@Lukasa Thanks for pointing to the right direction.

PR #3109, opened 2016-04-21
HTTPAdapter now updates its PoolManager connection_pool_kw

### LINKED ISSUE #2863: Multiple requests on a single Session with different client certs behave incorrectly
Consider the following code:

``` py
import requests

s = requests.session()
s.get("https://host", cert="path/to/cert1.pem")
s.get("https://host", cert="path/to/cert2.pem")
```

`requests` will silently reuse the same connection for both connections, which is incorrect.

jeremycline:
This is not particularly polished yet, and depends on the as-yet unreleased
urllib3 version 1.16. Still, I'd greatly appreciate some early feedback on how
(in)sane this approach is. 

With the addition of https://github.com/shazow/urllib3/pull/830 requests
should update the connection_pool_kw on the PoolManager so that new
ConnectionPools get created when TLS/SSL settings change. This ensures
that users can update the CA certificates used to verify servers as well
as the client certificate and key it uses to authenticate with servers.

This fixes issue #2863

---

Closes #2863

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -10,6 +10,10 @@ and maintain connections.
 
 import os.path
 import socket
+try:
+    from threading import RLock
+except ImportError:  # threading is an optional module and may not be present.
+    from dummy_threading import RLock
 
 from .models import Response
 from .packages.urllib3.poolmanager import PoolManager, proxy_from_url
@@ -119,6 +123,7 @@ class HTTPAdapter(BaseAdapter):
         self._pool_connections = pool_connections
         self._pool_maxsize = pool_maxsize
         self._pool_block = pool_block
+        self._pool_kw_lock = RLock()
 
         self.init_poolmanager(pool_connections, pool_maxsize, block=pool_block)
 
@@ -131,6 +136,7 @@ class HTTPAdapter(BaseAdapter):
         # self.poolmanager uses a lambda function, which isn't pickleable.
         self.proxy_manager = {}
         self.config = {}
+        self._pool_kw_lock = RLock()
 
         for attr, value in state.items():
             setattr(self, attr, value)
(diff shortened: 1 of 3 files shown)


Lukasa:
Thanks for this @jeremycline!

In principle this is fine. However, it changes the interface of the HTTPAdapter which is IMO part of our public interface, so I'd want to sit on this until 3.0.0. Happily, 3.0.0 shouldn't be too far away now so we can progress with this work. =)

sigmavirus24:
I'm confused why we would remove `cert_verify` as part of this.

Lukasa:
@sigmavirus24 Because it sets things on the connection pool directly, which is Not Right. We want to get a different pool if those are different.

jeremycline:
I thought about leaving the method name as `cert_verify`, but did it this way because the behaviour changed a great deal. I am, of course, not apposed to alternate approaches. I just wanted to make sure it was clear the way certificates are paired up with connections changed.

jeremycline:
Since urllib-1.16 is out now, I suppose I should get this ready to go :smile:

Other than rebasing once 1.16 is pulled in, is there anything that jumps out as lacking?

Lukasa:
Nothing leaps out, no. =)

jeremycline:
I've rebased onto 2.11 and all the tests are passing now.

sigmavirus24:
@jeremycline can you rebase once more? I promise we won't lose track of this again. =(

jeremycline:
@sigmavirus24 no problem, rebased!

sigmavirus24:
wait a second, @Lukasa can this not go into a pre-3.0 release?

Lukasa:
@sigmavirus24 Nope, we removed cert_verify which is public on the HTTPAdapter.

jeremycline:
@sigmavirus24 Should I add a 3.0 changelog entry or is that something you'd prefer to handle?

requests/adapters.py
```diff
@@ -10,6 +10,10 @@
 
 import os.path
 import socket
+try:
+    from threading import RLock
+except ImportError:  # threading is an optional module and may not be present.
```
sigmavirus24, line 15: I don't understand this comment. Also, I don't understand why if this is a problem we've never run into it before given the fact that Urllib3 uses the threading library in several places.

jeremycline: It looks like this [was a problem](https://github.com/shazow/urllib3/issues/289) in urllib3 which is where I originally learned threading is optional. I used [dummy_threading](https://docs.python.org/3.6/library/dummy_threading.html#module-dummy_threading) since that's what the Python documentation recommended, although it looks like urllib3 just defines the required interface itself.

sigmavirus24:
Looks good to me then!

PR #3132, opened 2016-04-26
utils: let select_proxy not raise an exception when url has no hostname

### LINKED ISSUE #3131: utils.select_proxy raises an exception when URL has no hostname
URLs sometimes have no hostname (e.g. file URLs, or unix socket URLs). The parsed URL `hostname` attribute returns `None` in this case, making `utils.select_proxy` complain about concat'ing a `str` with `None`.

chipaca:
Fixes #3131.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -579,7 +579,10 @@ def select_proxy(url, proxies):
     """
     proxies = proxies or {}
     urlparts = urlparse(url)
-    proxy = proxies.get(urlparts.scheme+'://'+urlparts.hostname)
+    if urlparts.hostname is None:
+        proxy = None
+    else:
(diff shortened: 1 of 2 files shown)


Lukasa:
While we're here, we should probably also handle the possibility that there's no scheme in the URL as well.

chipaca:
scheme defaults to `''`, though. Well, more or less:

> The scheme argument gives the default addressing scheme, to be used only if the URL does not specify one. It should be the same type (text or bytes) as urlstring, except that the default value '' is always allowed, and is automatically converted to b'' if appropriate.

Lukasa:
Urgh, `urlparse` is such an inconsistent function. Good spot.

Lukasa:

PR #3136, opened 2016-04-28
Update readthedocs links.

Natim:


Lukasa:
Thanks so much @Natim!

PR #3171, opened 2016-05-02
docs: Add a note about SSL c_rehash

luv:
Just adding a quick note to inform users that it's not enough to point to a directory with the certificates but the directory needs to be "indexed". It took me some time to figure this out ... this simple note might save time to others in the future :).


Lukasa:
Thanks for this @luv! :sparkles: :cake: :sparkles:

PR #3176, opened 2016-05-04
add ability to control the headers that will be deleted in case of redirect

RonenHoffer:
In some cases the redirect is done to a different host.
when the {"Host": "www.example.com"} is set in the headers and the request is redirected to another host, let's say www.another-example.com, the next request will be sent to a new URLwith a wrong Host header.


Lukasa:
This is up to @kennethreitz, but I'm disinclined to want to add this: I think I'd rather factor out the logic in `SessionRedirectMixin` as we've done with a few other bits to allow easier subclassing, rather than add further keyword arguments.

sigmavirus24:
I agree with @Lukasa. I don't think we need this as an extra parameter to this method. I also don't think we need to keep track of which headers we're sanitizing for a user. I also agree that we might be able to better serve you by refactoring things so you can override these with subclassing.

kennethreitz:
👍

ghost:
@sigmavirus24, @Lukasa thanks for the comments.
Inheriting `SessionRedirectMixin` was my first idea but then I've seen so much logic in the `resolve_redirects` function so it became a sort of code duplication.
By saying this, I agree with you guys, factoring out the logic might come very handy.
So how should we proceed?

Lukasa:
@adaronen If you'd like to supply a PR that factors out this logic to a method (`resolve_*`) then we'll happily review that. =)

ghost:
Cool, I'll do my best.

Lukasa:
Closing due to inactivity. Please feel free to reopen if you have the time to make more forward progress.

Issue #3177, opened 2016-05-05
Encode JSON to bytes per urllib3 expectations


haikuginger:
urllib3 expects to receive request bodies as bytes-like objects. While sending Unicode strings may work, it's suboptimal and can result in unexpected behavior (see shazow/urllib3#855). While we shouldn't encode user-supplied request bodies, we should encode requests-produced JSON strings.

PR #3178, opened 2016-05-05
Encoding JSON requests to bytes for urllib3 to handle

haikuginger:
Implements #3177.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -423,8 +423,10 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             # When urllib3 uses pyOpenSSL, it can only resume large uploads
             # properly if receiving a bytes-like object. In Python 2, json.dumps()
             # returns just that, but Python 3 returns a Unicode string.
-            content_type = 'application/json; charset=utf-8'
-            body = complexjson.dumps(json).encode('utf-8')
+            content_type = 'application/json'
+            body = complexjson.dumps(json)
+            if not isinstance(body, bytes):
+                body = body.encode('utf-8')
 
         is_stream = all([
             hasattr(data, '__iter__'),
--- a/requests/models.py
+++ b/requests/models.py
@@ -420,9 +420,8 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         length = None
 
         if not data and json is not None:
-            # When urllib3 uses pyOpenSSL, it can only resume large uploads
-            # properly if receiving a bytes-like object. In Python 2, json.dumps()
-            # returns just that, but Python 3 returns a Unicode string.
+            # urllib3 requires a bytes-like body. Python 2's json.dumps
+            # provides this natively, but Python 3 gives a Unicode string.
             content_type = 'application/json'
             body = complexjson.dumps(json)
             if not isinstance(body, bytes):
--- a/requests/models.py
+++ b/requests/models.py
@@ -420,8 +420,11 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         length = None
 
         if not data and json is not None:
-            content_type = 'application/json'
-            body = complexjson.dumps(json)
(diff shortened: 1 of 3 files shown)


requests/models.py
```diff
@@ -420,8 +420,11 @@ def prepare_body(self, data, files, json=None):
         length = None
 
         if not data and json is not None:
-            content_type = 'application/json'
-            body = complexjson.dumps(json)
+            # When urllib3 uses pyOpenSSL, it can only resume large uploads
+            # properly if receiving a bytes-like object. In Python 2, json.dumps()
+            # returns just that, but Python 3 returns a Unicode string.
+            content_type = 'application/json; charset=utf-8'
```
Lukasa: Heh, little-known fact: the content-type for JSON (`application/json`) does not take a `charset` parameter, and anyone who sets it is doing it wrong. Mind removing the charset parameter?

requests/models.py
```diff
@@ -420,8 +420,11 @@ def prepare_body(self, data, files, json=None):
...
+            body = complexjson.dumps(json).encode('utf-8')
```
Lukasa: This is too aggressive: it encodes to UTF-8 even when `json` returns a bytestring, which it does on Python 2: that means that this line of code can cause UnicodeDecodeErrors when run if the user's default encoding is no good. I think we need a test on the type here.

Lukasa:
@haikuginger Thanks for this! I've left some notes inline.

requests/models.py
haikuginger: Fascinating!

> 1.  Encoding
>    
>    JSON text SHALL be encoded in Unicode.  The default encoding is UTF-8.
>    
>    Since the first two characters of a JSON text will always be ASCII
>    characters [RFC0020], it is possible to determine whether an octet
>    stream is UTF-8, UTF-16 (BE or LE), or UTF-32 (BE or LE) by looking
>    at the pattern of nulls in the first four octets.
>    
>    ```
>       00 00 00 xx  UTF-32BE
>       00 xx 00 xx  UTF-16BE
>       xx 00 00 00  UTF-32LE
>       xx 00 xx 00  UTF-16LE
>       xx xx xx xx  UTF-8
>    ```

Fix incoming.

Lukasa: This can be your weird TIL for the day. =D

requests/models.py
```diff
@@ -420,8 +420,13 @@ def prepare_body(self, data, files, json=None):
         length = None
 
         if not data and json is not None:
+            # When urllib3 uses pyOpenSSL, it can only resume large uploads
```
sigmavirus24: What does resuming large uploads have to do with urllib3 requiring a bytes object?

haikuginger: You're right; it's not relevant. I was in the mindset of urllib3 still, but in the context of requests, "the urllib3 API specifies a bytes-like body" is enough.

haikuginger:
Bump! 🙂

sigmavirus24:
Sorry @haikuginger! GitHub doesn't ping me when you push new changes. :(

sigmavirus24:
I'm :+1: on this if @Lukasa is =D

Lukasa:
Let's do it! Thanks @haikuginger, you're doing awesome work! :sparkles: :cake: :sparkles:

requests/models.py
```diff
@@ -420,8 +420,12 @@ def prepare_body(self, data, files, json=None):
         length = None
 
         if not data and json is not None:
+            # urllib3 requires a bytes-like body. Python 2's json.dumps
+            # provides this natively, but Python 3 gives a Unicode string.
```
kennethreitz, line 424: Should be avoiding mentioning urllib3 in this part of the codebase. I'll correct tomorrow, if I remember.

kennethreitz, line 424: Also, I hope json.dump() is utf-8 on python3.

kennethreitz, line 424: Also, I think it's hilariously telling it's taken someone this long to notice this, so far.

PR #3179, opened 2016-05-05
Fix TypeError when get json-encoded content of a response

messense:
[``self.content``](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L728) could be ``None``, so ``len(self.content)`` may raise ``TypeError: object of type 'NoneType' has no len()``

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -792,7 +792,7 @@ class Response(object):
         :param \*\*kwargs: Optional arguments that ``json.loads`` takes.
         """
 
-        if not self.encoding and len(self.content) > 3:
+        if not self.encoding and self.content and len(self.content) > 3:
             # No encoding set. JSON RFC 4627 section 3 states we should expect
             # UTF-8, -16 or -32. Detect which one to use; If the detection or
             # decoding fails, fall back to `self.text` (using chardet to make
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1227,6 +1227,17 @@ class TestRequests:
         proxies['one'].clear.assert_called_once_with()
         proxies['two'].clear.assert_called_once_with()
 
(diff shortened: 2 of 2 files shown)


Lukasa:
Huh. I feel like maybe we should just change `Response.content` to never be `None`. It's certainly surprising to me that it can be. @kennethreitz @sigmavirus24?

sigmavirus24:
@Lukasa so we set it to `None` in [the descriptor](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L728) in (2) [error](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L739) [cases](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L744). 

I don't think we can just change it to never be `None` before 3.0 but that's my hesitancy to break user-level code speaking.

Lukasa:
@sigmavirus24 Yeah, that seems reasonable. In that case, clearly the `json()` code needs to handle it too.

That means I think this patch is fine. @messense are you interested in adding a test to validate that this works as expected?

pcurry:
This came up yesterday in a work context.  The content of a response CAN BE None, and that would be correct, in the case of a 204 response code.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

I agree it would be better if response.json() handled that gracefully rather than dying of a JSON parsing error.

messense:
I just realized that fix this will make it raise `ValueError: No JSON object could be decoded` instead because of `json.loads('')`

Lukasa:
Hurrah, this looks great to me! I'd argue the previous behaviour was a bug so we have no API compatibility concerns to worry about here, but I want @sigmavirus24 to ACK that before merging.

sigmavirus24:
This looks fine to me :)

PR #3181, opened 2016-05-06
Potential fix for #3066 (Transfer-Encoding and Content-Length headers both being set)

davidsoncasey:

Issue #3066, opened 2016-03-25
auth and data set results in invalid http header


julian-r:
https://stackoverflow.com/questions/36216274/uploading-a-0-bytes-file-to-owncloud-with-python-requests-hangs/36217781#36217781

When auth is set and a file object is passed to data, the header contains a content-length AND a Transfer-Encoding: chunked is set in the header and additionally the last chunk is not send.

Lukasa:
Yup, this is a real bug. `prepare_auth` unconditionally calls `prepare_content_length`, which does the wrong thing for situations where it doesn't know the content-length.

Lukasa:
Working out exactly how best to fix this is tricky: `prepare_content_length` does weirdly quite a lot, and it's not entirely apparent what parts of its logic are actually mandatory.

Lukasa:
Ok, I think the best way to do this is to have `prepare_content_length` short-circuit out if there's a `Transfer-Encoding` header present in the headers already: if we already concluded we want to send via chunked transfer encoding, we probably shouldn't change our minds about that.

Lukasa:
In fact, we maybe want to go higher level than that: perhaps `prepare_auth` should choose not to delegate to `prepare_content_length` when `Transfer-Encoding` is present in the headers.

davidsoncasey:
@Lukasa has anyone started work on this? I have some free time today, I was thinking I'd take a look at it. If I come up with something I can put together a PR.

Lukasa:
@davidsoncasey I don't believe anyone has taken this up yet, no.

davidsoncasey:
@Lukasa great, I'm working now on writing a test to replicate the issue. Once I've got that, I may ask you to take a look to verify that I'm correctly capturing the issue, and then I'll proceed with a fix.

davidsoncasey:
@Lukasa I just made PR #3181, with the solution to not call `prepare_content_length` if the `Transfer-Encoding` header is present. I went back and forth on whether to do the check there or to short-circuit out of `prepare_content_length` and I'm still not sure which I like better, so if you take a look and think it would be better to do within `prepare_content_length` then by all means.

I was thinking it may be more clear to refactor `prepare_body` slightly to not set the `Content-Length`  header at all in `prepare_body` (line 442 of models.py), and to call `prepare_content_length` regardless of if the data is a stream. I think then it could be a bit more explicit that the two headers should never coexist in the same request. Let me know if you have thoughts.

[...]

Issue #1648, opened 2013-10-04
Header "Transfer-Encoding: chunked" set even if Content-Length is provided which causes body to not actually get chunked


ysangkok:
## Test script

```
import requests
import time

def f():
    yield b"lol"
    time.sleep(2)
    yield b"man"

requests.post('http://127.0.0.1:8801/', data=f(), headers={"Content-Length": 6})
```
## Actual result

Received on the server:

```
$ nc -p 8801 -l
POST / HTTP/1.1
Host: 127.0.0.1:8801
User-Agent: python-requests/2.0.0 CPython/3.3.1 Linux/3.11.0-031100rc4-generic
Accept: */*
Transfer-Encoding: chunked
Content-Length: 6
Accept-Encoding: gzip, deflate, compress

lolman
```
## Expected result

Did not expect "Transfer-Encoding: chunked" since I provided the Content-Length. If requests insists on doing chunked transfer encoding, it should disregard the content length and actually chunk the content (as it does if there is not Content-Length header given).

Lukasa:
I agree, we should pick one of those two options. =)

sigmavirus24:
So commenting here as I should have done >6 hours ago:

>  If requests insists on doing chunked transfer encoding, it should disregard the content length and actually chunk the content (as it does if there is not Content-Length header given).

You're implying that in this instance we didn't actually chunk the data. Do you have actual evidence of that? The output alone is not nearly enough to demonstrate that.

Furthermore running your example above in 2.0.0 does not work: I get this traceback:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/icordasc/virtualenv/vcr/lib/python2.7/site-packages/requests/api.py", line 88, in post
    return request('post', url, data=data, **kwargs)
  File "/Users/icordasc/virtualenv/vcr/lib/python2.7/site-packages/requests/api.py", line 44, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/icordasc/virtualenv/vcr/lib/python2.7/site-packages/requests/sessions.py", line 357, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/icordasc/virtualenv/vcr/lib/python2.7/site-packages/requests/sessions.py", line 460, in send
    r = adapter.send(request, **kwargs)
  File "/Users/icordasc/virtualenv/vcr/lib/python2.7/site-packages/requests/adapters.py", line 319, in send
    timeout=timeout
[...]

PR #3184, opened 2016-05-11
Refactor prepare_body

davidsoncasey:
@Lukasa Here's an alternative fix to #3066 (discussed in PR #3181), where I refactored `PreparedRequest.prepare_body` and `PreparedRequest.prepare_content_length` so that `prepare_content_length` is always called.

My only question is in the case when the body is a stream (which it is in the case that brought up this issue) that the current position will always be 0 when the length is calculated. Previously, in cases where `prepare_auth` was not called, the content length was being calculated with `super_len`, now it would be calculated using `self.headers['Content-Length'] = builtin_str(max(0, end_pos - curr_pos))`. As far as I can tell, this will give the same result (as long as the current position is 0), and this is how it's been computed in cases where authorization is being used. But I'm curious if you have thoughts about this.

### DIFF
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -101,8 +101,8 @@ class RetryError(RequestException):
 class ConflictingHeaderError(RequestException):
     """Mutually exclusive request headers set"""
 
-class UnreachableCodeError(RequestException, RuntimeError):
-    """Unreachable code block reached"""
+class InvalidBodyError(RequestException, ValueError):
+    """An invalid request body was specified"""
 
 # Warnings
 
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -98,6 +98,9 @@ class StreamConsumedError(RequestException, TypeError):
 class RetryError(RequestException):
     """Custom retries logic failed"""
 
+class ConflictingHeaderError(RequestException):
+    """Mutually exclusive request headers set"""
+
 
 # Warnings
 
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -101,6 +101,8 @@ class RetryError(RequestException):
 class ConflictingHeaderError(RequestException):
     """Mutually exclusive request headers set"""
 
+class UnreachableCodeError(RequestException, RuntimeError):
+    """Unreachable code block reached"""
 
 # Warnings
 
--- a/requests/models.py
+++ b/requests/models.py
@@ -470,6 +470,8 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
                 self.headers['Content-Length'] = builtin_str(length)
             elif is_stream and not length:
                 self.headers['Transfer-Encoding'] = 'chunked'
+            else:
+                assert False, "If body is not null, it must either have a length or be streamable"
         elif (self.method not in ('GET', 'HEAD')) and (self.headers.get('Content-Length') is None):
             self.headers['Content-Length'] = '0'
 
--- a/requests/models.py
+++ b/requests/models.py
@@ -23,11 +23,11 @@ from .packages.urllib3.exceptions import (
     DecodeError, ReadTimeoutError, ProtocolError, LocationParseError)
 from .exceptions import (
     HTTPError, MissingSchema, InvalidURL, ChunkedEncodingError,
-    ContentDecodingError, ConnectionError, StreamConsumedError)
+    ContentDecodingError, ConnectionError, StreamConsumedError, ConflictingHeaderError)
 from .utils import (
     guess_filename, get_auth_from_url, requote_uri,
     stream_decode_response_unicode, to_key_val_list, parse_header_links,
-    iter_slices, guess_json_utf, super_len, to_native_string)
+    iter_slices, guess_json_utf, super_len, to_native_string, determine_if_stream)
 from .compat import (
     cookielib, urlunparse, urlsplit, urlencode, str, bytes, StringIO,
     is_py2, chardet, builtin_str, basestring)
@@ -423,15 +423,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             content_type = 'application/json'
             body = complexjson.dumps(json)
 
-        is_stream = all([
-            hasattr(data, '__iter__'),
-            not isinstance(data, (basestring, list, tuple, dict))
-        ])
-
-        try:
-            length = super_len(data)
-        except (TypeError, AttributeError, UnsupportedOperation):
-            length = None
+        is_stream = determine_if_stream(data)
 
         if is_stream:
             body = data
@@ -439,10 +431,6 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             if files:
                 raise NotImplementedError('Streamed bodies and files are mutually exclusive.')
 
-            if length:
-                self.headers['Content-Length'] = builtin_str(length)
-            else:
-                self.headers['Transfer-Encoding'] = 'chunked'
         else:
             # Multi-part file uploads.
             if files:
@@ -455,27 +443,38 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
                     else:
                         content_type = 'application/x-www-form-urlencoded'
 
-            self.prepare_content_length(body)
-
             # Add content-type if it wasn't explicitly provided.
             if content_type and ('content-type' not in self.headers):
                 self.headers['Content-Type'] = content_type
 
+        self.prepare_content_length(body)
         self.body = body
 
     def prepare_content_length(self, body):
-        if hasattr(body, 'seek') and hasattr(body, 'tell'):
-            curr_pos = body.tell()
-            body.seek(0, 2)
-            end_pos = body.tell()
-            self.headers['Content-Length'] = builtin_str(max(0, end_pos - curr_pos))
-            body.seek(curr_pos, 0)
-        elif body is not None:
-            l = super_len(body)
-            if l:
-                self.headers['Content-Length'] = builtin_str(l)
+        """Prepares Content-Length header.
+        
+        If the length of the body of the request can be computed, Content-Length is set using
+        super_len. If user has manually set either a Transfer-Encoding or Content-Length header
+        when it should not be set (they should be mutually exclusive) an ConflictingHeaderError
+        error will be raised.
+        """
+        if body is not None:
+            is_stream = determine_if_stream(body)
+
+            try:
+                length = super_len(body)
+            except (TypeError, AttributeError, UnsupportedOperation):
+                length = None
+
+            if length:
+                self.headers['Content-Length'] = builtin_str(length)
+            elif is_stream and not length:
+                self.headers['Transfer-Encoding'] = 'chunked'
         elif (self.method not in ('GET', 'HEAD')) and (self.headers.get('Content-Length') is None):
             self.headers['Content-Length'] = '0'
+
+        if 'Transfer-Encoding' in self.headers and 'Content-Length' in self.headers:
+            raise ConflictingHeaderError('Transfer-Encoding and Content-Length headers both set')
 
     def prepare_auth(self, auth, url=''):
         """Prepares the given HTTP auth data."""
--- a/requests/models.py
+++ b/requests/models.py
@@ -471,7 +471,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             elif is_stream and not length:
                 self.headers['Transfer-Encoding'] = 'chunked'
             else:
-                assert False, "If body is not null, it must either have a length or be streamable"
+                raise UnreachableCodeError("Non-null body must have length or be streamable")
         elif (self.method not in ('GET', 'HEAD')) and (self.headers.get('Content-Length') is None):
             self.headers['Content-Length'] = '0'
 
--- a/requests/models.py
+++ b/requests/models.py
@@ -471,7 +471,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             elif is_stream and not length:
                 self.headers['Transfer-Encoding'] = 'chunked'
             else:
-                raise UnreachableCodeError("Non-null body must have length or be streamable")
+                raise InvalidBodyError("Non-null body must have length or be streamable")
         elif (self.method not in ('GET', 'HEAD')) and (self.headers.get('Content-Length') is None):
             self.headers['Content-Length'] = '0'
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -731,6 +731,6 @@ def urldefragauth(url):
 def determine_if_stream(data):
     """Given data, determines if it should be sent as a stream.
     """
-    is_iterable = hasattr(data, '__iter__')
+    is_iterable = getattr(data, '__iter__', False)
     is_io_type = not isinstance(data, (basestring, list, tuple, dict))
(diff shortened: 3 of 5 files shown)


Lukasa:
So, I think in general I like this better. However, your point about `super_len` is a real one.

I think the solution here is to rewrite `prepare_content_length` to have it always call `super_len`. `super_len` does use `tell` to do a little dance here, so I think it's probably going to be good enough: @sigmavirus24, does that seem right to you?

davidsoncasey:
@Lukasa in that case, could we get rid of the first elif statement and just go to the case where `body is not None`? Looking at `super_len`, I think it would calculate length in the same way. So in what I'm proposing, prepare content length would look like this:

``` python
    def prepare_content_length(self, body):
        if 'Transfer-Encoding' in self.headers:
            return
        elif body is not None:
            l = super_len(body)
            if l:
                self.headers['Content-Length'] = builtin_str(l)
        elif (self.method not in ('GET', 'HEAD')) and (self.headers.get('Content-Length') is None):
            self.headers['Content-Length'] = '0'
```

All tests pass with this change.

Lukasa:
Ok, I think I'm happy with this, though I'd like @sigmavirus24 to review before we merge. =)

sigmavirus24:
I'm not sure of the right way to do this, but this will allow someone to do:

``` py
r = requests.post(url, headers={'Transfer-Encoding': 'chunked'}, data='foo')
```

Which will generate an invalid request to the server. I would much rather forcibly strip the TE header from something where we can determine the length than ignore adding the Content-Length header if a TE is present.

Lukasa:
@sigmavirus24 Arg, good spot. Hrm. We need to have some extra logic here to get this right.

davidsoncasey:
@Lukasa @sigmavirus24 it looks to me like the try/except of using `super_len` to calculate the length of the body could be moved to `prepare_content_length` - and then both the `Content-Length` and `Transfer-Encoding` headers would be set there. If content length can be calculated, then we strip the TE header, and otherwise, we strip the CL header. I think this would be the most clear, and would ensure that the headers are mutually exclusive, regardless of what a user enters.

Lukasa:
@davidsoncasey I think we want to keep the `try...except` in `super_len`: that method gets used outside requests quite frequently. It won't hurt to do the check twice.

davidsoncasey:
@Lukasa sounds good. I'll see if I can put something together for that. Thanks for helping work through this.

davidsoncasey:
@Lukasa @sigmavirus24 I updated this PR to make the Content-Length and Transfer-Encoding headers mutually exclusive, regardless of whether a user manually provides a value, as @sigmavirus24 showed in his comment. While working on this, I stumbled upon #1648. I hadn't realized that this had been a subject of so much discussion, and that people have differing opinions about how this should work. This is obviously related to that issue, so I understand if you decide that this isn't the fix that you're looking for. It could be refactored to raise an error if a user has provided a CL header when TE is set, and vice versa, instead of silently removing the unneeded header. Let me know what you think.

requests/models.py
```diff
@@ -439,9 +439,7 @@ def prepare_body(self, data, files, json=None):
             if files:
                 raise NotImplementedError('Streamed bodies and files are mutually exclusive.')
 
-            if length:
-                self.headers['Content-Length'] = builtin_str(length)
-            else:
+            if not length:
```
sigmavirus24: /me idly wonders if length can be 0

requests/models.py
```diff
@@ -455,27 +453,32 @@ def prepare_body(self, data, files, json=None):
...
-            self.headers['Content-Length'] = '0'
+        """Prepares Content-Length header.
+        
+        If Transfer-Encoding header has been set, we want to make sure that Content-Length
+        header is not set. If setting Content-Length, we will strip Transfer-Encoding header.
+        """
+
+        if 'Transfer-Encoding' in self.headers:
+            if 'Content-Length' in self.headers:
+                del headers['Content-Length']
```
sigmavirus24: Which of your tests exercise this?

sigmavirus24:
I'm not sure which behaviour I want (if we **must** merge something before 3.0). I think I'd rather raise an exception though.

requests/models.py
Lukasa: We very deliberately decided, several months ago, to handle things that appear to have zero length by chunk-encoding them. This protects against the possibility that we computed a zero length for something that was in fact length undefined.

sigmavirus24: Ah right. Okay.

Lukasa:
I'd be inclined to want to raise an exception, but I also think we should consider discussing this at PyCon.

requests/models.py
davidsoncasey: @sigmavirus24 I added a test at the start of this issue to ensure that both headers are not set in the 0 length case, but I'll add another couple tests to ensure that it works correctly in the case that a user specifies a header that would be invalid.

davidsoncasey:
@Lukasa @sigmavirus24 I'll go ahead and alter it to raise an exception for now (perhaps `InvalidHeadersError`? Or is there an existing exception that would make sense to raise?). And then we can leave this PR open until it's decided what the best behavior is.

Also, are you planning on coming to Portland for PyCon? I didn't get a ticket in time, but I live in Portland and it would be great to meet up and discuss the project.

Lukasa:
Yup, we'll be in Portland: we're planning to have a "state of Requests" chat at some point that'll be able to tackle things like these.

davidsoncasey:
Fantastic, I'd love to participate.

davidsoncasey:
@Lukasa @sigmavirus24 I made a few more changes and a bit more refactoring of `prepare_body` and `prepare_content_length`. These include:
- Moving the logic set `Transfer-Encoding` header to the `prepare_content_length` method. This makes it more explicit that the headers should be mutually exclusive, and does not rely on implicitly connected logic in two methods.
- Raises an `InvalidHeaderError` if a user manually sets either header when it should not be set (let me know if you think that there's a preexisting exception that could be raised instead).
- Added tests to check different cases of when each header should be set.

I understand we may need to wait to decide if this is the desired behavior, so no need to review or merge this right away. We can leave this PR open until you've had a chance to discuss and agree upon the desired behavior.

requests/models.py
```diff
@@ -455,28 +443,39 @@ def prepare_body(self, data, files, json=None):
...
+
+            try:
+                length = super_len(body)
+            except (TypeError, AttributeError, UnsupportedOperation):
+                length = None
+
+            if length:
+                self.headers['Content-Length'] = builtin_str(length)
+            elif is_stream and not length:
+                self.headers['Transfer-Encoding'] = 'chunked'
```
Lukasa, line 472: So, what happens here if `is_stream` is `False` and `length` is Falsy? That is, for a file whose length we can't determine, for example? Is there any reason not to want to set `Transfer-Encoding: chunked` in that case?

Lukasa:
I've got a note inline, but I don't mind this change: it's definitely backward-incompatible though, so we'll need to release this as part of 3.0.0.

requests/models.py
sigmavirus24, line 472: Yeah, if we don't set either Content-Length or Transfer-Encoding, this would be an invalid request.

requests/exceptions.py
```diff
@@ -98,6 +98,9 @@ class StreamConsumedError(RequestException, TypeError):
 class RetryError(RequestException):
     """Custom retries logic failed"""
 
+class InvalidHeaderError(RequestException):
```
sigmavirus24: This should be `ConflictingHeaderError`. The headers themselves are not invalid, they're an invalid combination, which you even mention in the docstring. Let's not tempt people to start using this to validate headers or any other such nonsense.

requests/utils.py
```diff
@@ -726,3 +726,10 @@ def urldefragauth(url):
     netloc = netloc.rsplit('@', 1)[-1]
 
     return urlunparse((scheme, netloc, path, params, query, ''))
+
```
sigmavirus24, line 729: Please add another newline here.

sigmavirus24:
Once we've cleaned this up, can we squash some of these commits together and send this to proposed/3.0.0?

requests/exceptions.py
davidsoncasey: Sounds good

requests/models.py
davidsoncasey, line 472: I was trying to match the previous behavior here, which as far as I could tell only set `Transfer-Encoding` in the case that `is_stream` is true. As far as a file whose length can't be determined, that should all be handled by the `encode_files` method - I don't believe that these changes should affect that behavior. I'll try to write another test for this case.

davidsoncasey:
@sigmavirus24 @Lukasa sounds good. I'll make those small cleanup changes and squash extraneous commits.

requests/models.py
Lukasa, line 472: Maybe I'm wrong, but I think `encode_files` is only called if the `files` kwarg is used. I'm talking about the situation where a file object is passed directly to the `data` argument.

davidsoncasey, line 472: Ah I misunderstood. In that case  `is_stream` is would be true (perhaps the name is misleading), so if length is Falsy then it would set `Transfer-Encoding: chunked`. I think this was actually the case that got me working on this initially - a file like object with length 0 was being passed as the `data` argument.

sigmavirus24, line 472: Our point is that what happens when it's something without a length, that is not streaming. That seems to be a corner case not presently handled (at least not explicitly). We kind of handle it in the next elif but only for certain methods and I'm starting to think we just need a comment explaining why that handling is okay.

davidsoncasey, line 472: I see what you're saying. Do you think it would make sense to change the variable name for `is_stream`? Or do you think just adding a comment is good?

Lukasa, line 472: I think mostly we need to just understand why the code potentially allows dropping out of this function without setting either a `Content-Length` or `Transfer-Encoding` header. That can be a comment, that can be refactoring the variable names, or it can be as much as adding an `else: assert False, "Unreachable code"` kinda section (except one that is a bit clearer in its error message).

davidsoncasey, line 472: I like the idea (at least initially) of raising an exception in that case, if we agree that one header or the other should always be set. Right now the obvious case that would skip all logic in the method is if `body = None` and the request method is either GET or HEAD. I guess I'm not sure what the desired behavior is in this case - it's interesting that a body of `None` is treated differently than a body that is an empty string.

Lukasa, line 472: The only case where it's acceptable to set neither header is on GET/HEAD requests, which by default have no bodies.

davidsoncasey:
I squashed some commits together and it looks like there's a conflict now. I'll get a fresh pull of the master branch and get that resolved.

davidsoncasey:
@Lukasa @sigmavirus24 alright, it took me a bit to get back to this, but I've cleaned it up a bit and it should be ready to merge into the proposed branch. The conlflict is in AUTHORS.rst. Let me know if there's anything else here you'd like to see changed, otherwise I'll try to find another issue to get started on.

requests/utils.py
```diff
@@ -726,3 +726,11 @@ def urldefragauth(url):
...
+
+def determine_if_stream(data):
+    """Given data, determines if it should be sent as a stream.
+    """
+    is_iterable = hasattr(data, '__iter__')
```
ueg1990: i am not sure if it is relevant in this case, but maybe using getattr is a better option here since hasattr has some issues in Python 2: https://hynek.me/articles/hasattr/

requests/models.py
```diff
@@ -455,28 +443,41 @@ def prepare_body(self, data, files, json=None):
...
+            else:
+                assert False, "If body is not null, it must either have a length or be streamable"
```
Lukasa: Rather than assert, let's define an exception we can raise here. That'll at least be somewhat cleaner. Or maybe we can just use a `RuntimeError`.

requests/utils.py
Lukasa: Probably doing the `__getattr__` dance from @hynek's blog post is a good idea here, thanks @ueg1990.

Lukasa:
Thanks @davidsoncasey, I've left a few more notes.

requests/models.py
davidsoncasey: @Lukasa any preference between the two? I'd probably lean towards defining a new exception, but I'm not sure if you'd rather keep the exceptions file a bit shorter and not define and exception that's only used once.

requests/utils.py
davidsoncasey: Sounds good!

requests/models.py
Lukasa: I'd be happy to define a new exception if it inherits appropriately.

davidsoncasey: Sounds good. `UnreachableCodeError` sound like a good name?

requests/exceptions.py
```diff
@@ -101,6 +101,8 @@ class RetryError(RequestException):
 class ConflictingHeaderError(RequestException):
     """Mutually exclusive request headers set"""
 
+class UnreachableCodeError(RequestException, RuntimeError):
+    """Unreachable code block reached"""
```
Lukasa: Heh, while I like this name I don't think it's a good one. It's too general. =)

Let's instead define a new exception that is more like this:

``` python
class InvalidBodyError(RequestException, ValueError):
    """An invalid request body was specified."""
```

davidsoncasey: Sounds good. I just made that change.

Lukasa:
Ok, good by me. @sigmavirus24?

sigmavirus24:
So this looks great, but:
1. Should this be against proposed/3.0?
2. Can we resolve the merge conflicts?

requests/models.py
```diff
@@ -455,28 +443,41 @@ def prepare_body(self, data, files, json=None):
...
             if content_type and ('content-type' not in self.headers):
                 self.headers['Content-Type'] = content_type
 
+        self.prepare_content_length(body)
         self.body = body
 
     def prepare_content_length(self, body):
-        if hasattr(body, 'seek') and hasattr(body, 'tell'):
-            curr_pos = body.tell()
-            body.seek(0, 2)
```
kennethreitz, line 469: if all of this seek/tell behavior perfectly replicated in super_len?

Lukasa, line 469: No, it's handled differently but better. Regardless, it needs to be reconciled: we were inconsistently using either this logic or `super_len` prior to this patch, which means that in unlikely scenarios we could get different answers to the content-length of the same thing.

kennethreitz, line 469: This section should have been better commented to understand why these steps were added where/when they were. I'd hate for something to fall through the cracks.

I'm excited about the simplification, given that it isn't lossy.

davidsoncasey, line 469: My impression at the time was that it would produce the same calculation results, but the calculation is done slightly differently. @kennethreitz you raise a good point - using `len(o.getvalue())` in `super_len` seems like it could be less efficient than using seek/tell to calculate the length. I'm not sure if it would make sense to change this in `super_len`? Seems like that should be a separate PR though.

davidsoncasey:
@sigmavirus24 I'll open a new PR into proposed/3.0. As far as the conflict, that was only in the contributors.rst, and I was thinking it would be a little cleaner if one of you resolved it on merge, instead of back merging master into this branch and fixing it there. I'm happy to do that if you like though.

kennethreitz:
needs a rebase

kennethreitz:
closing due to inactivity (not sure if that's our fault or not). please re-open if you're still interested in getting this in!

PR #3185, opened 2016-05-11
Support ALL_PROXY environment variable

### LINKED ISSUE #3183: Feature request: support ALL_PROXY environment variable
As of 2.10.0, I can set `HTTP_PROXY=socks5://example.com:port`, and then requests will use that SOCKS proxy. However, other tools (e.g. Docker) currently require the use of `ALL_PROXY` instead, and get confused if you set `HTTP_PROXY` to a URL that begins with `socks5://`. I think Docker should support the use of `HTTP_PROXY`, but requests should also support the use of `ALL_PROXY`, especially with the recent (thank you thank you thank you) addition of SOCKS support.

I thought that this might need to go through urllib3, but it seems like it would be a relatively simple change to [sessions.py], to look at the `'all'` key before checking the scheme. But, then I read your contributor's guide and thought it best to post an issue before just doing it. :smile:

Please let me know what you think.

[sessions.py]: https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L212

brettdh:
Closes #3183.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -580,11 +580,20 @@ def select_proxy(url, proxies):
     proxies = proxies or {}
     urlparts = urlparse(url)
     if urlparts.hostname is None:
-        proxy = None
-    else:
-        proxy = proxies.get(urlparts.scheme+'://'+urlparts.hostname)
-    if proxy is None:
-        proxy = proxies.get(urlparts.scheme)
+        return proxies.get('all', proxies.get(urlparts.scheme))
+
+    proxy_keys = [
+        'all://' + urlparts.hostname,
+        'all',
+        urlparts.scheme + '://' + urlparts.hostname,
+        urlparts.scheme,
+    ]
+    proxy = None
+    for proxy_key in proxy_keys:
+        if proxy_key in proxies:
+            proxy = proxies[proxy_key]
+            break
+
     return proxy
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,7 @@ Jinja2==2.8
 MarkupSafe==0.23
 py==1.4.31
 Pygments==2.1.1
+PySocks==1.5.6
 pytest==2.8.7
 pytest-cov==2.2.1
 pytest-httpbin==0.2.0
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -12,7 +12,7 @@ def test_chunked_upload():
     """can safely send generators"""
     close_server = threading.Event()
     server = Server.basic_response_server(wait_to_close_event=close_server)
-    data = (i for i in [b'a', b'b', b'c'])
+    data = iter([b'a', b'b', b'c'])
 
     with server as (host, port):
         url = 'http://{0}:{1}/'.format(host, port)
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -1,14 +1,18 @@
+import os
+import pytest
 import threading
 import requests
 
 from tests.testserver.server import Server
+
+from .utils import override_environ
 
 
 def test_chunked_upload():
     """can safely send generators"""
     close_server = threading.Event()
     server = Server.basic_response_server(wait_to_close_event=close_server)
-    data = (i for i in [b'a', b'b', b'c']) 
+    data = (i for i in [b'a', b'b', b'c'])
 
     with server as (host, port):
         url = 'http://{0}:{1}/'.format(host, port)
@@ -17,3 +21,36 @@ def test_chunked_upload():
 
     assert r.status_code == 200
     assert r.request.headers['Transfer-Encoding'] == 'chunked'
+
+
+_schemes_by_var_prefix = [
+    ('http', ['http']),
+    ('https', ['https']),
+    ('all', ['http', 'https']),
+]
+
+_proxy_combos = []
+for prefix, schemes in _schemes_by_var_prefix:
+    for scheme in schemes:
+        _proxy_combos.append(("{0}_proxy".format(prefix), scheme))
+
+_proxy_combos += [(var.upper(), scheme) for var, scheme in _proxy_combos]
+
+
+@pytest.mark.parametrize("var,scheme", _proxy_combos)
+def test_use_proxy_from_environment(httpbin, var, scheme):
+    url = "{0}://httpbin.org".format(scheme)
+    fake_proxy = Server()  # do nothing with the requests; just close the socket
+    with fake_proxy as (host, port):
+        proxy_url = "socks5://{0}:{1}".format(host, port)
+        kwargs = {var: proxy_url}
+        with override_environ(**kwargs):
+            # fake proxy's lack of response will cause a ConnectionError
+            with pytest.raises(requests.exceptions.ConnectionError):
+                requests.get(url)
+
+        # the fake proxy received a request
+        assert len(fake_proxy.handler_results) == 1
+
+        # it had actual content (not checking for SOCKS protocol for now)
+        assert len(fake_proxy.handler_results[0]) > 0
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -30,6 +30,7 @@ from requests.models import urlencode
 from requests.hooks import default_hooks
 
 from .compat import StringIO, u
+from .utils import override_environ
 
 # Requests to this URL should always fail with a connection timeout (nothing
 # listening on that port)
@@ -1546,21 +1547,6 @@ def test_requests_are_updated_each_time(httpbin):
         assert response.request.method == 'GET'
         send_call = SendCall((response.request,), default_keyword_args)
         assert session.calls[-1] == send_call
-
-
-@contextlib.contextmanager
-def override_environ(**kwargs):
-    save_env = dict(os.environ)
-    for key, value in kwargs.items():
-        if value is None:
-            del os.environ[key]
-        else:
-            os.environ[key] = value
-    try:
-        yield
-    finally:
-        os.environ.clear()
-        os.environ.update(save_env)
 
 
 @pytest.mark.parametrize("var,url,proxy", [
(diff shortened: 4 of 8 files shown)


brettdh:
ping @Lukasa

brettdh:
Agh, Python3 (iteritems). Pushing fix with six shortly.

On a related note - has there ever been talk of moving to [tox](http://tox.readthedocs.io/en/latest/) for testing?

tests/test_requests.py
```diff
@@ -1548,6 +1549,42 @@ def test_requests_are_updated_each_time(httpbin):
         assert session.calls[-1] == send_call
 
 
+@contextlib.contextmanager
+def override_environ(**kwargs):
+    save_env = dict(os.environ)
+    for key, value in iteritems(kwargs):
```
sigmavirus24: Please just use `dict.items`

brettdh:
Found that this doesn't actually work yet, as evidenced by the new, failing functional test. Will investigate further (and fix the other broken test) tomorrow.

Also, the jenkins setup doesn't seem to have PySocks installed. What's the best way to make sure that's present in the CI environment? Just add it to the requirements.txt?

Lukasa:
@brettdh Yeah, I think that would be fine. =)

brettdh:
Alrighty, I think it's good now.

requests/utils.py
```diff
@@ -579,12 +579,19 @@ def select_proxy(url, proxies):
...
+            'all://' + urlparts.hostname,
+            'all',
+            urlparts.scheme + '://' + urlparts.hostname,
+            urlparts.scheme,
+        ]
+        for proxy_key in proxy_keys:
+            if proxy_key in proxies:
+                proxy = proxies[proxy_key]
+                break
+
```
Lukasa, line 596: This code has logically changed. Before this PR, if the hostpart of the url was `None`, we'd lookup the proxy by scheme alone. Now we don't do that: if it's `None`, we return no proxy at all because we never enter the `if` block. Care to refactor this?

tests/test_lowlevel.py
```diff
@@ -17,3 +21,36 @@ def test_chunked_upload():
 
     assert r.status_code == 200
     assert r.request.headers['Transfer-Encoding'] == 'chunked'
+
+
+_schemes_by_var_prefix = {
+    'http': ['http'],
+    'https': ['https'],
+    'all': ['http', 'https'],
+}
```
Lukasa: There's nothing gained by this being a dictionary because we never do a lookup in it. Mind changing to a list of two-tuples and then remove the call to `.items` lower down?

requests/utils.py
brettdh, line 596: Whoops. Sure, will do.

tests/test_lowlevel.py
```diff
@@ -17,3 +21,36 @@ def test_chunked_upload():
...
+@pytest.mark.parametrize("var,scheme", _proxy_combos)
+def test_use_proxy_from_environment(httpbin, var, scheme):
+    url = "{0}://httpbin.org".format(scheme)
+    fake_proxy = Server()  # do nothing with the requests; just close the socket
+    with fake_proxy as (host, port):
+        proxy_url = "socks5://{0}:{1}".format(host, port)
+        kwargs = {var: proxy_url}
+        with override_environ(**kwargs):
+            with pytest.raises(requests.exceptions.ConnectionError):
+                print("Expecting requests to pick up env var: {0}={1}".format(var, os.environ[var]))
```
Lukasa: Can we not have prints in the tests?

tests/testserver/server.py
```diff
@@ -79,13 +81,33 @@ def _create_socket_and_bind(self):
...
+    def _close_server_sock_ignore_errors(self):
+        try:
+            self.server_sock.close()
+        except IOError:
+            pass
+
     def _handle_requests(self, server_sock):
-        for _ in range(self.requests_to_handle):
-            sock = server_sock.accept()[0]
+        for i in range(self.requests_to_handle):
```
Lukasa: Why have you added an unused variable here?

tests/testserver/server.py
```diff
@@ -79,13 +81,33 @@ def _create_socket_and_bind(self):
...
+    def _accept_connection(self, server_sock):
+        try:
+            ready, _, _ = select.select([server_sock], [], [])
+            if not ready:
+                return None
+
+            return server_sock.accept()[0]
+        except (select.error, socket.error) as e:
+            print("Failed to accept connection: {0}".format(e))
+            return None
```
Lukasa: Why are we suppressing these errors instead of letting them kill the test?

tests/testserver/server.py
brettdh: It was used in a debug print that I later removed. Will revert.

tests/testserver/server.py
brettdh: I wanted to be able to interrupt the server thread if an exception occurred in the main thread. In the instance that this fails, something is going wrong elsewhere, and I feel that the stack traces from the failed select/accept will just obscure the real failure.

requirements.txt
```diff
@@ -10,6 +10,7 @@ Jinja2==2.8
 MarkupSafe==0.23
 py==1.4.31
 Pygments==2.1.1
+PySocks==1.5.6
```
sigmavirus24, line 13: Why is this necessary?

tests/testserver/server.py
```diff
@@ -25,10 +25,11 @@ class Server(threading.Thread):
     """Dummy server using for unit testing"""
     WAIT_EVENT_TIMEOUT = 5
 
-    def __init__(self, handler, host='localhost', port=0, requests_to_handle=1, wait_to_close_event=None):
+    def __init__(self, handler=None, host='localhost', port=0,
```
sigmavirus24: Please do not reformat lines randomly.

requirements.txt
brettdh, line 13: https://github.com/kennethreitz/requests/pull/3185#issuecomment-218884018

The new functional test tries to connect to a (fake) socks proxy, which requires PySocks.

tests/testserver/server.py
```diff
@@ -39,6 +40,10 @@ def __init__(self, handler, host='localhost', port=0, requests_to_handle=1, wait
         self.ready_event = threading.Event()
         self.stop_event = threading.Event()
 
+        self.server_sock = self._create_socket_and_bind()
```
sigmavirus24: Why did you move this into `__init__`?

tests/testserver/server.py
sigmavirus24: > I feel that the stack traces from the failed select/accept will just obscure the real failure.

I don't agree. Do you have evidence for this or is it purely intuition that is causing you to think this? If you have the former, I'd rather see this addressed separately with that evidence provided.

tests/testserver/server.py
brettdh: Added an argument briefly (which made this rather long, though now I remember that one line in the contributor's guide), then removed it, and forgot to revert the formatting change. Apologies. Will revert.

tests/testserver/server.py
```diff
@@ -79,13 +81,33 @@ def _create_socket_and_bind(self):
...
-            sock = server_sock.accept()[0]
+            sock = self._accept_connection(server_sock)
+            if not sock:
+                break
+
             handler_result = self.handler(sock)
 
             self.handler_results.append(handler_result)
 
+    def _accept_connection(self, server_sock):
```
sigmavirus24: This is an instance method that doesn't use `self` and accepts as a value something stored on `self`, yes?

sigmavirus24:
It's rather difficult to review this with the amount of refactoring of the testserver. It's distracting from the actual feature and reviewing that.

brettdh:
@sigmavirus24 The test server refactoring was mainly done as I added a failing functional test for the feature, in order to prevent that test failure from hanging the server (because it was waiting for a connection that wasn't coming).

Now that the test is passing, I'm happy to separate the test server refactoring into a separate PR if that helps.

tests/testserver/server.py
brettdh: The former. Will detail when giving the test server refactoring its own PR.

tests/testserver/server.py
brettdh: Yes, as is `_handle_requests` (well, the latter, anyway). Will tidy both.

tests/testserver/server.py
brettdh: `server_sock` was not previously an instance variable, and I wanted to be able to access it outside of `_handle_requests` (e.g. `__exit__`).

brettdh:
Ok, I split off all but the minimum changes needed for the new test. I'll make a new PR for the test server refactoring soon.

Lukasa:
Cool, @sigmavirus24 do you want to re-review?

tests/test_lowlevel.py
```diff
@@ -1,14 +1,18 @@
...
 
+from .utils import override_environ
+
 
 def test_chunked_upload():
     """can safely send generators"""
     close_server = threading.Event()
     server = Server.basic_response_server(wait_to_close_event=close_server)
-    data = (i for i in [b'a', b'b', b'c']) 
+    data = (i for i in [b'a', b'b', b'c'])
```
sigmavirus24: Why not just `iter([b'a', b'b', b'c'])`?

sigmavirus24:
Eh I realize you just trimmed that whitespace, but you drew my attention to it ;). Otherwise, this looks okay to me.

brettdh:
@sigmavirus24 Yeah, sorry; I forgot that I had vim configured to always trim trailing whitespace on save. Thought I had reverted those before committing, but clearly I missed one. :-)

sigmavirus24:
@brettdh no need to put trailing whitespace back, just change that line to use iter instead =D

Lukasa:
Ok, :+1: from me. @sigmavirus24?

sigmavirus24:
Looks good to me. Thanks @brettdh!

PR #3192, opened 2016-05-14
Allow graceful interruption of testserver.Server

brettdh:
So that failing tests don't cause the server thread to hang
indefinitely, waiting for connections that will never come.

Rationale for suppressing error/traceback from interrupted
_accept_connection in testserver.Server:
https://gist.github.com/brettdh/b6e741227b2297f19d2118077f14dfa5

### DIFF
--- a/tests/test_testserver.py
+++ b/tests/test_testserver.py
@@ -149,3 +149,14 @@ class TestTestServer:
 
         # if the server thread fails to finish, the test suite will hang
         # and get killed by the jenkins timeout.
+
+    def test_server_finishes_when_no_connections(self):
+        """the server thread exits even if there are no connections"""
+        server = Server.basic_response_server()
+        with server:
+            pass
+
+        assert len(server.handler_results) == 0
+
+        # if the server thread fails to finish, the test suite will hang
+        # and get killed by the jenkins timeout.
--- a/tests/test_testserver.py
+++ b/tests/test_testserver.py
@@ -139,11 +139,9 @@ class TestTestServer:
     def test_server_finishes_on_error(self):
         """the server thread exits even if an exception exits the context manager"""
         server = Server.basic_response_server()
-        try:
+        with pytest.raises(Exception):
             with server:
                 raise Exception()
-        except Exception:
-            pass
 
         assert len(server.handler_results) == 0
 
--- a/tests/test_testserver.py
+++ b/tests/test_testserver.py
@@ -135,3 +135,17 @@ class TestTestServer:
             sock.close()
             
         assert server.handler_results[0] == data
+
+    def test_server_finishes_on_error(self):
+        """the server thread exits even if an exception exits the context manager"""
+        server = Server.basic_response_server()
+        try:
+            with server:
+                raise Exception()
+        except Exception:
+            pass
+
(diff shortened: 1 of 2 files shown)


brettdh:
ping @Lukasa @sigmavirus24 

I think I addressed the comments from #3185.

Oddly, I cannot make the test suite hang indefinitely when I run it locally, but that's what's happening in jenkins. :confused: Maybe you can spot something I'm missing?

tests/test_testserver.py
```diff
@@ -135,3 +135,28 @@ def test_request_recovery_with_bigger_timeout(self):
...
         assert server.handler_results[0] == data
+
+    def test_server_finishes_on_error(self):
+        """the server thread exits even if an exception exits the context manager"""
+        server = Server.basic_response_server()
+        try:
+            with server:
+                raise Exception()
+        except Exception:
+            pass
```
Lukasa: Can we assert that the raise makes it out? `with pytest.raises()` would do that, rather than a `try...except`.

tests/testserver/server.py
```diff
@@ -39,6 +39,10 @@ def __init__(self, handler, host='localhost', port=0, requests_to_handle=1, wait
         self.ready_event = threading.Event()
         self.stop_event = threading.Event()
 
+        self.server_sock = self._create_socket_and_bind()
```
Lukasa: I'm _really_ nervous about this being in `__init__`. Is there any reason it's here?

tests/testserver/server.py
```diff
@@ -79,13 +80,32 @@ def _create_socket_and_bind(self):
...
+            if not sock:
+                break
+
             handler_result = self.handler(sock)
 
             self.handler_results.append(handler_result)
 
+    def _accept_connection(self):
+        try:
+            ready, _, _ = select.select([self.server_sock], [], [])
```
Lukasa: Should this have a timeout on it? Right now this should block forever unless the listening socket is closed.

tests/testserver/server.py
sigmavirus24: I asked this on the last PR and didn't get an answer.

tests/testserver/server.py
brettdh: Yes! Thank you. That might be the reason for the failure during CI. Will add a timeout.

tests/testserver/server.py
brettdh: I gave an answer; it was just a rather unsatisfying one about setting instance variables in the constructor. :smile:

I think I was not seeing the synchronization between `__enter__` and `run` earlier, which made me want to set this before starting the server thread. I'll go ahead and move this back to `run`.

tests/testserver/server.py
brettdh: Great; that seems to have fixed the CI timeout.

Lukasa:
Ok, cool, I'm happy with this. Assuming @sigmavirus24 is as well, he may merge.

sigmavirus24:
I'm going to squash merge because the title of 5cfb691 is not very helpful and I don't want it in our commit history if we go to bisect things.

sigmavirus24:
Thanks @brettdh!

PR #3195, opened 2016-05-16
Fix to check for Plain ip notations in no_proxy settings if not CIDR

kumarvaradarajulu:
For ipv4 addresses no_proxy is not being honored. 

[This line ](https://github.com/kennethreitz/requests/blob/master/requests/utils.py#L539) checks for `cidr notation` but `plain ip` notation is not considered, due to which the request is always routed to proxy server. This results in error ReadTimeoutError

Sample `proxy` configuration

    no_proxy=192.168.1.1,192.168.1.3,*.example.com
    http_proxy=http://192.168.1.100:8080
    https_proxy=http://192.168.1.100:8080

>When request is made to `192.168.1.1`, the request should not be made to proxy server, without this fix it is being made to proxy server `192.168.1.100`

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -131,42 +131,6 @@ class TestGetEnvironProxies:
         assert get_environ_proxies(url) != {}
 
 
-class TestShouldBypassProxies:
-    """
-    Tests for should_bypass_proxies function
-    """
-
-    @pytest.mark.parametrize(
-        'url, expected', (
-                ('http://192.168.0.1:5000/', True),
-                ('http://192.168.0.1/', True),
-                ('http://172.16.1.1/', True),
-                ('http://172.16.1.1:5000/', True),
-                ('http://localhost.localdomain:5000/v1.0/', True),
-        ))
-    def test_should_bypass_proxies(self, url, expected, monkeypatch):
-        """
-        Test to check if proxy is bypassed
-        """
-        monkeypatch.setenv('no_proxy', '192.168.0.0/24,127.0.0.1,localhost.localdomain,172.16.1.1')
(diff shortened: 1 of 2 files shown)


Lukasa:
This seems good to me. Can we have a test to validate that this function works as expected?

kumarvaradarajulu:
@Lukasa added tests, please check and merge.

tests/test_utils.py
```diff
@@ -131,6 +131,42 @@ def test_not_bypass(self, url):
...
+                ('http://google.com:5000/v1.0/', False),
+        ))
+    def test_should_bypass_proxies(self, url, expected, monkeypatch):
+        """
+        Test to check if proxy is not bypassed
+        """
+        monkeypatch.setenv('no_proxy', '192.168.0.0/24,127.0.0.1,localhost.localdomain,172.16.1.1')
+        monkeypatch.setenv('NO_PROXY', '192.168.0.0/24,127.0.0.1,localhost.localdomain,172.16.1.1')
+        assert should_bypass_proxies(url) == expected
+
```
Lukasa: Is there any reason not to make these two tests into one test? Their only difference is whether they expect the bypass to work, which is a parametrized field anyway.

kumarvaradarajulu: I was thinking that, but wanted to have a difference, have combined to one test now. Please check

kumarvaradarajulu:
@Lukasa addressed comments, pls check

Lukasa:
Cool, I'm :+1: on this for now. I'd like another :+1: from @sigmavirus24 before we merge, but I think this is good.

sigmavirus24:
Looks good to me. Thanks @kumarvaradarajulu

Issue #2645, opened 2015-06-19
[RFC] Change behaviour of Response.ok in 3.0


sigmavirus24:
# Preamble

Please note that this a request for comments, not something that we will _definitely_ do in Requests.
# Problem

Right now we have an attribute defined on a Response object, `ok`. This attribute [currently](https://github.com/kennethreitz/requests/blob/9bbab338fdbb562b923ba2d8a80f0bfba697fa41/requests/models.py#L617..L623) calls `self.raise_for_status()` and catches the [exception](https://github.com/kennethreitz/requests/blob/9bbab338fdbb562b923ba2d8a80f0bfba697fa41/requests/models.py#L825..L837) raised. `raise_for_status` appropriately only raises an exception for status codes in the 4xx or 5xx range. This means that a Response with a status code in the 2xx and 3xx range will be "ok". The problem is that "ok" has a certain association in HTTP with 2xx responses, specifically 200 responses. This _may_ (I haven't looked to see if anyone has had problems with this) be misleading, especially if the user is combining their usage of the `ok` attribute with `allow_redirects=False`.
# Proposed Solution

Instead of using `raise_for_status` to determine the "ok-ness" of a response, we should compare the status code of the response directly. This will do two things:
1. It will narrow the definition of `Response.ok`
2. It will make using `Response.ok` faster. Currently we add a new stack, potentially throw and catch an exception, and then return. With a simple comparison, we could just do: `return 200 <= self.status_code < 300` which is much faster. This argument, however, doesn't hold much weight for me personally.

I'd really love @Lukasa and @kennethreitz's opinions here as well as anyone else willing to share their opinion publicly.

Lukasa:
I think that `ok` is just poorly named. I'd be more interested in having predicates that cover the major response ranges: `response.success` , `response.redirect`, `response.client_error`, `response.server_error`. We could then have overlap fields, like `response.error` (a superset of `client_error` and `server_error`).

That said, if we think that's too heavyweight, I'm still in favour of reducing the scope of `response.ok`.

sigmavirus24:
Yeah I agree that `ok` is poorly named. I'm not sure a bunch of new attributes is the right way to do this, but I'm not against it. We already have `response.is_redirect()` which, to be fair, could/should be changed to an attribute.

kennethreitz:
Against a bunch of new attributes. I'm all in favor of people doing a direct comparison themselves.

[...]

PR #3198, opened 2016-05-16
Issue #2645 - Change behavior of Response.ok

AlexPHorta:
As discussed on issue #2645, I added a success attribute to
requests.codes. Now, Response.ok checks if the response's status code
is listed on requests.codes.success.

### DIFF
--- a/requests/status_codes.py
+++ b/requests/status_codes.py
@@ -89,3 +89,18 @@ for code, titles in _codes.items():
         setattr(codes, title, code)
         if not title.startswith('\\'):
             setattr(codes, title.upper(), code)
+
+_success = {
+    200,
+    201,
+    202,
+    203,
+    204,
+    205,
+    206,
+    207,
+    208,
+    226,
+}
+
+setattr(codes, 'success', _success)
--- a/requests/status_codes.py
+++ b/requests/status_codes.py
@@ -90,7 +90,7 @@ for code, titles in _codes.items():
         if not title.startswith('\\'):
             setattr(codes, title.upper(), code)
 
-_success = {
+success = set([
     200,
     201,
     202,
@@ -101,6 +101,4 @@ _success = {
     207,
     208,
     226,
-}
-
-setattr(codes, 'success', _success)
+])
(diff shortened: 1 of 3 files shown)


requests/status_codes.py
```diff
@@ -89,3 +89,18 @@
...
+    203,
+    204,
+    205,
+    206,
+    207,
+    208,
+    226,
+}
+
+setattr(codes, 'success', _success)
```
sigmavirus24: I think this is over-engineering the solution. This also isn't the appropriate object to tack "success" onto when it's not a status code and the attribute you're setting is not an integer (like everything else on codes has been historically).

I'm okay having something like

``` py
# requests/status_codes.py
success_set = {
    # ...
}
```

``` py
# requests/models.py
from . import status_codes

class Response(..):
    @property
    def ok(self):
        return self.status_code in status_codes.success_set
```

Lukasa: :+1:

sigmavirus24: Oh, also note that on Py26, you can't use set literal notation. You have to do

``` py
success_set = set([
    200,
    201,
    # ...
])
```

AlexPHorta:
I get it, it's just that I thought the agreed solution was to add the success attribute to requests.codes. I'll update the commit right now.

Lukasa:
This LGTM. We can't merge until we're ready to do 3.0.0, though.

kennethreitz:
An additional check should be added for entire 2xx range after the registered codes are checked (or perhaps the '-1 index +1 :' slice). Both for future-proofing and a server might be using an esoteric number honoring the numbering scheme.

kennethreitz:
Also, I've shared some thoughts about this on https://github.com/kennethreitz/requests/issues/2645.

Basically, no behavioral modifications to `ok`. A new attribute can be added for the 200-only functionality, if this is still desired. 

I will need to think for a bit about the naming. feel free to use `is_successful` as a placeholder for now.

---

Maintainers: **do not merge** without my approval.

kennethreitz:
P.S. everybody gets cake :sparkles: :cake: ✨

PR #3200, opened 2016-05-19
Document the cert pooling bug

zestyping:
Hi there.  I just ran into https://github.com/kennethreitz/requests/issues/2863 and spent a lot of time trying to figure out what was wrong.  It would be really nice for future people who might encounter this bug to be informed up front, so here is a note to add to the User Guide and the API documentation.

I understand that people are working on fixing this bug, but it's not fixed yet and it's been open for about 6 months.  I think it's valuable to make this quick change to the docs in the meantime, and then when the bug is fixed we can remove these warnings.

Thanks!

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -237,6 +237,13 @@ If you specify a wrong path or an invalid cert, you'll get a SSLError::
 .. warning:: The private key to your local certificate *must* be unencrypted.
    Currently, requests does not support using encrypted keys.
 
+.. warning:: The ``cert`` option is also supported on
+   :class:`Session <requests.Session>` methods, but it should not be used.
+   Unfortunately, :class:`Session <requests.Session>` objects currently
+   pool and reuse HTTP connections regardless of the ``cert`` option, so
+   subsequent requests on the same :class:`Session <requests.Session>`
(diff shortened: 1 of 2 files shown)


Lukasa:
Thanks for this suggestion! However, I don't think the proposed documentation change is good: in fact, I think it is dangerously misleading. The bug occurs regardless of how `cert` is set: so long as it has been set and a session is being used it is possible for it to be effectively incorrectly set. 

However, I also think there is no urgency to document this. A fix has been merged into urllib3 (shazow/urllib3#830), and a new urllib3 release will be out in the next few weeks. I doubt this will remain unfixed for more than another few weeks.

zestyping:
Thanks, I didn't know a fix was imminent.  I'm looking forward to it!

I don't understand what's wrong with the advice to set `cert` immediately on creation and never changing it—only asking for clarification in case it means I don't understand what the bug is.  If you set `cert` before making any requests and never change it, then how can you encounter the bug?

Lukasa:
Nothing's wrong with that advice: it's just not what your documentation says. =) Your documentation change says _don't use it at all_, which is not necessarily right.

You're right though: if you set `cert` before using the Session and never want to change or override it, you won't hit the bug.

PR #3216, opened 2016-05-23
Make BaseAdapter describe the mandatory adapter interface

nanonyme:
BaseAdapter should document the mandatory interfaces for implementing your own adapter from scratch for a different HTTP library. Currently requests requires all the parameters from HTTPAdapter implementation of send to be defined so copying them and relevant documentation over to the base adapter. Also adding relevant parts of documentation on what close is to the base class

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -54,7 +54,8 @@ class BaseAdapter(object):
     def __init__(self):
         super(BaseAdapter, self).__init__()
 
-    def send(self):
+    def send(self, request, stream=False, timeout=None, verify=True,
+             cert=None, proxies=None):
         raise NotImplementedError
 
     def close(self):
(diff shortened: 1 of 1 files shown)


requests/adapters.py
```diff
@@ -54,10 +54,24 @@ class BaseAdapter(object):
...
+            read timeout) <timeouts>` tuple.
+        :type timeout: float or tuple
+        :param verify: (optional) Whether to verify SSL certificates.
+        :param cert: (optional) Any user-provided SSL certificate to be trusted.
+        :param proxies: (optional) The proxies dictionary to apply to the request.
+        """
         raise NotImplementedError
 
     def close(self):
+        """Disposes of any internal state."""
```
sigmavirus24: This isn't quite true. There is typically some amount of internal state that isn't disposed of.

This would more accurately be:

``` py
Cleans up adapter specific items.
```

Lukasa:
There's one small note from @sigmavirus24 but I'd be happy to merge this when that's resolved.

PR #3225, opened 2016-05-29
Convert readthedocs link for their .org -> .io migration for hosted projects

adamchainz:
As per their email ‘Changes to project subdomains’:

> Starting today, Read the Docs will start hosting projects from subdomains on the domain readthedocs.io, instead of on readthedocs.org. This change addresses some security concerns around site cookies while hosting user generated data on the same domain as our dashboard.

Test Plan: Manually visited all the links I’ve modified.


adamchainz:
Just realized these are all in `urllib3`, replaced with shazow/urllib3#882

PR #3236, opened 2016-05-29
Use xfail marker for a test expected to fail without Internet connection

eriol:
This is only a minor improvement on the great work of https://github.com/kennethreitz/requests/pull/2859 that permits to run tests on hosts without Internet connection without failures.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1107,6 +1107,7 @@ class TestRequests:
             preq = req.prepare()
             assert test_url == preq.url
 
+    @pytest.mark.xfail(raises=ConnectionError)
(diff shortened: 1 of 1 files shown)


kennethreitz:
Not sure why no-one's gotten back to you on this yet — my apologies. 

This looks like a great improvement. Thank you.

sigmavirus24:
@eriol I'm so sorry. I think I lost track of this while I was travelling to PyCon. I expect the same happened to @Lukasa

eriol:
Sorry for my late reply... No need to apologize! I was aware of the PyCon during those days and I was sure this would not be lost... so just thanks for merging it! :)

Issue #3238, opened 2016-05-31
Support proxy protocol http_no_tunnel


likel:
Hi,
Is there any way to support 'http_no_tunnel' protocol in proxy? I want to move from httplib2 to requests, but I didn't find any way to archive it. Thanks.

birm:
what happens when you preface your hostname with http:// and run? Any difference?

PR #3274, opened 2016-06-05
Try to allow the socks patch to use http_no_tunnel

birm:
re #3238


Lukasa:
Thanks for this. Unfortunately, we can't merge this patch: it is entirely in urllib3 (a third-party library that we don't merge patches for). It's also not correct: it uses a constant that doesn't exist and handles a weirdly scoped URL rather than looking for environment variables.

birm:
Thank you. I'll see if I can find another way to add this functionality.

PR #3286, opened 2016-06-06
Load all unbundled urllib3 modules

jayvdb:


Lukasa:
I am _extremely_ nervous about applying further upstream patches to handle downstream unbundling behaviours. In particular, it's not clear to me what exact problem you're experiencing in #3287.

For the moment however I'm inclined to call this a Debian bug unless told otherwise.

jayvdb:
The problem is that pyopenssl injection doesnt work, despite having all of the necessary dependencies installed, and despite the pyopenssl injection function appearing to be successful (i.e. no `ImportError` is raised).  i.e. Debian's requests 2.10 doesnt have OpenSSL & SNI support, at least in the combinations I've tested using.

It looks like there is no tests covering this unbundling support code, so I'll set up unit tests to verify this isnt a Debian only problem.

fwiw, I have no objection to the unbundling support code being removed from `requests.packages.__init__`, but it is currently there, and doesn't work, but has lots of documentation in the module that gives packagers the false expectation that it works.

sigmavirus24:
So pip does [something similar](https://github.com/pypa/pip/blob/master/pip/_vendor/__init__.py#L79) but in a far more intelligent way than what is happening here.

That said, I don't think we have evidence that what pip is doing works either.

I also don't agree with your all or nothing mentality @jayvdb. It's not productive.

eriol:
Hello,
sorry for the late reply, I was on trip and once returned I had to work on the backport of betamax for the stable release due the sheduled upload on OpenStack (I don't remember which one).

I definitively agree with @Lukasa here: the problem is on the Debian side.
So, as soon I complete the backport of betamax I will work on this.
This is my plan:
1. use the same pip's `vendored` function to patch requests.packages.**init**;
2. add some tests on Debian CI infrastructure over this specific issue;
3. upload this new revision to experimental suite;
4. make a call for test;

@jayvdb can you share your tests about OpenSSL & SNI support on Debian? Thanks!

eriol:
@jayvdb never mind, you were talking about https://github.com/kennethreitz/requests/issues/3287.

kennethreitz:
I agree with @Lukasa — this part of the codebase is absolutely abhorrent and should not even exist — however, we have chosen to do so to improve the user experience of our very unfortunate distro-installed users. 

We've already done more than I think we should have. Doing even more, without extremely strong beyond-a-reasonable-doubt purpose, is out of the question.

Issue #3288, opened 2016-06-06
Minor session object documentation issues


jayvdb:
I suspect this is a documentation issue, for expected and desired behaviour, rather than an unintended bug...

http://docs.python-requests.org/en/master/user/advanced/#session-objects states that 

> The Session object ... will use urllib3's connection pooling. So if you're making several requests to the same host, the underlying TCP connection will be reused, which can result in a significant performance increase (see HTTP persistent connection).
> ...
> Note, however, that method-level parameters will not be persisted across requests, even if using a session. This example will only send the cookies with the first request, but not the second: ..."

There is a minor conflict between those, in that exceptions to the rule include the `verify` parameter and likely other connection related parameters (`cert`, `timeout`, ...?), which do in fact persist across the requests.  This is strongly implied by the opening paragraph, which says that the connection is reused.

Also "several requests to the same host" is not quite correct, I believe, as different protocols will result in different connections, naturally.

jayvdb:
Note #3289 shows the `verify` parameter being persistent between requests in the same session.

jayvdb:
What would be interesting to test is whether a connection reset of some sort causes the `verify` parameter in the session to the same protocol/host to be lost.  Maybe there is a test in `urllib3` that already covers that scenario.

Lukasa:
This is a known issue (see #2863) and will be fixed in an upcoming release, having had the underlying urllib3 issue fixed (shazow/urllib3#830).

PR #3289, opened 2016-06-06
Test verify parameter

jayvdb:


tests/test_requests.py
```diff
@@ -606,6 +621,115 @@ def test_unicode_header_name(self, httpbin):
...
+            session.get('https://testssl-expire-r2i2.disig.sk/index.en.html')
+
+        with pytest.raises(SSLError):
+            session.get('https://testssl-expire-r2i2.disig.sk/index.en.html')
+
+        with pytest.raises(SSLError):
+            session.get('https://self-signed.badssl.com/')
+
+        with pytest.raises(SSLError):
+            session.get('https://self-signed.badssl.com/')
```
Lukasa: Requests is as much as possible moving to offline testing: that is, testing where we do not need to hit real URLs. That's because this kind of testing is flaky and requires a good network floating around. It would be better if we could create endpoints locally that exhibit similar problems.

jayvdb: Yea, it would be good if pytest-httpbin (or similar) had recipes for servers with different ssl configs. Any suggestions on where to add this functionality? Im guessing it should be reusable for urllib3, and vcr, etc.

Lukasa: I don't believe there's any need to have it be re-useable. urllib3 already has its socket-level tests which are well equipped for this kind of testing, and you could use requests' own socket-level tests to reproduce this function as well.

jayvdb:
I've removed the `SSLError` tests, and session tests due to what you've explained in #3288, and focused on the `SubjectAltNameWarning` which should always occur with `pytest-httpbin`.

This is actually better at showing the problem in #3287 , as it needs to occur in all combinations of dependencies and Python versions.

tests/test_requests.py
```diff
@@ -32,10 +33,29 @@
 from .compat import StringIO, u
 from .utils import override_environ
 
+from requests.packages.urllib3.exceptions import SNIMissingWarning
+
+# urllib3 sets this to only go off once, but we need it to
+# always fire for test_https_warnings to work
+warnings.simplefilter('always', SNIMissingWarning)
```
Lukasa: I feel like this risks warning spam in a pretty big way on older systems.

Lukasa:
This looks reasonable to me aside from my one note about warning spam. @sigmavirus24, what do you think?

tests/test_requests.py
sigmavirus24: Maybe set this in a context manager around the single test that needs this. E.g.,

``` py
@contextlib.contextmanager
def enable_sni_missing():
    warnings.simplefilter('always', SNIMissingWarning)
    yield
    warnings.simplefilter('once', SNIMissingWarning)
```

(Pseudo-code warning)

jayvdb: No. After it has been done **once** somewhere else, it is gone.

sigmavirus24:
This is a good start. I think we can mitigate the spam.

jayvdb:
There are also other bugs in pytests handling of warnings (see https://github.com/pytest-dev/pytest/issues/840)
An alternative approach to address the warnings problem is to redirect them through the logging layer for the test suite, using `captureWarnings` and backports or some pytest extension which provides something similar.  py.test itself doesnt [appear](https://pytest.org/latest/recwarn.html) to have anything particularly fancy for managing warnings in the global test context.  There are pytest extensions for the logging layer.  Let me know if you want me to do a little analysis down that path.

tests/test_requests.py
Lukasa: Yeah, that's why I said it risks warning spam but didn't strictly say to remove it: I'm a bit aware that warning filters are awkwardly global. Is there nothing we can do about the spam at all?

jayvdb: Could you explain your concern a bit more?
py.test hides the spam by default, so it will only appear for test that fail, and only for https tests, which there are not many of at the moment any way.

Lukasa: Oh, I didn't know that py.test hid warnings by default. That alleviates my primary concern.

tests/test_requests.py
```diff
@@ -606,6 +626,27 @@ def test_unicode_header_name(self, httpbin):
     def test_pyopenssl_redirect(self, httpbin_secure, httpbin_ca_bundle):
         requests.get(httpbin_secure('status', '301'), verify=httpbin_ca_bundle)
 
+    def test_https_warnings(self, httpbin_secure, httpbin_ca_bundle):
+        """warnings are emitted with requests.get"""
+        if HAS_MODERN_SSL or HAS_PYOPENSSL:
+            warnings_expected = ('SubjectAltNameWarning', )
```
jayvdb, line 626: I have one concern about the patch (and series in general), and that is whether this warning should appear every time, or only the first time for a connection , and is the connection to httpbin_secure being cached.
My reading of VerifiedHTTPSConnection suggests it should only happen once per connection, and I am not sure if httpbin_secure is cached.

Lukasa, line 626: It should happen once per connection. `httpbin_secure` does no caching, it literally returns a URL. Any caching would happen in `requests.get`, and `requests.get` explicitly does not connection pool, so this test should be fine.

tests/test_requests.py
jayvdb: Hmm. Let me check that, as I cant see where in py.test that is happening..

tests/test_requests.py
jayvdb, line 626: Thanks for confirming that.

tests/test_requests.py
jayvdb: OK, so it doesnt have any specific code for warnings, it is just the normal --capture=sys/fd , which is capturing and hiding the stream, warnings included, by default.

`py.test -s` or `--capture=no` , or config is needed to override that default.

tests/test_requests.py
```diff
@@ -606,6 +626,27 @@ def test_unicode_header_name(self, httpbin):
...
+            warnings_expected = ('SNIMissingWarning',
+                                 'InsecurePlatformWarning',
+                                 'SubjectAltNameWarning', )
+
+        with pytest.warns(None) as warning_records:
+            warnings.simplefilter('always')
+            requests.get(httpbin_secure('status', '200'),
+                         verify=httpbin_ca_bundle)
+
+        warning_records = [item for item in warning_records
```
jayvdb, line 637: Note that there were two ResourceWarning's captured here, at the end of the list.  I would have expected only one; the usual one caused by using `requests.get`.

jayvdb:
Any objections to the warnings init moving into `tests.__init__` like so...
https://github.com/jayvdb/requests/blob/unbundling_walkthrough-v2/tests/__init__.py ? It makes the results more stable.

Lukasa:
I'm ok with it.

jayvdb:
Sorry for the delay; been recovering from hard disk loss

Lukasa:
Ok I'm good with this, leaving it up to @sigmavirus24 to merge when he's happy.

jayvdb:
@sigmavirus24 , ping ... ?

sigmavirus24:
Sorry. Was distracted with other things. This looks fine to me.

PR #3297, opened 2016-06-09
Note how HTTPErrors are raised

davidfischer:
After looking through the code line, `requests.exceptions.HTTPError`s are only raised by `raise_for_status` but the docs made it seem like that error could happen through other means. This PR makes the docs explicit.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -492,8 +492,9 @@ Errors and Exceptions
 In the event of a network problem (e.g. DNS failure, refused connection, etc),
 Requests will raise a :class:`~requests.exceptions.ConnectionError` exception.
 
-In the rare event of an invalid HTTP response, Requests will raise an
-:class:`~requests.exceptions.HTTPError` exception.
+:meth:`Response.raise_for_status() <requests.Response.raise_for_status>` will
+raise an :class:`~requests.exceptions.HTTPError` if the HTTP request
+returned an unsuccessful status code.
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
I believe the original text is indeed correct, that error can occur when an HTTPError (not related to status code) is raised. I'm not sure, off the top of my head, when that could be anymore. It's possible that's a remnant from when requests ran on urllib2. 

Regardless, the _additions_ to the text are most certainly welcome.

This PR may be perfect as-is, but will need double checking, given the notes in my first paragraph.

davidfischer:
Urllib3 has an exception with the same name, but it is not actually related to `requests.exceptions.HTTPError`. I [searched this repository](https://github.com/kennethreitz/requests/search?utf8=%E2%9C%93&q=httperror) (well, I used grep) for HTTPError and the only place I can see where requests itself uses it is with `raise_for_status`.

I had a discussion with @Lukasa briefly on IRC about this. As far as I can tell, `requests.exceptions.HTTPError` is only used for this purpose. However, if you think you'll use it for other purposes, another alternative is to have something like an `HTTPStatusCodeError` (a subclass of HTTPError) that is specifically for `raise_for_status`.

kennethreitz:
Okay then, that is, in fact, a remnant from running on urllib2, and your improved wording is welcome. Thanks for the due diligence :)

kennethreitz:
`HTTPStatusCodeError` would be a better name for this, but it is too late to change it now. Behavior was a bit different when this was originally crafted :)

PR #3298, opened 2016-06-09
Update a note on AppEngine

davidfischer:
This is simply an update on AppEngine support to point people in the right direction.


kennethreitz:
thanks!

PR #3339, opened 2016-06-16
Use seek from end rather than getvalue

jseabold:
Avoids having the (potentially large) object in memory twice.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -84,7 +84,8 @@ def super_len(o):
             # is actually a special file descriptor like stdin. In this
             # instance, we don't know what the length is, so set it to zero and
             # let requests chunk it instead.
-            current_position = total_length
+            if total_length is not None:
+                current_position = total_length
 
     if hasattr(o, 'seek') and total_length is None:
         # StringIO has a notimplemented fileno
@@ -98,6 +99,9 @@ def super_len(o):
 
         # seek back to current position to support partially read file-like
         o.seek(current_position or 0)
+
+    if total_length is None:
+        total_length = 0
 
     return max(0, total_length - current_position)
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -46,7 +46,7 @@ def dict_to_sequence(d):
 
 
 def super_len(o):
-    total_length = 0
+    total_length = None
     current_position = 0
 
     if hasattr(o, '__len__'):
@@ -54,10 +54,6 @@ def super_len(o):
 
     elif hasattr(o, 'len'):
         total_length = o.len
-
-    elif hasattr(o, 'getvalue'):
-        # e.g. BytesIO, cStringIO.StringIO
-        total_length = len(o.getvalue())
 
     elif hasattr(o, 'fileno'):
         try:
@@ -89,6 +85,13 @@ def super_len(o):
             # instance, we don't know what the length is, so set it to zero and
             # let requests chunk it instead.
             current_position = total_length
+
+    if hasattr(o, 'seek') and total_length is None:
+        # StringIO has a notimplemented fileno
+        # BytesIO has seek and not fileno
+        total_length = o.seek(0, 2)
+        # seek back to current position to support partially read file-like
+        o.seek(current_position)
 
     return max(0, total_length - current_position)
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -90,8 +90,14 @@ def super_len(o):
         # StringIO has a notimplemented fileno
         # BytesIO has seek and not fileno
         total_length = o.seek(0, 2)
+        # StringIO objects in py2 returns None
+        if total_length is None and hasattr(o, 'tell'):
+            total_length = o.tell()
+        elif total_length is None:
+            total_length = 0
+
         # seek back to current position to support partially read file-like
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this patch!

As composed, however, this patch doesn't make sense. The code block tests for `getvalue()` but then uses two unrelated methods.

It might be better to have a separate block that tests for `seek()` and `tell()` and then use those in that block, considering that a higher priority than `getvalue()`. Does that sound sensible?

jseabold:
`seek` and `tell` are necessary but not sufficient conditions for it being a BytesIO object, no? That would also capture the file-objects that are the last thing checked for and use the filesystem.

The whole thing strikes me as a bit funny, but I didn't have enough context here to make wholesale changes. Why not just use `isinstance` be explicit about what's what? Python 2/3 issues? `getvalue` is a sufficient condition for BytesIO though.

Thoughts?

Lukasa:
The main reason not to use `isinstance` is because we want to support duck-typed objects: things that behave _like_ a BytesIO object but are not BytesIO objects, for example. This is particularly important with file-like objects, as there is _technically_ a protocol to which they conform but _in practice_ many file-like objects support only part of the API.

In this case, we may be able to resolve the problem by changing the logic. Rather than have one gigantic `if...elif` chain, we could change it to be more like this:

``` python
total_length = None
if x:
    total_length = whatever()

if total_length is None and y:
    total_length = whatever2()

if total_length is None and z:
    total_length = whatever3()
```

That would allow us to re-order things to do the `fileno` block _first_ and then, if we drop out of the `fileno` block, fall back on seek() and tell(). How does that sound?

jseabold:
Makes sense. Let me give it another go.

jseabold:
See what you think about this. Added some code comments about why it's implemented this way.

requests/utils.py
```diff
@@ -45,8 +45,11 @@ def dict_to_sequence(d):
     return d
 
 
+def _file_length(o):
+    pass
+
```
Lukasa: What's this for?

jseabold:
I suppose we could run into a case where `current_position` becomes None in the 'tell' block, but I'm not sure under what circumstances.

jseabold:
Well, looks like that happens on Python 2.

jseabold:
🍝

requests/utils.py
jseabold: Removed

requests/utils.py
```diff
@@ -90,6 +86,19 @@ def super_len(o):
...
+        # StringIO objects in py2 returns None
+        if total_length is None and hasattr(o, 'tell'):
+            total_length = o.tell()
+        elif total_length is None:
+            total_length = 0
+
+        # seek back to current position to support partially read file-like
+        o.seek(current_position or 0)
+
     return max(0, total_length - current_position)
```
Lukasa, line 106: So with this patch we can fall down to here with no `total_length` and then do some arithmetic on it, which will blow up. We need a way to default `total_length` to 0 right before this. The same problem applies to `current_position`, potentially.

requests/utils.py
```diff
@@ -90,6 +86,19 @@ def super_len(o):
...
+    if hasattr(o, 'seek') and total_length is None:
+        # StringIO has a notimplemented fileno
+        # BytesIO has seek and not fileno
+        total_length = o.seek(0, 2)
+        # StringIO objects in py2 returns None
+        if total_length is None and hasattr(o, 'tell'):
+            total_length = o.tell()
+        elif total_length is None:
+            total_length = 0
+
```
Lukasa, line 99: This is...odd. Surely `seek` either does or does not return a number, why are we now overriding it?

jseabold, line 99: In Python 2, it does not apparently. That's why my tests failed in 2.

```
In [3]: StringIO.StringIO.seek?
Signature: StringIO.StringIO.seek(self, pos, mode=0)
Docstring:
Set the file's current position.

The mode argument is optional and defaults to 0 (absolute file
positioning); other values are 1 (seek relative to the current
position) and 2 (seek relative to the file's end).

There is no return value.
```

```
In [6]: io.StringIO.seek?
Docstring:
Change stream position.

Seek to character offset pos relative to position indicated by whence:
    0  Start of stream (the default).  pos should be >= 0;
    1  Current position - pos must be 0;
    2  End of stream - pos must be 0.
Returns the new absolute position.
```

requests/utils.py
jseabold, line 106: Wrote some logic for this case, though I don't know how to produce it.

requests/utils.py
Lukasa, line 99: Ah, ok. In that case, let's not rely on this at all because it confuses the control flow a bit. Let's just use `tell()`.

jseabold:
I'm not going to have time to see this through. Closing in favor of other intrepid souls. Happy to have a look at subsequent work.

Lukasa:
Thanks for that @jseabold, it's helpful when people come back to tidy up when they no longer have the time to dedicate to the patch. All the best with whatever you're spending your time on, and I hope you hang around!

nateprewitt:
Thanks for contributing to Requests @jseabold!

PR #3342, opened 2016-06-17
fix PUT of empty (0 byte) stream

nlevitt:
I found #3051 but I'm offering this anyway. It's a simple fix for what appears to be a simple oversight in the code.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -443,7 +443,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             if files:
                 raise NotImplementedError('Streamed bodies and files are mutually exclusive.')
 
-            if length:
+            if length is not None:
                 self.headers['Content-Length'] = builtin_str(length)
             else:
                 self.headers['Transfer-Encoding'] = 'chunked'
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -129,6 +129,12 @@ class TestRequests:
                                    data=u"ööö".encode("utf-8")).prepare()
         assert isinstance(request.body, bytes)
 
+    def test_put_empty_stream(self):
+        request = requests.Request('PUT', 'http://example.com',
+                                   data=io.BytesIO(b'')).prepare()
+        assert not 'transfer-encoding' in request.headers
+        assert requests.headers['content-length'] == '0'
+
     @pytest.mark.parametrize('scheme', ('http://', 'HTTP://', 'hTTp://', 'HttP://'))
     def test_mixed_case_scheme_acceptable(self, httpbin, scheme):
         s = requests.Session()


Lukasa:
Thanks for this!

This is not an oversight, it's entirely deliberate. We do this to handle the case that we have a body with indeterminate length (e.g. stdin) which reports its length as zero or throws errors when we try to work out how long it is. In this case, defaulting to transfer encoding chunked is sensible.

All servers should be able to tolerate this. Have you encountered an actual problem with this?

nlevitt:
Thanks for the quick response. I was still looking into the test failures when I saw it.

It seems like the length should be `None` rather than zero in case of a body with indeterminate length, don't you think? :) 

I do have a real world problem with this. The archive.org s3-like api doesn't support chunked uploads and responds to a PUT without content-length with a "411 Length Required". 
https://archive.org/help/abouts3.txt

For example:

```
$ curl -LsSv --request PUT --upload-file /dev/null --header x-archive-meta-collection:test_collection  --header "authorization: LOW xxx:xxx" https://s3.us.archive.org/nlevitt-test-2016-06-17/
*   Trying 207.241.224.50...
* Connected to s3.us.archive.org (207.241.224.50) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate: *.s3.us.archive.org
* Server certificate: Go Daddy Secure Certificate Authority - G2
* Server certificate: Go Daddy Root Certificate Authority - G2
> PUT /nlevitt-test-2016-06-17/null HTTP/1.1
> Host: s3.us.archive.org
> User-Agent: curl/7.43.0
> Accept: */*
> Transfer-Encoding: chunked
> x-archive-meta-collection:test_collection
> authorization: LOW xxx:xxx
> Expect: 100-continue
> 
< HTTP/1.1 411 Length Required
< Date: Fri, 17 Jun 2016 19:34:15 GMT
< Server: Apache/2.4.7 (Ubuntu)
< Content-Length: 320
< Connection: close
< Content-Type: text/html; charset=iso-8859-1
< 
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>411 Length Required</title>
</head><body>
<h1>Length Required</h1>
<p>A request of the requested method PUT requires a valid Content-length.<br />
</p>
<hr>
<address>Apache/2.4.7 (Ubuntu) Server at s3.us.archive.org Port 80</address>
</body></html>
* Closing connection 0
```

Lukasa:
The reason we fall back to zero is a side effect of the way lengths are calculated. If you dive back into bug reports you'll see that we had bugs where we sent bodies with Content-Length: 0: this behaviour is much nicer. =D

Archive.org in this case is wildly violating the HTTP RFCs, but you can force the correct behaviour by using the prepared request flow and setting up the headers and body as you need them.

nlevitt:
Ok. Thanks.

kennethreitz:
The archive.org s3-like API is extremely important, despite the fact that it is poorly implemented. Personally, I have utilized it successfully with Boto (which uses Requests internally, for some things). You may find the use of that library more appropriate. 

PreparedRequest is indeed the solution, if not, though.

PR #3356, opened 2016-06-21
Do not attempt to decode to unicode twice

zachmullen:
It's possible I'm missing something here, but it seems like we should not be trying to decode via the underlying urllib3 method since `iter_content` is responsible for the decision of whether or not to decode.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -665,7 +665,7 @@ class Response(object):
             # Special case for urllib3.
             if hasattr(self.raw, 'stream'):
                 try:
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this!

The urllib3 decoding is about decoding compressed content-encodings like gzip and deflate. We do always want that on. =) iter_content's decoding is about decoding to unicode, an unrelated concern.

zachmullen:
Ah, thank you for the explanation :)

PR #3358, opened 2016-06-21
Update list of supported Python versions in todo.rst

petedmarsh:
Extremely minor change; I happened to notice that the list of supported Python versions in todo.rst was out-of-sync with the Python version classifiers in setup.py, this just puts it back in sync.


Lukasa:
Thanks @petedmarsh! :sparkles: :cake: :sparkles:

PR #3361, opened 2016-06-23
Allow sending an unprepared request via session.

jamielennox:
This can be considered a request in the form of a PR because it was easier to make the code change than explain it. 

I have multiple places where i am passing around method, url, json etc before it makes it to the requests level. Now these can all be bundled up into a single request object which is much easier to pass around, but then i have to manually call

prep = session.prepare_request(req)
settings = session.merge_environment_settings(....) 
settings['timeout'] = timeout
session.send(prep, **settings)

Now it's not super hard, but it's something that happens by default doing request() and it's certainly not intuitive. If we add a send_request() which does exactly the same as request() just on an existing request object it makes it much easier to correctly use Request objects.

Note: If you like the idea but not the implementation and have something equivalent - please just do it your way. I don't care if this particular PR gets accepted or not if the idea gets up. I proposed this against 3.0 thinking it was going to be a more breaking change, but it seems pretty easy if you'd prefer i propose against master.

Thanks

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -471,23 +471,14 @@ class Session(SessionRedirectMixin):
             cookies = cookies,
             hooks = hooks,
         )
-        prep = self.prepare_request(req)
 
-        proxies = proxies or {}
-
-        settings = self.merge_environment_settings(
-            prep.url, proxies, stream, verify, cert
-        )
-
-        # Send the request.
-        send_kwargs = {
-            'timeout': timeout,
-            'allow_redirects': allow_redirects,
-        }
-        send_kwargs.update(settings)
-        resp = self.send(prep, **send_kwargs)
-
-        return resp
+        return self.send_request(req,
+            proxies = proxies,
+            timeout = timeout,
+            allow_redirects = allow_redirects,
+            stream = stream,
+            verify = verify,
+            cert = cert)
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
PreparedRequests should be suitable for accomplishing what you're trying to accomplish.

I see where you're coming from, and thank you for your contribution. However, in short, no. :)

@sigmavirus24 & @Lukasa are welcome to add additional thoughts, if desired.

sigmavirus24:
I once thought about this too. Then I realized that you're introducing yet another way of doing things when there already exist more than one way to do it. You can do:

``` py
#: 1
requests.get(...)
#: 2
requests.request('GET', ...)
#: 3
my_session.get(...)
#: 4
my_session.request('GET', ...)
#: 5
request = requests.Request(method='GET', ...)
prepared = my_session.prepare_request(request)
settings = my_session.merge_environment_settings(....) 
# modify settings
my_session.send(prepared, **settings)
```

We really don't need yet another way of doing roughly the same thing. That's too much API surface area for very little extra benefit. If you're using the PreparedRequest flow, your proposal is only saving a tiny fraction of that workflow and diverging the typical code-paths further. Frankly, I don't think it's worth the extra surface area. In my experience, people who think the PreparedRequest flow is too complicated or kludgy, don't actually need it.

PR #3365, opened 2016-06-27
added in type check for chunk_size

nateprewitt:
This should raise a type error if a non-integer chunk_size is passed. I'm assuming we're not concerned about a chunk_size of type long, but that may be wrong.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -685,6 +685,8 @@ class Response(object):
 
         if self._content_consumed and isinstance(self._content, bool):
             raise StreamConsumedError()
+        elif not isinstance(chunk_size, int):
+            raise TypeError("chunk_size must be an int, it is instead a %s." % type(chunk_size))
         # simulate reading small chunks of the content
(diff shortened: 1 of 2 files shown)


requests/models.py
```diff
@@ -685,6 +685,8 @@ def generate():
 
         if self._content_consumed and isinstance(self._content, bool):
             raise StreamConsumedError()
+        elif not isinstance(chunk_size, int):
+            raise TypeError("Value of chunk_size must be of type int.")
```
sigmavirus24: More simply:

> "chunk_size must be an int, it is instead a %s" % type(chunk_size)

nateprewitt: Ok, things should be reworded. I didn't line wrap, as per the contributing docs, but let me know if it needs any other formatting tweaks.

Lukasa:
Very nice @nateprewitt! Thanks! :sparkles: :cake: :sparkles:

Issue #2947, opened 2015-12-28
No protection against line breaks in header values


dsavenko:
Hi,

The following call leads to 'another-header' appearing in the headers sent to the server:

```
>>> r = requests.get('https://httpbin.org/get', params={'q': 'query'}, headers={'my-header': 'a\r\nanother-header: b'})
```

I expected either the line break to be escaped in some way, or an exception to be raised. A new header silently appearing in the request doesn't look right. 

Requests version is 2.9.1, installed with pip.

Lukasa:
Ugh, yeah, good spot. This isn't really ideal: while what occurs here is kinda exactly what you'd expect if you have an understanding of the system, it provides a nice shiny footgun for developers who aren't expecting it, potentially allowing for header injection if they're passing user-provided data directly to a different service. We should probably remove it.

In general, I'm inclined to want to throw errors here, resisting the temptation to guess, but @sigmavirus24 or @kennethreitz may disagree and instead want us to simply strip line terminating characters from headers.

sigmavirus24:
Ugh this is tough. I want to raise exceptions here as well. I'm a bit torn as to when we should introduce those exceptions though. Let's look at it like this: I _think_ most people who are using requests will not be affected by us introducing a new exception around this problem because I _think_ most people aren't going to encounter this. That said, it's still a backwards incompatible change, and the above thinking is what has caused urllib3 to break requests a couple times this year and both of the projects to break other projects (e.g., OpenStack). (That is not to say that I think we're going to break OpenStack with this change, just that we've been fairly confident in the past year that changing X wouldn't hurt anyone and then it does.)

Lukasa:
Yeah, so we need to decide whether we can either a) break our normal release policy in order to throw exceptions here, or b) treat this as a strong impetus to move to 3.0.0 in the relatively short term.

sigmavirus24:
I think we can but we shouldn't (break our normal release policy).

Lukasa:
So @sigmavirus24, are you thinking that we should attempt to release 3.0.0 sometime in January including a fix for this issue?

jonathan-s:
When I try the request as described above I get the following traceback. So the request does not succeed, so is this issue still valid? Or would we want to turn the `ValueError` into an exception thrown by the request library? I used requests 2.9.1. 

```
  File "/Users/jonathan/.virtualenvs/test/lib/python2.7/site-packages/requests/adapters.py", line 376, in send
    timeout=timeout
[...]

PR #3366, opened 2016-06-29
check for headers containing return characters

nateprewitt:
Addresses issue raised in #2947.

This likely needs some tweaking. I added the check in `adapters.py` because it was the lowest level I could find header modification being done before sending. I'm not sure it actually needs to be this low though since it doesn't seem easily reachable with a non-`PreparedRequest` object.

Alternatively, we could move the check to the [`prepare_headers`](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L406) method in `PreparedRequest`.

I also ignored `Accept` headers in the check because I the requirements in [RFC7230](https://tools.ietf.org/html/rfc7230#page-24) seems a bit vague.

  >This specification deprecates such line folding except within the message/http media type
   (Section 8.3.1).  A sender MUST NOT generate a message that includes line folding (i.e., that has any    field-value that contains a match to the obs-fold rule) unless the message is intended for packaging
   within the message/http media type.

This may be a poor interpretation on my part though.

### DIFF
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -80,7 +80,11 @@ class InvalidSchema(RequestException, ValueError):
 
 
 class InvalidURL(RequestException, ValueError):
-    """ The URL provided was somehow invalid. """
+    """The URL provided was somehow invalid."""
+
+
+class InvalidHeader(RequestException, ValueError):
+    """The header value provided was somehow invalid."""
 
 
 class ChunkedEncodingError(RequestException):
--- a/requests/models.py
+++ b/requests/models.py
@@ -27,7 +27,8 @@ from .exceptions import (
 from .utils import (
     guess_filename, get_auth_from_url, requote_uri,
     stream_decode_response_unicode, to_key_val_list, parse_header_links,
-    iter_slices, guess_json_utf, super_len, to_native_string)
+    iter_slices, guess_json_utf, super_len, to_native_string, 
+    check_header_validity)
 from .compat import (
     cookielib, urlunparse, urlsplit, urlencode, str, bytes, StringIO,
     is_py2, chardet, builtin_str, basestring)
@@ -403,10 +404,13 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
     def prepare_headers(self, headers):
         """Prepares the given HTTP headers."""
 
+        self.headers = CaseInsensitiveDict()
         if headers:
-            self.headers = CaseInsensitiveDict((to_native_string(name), value) for name, value in headers.items())
-        else:
-            self.headers = CaseInsensitiveDict()
+            for header in headers.items():
+                # Raise exception on invalid header value.
+                check_header_validity(header)
+                name, value = header
+                self.headers[to_native_string(name)] = value
 
     def prepare_body(self, data, files, json=None):
         """Prepares the given HTTP body data."""
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -27,7 +27,7 @@ from .compat import (quote, urlparse, bytes, str, OrderedDict, unquote, is_py2,
                      basestring)
 from .cookies import RequestsCookieJar, cookiejar_from_dict
 from .structures import CaseInsensitiveDict
-from .exceptions import InvalidURL, FileModeWarning
+from .exceptions import InvalidURL, InvalidHeader, FileModeWarning
 
 _hush_pyflakes = (RequestsCookieJar,)
 
@@ -732,6 +732,24 @@ def to_native_string(string, encoding='ascii'):
 
     return out
 
+# Moved outside of function to avoid recompile every call
+_CLEAN_HEADER_REGEX_BYTE = re.compile(b'^\\S[^\\r\\n]*$|^$')
+_CLEAN_HEADER_REGEX_STR = re.compile(r'^\S[^\r\n]*$|^$')
+
+def check_header_validity(header):
+    """Verifies that header value doesn't contain leading whitespace or
+    return characters. This prevents unintended header injection.
+
+    :param header: tuple, in the format (name, value).
+    """
+    name, value = header
+
+    if isinstance(value, bytes):
+        pat = _CLEAN_HEADER_REGEX_BYTE
+    else:
+        pat = _CLEAN_HEADER_REGEX_STR
+    if not pat.match(value):
+        raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
 
 def urldefragauth(url):
     """
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -166,3 +166,4 @@ Patches and Suggestions
 - Dmitry Dygalo (`@Stranger6667 <https://github.com/Stranger6667>`_)
 - piotrjurkiewicz
 - Jesse Shapiro <jesse@jesseshapiro.net> (`@haikuginger <https://github.com/haikuginger>`_)
+- Nate Prewitt <nate.prewitt@gmail.com> (`@nateprewitt <https://github.com/nateprewitt>`_)
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -23,7 +23,7 @@ from requests.cookies import cookiejar_from_dict, morsel_to_cookie
 from requests.exceptions import (
     ConnectionError, ConnectTimeout, InvalidSchema, InvalidURL,
     MissingSchema, ReadTimeout, Timeout, RetryError, TooManyRedirects,
-    ProxyError)
+    ProxyError, InvalidHeader)
 from requests.models import PreparedRequest
 from requests.structures import CaseInsensitiveDict
 from requests.sessions import SessionRedirectMixin
@@ -1127,6 +1127,47 @@ class TestRequests:
         # we go.
         assert 'unicode' in p.headers.keys()
         assert 'byte' in p.headers.keys()
+
+    def test_header_validation(self,httpbin):
+        """Ensure prepare_headers regex isn't flagging valid header contents."""
+        headers_ok = {'foo': 'bar baz qux',
+                      'bar': '1',
+                      'baz': '',
+                      'qux': str.encode(u'fbbq')}
+        r = requests.get(httpbin('get'), headers=headers_ok)
+        assert r.request.headers['foo'] == headers_ok['foo']
+
+    def test_header_no_return_chars(self, httpbin):
+        """Ensure that a header containing return character sequences raise an
+        exception. Otherwise, multiple headers are created from single string.
+        """
+        headers_ret = {'foo': 'bar\r\nbaz: qux'}
+        headers_lf = {'foo': 'bar\nbaz: qux'}
+        headers_cr = {'foo': 'bar\rbaz: qux'}
+
+        # Test for newline
+        with pytest.raises(InvalidHeader):
+            r = requests.get(httpbin('get'), headers=headers_ret)
+        # Test for line feed
+        with pytest.raises(InvalidHeader):
+            r = requests.get(httpbin('get'), headers=headers_lf)
+        # Test for carriage return
+        with pytest.raises(InvalidHeader):
+            r = requests.get(httpbin('get'), headers=headers_cr)
+
+    def test_header_no_leading_space(self, httpbin):
+        """Ensure headers containing leading whitespace raise
+        InvalidHeader Error before sending.
+        """
+        headers_space = {'foo': ' bar'}
+        headers_tab = {'foo': '   bar'}
+
+        # Test for whitespace
+        with pytest.raises(InvalidHeader):
+            r = requests.get(httpbin('get'), headers=headers_space)
+        # Test for tab
+        with pytest.raises(InvalidHeader):
+            r = requests.get(httpbin('get'), headers=headers_tab)
 
     @pytest.mark.parametrize('files', ('foo', b'foo', bytearray(b'foo')))
     def test_can_send_objects_with_files(self, httpbin, files):


Lukasa:
We should forbid the newline in Accept headers too: that is a misreading of RFC 7230. =) Otherwise, this is basically the right approach. Some notes:
- Can we move the logic into a underscore-prefixed method on the HTTPAdapter? That will let people override the method if they need to.
- Can we also search for line feed characters (`'\r'`)?
- Can we also also confirm the header field does not begin with linear whitespace (e.g. space, tab)?

This should handle most of the header smuggling risks. At this point we probably just want to match a regex against the header.

kennethreitz:
I understand the reasoning behind putting it in adapters, but I do feel like it should be in the prepare_headers section instead, if possible. PreparedRequests should allow users to do stupid things, like this. No need to have saftey tape at that level, imo. 

I won't be surprised if @Lukasa disagrees :)

nateprewitt:
If we're treating this as a security issue and trying to prevent injection on dynamically generated headers, I think @Lukasa's internal method allows flexibility while maintaining the smallest surface area for issues. If we're just looking to prevent unintentional foot shooting, prepare_headers seems "cleaner" to me.

kennethreitz:
Agreed. I feel like it's the latter. Am I incorrect?

nateprewitt:
No, I was just checking I understood the intent of the patch :)

Working on moving things up to `prepare_headers` with Lukasa's proposed changes now.

requests/models.py
```diff
@@ -403,10 +404,13 @@ def prepare_url(self, url, params):
     def prepare_headers(self, headers):
         """Prepares the given HTTP headers."""
 
+        self.headers = CaseInsensitiveDict()
         if headers:
-            self.headers = CaseInsensitiveDict((to_native_string(name), value) for name, value in headers.items())
-        else:
-            self.headers = CaseInsensitiveDict()
+            pat = re.compile("^\S([^\r\n]+)?$")
```
kennethreitz: comment here would be good

nateprewitt: Good call on that, I hadn't thought out the regex completely. It was missing strings of length 0. Comment and test updated.

Lukasa:
Much to @kennethreitz's surprise, I agree with him. =)

I'm worried that users who are unsuspectingly using our high-level APIs aren't subject to risk, but if you're going to drop down to monkeying with PreparedRequests then you should be able to do more-or-less what you like. =)

requests/models.py
```diff
@@ -403,10 +404,15 @@ def prepare_url(self, url, params):
...
         """Prepares the given HTTP headers."""
 
+        self.headers = CaseInsensitiveDict()
         if headers:
-            self.headers = CaseInsensitiveDict((to_native_string(name), value) for name, value in headers.items())
-        else:
-            self.headers = CaseInsensitiveDict()
+            #Regex checks that string doesn't begin with whitespace or contain return characters.
+            #Otherwise, it checks if it's a blank string.
+            pat = re.compile("^\S[^\r\n]*$|^$")
```
Lukasa: We need to bring `to_native_string` up to before the match on this regex. This is because we allow users to provide headers as unicode and byte strings, which will misbehave with a native string regex.

requests/models.py
```diff
@@ -403,10 +404,15 @@ def prepare_url(self, url, params):
...
+            for name, val in headers.items():
+                if not pat.match(val):
```
Lukasa: To be clear, `to_native_string` should probably go above this line.

requests/models.py
```diff
@@ -403,10 +404,15 @@ def prepare_url(self, url, params):
...
+                    raise InvalidHeader("Invalid return character or space in header: %s" % (name))
```
Lukasa: Can we say 'leading space', not just 'space'?

Lukasa:
I've got some small notes about code structure, but this is looking really good!

requests/models.py
nateprewitt: Moved the `to_native_string` call up and added it for `val` too, since that's what we're actually matching the regex on.

nateprewitt:
Thanks, I think all of the changes you asked for should be in place. If you've got any additional notes, let me know. Always open to feedback :)

requests/models.py
```diff
@@ -403,10 +404,16 @@ def prepare_url(self, url, params):
...
+        self.headers = CaseInsensitiveDict()
         if headers:
-            self.headers = CaseInsensitiveDict((to_native_string(name), value) for name, value in headers.items())
-        else:
-            self.headers = CaseInsensitiveDict()
+            #Regex checks that string doesn't begin with whitespace or contain return characters.
+            #Otherwise, it checks if it's a blank string.
+            pat = re.compile(r'^\S[^\r\n]*$|^$')
+            for name, val in headers.items():
+                name, val = to_native_string(name), to_native_string(val)
```
Lukasa: I'm a little bit worried about this, actually. `val` could be an opaque bytestring that doesn't decode well on Python 3. Let's change this to use a bytestring regex (prefixed with `rb`) and then elide the call to `to_native_string` on `val`.

requests/models.py
```diff
@@ -403,10 +404,16 @@ def prepare_url(self, url, params):
     def prepare_headers(self, headers):
         """Prepares the given HTTP headers."""
 
+        self.headers = CaseInsensitiveDict()
         if headers:
-            self.headers = CaseInsensitiveDict((to_native_string(name), value) for name, value in headers.items())
-        else:
-            self.headers = CaseInsensitiveDict()
+            #Regex checks that string doesn't begin with whitespace or contain return characters.
+            #Otherwise, it checks if it's a blank string.
```
Lukasa: One space after the `#` character, please. =)

requests/exceptions.py
```diff
@@ -80,8 +80,10 @@ class InvalidSchema(RequestException, ValueError):
 
 
 class InvalidURL(RequestException, ValueError):
-    """ The URL provided was somehow invalid. """
+    """The URL provided was somehow invalid. """
 
+class InvalidHeader(RequestException, ValueError):
+    """The header value provided was somehow invalid. """
```
Lukasa: These need to be separated from each other by two newlines, not one.

Lukasa:
Sorry, some more smaller notes. =) This should be the last review though!

requests/models.py
nateprewitt: @Lukasa Just to clarify, do we want to leave `name = to_native_string(name)` on line 413 to prevent a byte string from being passed to the exception, or did you want it moved back to where it was originally on line 416?

nateprewitt: @Lukasa Also the rb'string' syntax doesn't appear to work for python 2.6 or 2.7.

Lukasa: I don't really mind if it gets moved. =)

As to the rb string syntax, in that case I recommend not using a raw string at all and just escaping the backslashes and using bytestrings.. It's annoying, but works on all platforms.

nateprewitt: Hmm, so that worked fine on my python2 testing but it looks like it fails for python3 with the following error:

`TypeError: can't use a bytes pattern on a string-like object`

I'm digging around for a more universal option.

Lukasa:
To be clear, you need to remove `to_native_string(val)` when using the bytes pattern.

Lukasa:
Hrgh, actually, this is a bit trickier than that. We need two regexes, one for unicode strings and one for bytes strings. And at this point, we probably want to turn this into a full-blown utility function.

nateprewitt:
Hmm, yeah. So you're suggesting something along the lines of this?

```

def validate_header(val):
    if isinstance(val, str):
        regex = str_reg
    else:
        regex = byte_reg
    if regex.match(val):
         raise InvalidHeader()
    return val

---

def prepare_headers(self, headers) 
    self.headers = CaseInsensitiveDict()
    if headers:
        for name, val in headers.items():
                self.headers[to_native_string(name)] = validate_header(val)

```

Lukasa:
Yup, something like that, yeah.

nateprewitt:
@Lukasa Alright, there's rev 1 for a utility function.

requests/models.py
```diff
@@ -9,6 +9,7 @@
 
 import collections
 import datetime
+import re
```
Lukasa: We don't need this import any more.

requests/utils.py
```diff
@@ -732,6 +732,22 @@ def to_native_string(string, encoding='ascii'):
...
+    name, value = header
+
+    if isinstance(value, bytes):
+        pat = re.compile(b'^\S[^\\r\\n]*$|^$')
+    else:
+        pat = re.compile(r'^\S[^\r\n]*$|^$')
+    if not pat.match(value):
+        raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
+
+    return (to_native_string(name), value)
```
Lukasa: Frankly, I don't love that this hides the call to `to_native_string`. I understand why it's here, but I'd rather that it didn't. Can we just turn this into a function that we can pass to `filter`?

requests/utils.py
```diff
@@ -732,6 +732,22 @@ def to_native_string(string, encoding='ascii'):
...
+    return characters. This prevents unintended header injection.
+
+    header is a tuple, in the format (name, value)
+    """
+    name, value = header
+
+    if isinstance(value, bytes):
+        pat = re.compile(b'^\S[^\\r\\n]*$|^$')
+    else:
+        pat = re.compile(r'^\S[^\r\n]*$|^$')
```
Lukasa: So using `re.compile` is great here, but we lose the advantage because these are done in the function. Can you hoist both these REs up to module scope and give them neat names, and then have this function choose between them?

nateprewitt: That's what I was thinking too (with performance), I just wasn't sure where to declare them. You're suggesting throwing something like this at the top of `utils.py`, yeah?

```
CLEAN_HEADER_REGEX_BYTES = re.compile(b'^\S[^\\r\\n]*$|^$')
CLEAN_HEADER_REGEX_STR = re.compile(r'^\S[^\r\n]*$|^$')
```

Lukasa: Or even just hoist them directly above this function. Let's prefix their names with underscores just to discourage people relying on them.

requests/models.py
```diff
@@ -404,7 +404,8 @@ def prepare_headers(self, headers):
         """Prepares the given HTTP headers."""
 
         if headers:
-            self.headers = CaseInsensitiveDict(validate_header(i) for i in headers.items())
+            valid_headers = filter(is_valid_header, headers.items())
+            self.headers = CaseInsensitiveDict((to_native_string(name), val) for name, val in valid_headers)
```
nateprewitt: I'm not particularly fond of having to iterate over the header list twice, but I couldn't devise a way to wrap name in to_native_string() while using `is_valid_header` as a filter function. Perhaps I'm missing an easy solution though.

Lukasa: I mean, `filter` is kind of a lie here (it's really `map`, and even there we aren't transforming the data), so you could have just one loop that calls `is_valid_header` then sets the kvp in the CaseInsensitiveDict.

requests/models.py
```diff
@@ -403,11 +403,13 @@ def prepare_url(self, url, params):
...
 
+        self.headers = CaseInsensitiveDict()
         if headers:
-            valid_headers = filter(is_valid_header, headers.items())
-            self.headers = CaseInsensitiveDict((to_native_string(name), val) for name, val in valid_headers)
-        else:
-            self.headers = CaseInsensitiveDict()
+            for header in headers.items():
+                # Raise exception on invalid header value.
+                is_valid_header(header)
```
nateprewitt: This seems slightly kludgy to me, since we're not doing a transform or using this as a conditional. Maybe it's just a naming issue though. Renaming it `check_header_validity` and removing the `return True` may be a better option to describe it's current functionality. Otherwise, I'll leave it in, as is, to use as a generalized utility check elsewhere if needed.

Lukasa: Yeah, let's do that.

nateprewitt: Alright, updated. If it's looking ok, I'll squash everything back down.

requests/utils.py
```diff
@@ -732,6 +732,24 @@ def to_native_string(string, encoding='ascii'):
 
     return out
 
+# Moved outside of function to avoid recompile every call
+_CLEAN_HEADER_REGEX_BYTE = re.compile(b'^\S[^\\r\\n]*$|^$')
```
Lukasa: I think _strictly_ we're missing a backslash in this regex. You have `\S`, but I think you want `\\S`.

Lukasa:
Ok, one tiny string fix and I'll finally be done complaining. =D

requests/utils.py
nateprewitt: Right on, completely missed that. Thanks! Pushed up the additional \ and will squash the commits down if this is looking ok.

Lukasa:
Cool, LGTM. Rebase away. =D

nateprewitt:
Squashed and passing. Thanks for taking the time to work through this with me @Lukasa :)

Lukasa:
Hooray! I am happy. =)

Lukasa:

kennethreitz:
@Lukasa, stealing my emoji chain hahaha

saper:
Just to let you know that we bumped into this recently here: https://github.com/jjjake/internetarchive/issues/151 - we were just sending integer 1 or 0 expecting it to be coerced to "1" or "0".
This is was simple to fix but I think this should be marked as a breaking change.

Regarding the implementation: using `isinstance()` checks is kind of [un-pythonic](http://canonical.org/~kragen/isinstance/), it forbids the caller to ship instances of objects that should behave like strings or return rendered strings when asked.

Lukasa:
> Regarding the implementation: using isinstance() checks is kind of un-pythonic, it forbids the caller to ship instances of objects that should behave like strings or return rendered strings when asked.

There is no way in Python for something to behave "like a string" without being a string. You can ask something to be a string via `str()`, but because`str()` falls back to `repr()` it cannot be used as a sensible screening mechanism.

What would be your proposed alternative?

saper:
That probably means we have no choice but to depend on `str()` ? The original bug was about embedded newlines only I think.

Lukasa:
We _cannot_ depend on `str()`, its behaviours are consistently and completely unacceptable. The worst of them being the fact that, on Python 3, `str(b'hello') == "b'hello'"`, meaning that anyone passing a bytestring header value on Python 3 would get their header value totally mangled, despite the fact that passing us a bytestring is actually the _simplest_ case from the perspective of our logic: we pass it through, after checking that no ASCII bytes for newlines or leading/trailing whitespace are present.

Unconditionally casting data to strings does not behave well in Python, and the biggest problem is that it leads to subtle, difficult-to-discover bugs. The bug you hit, @saper, fails loudly and clearly. It is easy to understand what the problem is. However, if we adjust to your proposed behaviour, the failure mode is that someone has passed us something like a `datetime`, leading to Requests sending a header that reads `Date: datetime(days=52, hours=23, minutes=12, seconds=0)`, which is a _clearly_ incorrect header value. That failure mode is much harder to debug: all that you see in your code is a 400 Client Error that needs to be chased down, requiring you to examine everything about your request.

Subtle failure modes are almost always bad. I appreciate that this is an inconvenience for you, but we never intended your use-case to work, and historically it did not. It was by accident that it became acceptable several releases ago, and we have now taken action to formally prevent that use-case from being allowed again.

saper:
Question: Did implicit stringification work correctly in both Python releases? (Before we needed an explicit one to cut away newlines?). Except for the `repr()` case of course?

I've seen [{k: str(v) for k, v in headers.items()}](https://github.com/jjjake/internetarchive/pull/150/files#diff-562526d0e51ec94fea4ddda8c54853e5R155) proposed as a fix and I understand from what you are saying that explicit stringification is no good.

Lukasa:
That depends on your definition of "correctly", but basically: no. All my complaints about the weirder datatypes remain in place.

saper:
Thank you, that makes sense. But wouldn't be good to mark it as a breaking change? A relatively innocent implicit `int` conversion just stopped to work. Not sure if there were people implementing `__str__()` to have their HTTP headers formatted nicely.

Lukasa:
That's a reasonable question, and we discussed it. The conclusion was that, given that we had documented that header were strings, that we were covered by that restriction. I felt less strongly on that point than the other core maintainers, but the call was made. =)

saper:
Well, let it be then, thanks!

Lukasa:
Thanks for the understanding @saper! =)

PR #3368, opened 2016-07-01
Allow None value for chunk_size again

joyzheng:
Passing in `chunk_size=None` used to be fine, but started raising a `TypeError` following the addition of the type check (#3365)

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -685,7 +685,7 @@ class Response(object):
 
         if self._content_consumed and isinstance(self._content, bool):
             raise StreamConsumedError()
-        elif not isinstance(chunk_size, int):
+        elif chunk_size is not None and not isinstance(chunk_size, int):
             raise TypeError("chunk_size must be an int, it is instead a %s." % type(chunk_size))
         # simulate reading small chunks of the content
         reused_chunks = iter_slices(self._content, chunk_size)
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -987,14 +987,19 @@ class TestRequests:
         chunks = r.iter_content(decode_unicode=True)
         assert all(isinstance(chunk, str) for chunk in chunks)
 
-    def test_response_chunk_size_int(self):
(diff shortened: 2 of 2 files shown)


sigmavirus24:
Thanks @joyzheng! :cake:

nateprewitt:
It may be helpful to reword the docstring on iter_content to fit this functionality. Right now it reads as though the method is intended to avoid reading the contents all at once.

An unintended side effect of `chunk_size=None` is a failure in `iter_slices()`. The two code snippets below will both trigger [`reused_chunks` being chosen](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L695) as the return value of `iter_content`. The generator is produced by `iter_slices` which currently requires the chunk_size to be an integer, otherwise it raises a TypeError when [trying to add an int and NoneType](https://github.com/kennethreitz/requests/blob/master/requests/utils.py#L388).

I was going to open a pull request with:

```
if slice_length is None: 
    slice_length = 0
````

but this creates an infinite loop, since pos never increments. An alternative would be to set slice_length to 1, but I'm not sure if that'll be an IO issue for larger files. It also kind of defeats the point of setting `chunk_size=None` in the first place.


#### Examples ####
```
def test_stream():
    s = requests.Session()
    req = requests.Request('GET', 'https://httpbin.org/get').prepare()
    resp = s.send(req, stream=False)
    next(resp.iter_content(None))

def test_decode():
    r = requests.Response()
    r.raw = io.BytesIO(b'the content')
    all_content = r.content
    decoded_chunk = next(r.iter_content(None, True))
```

nateprewitt:
It may also be worth updating the `iter_content` docstring to reflect the "readall" behaviour with `iter_content(None)`.

joyzheng:
Yeah, good point. I'm not actually sure about all the behavior cases for `iter_content(None)`, since it doesn't always result in reading the entire input; the primary one I've used it for is chunked encoding with `stream=True`, where setting `chunk_size=None` iterates chunk-by-chunk.

Issue #3369, opened 2016-07-01
TypeError in iter_slices() when slice_length=None


nateprewitt:
An unintended side effect of using `chunk_size=None` (as referenced in PR #3368) is a failure in `iter_slices()`. The two code snippets below will both trigger [`reused_chunks` being chosen](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L695) as the return value of `iter_content`. The generator is produced by `iter_slices` which currently requires the chunk_size to be an integer, otherwise it raises a TypeError when [trying to add an int and NoneType](https://github.com/kennethreitz/requests/blob/master/requests/utils.py#L388).

Perhaps these are obscure enough edge cases, that it's not an issue. I just wanted to make a quick note of it.

#### Examples ####
```
def test_stream():
    s = requests.Session()
    req = requests.Request('GET', 'https://httpbin.org/get').prepare()
    resp = s.send(req, stream=False)
    next(resp.iter_content(None))

def test_decode():
    r = requests.Response()
    r.raw = io.BytesIO(b'the content')
    all_content = r.content
    decoded_chunk = next(r.iter_content(None, True))
```

Lukasa:
Why does `iter_slices` with `slice_length = None` not trigger a `TypeError`?

nateprewitt:
Sorry, I'm not sure I understand the question. I'm currently seeing the following error returned with both python2.7 and python3.5.

`TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'`

A more direct code snippet for reproduction would be:

`next(requests.utils.iter_slices('my string', None))`

Lukasa:
Ok, that's good, I just didn't read the report clearly enough I guess. =)

PR #3370, opened 2016-07-01
adding in slice_length fix and test for chunk_size=None

nateprewitt:
This fixes the issue discussed in #3369 but may not be the best way. It adds a certain amount of additional complexity to iter_slices that may be better solved by using a try/except. If we throw an exception requiring an int for slice_length, we could pass `chunk_size or len(self._content)` [here](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L691) and solve the issue that way as well.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -384,6 +384,8 @@ def stream_decode_response_unicode(iterator, r):
 def iter_slices(string, slice_length):
     """Iterate over slices of a string."""
     pos = 0
+    if slice_length is None or slice_length <= 0: 
+        slice_length = len(string)
     while pos < len(string):
         yield string[pos:pos + slice_length]
         pos += slice_length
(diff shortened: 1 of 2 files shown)


Lukasa:
I'm happy with this, but would like @sigmavirus24 to take a look as well.

nateprewitt:
Actually one quick note on this, it looks like we actually should have an integer > 0 as a requirement. slice_length <= 0 produces an infinite generator. I can modify the check if that's something you think should be caught (I do).

Lukasa:
Sure thing =)

Lukasa:
Ok, I'm still happy, but would still like @sigmavirus24 to weigh in.

sigmavirus24:
:+1: Looks fine to me. Thanks @nateprewitt!

PR #3382, opened 2016-07-01
updating iter_content docstring to match functionality

nateprewitt:
Sorry to be bombarding you guys with PRs. This one is pretty minor but explains what I would say was non-obvious behaviour of the function before PR #3368 and PR #3365.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -657,6 +657,12 @@ class Response(object):
         read into memory.  This is not necessarily the length of each item
         returned as decoding can take place.
 
+        chunk_size must be of type int or None. A value of None will
+        function differently depending on the value of `stream`.
+        stream=True will read data as it arrives in whatever size the
+        chunks are recieved. If stream=False, data is returned as
+        a single chunk.
+
         If decode_unicode is True, content will be decoded using the best
         available encoding based on the response.
         """


Lukasa:
So this isn't quite right. In particular, `iter_content` with `stream=True` and a chunk encoded response will iterate in chunk size. So it would be better for this to say that it iterates in chunks of undefined size.

nateprewitt:
Looks like I failed to set stream=True in my test case. I was misinterpreting the call to `r.content` as the stream read. Thanks @Lukasa, string updated.

requests/models.py
```diff
@@ -657,6 +657,9 @@ def iter_content(self, chunk_size=1, decode_unicode=False):
         read into memory.  This is not necessarily the length of each item
         returned as decoding can take place.
 
+        chunk_size must be of type int or None. A value of None implies
+        reading data in chunks of an undefined/inconsistent size.
+
```
Lukasa, line 665: Now that we're here, we may as well be clear about exactly what this does. It'll take quite a few more LoC, but it'll help us by being a better explanation.

nateprewitt, line 665: Alright, it's a bit more verbose now.

Lukasa:
Cool, I'm happy with this! Thanks @nateprewitt! :sparkles: :cake: :sparkles:

PR #3383, opened 2016-07-01
Change exception and variable names so that tests will run.

davidsoncasey:
@Lukasa @sigmavirus24 

I tried running the tests on the proposed/3.0.0 branch, and found that they did not run. So I changed a couple of variable names to fix the obvious errors. If you prefer InvalidSchema and MissingSchema I can change those back, I changed the imports to match the names of the exceptions as they were defined.

After these changes, there is still one failing test. I'll try to take a look and see if I can debug it. But this was low hanging fruit.


Lukasa:
Looks good to me, thanks!

Issue #3386, opened 2016-07-05
The keywords headers must be string or buffer in new version 2.10.0?


wut0n9:
The keywords `headers` value must be string or buffer?
If the  keywords `dnt` and `upgrade-insecure-requests`  value is  **string** or **buffer**, it's right,but the value is int,it's wrong.Why is that?
Right:
```
   'dnt': '1',
    'upgrade-insecure-requests': '1',
```
Wrong:
```
   'dnt': 1,
    'upgrade-insecure-requests': 1,
```
Why?

Lukasa:
@wut0n9 This behavioural change is not in v2.10.0, it's in the current master branch.

However, that's a real bug: #3366 has regressed this. @nateprewitt, are you interested in trying to update with a fix for this?

sigmavirus24:
Headers should always have been string. I don't think this is a significant regression if a regression at all. We've always documented that header values should be strings.

Lukasa:
Regardless of what we documented, this used to work and now it doesn't. We'll break a _lot_ of code if we don't change this back.

nateprewitt:
Yeah, I'll get right on this @Lukasa.

sigmavirus24:
See also https://github.com/kennethreitz/requests/issues/865 and https://github.com/kennethreitz/requests/pull/866

I don't know how much a "lot" actually is in this case.

nateprewitt:
So I have a working PR ready to go, I'm just trying work through a couple of questions. Would you rather I put them here or post the PR in its current state and work through it with you guys there?

Lukasa:
@sigmavirus24 has linked to the relevant earlier opinions, which suggest that actually we don't allow non-string header values. So that means this is not a bug: it was us making a revision that is within the scope of the API definition.

sigmavirus24:
@Lukasa right, I'm frankly surprised this didn't break earlier. The meaning of non-bytes/str as a header value is undefined as far as I'm concerned.

PR #3387, opened 2016-07-05
fixing #3366 for non-str header values

nateprewitt:
This will fix issue #3386 with non-string/buffer header values, but there may be some more discussion to be had here. This is a passthrough for non-string/buffers which means all other datatypes will skip the regex check. I wasn't able to create the header split issue with most of the other standard datatypes I checked (lists, sets, dicts), so this *should* be safe with the passthrough. 

I think an alternative, and possibly better solution is to cast everything that isn't a byte string as a `str`, whether that be standard, or `encode('latin-1')`. This will ensure the check still runs and doesn't modify the value in the actual `Requests.headers` value. This is probably closer to what @sigmavirus24 already did [here](https://github.com/kennethreitz/requests/pull/866/files#diff-5956087d5835a57d9ef6fff974f6fd9bR273).

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1145,9 +1145,10 @@ class TestRequests:
     def test_header_validation(self,httpbin):
         """Ensure prepare_headers regex isn't flagging valid header contents."""
         headers_ok = {'foo': 'bar baz qux',
-                      'bar': '1',
+                      'bar': u'fbbq'.encode('utf8'),
                       'baz': '',
-                      'qux': str.encode(u'fbbq')}
+                      'qux': 3,
+                      'f': '1'}
         r = requests.get(httpbin('get'), headers=headers_ok)
         assert r.request.headers['foo'] == headers_ok['foo']
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -748,8 +748,10 @@ def check_header_validity(header):
 
     if isinstance(value, bytes):
         pat = _CLEAN_HEADER_REGEX_BYTE
-    else:
+    elif isinstance(value, str):
         pat = _CLEAN_HEADER_REGEX_STR
+    else:
+        return 
     if not pat.match(value):
         raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
 


tests/test_requests.py
```diff
@@ -1145,9 +1145,10 @@ def test_header_keys_are_native(self, httpbin):
     def test_header_validation(self,httpbin):
         """Ensure prepare_headers regex isn't flagging valid header contents."""
         headers_ok = {'foo': 'bar baz qux',
-                      'bar': '1',
+                      'bar': str.encode(u'fbbq'),
```
sigmavirus24: Why `str.encode` and not `u'fbbq'.encode('utf-8')`?

sigmavirus24:
@nateprewitt that pull request was _rejected_ precisely because integers are not defined behaviour within requests and haven't been for _years_.

tests/test_requests.py
nateprewitt: It should be the latter, str.encode() was just what I had in the test previously.

nateprewitt:
Sorry, @sigmavirus24, I was suggesting something along this line for the check since it wouldn't be transforming the value in this case. I was trying to use this as a template example. I realize the documentation says that headers must be strings, and that was the assumption I was operating off of when I wrote #3366. I submitted the passthrough as the initial pull request here because it avoids dealing with any defining behaviour by ignoring non-strings.

nateprewitt:
Sorry, @sigmavirus24, I was suggesting something along this line for the check since it wouldn't be transforming the value in this case. I was trying to use this as a template example. I realize the documentation says that headers must be strings, and that was the assumption I was operating off of when I wrote #3366. I submitted the passthrough as the initial pull request here because it avoids dealing with any defining behaviour by ignoring non-strings.

I just wanted to make sure both options were at least briefly discussed.

Lukasa:
Yeah, so the question boils down to whether we handle non-string headers. Clearly we've oscillated around for a while: they didn't work, and then they did, and now they don't again.

However, what I failed to note is Kenneth's original response in #865. With that in mind, I'm inclined to want to continue to respect his wishes and say that non-string headers are not acceptable.

Sorry for having you do this work @nateprewitt!

nateprewitt:
No problem, glad to have the question resolved.

PR #3388, opened 2016-07-06
updating documentation to reflect decision of #3386

nateprewitt:
These are a couple of minor doc tweaks to document the decision to enforce strings in header values. Since requests previously operated with non-string header values, I figured it was best to make note that an exception will now be thrown in the event of a non-string value.

There may be some formatting and wording changes needed, this iteration is mainly to get the conversation going.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -258,6 +258,10 @@ Behavioural Changes
 
 * Keys in the ``headers`` dictionary are now native strings on all Python
   versions, i.e. bytestrings on Python 2 and unicode on Python 3. If the
-  keys are not native strings (unicode on Python2 or bytestrings on Python 3)
+  keys are not native strings (unicode on Python 2 or bytestrings on Python 3)
   they will be converted to the native string type assuming UTF-8 encoding.
 
+* Values in the ``headers`` dictionary should always be strings. This has
+  been the project's position since before 1.0 but a recent change
+  (since version 2.11.0) enforces this more strictly. It's advised to avoid
+  passing header values as unicode when possible.
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -211,6 +211,7 @@ Note: Custom headers are given less precedence than more specific sources of inf
 
 Furthermore, Requests does not change its behavior at all based on which custom headers are specified. The headers are simply passed on into the final request.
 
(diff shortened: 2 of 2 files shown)


docs/user/quickstart.rst
```diff
@@ -211,6 +211,7 @@ Note: Custom headers are given less precedence than more specific sources of inf
 
 Furthermore, Requests does not change its behavior at all based on which custom headers are specified. The headers are simply passed on into the final request.
 
+Note: All header values must be a ``string``, bytestring (on Python 2), or unicode (on Python 3).
```
Lukasa: This isn't quite accurate: they're allowed to be bytestrings or unicode on any version of Python, it's just that passing unicode is rarely a good idea. =)

nateprewitt: Tweaked the wording to avoid version listing, and added a suggestion to avoid passing unicode values in both places.

Lukasa:
Cool, LGTM! :sparkles: @sigmavirus24?

nateprewitt:
While you're here @sigmavirus24 and @Lukasa, I threw together [another commit](https://github.com/nateprewitt/requests/commit/be31a90906deb5553c2e703fb05cf6964ee23ed5) related to this to encapsulate the type error thrown by `re` when non-strings are passed to `check_header_validity`. This is to make it clear that the behaviour is to be expected, but may be overkill with this documentation now. Any thoughts on if this would be worth opening?

docs/api.rst
```diff
@@ -258,6 +258,9 @@ Behavioural Changes
 
 * Keys in the ``headers`` dictionary are now native strings on all Python
   versions, i.e. bytestrings on Python 2 and unicode on Python 3. If the
-  keys are not native strings (unicode on Python2 or bytestrings on Python 3)
+  keys are not native strings (unicode on Python 2 or bytestrings on Python 3)
   they will be converted to the native string type assuming UTF-8 encoding.
 
+* Values in the ``headers`` dictionary are now required to be passed as strings,
```
sigmavirus24: I would rather say

```
Values in the ``headers`` dictionary should always be strings. This has been the 
project's position since before 1.0 but a recent change (since version 2.11.0) enforces 
this more strictly. It's advised to avoid passing header values as unicode when possible.
```

nateprewitt: I like the sound of that better too, docs are updated.

sigmavirus24:
Thanks @nateprewitt 🍮 🍰 ☕

PR #3390, opened 2016-07-07
Defining header value type requirements and tests

nateprewitt:
Currently a non-string/bytes value will cause `pat.match(value)` to raise a TypeError from `re`. I'm proposing catching this exception and raising it as a more descriptive `InvalidHeader` instead, so that it's clear we're intending this to happen.


sigmavirus24:
👍 from me. @Lukasa ?

Lukasa:
LGTM, let's do it! :sparkles:

PR #3401, opened 2016-07-14
in python3.x not have StringIO

Zeusw:
I think, should such an amendment.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -132,7 +132,7 @@ For example, to create an image from binary data returned by a request, you can
 use the following code::
 
     >>> from PIL import Image
-    >>> from StringIO import StringIO
+    >>> from io import StringIO
 
     >>> i = Image.open(StringIO(r.content))
 


Lukasa:
Thanks! :sparkles:

kennethreitz:
AFAIK (I haven't checked), all of our documentation is written with 2.x in mind. If this holds true, I believe this is a slightly user-hostile change, as the majority of user userbase is not using Python 3.

If we want to be subtly suggestive that they should be using 3, we need to make sure it's happening everywhere, not just in this one spot. 

Again, I haven't checked.

Lukasa:
@kennethreitz Python 2.7 also has the `io` module: this change keeps the code working as expected on Python 2.

kennethreitz:
Ah yes, I forgot a `StringIO` class was also available under `io` in 2.7. Obviously, it's fine to not pander to 2.6 users :)

Either way, my above comment is generally relevant. We may want to start making our examples 3.x soon, if we want to "move the needle". I don't think there's any rush to do so, though.

sigmavirus24:
> Either way, my above comment is generally relevant. We may want to start making our examples 3.x soon, if we want to "move the needle". I don't think there's any rush to do so, though.

@kennethreitz have you looked at the downloads for requests recently? I'm not sure which version is "winning" but regardless, I think we can write the examples to work across all supported versions without having to have an opinion on whether they're all Py2 or Py3 or moving any needle.

kennethreitz:
I have, yes.

That would be my preferred method too. I'm speaking more if we'd like to make a more political move within our documentation, akin to what the Django project has done.

I don't think we need to, but am brining it up as a possibility.

PR #3415, opened 2016-07-15
removing redundant test

nateprewitt:
I was writing up a new test and found this comment block. Looks like it was commented out as part of a cleanup effort, but never removed. Test functionality is fulfilled by [this test](https://github.com/kennethreitz/requests/blob/master/tests/test_requests.py#L196).


Lukasa:

Issue #2939, opened 2015-12-19
Multiple Location headers in response are concatenated


L2501:
tested with requests 2.9
when a http server resonds with multiple Location headers in 301 redirect the locations are concatenated 
eg
Location: http://foo.com
Location: http://foo.com

results in a redirect to 
"http://foo.com,http://foo.com"

pretty sure this behavior is not present in any other http clients

Lukasa:
A web server MUST NOT respond with multiple location header fields. From [RFC 7320 Section 3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2):

> A sender MUST NOT generate multiple header fields with the same field name in a message unless either the entire field value for that header field is defined as a comma-separated list [i.e., #(values)] or the header field is a well-known exception (as noted below).

There is exactly _one_ well-known exception: `Set-Cookie`. `Location` is not `Set-Cookie`, so that doesn't apply. So it would only be acceptable to send multiple `Location` headers if the syntax allowed multiple comma-separated values. `Location` is defined in [RFC 7231 Section 7.1.2](https://tools.ietf.org/html/rfc7231#section-7.1.2):

>    Location = URI-reference
> 
> The field value consists of a single URI-reference.

The server in this example is violating the HTTP specification. In the face of such violation, a user-agent is simply not required to take any specific action. In particular, while the example you provided above has a very simple course of action (both URLs are the same), if both URLs were _not_ the same requests would be unable to take any sensible action: the situation is too ambiguous to allow a correct decision.

At the level of requests actually catching this and treating it as an error condition is tricky, because requests only sees the header string `http://foo.com,http://foo.com`, which perversely does parse as a valid URL in the case of most Python libraries because in principle the comma is an allowed character in the host part of the URL. This puts us in a bit of a bind. However, `urllib3`'s `HTTPHeaderDict` structure could enforce this requirement and treat this case as an error on receiving it. That might be a sensible way to go, but it's us going out of our way to provide a slightly nicer error when a server does something it really really shouldn't do. @sigmavirus24, thoughts?

L2501:
I guess then the web server is doing this to break python-urllib clients as it does not seem to affect web browsers or curl

Lukasa:
[...]

PR #3417, opened 2016-07-15
raise InvalidHeader on multiple Location values

nateprewitt:
Addresses issue raised in #2939 with a fix for multiple Location headers in response. This currently breaks on `test_requests_are_updated_each_time` because the [`RedirectSession`](https://github.com/kennethreitz/requests/blob/proposed/3.0.0/tests/test_requests.py#L1564) populates it's `raw` attribute with `StringIO` rather than a urllib3 `HTTPResponse`. I can convert the `_build_raw` method to fix the test, if this looks like a workable solution.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -20,7 +20,8 @@ from .models import Request, PreparedRequest, DEFAULT_REDIRECT_LIMIT
 from .hooks import default_hooks, dispatch_hook
 from .utils import to_key_val_list, default_headers, to_native_string
 from .exceptions import (
-    TooManyRedirects, InvalidScheme, ChunkedEncodingError, ContentDecodingError)
+    TooManyRedirects, InvalidScheme, ChunkedEncodingError,
+    ContentDecodingError, InvalidHeader)
 from .packages.urllib3._collections import RecentlyUsedContainer
 from .structures import CaseInsensitiveDict
 
@@ -98,6 +99,10 @@ class SessionRedirectMixin(object):
         request = response.request
 
         while response.is_redirect:
+            if len(response.raw.headers.getlist('location')) > 1:
+                raise InvalidHeader('Response contains multiple Location headers. '
+                                    'Unable to perform redirect.')
+
             prepared_request = request.copy()
 
             if redirect_count > 0:
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -29,7 +29,7 @@ from .adapters import HTTPAdapter
 
 from .utils import (
     requote_uri, get_environ_proxies, get_netrc_auth, should_bypass_proxies,
-    get_auth_from_url
+    get_auth_from_url, is_valid_location
 )
 
 from .status_codes import codes
@@ -99,7 +99,7 @@ class SessionRedirectMixin(object):
         request = response.request
 
         while response.is_redirect:
-            if len(response.raw.headers.getlist('location')) > 1:
+            if not is_valid_location(response):
                 raise InvalidHeader('Response contains multiple Location headers. '
                                     'Unable to perform redirect.')
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -665,6 +665,17 @@ def parse_header_links(value):
 
     return links
 
+def is_valid_location(response):
+    """Verify that multiple Location headers weren't
+    returned from the last response.
+    """
+    headers = getattr(response.raw, 'headers', None)
+    if headers is not None:
+        getlist = getattr(headers, 'getlist', None)
+        if getlist is not None:
+            return len(getlist('location')) <= 1
+    # If response.raw isn't urllib3-like we can't reliably check this
+    return True  
 
 # Null bytes; no need to recreate these on each call to guess_json_utf
 _null = '\x00'.encode('ascii')  # encoding to ASCII for Python 3
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -83,6 +83,10 @@ class InvalidURL(RequestException, ValueError):
     """ The URL provided was somehow invalid. """
 
 
+class InvalidHeader(RequestException, ValueError):
+    """The header value provided was somehow invalid."""
+
+
 class ChunkedEncodingError(RequestException):
     """The server declared chunked encoding but sent an invalid chunk."""
 
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1626,7 +1626,7 @@ def test_requests_are_updated_each_time(httpbin):
     r0 = session.send(prep)
     assert r0.request.method == 'POST'
     assert session.calls[-1] == SendCall((r0.request,), {})
-    redirect_generator = session.resolve_redirects(r0, prep)
+    redirect_generator = session.resolve_redirects(r0)
     default_keyword_args = {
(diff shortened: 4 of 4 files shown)


Lukasa:
@nateprewitt Thanks for starting this!

Unfortunately, quite a lot of people use non-urllib3 file-like objects to back requests, and I'd like to avoid breaking that if we can. Can you rearrange this to support urllib3 if possible but otherwise fallback to some other behaviour, [per my comment](https://github.com/kennethreitz/requests/issues/2939#issuecomment-166003526).

requests/sessions.py
```diff
@@ -98,6 +99,10 @@ def resolve_redirects(self, response, stream=False, timeout=None,
         request = response.request
 
         while response.is_redirect:
+            if len(response.raw.headers.getlist('location')) > 1:
```
Lukasa: This will need to include a `getattr` check. Frankly, it's a bit implementation detaily, so we may want to hide this in a utility function.

sigmavirus24:
This looks good to me.

sigmavirus24:
Sadly we have no CI for this though, so one or both of us should pull this down and test it.

nateprewitt:
Sorry, @sigmavirus24, I just pushed a new copy with partially finished changes. I have a couple of questions on how we're expecting the code to function that I was going to annotate for discussion.

requests/utils.py
```diff
@@ -665,6 +665,21 @@ def parse_header_links(value):
...
+    """
+    if hasattr(response.raw, 'headers'):
+        if hasattr(response.raw.headers, 'getlist'):
+            return len(response.raw.headers.getlist('location')) <= 1
+        else:
+            pass # Do something with header value
+    elif hasattr(response.raw, 'read'):
+        response.raw.seek(0) # Move to beginning of file
+        data = response.raw.read()
+        # parse_headers(data) # try to parse headers out of raw response
```
nateprewitt: Going through a few libraries that implement different adapters, the `raw` field seems considerably varied in it's use. 

In the event we end up with something file-like, such as `StringIO`, we can parse the headers out, but only if it's the complete response. In the limited testing I did though, this was typically just the response body. I'm not sure if we can do anything useful in that case. Also if it is the response body and happens to contain contents mimicking an HTTP response, then we could in theory flag this incorrectly.

I'm just seeing a lot of unknowns here that I'm not sure have a clean fix. Perhaps just returning True when we can't make a reasonable discernment would be best.

requests/utils.py
```diff
@@ -665,6 +665,21 @@ def parse_header_links(value):
...
 
+def is_valid_location(response):
+    """Verify that multiple Location headers weren't
+    returned from the last response.
+    """
+    if hasattr(response.raw, 'headers'):
+        if hasattr(response.raw.headers, 'getlist'):
+            return len(response.raw.headers.getlist('location')) <= 1
+        else:
+            pass # Do something with header value
```
nateprewitt: Here we'll likely have a header value similar to `'http://foo.com, http://foo.com'` which could be parsed but probably not reliably. I recognize that URLs can contain unencoded `,` but header values are also defined to be comma separated, so I'm not sure how the case of a multi-value header with values containing commas is handled. A cursory google search wasn't terribly informative. 

Browsers seem to catch this problem at what would be our urllib3 layer before the field merge is ever done. Patching it there is null though, because here we're concerned about how to handle the issue for non-urllib3 responses. Perhaps this is another case we can't handle, but then that puts us back at ground-zero for the pull request (other than preventing the exception for StringIO). Perhaps that what @Lukasa was asking for though and I severely overthought the problem.

requests/utils.py
Lukasa: I think this is more than we need. If we can't get headers in the above case, don't worry about it. 

Mostly what I'm concerned about here is not exploding in the non-urllib3 case: we don't need to be too clever, or even really "handle" it in any meaningful way.

nateprewitt: Ok, great, that helps narrow the scope considerably. I'll shorten this down to just catching the urllib3 case then. Thanks!

nateprewitt:
Ok, I think this is probably ready for another glance. I'm getting an `AssertionError` from `test_requests_are_updated_each_time` on line 1641, but that appears to be pre-existing in the proposed/3.0.0 branch. I was able to reproduce it in a fresh virtualenv with `proposed/3.0.0` checked out with python 2.7.

nateprewitt:
Ok, I think this is probably ready for another glance. I'm getting an `AssertionError` from `test_requests_are_updated_each_time` on line 1641, but that appears to be pre-existing in the `proposed/3.0.0` branch. I was able to reproduce it in a fresh virtualenv with `proposed/3.0.0` checked out with python 2.7.

requests/utils.py
```diff
@@ -665,6 +665,15 @@ def parse_header_links(value):
 
     return links
 
+def is_valid_location(response):
+    """Verify that multiple Location headers weren't
+    returned from the last response.
+    """
+    if hasattr(response.raw, 'headers'):
+        if hasattr(response.raw.headers, 'getlist'):
```
Lukasa: So we shouldn't use hasattr, we should use getattr. This is because [of this](https://hynek.me/articles/hasattr/).

Lukasa: To be clear, I suggest the `getattr` pattern in this case.

kennethreitz: Per usual, I prefer the Python 2 behavior here hahaha

nateprewitt: Yeah, I'm onboard with this. I updated the PR with a somewhat verbose implementation of `getattr`.

Lukasa:
Ok, I'm happy with this. @sigmavirus24?

nateprewitt:
On a quick side note, I PR'd this against `proposed/3.0.0` because of comments in the issue, but is there a reason this shouldn't go straight against `master`?

Lukasa:
@nateprewitt Yeah, this is a breaking change. =)

nateprewitt:
@Lukasa @sigmavirus24 Not sure what SOP is for `proposed/3.0.0` merges, but let me know if I'm holding anything up :)

nateprewitt:
Hey @Lukasa @sigmavirus24, just wanted to check in on this. Let me know if you're waiting on anything from me.

Lukasa:
No problem from me. @sigmavirus24?

sigmavirus24:
Looks fine as a start. Thanks @nateprewitt 

@Lukasa since we're sans CI, have you pulled this and run the tests?

Lukasa:
@sigmavirus24 I haven't, would you like me to?

sigmavirus24:
I'll do it. Just wanted to make sure someone was doing it.

nateprewitt:
It's still failing on #3442 because I haven't rebased on to master since the merge. Let me do that really quickly.

sigmavirus24:
@nateprewitt I merged the original branch into the head of proposed/3.0.0 and it passed just fine. No need to rebase

sigmavirus24:
Oh I see you already did, effectively invalidating the tests. Let me run them again

nateprewitt:
Sorry, I'd already pushed it before the message.

sigmavirus24:
Tests passed both times.

Issue #3419, opened 2016-07-15
Requests connectionpool.py is spammy


springjools:
Connectionpool spams always when starting connection. It should be DEBUG log level  and not INFO. Thank you.

Lukasa:
This is a log coming from the urllib3 module, not from requests. Please open this bug report there. =)

PR #3420, opened 2016-07-15
Update connectionpool.py

springjools:
Change some spammy DEBUG logs to the proper log level.


Lukasa:
As mentioned in #3419, these logs come from the urllib3 project. Requests does not carry patches to that project: you'll need to submit this fix upstream.

PR #3426, opened 2016-07-20
converting hasattr usages to getattr

nateprewitt:
This spawned off of [an article](https://hynek.me/articles/hasattr/) @Lukasa brought up in another PR. These changes should fix the pitfalls with using `hasattr` by converting their logic to use `getattr` instead.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -419,9 +419,7 @@ class HTTPAdapter(BaseAdapter):
 
             # Send the request.
             else:
-                if hasattr(conn, 'proxy_pool'):
-                    conn = conn.proxy_pool
-
+                conn = getattr(conn, 'proxy_pool', conn)
                 low_conn = conn._get_conn(timeout=DEFAULT_POOL_TIMEOUT)
 
                 try:
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -120,8 +120,7 @@ def extract_cookies_to_jar(jar, request, response):
     :param request: our own requests.Request object
     :param response: urllib3.HTTPResponse object
     """
-    if not (hasattr(response, '_original_response') and
-            response._original_response):
+    if not getattr(response, '_original_response', None):
         return
     # the _original_response field is the wrapped httplib.HTTPResponse object,
     req = MockRequest(request)
@@ -305,8 +304,11 @@ class RequestsCookieJar(cookielib.CookieJar, collections.MutableMapping):
         remove_cookie_by_name(self, name)
 
     def set_cookie(self, cookie, *args, **kwargs):
-        if hasattr(cookie.value, 'startswith') and cookie.value.startswith('"') and cookie.value.endswith('"'):
-            cookie.value = cookie.value.replace('\\"', '')
+        try:
+            if cookie.value.startswith('"') and cookie.value.endswith('"'):
+                cookie.value = cookie.value.replace('\\"', '')
+        except AttributeError:
+            pass
         return super(RequestsCookieJar, self).set_cookie(cookie, *args, **kwargs)
 
     def update(self, other):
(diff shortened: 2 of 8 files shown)


requests/adapters.py
```diff
@@ -419,8 +419,9 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
 
             # Send the request.
             else:
-                if hasattr(conn, 'proxy_pool'):
-                    conn = conn.proxy_pool
+                proxy_pool = getattr(conn, 'proxy_pool', None)
+                if proxy_pool is not None:
+                    conn = proxy_pool
```
Lukasa: Ok, so now I'm looking closer.

We can translate this to `getattr(conn, 'proxy_pool', conn)`.

Lukasa: That avoids the need for the `if` block altogether.

requests/cookies.py
```diff
@@ -120,8 +120,7 @@ def extract_cookies_to_jar(jar, request, response):
     :param request: our own requests.Request object
     :param response: urllib3.HTTPResponse object
     """
-    if not (hasattr(response, '_original_response') and
-            response._original_response):
+    if getattr(response, '_original_response', None) is None:
```
Lukasa: This one doesn't need to be an `is None`, sorry: given that we're doing a bool test, it just needs to fall back to `None`.

requests/cookies.py
```diff
@@ -305,8 +304,9 @@ def __delitem__(self, name):
         remove_cookie_by_name(self, name)
 
     def set_cookie(self, cookie, *args, **kwargs):
-        if hasattr(cookie.value, 'startswith') and cookie.value.startswith('"') and cookie.value.endswith('"'):
-            cookie.value = cookie.value.replace('\\"', '')
+        if (isinstance(cookie.value, (str, bytes)) and cookie.value.startswith('"')
+            and cookie.value.endswith('"')):
+                cookie.value = cookie.value.replace('\\"', '')
```
Lukasa, line 309: I think `try...except AttributeError` is better than the type check here.

nateprewitt:
Ok, changes have been implemented.

kennethreitz:
What is the point of this? Because of the way properties are actually namespaced in Python 2?

Lukasa:
@kennethreitz Same reason: https://hynek.me/articles/hasattr/

kennethreitz:
Yeah, I read the article. I still don't see why we're making these changes.

Lukasa:
The short answer is: because it marginally reduces the likelihood of us encountering bugs and tripping over errors. They're not mandatory, but from my perspective they don't hurt.

kennethreitz:
Well, my opinion is that this is an extremely easy to comprehend built-in function that is not causing us any problems (obviously), and removing all use of it suddenly (and making the code ugly in the process) is a bit reactionary and either some strange form of premature optimization or simply silliness. 

The first change in this diff is the only one I like (in `adapters.py`).

nateprewitt:
This was thrown together mostly as a nice-to-have. I agree that `is not None` is kind of an ugly idiom in this context, but figured covering the accidental execution may be helpful. I'm fine either paring this down to the diff you like (could we get a link to that one?) or closing it, if the clutter is determined to outweigh the edge case.

kennethreitz:
P.S. there are no pitfalls to `hasattr`, it's doing exactly what it's supposed to do. Properties are weird in Python 2, and if you want to pin "pitfalls" on anything, it should be them. `hasattr` works exactly as expected, which can be seen if you use it against a property's actual getter or setter methods.

kennethreitz:
I'm -10, but I'm not the only one here. I'm just sort of shocked this got so far :)

nateprewitt:
I'll go ahead and close this for now. Let me know if there's any use in the `adapters.py` change and I'll rebase this commit.

Issue #3427, opened 2016-07-20
Code Consistency Work


nateprewitt:
This is a combination of two passes over the code to help cleanup some of the coding style differences that have creeped in over time. It's mostly cosmetic but I think it'll help reduce inconsistencies going forward by having the code be an example for future contributors. Per @Lukasa's request, I'm going to break this up into two pull requests. One involving docstrings and the other involving pep8 fixes. I don't know that these changes are necessarily "correct", I simply chose what appeared to be the most common occurrence in the code. I'm sure @kennethreitz will have opinions on which choices are best.

kennethreitz:
Is this supposed to be a pull request?

Lukasa:
@kennethreitz, I kinda want you to take the lead on code review for this: just flag changes you don't like and I'm sure @nateprewitt will undo it.

Lukasa:
Nah, it's an umbrella issue that'll cover two PRs.

kennethreitz:
👍

PR #3429, opened 2016-07-20
Docstring cleanup

nateprewitt:
This is adding in mostly cosmetic fixes to docstrings, including some sphinx additions. I think these will help catch docstring style choices that are determined by looking at the function above or below where you're adding your new one.

This is part two of two for #3427.

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -38,7 +38,6 @@ is at <http://python-requests.org>.
 
 :copyright: (c) 2016 by Kenneth Reitz.
 :license: Apache 2.0, see LICENSE for more details.
-
 """
 
 __title__ = 'requests'
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -957,8 +957,7 @@ class TestRequests:
         io.close()
 
     def test_response_decode_unicode(self):
-        """
-        When called with decode_unicode, Response.iter_content should always
+        """When called with decode_unicode, Response.iter_content should always
         return unicode.
         """
         r = requests.Response()
@@ -1049,9 +1048,7 @@ class TestRequests:
         assert r.status_code == 200
 
     def test_fixes_1329(self, httpbin):
-        """
-        Ensure that header updates are done case-insensitively.
-        """
+        """Ensure that header updates are done case-insensitively."""
         s = requests.Session()
         s.headers.update({'ACCEPT': 'BOGUS'})
         s.headers.update({'accept': 'application/json'})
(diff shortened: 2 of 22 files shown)


tests/test_requests.py
```diff
@@ -1533,28 +1520,24 @@ def test_expires_invalid_int(self, value, exception):
...
         """Test case where a valid max age in seconds is passed."""
-
         morsel = Morsel()
         morsel['max-age'] = 60
         cookie = morsel_to_cookie(morsel)
         assert isinstance(cookie.expires, int)
 
     def test_max_age_invalid_str(self):
         """Test case where a invalid max age is passed."""
-
```
kennethreitz, line 1557: I think this (the line of whitespace you're removing), makes the code more readable. There are many more examples above. 

We certainly can collapse them, but I don't think that's necessary, and I do think it will make the code more difficult to visually parse.

kennethreitz:
Left a note.

tests/test_requests.py
nateprewitt, line 1557: I agree it does provide a more visually appealing layout with the separation. I chose the removal because it appeared to be more common in the code base. Would you prefer I move things in the other direction (adding a blank line between docstrings and code start), or just leaving them alone entirely.

kennethreitz, line 1557: I don't think they always need to be there — but I wouldn't be surprised if there are instances where readability could be improved by adding it.

nateprewitt, line 1557: Alright, I'll drop c50eb62 for now and possibly make another pass at fixing any glaring space issues in a separate PR tied to the parent issue.

requests/__init__.py
```diff
@@ -38,7 +38,6 @@
 
 :copyright: (c) 2016 by Kenneth Reitz.
 :license: Apache 2.0, see LICENSE for more details.
-
```
nateprewitt, line 41: Do we want the spaces here for readability?

nateprewitt:
Commit removing whitespace between docstrings and code is removed. I did have another question regarding the acceptability of an extra blank line before the closing `"""` of a docstring.

[e.g.](https://github.com/kennethreitz/requests/pull/3429/files#diff-375a14e45cf81919afe573d3a1364c09L41)

requests/__init__.py
kennethreitz, line 41: I think this one's fine to remove

nateprewitt, line 41: Ok, great! I think things are set on my end unless you have any other tweaks.

sigmavirus24:
These all look rather straightforward.

kennethreitz:
:sparkles: :cake: :sparkles:

Issue #3430, opened 2016-07-22
better default exception strings


dhduvall:
I'm getting a bunch of stack traces from an app (various OpenStack componentry) which end with the following message

    ConnectionError: ('Connection aborted.', error(146, 'ECONNREFUSED'))

or something similar.  In all cases, the underlying problem is bad administrative configuration -- that is, not strictly a bug in either the app or in requests -- but the message is so vague that it's a huge pain in the ass to try to track down what it was trying to connect to when it failed.  Sometimes by looking at the stack trace I can figure out what it was trying to do, but not always.

I could go and patch the apps to catch exceptions from requests and have it print out the URL, but that's potentially a lot of places.  I was wondering if it would make sense to have a `__str__()` method for `RequestException` that would spit out the URL in addition to the basic junk from `IOError`.  I tried

    def __str__(self):
        s = super(RequestException, self).__str__()
        return "%s: %s" % (self.request.url, s)

and it did what I expected, at least for the one test case of connecting to an unlistened-on port.

I'm not looking for end-user-appropriate strings like in issue #2876, but just something to help out a reasonably savvy administrator unscared of things like stack traces and other gobbledygook, but short on time and patience to go figure out what the software should be telling her anyway.

Obviously more information might be necessary to diagnose certain problems, since the problem might be related to the request data, or other possibly large data, which you almost certainly would never want to show by default, but usually the requested URL is a useful point of information.

Thoughts?

sigmavirus24:
So, as a person who contributes to OpenStack, I'll say this:

The fact that you're seeing an exception from Requests is a serious problem. Whatever service you're using, might be logging these but should also be logging (with it) an explanation of what happened. Granted, it's OpenStack, so I really shouldn't expect that level of quality, but this is (in my opinion) more of a bug in whatever service caused you this grief than in Requests.

---

The above aside, I'm +0 on this. I really think our exception messages aren't ever meant for anyone other than the developer who is writing the application using requests, but I can imagine how this _might_ be useful. I'm not +1 on this because we've been trying to attach request objects to most of these errors as it is, so if developers are going to log Requests' exceptions they can display information about the request too.

Lukasa:
I'm not averse to this in principle. I suspect we'll find getting this right across things like redirects is a bit tricky, but we can worry about that when we do it. If you want to submit a patch, @dhduvall, I'll review it. =)

dhduvall:
[...]

PR #3439, opened 2016-07-22
Put the failing URL into the default RequestException string (issue 3430)

dhduvall:


Lukasa:
I'm happy enough with this: :+1:

sigmavirus24:
To be clear, I assigned @kennethreitz to this because I think this is something he's going to care about and have opinions about. I'm fine with this but I'll leave it to him to merge.

kennethreitz:
#3430

kennethreitz:
An example of what this would actually look like in practice would be nice. I'm not going to interrupt my workflow at the moment to download and see for myself. Maybe in a few days I will be.

dhduvall:
At some point between 2.7.0 and 2.8.0, doing something like `requests.Session().request("GET", "http://badhostname/")` went from dropping

```
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.7/vendor-packages/requests/sessions.py", line 464, in request
  File "/usr/lib/python2.7/vendor-packages/requests/sessions.py", line 576, in send
  File "/usr/lib/python2.7/vendor-packages/requests/adapters.py", line 415, in send
    raise ConnectionError(err, request=request)
requests.exceptions.ConnectionError: ('Connection aborted.', gaierror(8, 'node name or service name not known'))
```

to

```
  File "<string>", line 1, in <module>
  File "requests/sessions.py", line 471, in request
  File "requests/sessions.py", line 581, in send
  File "requests/adapters.py", line 481, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPConnectionPool(host='badhostname', port=80): Max retries exceeded with url: / (Caused by NewConnectionError('<requests.packages.urllib3.connection.HTTPConnection object at 0xed5ac0ec>: Failed to establish a new connection: [Errno 8] node name or service name not known',))
```

Same with connecting to a bad port on a good hostname, which was my original itch.  That's probably good enough for me, and just prepending the URL isn't super readable anymore, since it adds to an already very long line.  My apologies for making noise when an upgrade would have been sufficient (we've been on 2.6.0, but will be upgrading to 2.9 soon).

Lukasa:
Thanks for following up @dhduvall, and thanks for the work, even though we will probably no longer accept it. =)

PR #3443, opened 2016-07-27
Fix quantity of blank lines after code objects.

stepshal:


kennethreitz:
I don't think these particular changes are helpful. Thank you though!

PR #3444, opened 2016-07-27
Adding better logging for requests/responses

seemethere:
# Description
A logging mechanism to display important information regarding all requests and responses sent through the `requests` api.

# Why is this necessary?
At the company I work for we've been using this [logging mechanism](https://github.com/openstack/opencafe/blob/master/cafe/plugins/http/cafe/engine/http/client.py#L29) for a very long time and have had great success with it for testing numerous APIs. The level of logging provides the user with information they would normally have to dig for, and also provides a ready made log from which to draw relevant information. This is especially important for testing as it provides testers with sufficient logs to assess what is wrong, if a request should fail.

This information is even valuable for API developers as they can get immediate, easy to see, feedback on what their API is doing and returning.

# Try it for yourself!:
## Code:
```python
import logging;root=logging.getLogger();root.addHandler(logging.StreamHandler());root.setLevel(0);
import requests
requests.post('https://jsonplaceholder.typicode.com/posts', data="""{"id": 101, "title": "foo", "body": "bar", "userId": 1}""", headers={'content-type': 'application/json'})
```
## Logging output:
```
(<requests.sessions.Session object at 0x111000a90>,) {'data': '{"id": 101, "title": "foo", "body": "bar", "userId": 1}', 'url': 'https://jsonplaceholder.typicode.com/posts', 'method': 'post', 'headers': {'content-type': 'application/json'}, 'json': None}
Starting new HTTPS connection (1): jsonplaceholder.typicode.com
"POST /posts HTTP/1.1" 201 65

------------
REQUEST SENT
------------
request method..: POST
request url.....: https://jsonplaceholder.typicode.com/posts
request params..:
request headers.: {'Content-Length': '55', 'Connection': 'keep-alive', 'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'User-Agent': 'python-requests/2.10.0', 'content-type': 'application/json'}
request body....: {"id": 101, "title": "foo", "body": "bar", "userId": 1}


-----------------
RESPONSE RECEIVED
-----------------
response status..: <Response [201]>
response time....: 0.7399938106536865
response headers.: {'CF-RAY': '2c8d2175a6d20108-DFW', 'X-Content-Type-Options': 'nosniff', 'Etag': 'W/"41-1twMC34dpcMvmO2NLu+Kkg"', 'Pragma': 'no-cache', 'Server': 'cloudflare-nginx', 'Cache-Control': 'no-cache', 'Content-Type': 'application/json; charset=utf-8', 'Content-Length': '65', 'Set-Cookie': '__cfduid=dadf01d51aae9ede807ecf3e77439a3d51469591463; expires=Thu, 27-Jul-17 03:51:03 GMT; path=/; domain=.typicode.com; HttpOnly', 'X-Powered-By': 'Express', 'Via': '1.1 vegur', 'Vary': 'Origin, X-HTTP-Method-Override, Accept-Encoding', 'Date': 'Wed, 27 Jul 2016 03:51:03 GMT', 'Access-Control-Allow-Credentials': 'true', 'Expires': '-1', 'Connection': 'keep-alive'}
response body....: b'{\n  "id": 101,\n  "title": "foo",\n  "body": "bar",\n  "userId": 1\n}'
-------------------------------------------------------------------------------
```

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -8,6 +8,7 @@ This module provides a Session object to manage and persist settings across
 requests (cookies, auth, proxies).
 """
 import os
+import logging
 from collections import Mapping
 from datetime import datetime
 
@@ -27,7 +28,7 @@ from .adapters import HTTPAdapter
 
 from .utils import (
     requote_uri, get_environ_proxies, get_netrc_auth, should_bypass_proxies,
-    get_auth_from_url
+    get_auth_from_url, _log_transaction
 )
 
 from .status_codes import codes
@@ -36,6 +37,8 @@ from .status_codes import codes
 from .models import REDIRECT_STATI
 
 REDIRECT_CACHE_SIZE = 1000
+
+log = logging.getLogger(__name__)
 
 
 def merge_setting(request_setting, session_setting, dict_class=OrderedDict):
@@ -391,6 +394,7 @@ class Session(SessionRedirectMixin):
         )
         return p
 
+    @_log_transaction(log)
     def request(self, method, url,
         params=None,
         data=None,
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -12,11 +12,14 @@ import cgi
 import codecs
 import collections
 import io
+import logging
 import os
 import re
 import socket
+import six
 import struct
 import warnings
+from time import time
 
 from . import __version__
 from . import certs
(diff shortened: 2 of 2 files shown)


Lukasa:
Thanks for this! 

It's my view that this is overkill: generally speaking, users that require this function can implement it themselves, and the majority of users don't need it. But I'll leave that decision up to @kennethreitz.

sigmavirus24:
@Lukasa further, in v1.0 didn't @kennethreitz completely tear out what logging requests did provide because people were constantly trying to add tons more logging on top of what we provided? I feel like were were to accept this we'd be in the same situation.

Further, what's being logged here is a very clear overload of information for any logging that I would view as reasonable for requests. This seems to be working very well in Café as it is and I see little reason to include this in Requests.

I appreciate your offer of the code @seemethere, but I'm very strongly against this. If you want review on the code anyway, I'm happy to provide that in a separate forum.

Issue #3461, opened 2016-07-30
Closing streams before complete read leaks connections


ml31415:
```python
import requests
s = requests.Session()
adp = requests.adapters.HTTPAdapter(pool_maxsize=3,
    pool_connections=3,
    pool_block=True)
s.mount('http://', adp)

for _ in range(5):
    resp = s.get('http://google.com', stream=True)
#    resp.content
#    resp.raw.release_conn()
    resp.close()
    print "closed"
```
This hangs after the 3rd close, as the connection pool is depleted. `resp.close()` just closes the connection, when the data is not read completely, but no connection is returned to the pool:
https://github.com/kennethreitz/requests/blob/master/requests/models.py#L871

ml31415:
Thanks guys!

PR #3462, opened 2016-07-30
Close and then release the connection

sigmavirus24:
urllib3 closes the underlying connection when we call
urllib3.Response.close but does not release it back to the connection
pool. This can cause issues when users have a blocking connection pool
configured and connections are not readily returned to the pool.

Since the underlying connection is closed, we should be able to safely
return the connection to the connection pool, so to fix this issue we
merely need to not return after closing the response.

Closes gh-3461

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -868,6 +868,6 @@ class Response(object):
         *Note: Should not normally need to be called explicitly.*
         """
         if not self._content_consumed:
-            return self.raw.close()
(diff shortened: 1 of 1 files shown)


Lukasa:
Took a second to understand this patch, but when I saw it I felt pretty silly. This is an obvious win! Thanks! :sparkles: :cake: :sparkles:

Issue #3416, opened 2016-07-15
Unable to override cookie policy in Session.prepare_request


bdusell:
I would like to be able to override the [cookie policy](https://docs.python.org/2.7/library/cookielib.html#cookielib.CookiePolicy) used by a `Session` when making a request or when using a `Session` to prepare a request. However, overriding the cookie policy on the cookie jar of the `Session` or of the `Request` object has had no effect. This line of code explains why:

https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L377

The cookie jars of both the session and the request are merged into a newly constructed `RequestsCookieJar`, discarding the policies of both. I believe there should be a way to inject a non-default cookie policy at this point.

What I expected to be able to do, and what isn't working currently, is something like the following:

``` python
import requests
s = requests.Session()
s.cookies = requests.cookies.RequestsCookieJar(policy=MyCustomCookiePolicy())
s.get('http://httpbin.org/cookies/set?k2=v2&k1=v1') # Put some cookies in the jar
r = requests.Request('GET', 'http://httpbin.org/cookies')
pr = s.prepare_request(r) # Include certain cookies in the jar based on arbitrary rules
# Print the request data for debugging
s.send(pr)
```

nateprewitt:
I think it would make sense to at least preserve the cookie policy in `merge_cookies`. This could be pretty quickly accomplished updating [here](https://github.com/kennethreitz/requests/blob/master/requests/cookies.py#L488) with something like:

`cookiejar.set_policy(cookies.policy)`

nateprewitt:
I think it would make sense to at least preserve the cookie policy in `merge_cookies`. This could be pretty quickly accomplished adding a line [here](https://github.com/kennethreitz/requests/blob/master/requests/cookies.py#L488) with something like:

`cookiejar.set_policy(cookies._policy)`

bdusell:
AFAIK there is no public API for getting the policy of a cookie jar (it's stored under `_policy` in `cookielib.CookieJar`).

https://docs.python.org/2.7/library/cookielib.html#cookiejar-and-filecookiejar-objects

nateprewitt:
Another option could be to simply swap which dictionary is updated.

e.g.

```python
elif isinstance(cookies, cookielib.CookieJar):
        try:
            cookies.update(cookiejar)
        except AttributeError:
            for cookie_in_jar in cookiejar:
                cookies.set_cookie(cookie_in_jar)
        cookiejar = cookies
```

[...]

PR #3463, opened 2016-07-30
#3416 persisting cookie policy.

nateprewitt:
This will add in the option to persist cookie policies on requests when the session and request cookie jars are merged. Currently any custom cookie policies on the session are not sent as noted in #3416.

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -181,7 +181,15 @@ class RequestsCookieJar(cookielib.CookieJar, collections.MutableMapping):
     Unlike a regular CookieJar, this class is pickleable.
 
     .. warning:: dictionary operations that are normally O(1) may be O(n).
+
+    :param policy: (optional) custom ``CookiePolicy`` defining cookie use
+    :param persist_policy: (optional) boolean defining whether to persist
+        the ``_policy`` on CookieJar merges.
     """
+
+    def __init__(self, policy=None, persist_policy=False):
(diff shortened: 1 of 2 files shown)


Lukasa:
So I'm generally ok with this, but I'd like @sigmavirus24 and @kennethreitz to have a look.

nateprewitt:
@sigmavirus24 given your thoughts in #3595 on leaking cookies, I wanted to ping you on this again. Requests current implementation of `prepare_cookies` for Sessions silently deletes the cookie policy provided and gets rid of information like `allowed_domains` and `blocked_domains`.

Since I opened this, I ran across #2714, but this doesn't seem to fall under the work you've outlined there. There does seem to a bit of interaction between the code changes though. I'd appreciate any feedback you have.

nateprewitt:
Hey @sigmavirus24, I know you're busy. I just wanted to move this back up the stack again in case you find yourself with an extra moment to spare. Thanks!

kennethreitz:
Needs a rebase.

nateprewitt:
I'm actually going to propose we close this one unless there's consensus it's needed. There seems to be a lot of unresolved discussion on the state of Session cookies.

kennethreitz:
thanks @nateprewitt

nateprewitt:
Given @Lukasa's comment in #4000 do we want to revisit this?

The caveat of conflicting policies still seems a bit ugly but that would be a pretty rare case that requires manual user intervention to occur.

PR #3475, opened 2016-08-08
Prepare for 2.11 release.

Lukasa:
@kennethreitz, this is all we need from a boring administrative perspective. If you merge this PR, you only need to do three things to release:

1. Update the version number to 2.11.
2. Put the date you do this in the changelog instead of the XX.
3. Push the release.

@sigmavirus24 I'm inclined to want to let all currently open PRs remain open rather than try to rush merge any: they can always wait for a 2.11.1 or 2.12 without any risk.


sigmavirus24:
> I'm inclined to want to let all currently open PRs remain open rather than try to rush merge any: they can always wait for a 2.11.1 or 2.12 without any risk.

My opinion has been (for a while now) that release numbers are cheap. I'm on board with this. 👍 :shipit:

kennethreitz:
Released! ✨🍰✨

PR #3478, opened 2016-08-08
Fix python3.2 unicode literral error.

huseyinyilmaz:
This is a specific fix for python3.2.
When I install requests version v2.11.0, I get an invalid syntax error.
It turns out the reason for that error is because "u" prefix for unicode strings are not allowed in python3.2. So I removed the "u" prefix to fix the issue.

To test it ,I moved the request library into my codebase and get syntax error:

https://travis-ci.org/huseyinyilmaz/placebo/builds/150770392

Than I fixed the issue and ran my tests again:

https://travis-ci.org/huseyinyilmaz/placebo/builds/150770803

I realized that this is not a right way to test this. But I could not install python3.2 on my mac. So I decided to use travis-ci's python version.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -853,10 +853,10 @@ class Response(object):
             reason = self.reason
 
         if 400 <= self.status_code < 500:
-            http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
+            http_error_msg = '%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
 
         elif 500 <= self.status_code < 600:
-            http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
+            http_error_msg = '%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
 
         if http_error_msg:
             raise HTTPError(http_error_msg, response=self)


sigmavirus24:
We have not supported Python 3.2 for ... over a year? I don't think we'll accept this. @kennethreitz, thoughts?

huseyinyilmaz:
@sigmavirus24 . I did not realize that pyhon3.2 is not supported. Thanks for the explanation. I will close the pr.

remram44:
I'm running into the same issue, where this totally worked before, and stopped working with today's release. Looked through the changelog but found nothing, should this have been publicized?

remram44:
```
  File "/home/travis/virtualenv/python3.2.6/lib/python3.2/site-packages/requests/models.py", line 856
    http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
```

StewPoll:
Requests no longer supports Python 3.2. As referenced by @sigmavirus24 this has been the case for ever a year.

If you view the setup.py for requests it specifies 3.3 and higher, or 2.6 and higher.

drpoggi:
The setup.py specifies 3.3 or higher, however the documentation homepage and the PyPi page both say: 

> Requests supports Python 2.6 — 3.5, and runs great on PyPy.

These should be amended for clarity, if you're not looking closely this is a pretty innocent mistake.

drpoggi:
I've submitted a PR for this in #3480

Issue #3481, opened 2016-08-09
Error with decode_unicode=True on streamed response


jone:
With `requests==2.11.0` I get `AttributeError: 'NoneType' object has no attribute 'readline'` when trying to use `iter_lines` with `decode_unicode=True` on a streamed response.

```sh
$ ./bin/pip freeze
requests==2.11.0
wsgiref==0.1.2
```

```python
from contextlib import closing
import requests


with closing(requests.get('http://httpbin.org/stream/20', stream=True)) as response:
    for line in response.iter_lines(chunk_size=30, decode_unicode=True):
        print line
```

```python
$ ./bin/python case.py

Traceback (most recent call last):
  File "case.py", line 6, in <module>
    for line in response.iter_lines(chunk_size=30, decode_unicode=True):
  File "/Users/jone/temp/requests-stream/lib/python2.7/site-packages/requests/models.py", line 720, in iter_lines
    for chunk in self.iter_content(chunk_size=chunk_size, decode_unicode=decode_unicode):
  File "/Users/jone/temp/requests-stream/lib/python2.7/site-packages/requests/utils.py", line 374, in stream_decode_response_unicode
    for chunk in iterator:
  File "/Users/jone/temp/requests-stream/lib/python2.7/site-packages/requests/models.py", line 676, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
  File "/Users/jone/temp/requests-stream/lib/python2.7/site-packages/requests/packages/urllib3/response.py", line 353, in stream
    for line in self.read_chunked(amt, decode_content=decode_content):
  File "/Users/jone/temp/requests-stream/lib/python2.7/site-packages/requests/packages/urllib3/response.py", line 521, in read_chunked
    line = self._fp.fp.readline()
AttributeError: 'NoneType' object has no attribute 'readline'
```

When removing either `stream=True` or `decode_unicode=True` the script works well.

This issue may be related to #3174.

Lukasa:
Well this is weird: it's not immediately apparent to me how `decode_unicode` leads to this behaviour.

Lukasa:
Aha, got it.

`decode_unicode=True` was enhanced in the last release to perform the same auto-detection of encoding as `text` does for Requests. Unfortunately, that may involve a call to `apparent_encoding`, which streams in the entire request body and attempts to determine the encoding.

This was done as part of #3362 in an attempt to resolve #3359.

[...]

Issue #3359, opened 2016-06-22
Uncertain about content/text vs iter_content(decode_unicode=True/False)


mikepelley:
When requesting an application/json document, I'm seeing `next(r.iter_content(16*1024, decode_unicode=True))` returning bytes, whereas `r.text` returns unicode. My understanding was that both should return a unicode object. In essence, I thought "iter_content" was equivalent to "iter_text" when decode_unicode was True. Have I misunderstood something? I can provide an example if needed.

For reference, I'm using python 3.5.1 and requests 2.10.0.

Thanks!

birm:
what does (your response object).encoding return?

Lukasa:
There's at least one key difference: `decode_unicode=True` doesn't fall back to `apparent_encoding`, which means it'll never autodetect the encoding. This means if `response.encoding` is None it is a no-op: in fact, it's a no-op that yields bytes.

That behaviour seems genuinely bad to me, so I think we should consider it a bug. I'd rather we had the same logic as in `text` for this.

mikepelley:
`r.encoding` returns `None`.

On a related note, `iter_text` might be clearer/more consistent than `iter_content(decode_unicode=True)` if there's room for change in the APIs future (and `iter_content_lines` and `iter_text_lines` I guess), assuming you don't see that as bloat.

sigmavirus24:
@mikepelley The API is presently frozen so I don't think we'll be adding those three methods. Besides, `iter_text` likely wouldn't provide much extra value outside of calling `iter_content(decode_unicode=True)`.

nateprewitt:
I added a pull request #3362 that adds the same check used in `text()` to set the encoding type if there is none. It seems to solve the raw bytes issue. Let me know if there's any needed tweaking.

kennethreitz:
Glad we fixed this, this was definitely an oversight.

sentientcucumber:
I was reading through this and the related issues and PRs and I'd love to help out. However, I'm not sure what direction to head with it (or if it's something more suitable for the next version of `requests`).

Just so I understand correctly, we want to get it so `iter_content` and `text` both return Unicode strings. The original PR solved this by grabbing the `apparent_encoding`, which caused a problem when using streaming responses (`chardet.detect` consumes the entire stream).

[...]

PR #3486, opened 2016-08-09
reverting 3362

nateprewitt:
This addresses #3481 and will revert #3362 back to its prior state. We'll likely want to reopen #3359 since this will reintroduce the bytes/text issue.


sigmavirus24:
I'm good with this. How about you @Lukasa? (To be honest, I would have rathered let GitHub generate the revert than have it done artisanally.)

Lukasa:
I'm fine with this too.

nateprewitt:
@sigmavirus24 sorry, I tried the revert button and got an error so did it by hand.

PR #3489, opened 2016-08-09
adding in pickling tests for PreparedRequest

nateprewitt:
This is a continuation of @keyan's work in #2757 and addresses issue #1558 regarding pickling Request objects (specifically `PreparedRequest` objects in this PR).

These tests ensure that a standard `PreparedRequest`, a `PreparedRequest` with a file object as the body, and a `PreparedRequest` with hooks defined outside of the `locals` scope will all pickle properly.

This also works for data passed to the `json` parameter and I can add my test for that as well if it's deemed helpful.

Hooks defined inside of a method (such as in the [original test](https://github.com/kennethreitz/requests/pull/2757/files#diff-56c2d754173a4a158ce8f445834c8fe8R843) by @keyan) will fail because you can't pickle local objects. Is this a use case that needs to be handled?

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1053,63 +1053,51 @@ class TestRequests:
         assert r.request.url == pr.request.url
         assert r.request.headers == pr.request.headers
 
-    def test_prepared_request_is_pickleable(self):
+    def test_prepared_request_is_pickleable(self, httpbin):
         p = requests.Request('GET', httpbin('get')).prepare()
 
-        # Verify we can pickle the request.
-        assert pickle.dumps(p)
-        r = pickle.dumps(p)
+        # Verify PreparedRequest can be pickled and unpickled
+        r = pickle.loads(pickle.dumps(p))
+        assert r.url == p.url
+        assert r.headers == p.headers
+        assert r.body == p.body
 
-        # Verify we can use the pickled request.
-        assert pickle.loads(r)
-        r = pickle.loads(r)
-
-        # Verify we can use the unpickled request.
+        # Verify unpickled PreparedRequest sends properly
         s = requests.Session()
-        r = s.send(r)
+        resp = s.send(r)
+        assert resp.status_code == 200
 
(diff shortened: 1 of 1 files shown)


nateprewitt:
@Lukasa any thoughts here? We still have edge cases where if the `data` param is a generator or file-like object, it can't be pickled. Is there any utility in adding this subset of tests, or should #1558 be closed as can't fix?

tests/test_requests.py
```diff
@@ -1053,6 +1053,52 @@ def test_request_and_response_are_pickleable(self, httpbin):
...
+        resp = s.send(r)
+        assert resp.status_code == 200
+
+    def test_prepared_request_with_file_is_pickleable(self, httpbin):
+        files = {'file': open(__file__, 'rb')}
+        r = requests.Request('POST', httpbin('post'), files=files)
+        p = r.prepare()
+
+        # Verify PreparedRequest can be pickled and unpickled
+        r = pickle.loads(pickle.dumps(p))
```
Lukasa, line 1076: This doesn't actually test that a file object in the body is pickleable: `r.prepare()` will read the whole file object into memory as a string. I don't know that we actually do support having a file object as the body when pickling/unpickling, though it'd be interesting to see.

Lukasa:
I have no particular problems with merging this PR, though I'd like to confirm we're on the same page regarding file objects.

tests/test_requests.py
nateprewitt: Yeah, we 100% do not support that because it's not a capability of any version of pickle as far as I can tell. The only reason `files` works while `data` doesn't is because we consume the file object during `prepare_body`. It may be worth changing the name or adding a docstring here for clarification on why this works.

nateprewitt:
Yeah, I think we're on the same page. I'm also not terribly tied to the PR, I started it trying to wrap up #1558 and Keyan's work, since Kenneth felt there was some value in having these tests.

Issue #3490, opened 2016-08-09
Remove header in middle of redirection


saveman71:
Hey there,

I'm having a pretty specific issue while trying to reproduce the login sequence of a heavy professional website (lots of redirect / cookies involved).
The login sequence was tested working on Chrome.

I spent some time debugging and finally tracked down the issue, that is that the server will not accept the `Content-Type` header that was required on the first request on the first redirect, thus making the second request crash with a `500`.

First request shown in Chrome's dev inspector:
![screenshot from 2016-08-09 20-03-26](https://cloud.githubusercontent.com/assets/1615426/17528006/dec4dfe6-5e6d-11e6-8e67-9af08aa62426.png)

Second request (first redirect) show in Chrome's dev inspector:
![screenshot from 2016-08-09 20-15-02](https://cloud.githubusercontent.com/assets/1615426/17528125/5dee5aea-5e6e-11e6-8030-be6a37f4df0a.png)

Notice of how on the second request, no `Content-Type` header. The presence of a `Content-Type` header would make the request fail.

The code (URLs are mangled):

```python
#!/usr/bin/env python

import requests
import urllib.parse

headers = {
    'Origin': 'https://www.example.com',
    'Accept-Encoding': 'gzip, deflate, br',
    'Accept-Language': 'en-US,en;q=0.8,fr-FR;q=0.6,fr;q=0.4',
    'Upgrade-Insecure-Requests': '1',
    'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.82 Safari/537.36',
    'Content-Type': 'application/x-www-form-urlencoded',
    'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8',
    'Cache-Control': 'max-age=0',
    'Referer': 'https://www.example.com/',
    'Connection': 'keep-alive',
}

data = 'some_data'

s = requests.Session()

r = s.post('https://www.exampleonanotherdomain.com/openam/UI/Login?realm=front_office&service=EEService&goto=https://www.exampleonanotherdomain.com/ice/rest/aiguillagemp/redirect?dest=',
           headers=headers, data=data)
```

I worked it around by removing the `Content-Type` header at runtime on the 3rd send (first is the first request header, 2nd is the first request data) to confirm that it was the issue (it was!) 

Patch code added before the request.
```python
import http.client as http_client

def patch():
[...]

PR #3493, opened 2016-08-10
#3490 removing Content-Type and Transfer-Encoding headers on redirect

nateprewitt:
This will remove additional headers (`Content-Type` and `Transfer-Encoding`) from non-temporary/non-permanent redirects in `resolve_redirects` to address #3490.

This functionality isn't currently being tested but I'm happy to add a test or two for it if desired.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -140,9 +140,10 @@ class SessionRedirectMixin(object):
 
             # https://github.com/kennethreitz/requests/issues/1084
             if resp.status_code not in (codes.temporary_redirect, codes.permanent_redirect):
-                if 'Content-Length' in prepared_request.headers:
-                    del prepared_request.headers['Content-Length']
-
+                # https://github.com/kennethreitz/requests/issues/3490
+                purged_headers = ('Content-Length', 'Content-Type', 'Transfer-Encoding')
+                for header in purged_headers:
+                    prepared_request.headers.pop(header, None)
                 prepared_request.body = None
 
             headers = prepared_request.headers
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -221,6 +221,41 @@ class TestRequests:
         assert r.history[0].status_code == 303
         assert r.history[0].is_redirect
 
+    def test_header_and_body_removal_on_redirect(self, httpbin):
+        purged_headers = ('Content-Length', 'Content-Type')
+        ses = requests.Session()
+        req = requests.Request('POST', httpbin('post'), data={'test': 'data'})
+        prep = ses.prepare_request(req)
+        resp = ses.send(prep)
+
+        # Mimic a redirect response
+        resp.status_code = 302
+        resp.headers['location'] = 'get'
+
+        # Run request through resolve_redirects
+        next_resp = next(ses.resolve_redirects(resp, prep))
+        assert next_resp.request.body is None
+        for header in purged_headers:
+            assert header not in next_resp.request.headers
+
+    def test_transfer_enc_removal_on_redirect(self, httpbin):
+        purged_headers = ('Transfer-Encoding', 'Content-Type')
+        ses = requests.Session()
+        req = requests.Request('POST', httpbin('post'), data=(b'x' for x in range(1)))
+        prep = ses.prepare_request(req)
+        assert 'Transfer-Encoding' in prep.headers
+        resp = ses.send(prep)
+
+        # Mimic a redirect response
+        resp.status_code = 302
+        resp.headers['location'] = 'get'
+
+        # Run request through resolve_redirect
+        next_resp = next(ses.resolve_redirects(resp, prep))
+        assert next_resp.request.body is None
+        for header in purged_headers:
+            assert header not in next_resp.request.headers
+
     def test_HTTP_200_OK_GET_WITH_PARAMS(self, httpbin):
         heads = {'User-agent': 'Mozilla/5.0'}
 
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -244,11 +244,16 @@ class TestRequests:
         req = requests.Request('POST', httpbin('post'), data=(b'x' for x in range(1)))
         prep = ses.prepare_request(req)
         assert 'Transfer-Encoding' in prep.headers
-        resp = ses.send(prep)
+
+        # Create Response to avoid https://github.com/kevin1024/pytest-httpbin/issues/33
+        resp = requests.Response()
+        resp.raw = io.BytesIO(b'the content')
+        resp.request = prep
+        setattr(resp.raw, 'release_conn', lambda *args: args)
 
         # Mimic a redirect response
         resp.status_code = 302
-        resp.headers['location'] = 'get'
+        resp.headers['location'] = httpbin('get')
 
         # Run request through resolve_redirect
         next_resp = next(ses.resolve_redirects(resp, prep))


Lukasa:
I would like to see tests for this please! :sparkles:

saveman71:
I was on it too. Did not find a way to properly add tests though since httpbin does not seem to be able to post to an endpoint and respond with a 301/302. Am I wrong on this one?

Anyway since @nateprewitt is on it I will not open my PR ;)

saveman71:
On my version of the fix (https://github.com/kennethreitz/requests/compare/master...saveman71:master) I moved [this line](https://github.com/kennethreitz/requests/pull/3493/files#diff-28e67177469c0d36b068d68d9f6043bfR151) a little higher and removed the try/catch below as it seems the general way (at least in this file) is to check the existence of the key prior to the deletion.

nateprewitt:
@saveman71, sorry, I didn't mean to jump on your PR. I wasn't sure if anyone had started working on it.

@Lukasa I added a test to check the header removal and empty body.

nateprewitt:
@Lukasa @saveman71 Re: [this line](https://github.com/kennethreitz/requests/pull/3493/files#diff-28e67177469c0d36b068d68d9f6043bfR151). I would vote we completely remove the `headers = prepare_requests.headers` line since it isn't used anywhere else except the Cookie check. The we can change line 153 to:

```python
  if 'Cookie' in prepared_requests.headers:
      del prepared_requests.headers['Cookie']
```

Perhaps @sigmavirus24  will have an opinion on calling the attribute twice though.

nateprewitt:
@Lukasa re: [this line](https://github.com/kennethreitz/requests/pull/3493/files#diff-28e67177469c0d36b068d68d9f6043bfR151)

Do you have an opinion on moving the declaration farther up to avoid the multiple calls to the headers attribute on the prepare_request? Also as for the conversion from try/except, if that was done would you want it in a separate PR to avoid semi-unrelated changes?

saveman71:
@nateprewitt no worries at all, I should have stated I started working on it.
As for the line thing, I do not think it is a performance issue (I believe, but I might be mistaken) because well requests is a network library and "premature optimization is the root of all evil". (https://www.xkcd.com/1691/)

I was suggesting that merely because it's already been like that [here](https://github.com/nateprewitt/requests/blob/20a2b5ac71f1d6ce674ca2633172ef5964ec9c38/requests/sessions.py#L192) and [here](https://github.com/nateprewitt/requests/blob/20a2b5ac71f1d6ce674ca2633172ef5964ec9c38/requests/sessions.py#L223) for example.

On a side note, :+1: for a separate PR.

nateprewitt:
@saveman71, I only bring up the optimization because @sigmavirus24 has [brought it up before](https://github.com/kennethreitz/requests/pull/3362#discussion_r68398047). If I'm misunderstanding his original intent though, I vote we completely remove the declaration, in favor of `prepared_request.headers`. `headers` currently doesn't provide a lot of use in `resolve_redirects`.

Edit: Clarity and Link

saveman71:
Quick semi-unrelated question, what would be the timeline of a release on PyPI after this gets merged?

cc @Lukasa

Lukasa:
We'd be aiming to release this early next week if we can merge it this week.

Lukasa:
To be clear, this is on my queue, but I'll review tomorrow.

requests/sessions.py
```diff
@@ -140,8 +140,11 @@ def resolve_redirects(self, resp, req, stream=False, timeout=None,
 
             # https://github.com/kennethreitz/requests/issues/1084
             if resp.status_code not in (codes.temporary_redirect, codes.permanent_redirect):
-                if 'Content-Length' in prepared_request.headers:
-                    del prepared_request.headers['Content-Length']
+                # https://github.com/kennethreitz/requests/issues/3490
+                purged_headers = ('Content-Length', 'Content-Type', 'Transfer-Encoding')
+                for header in purged_headers:
+                    if header in prepared_request.headers:
```
sigmavirus24: I know I wrote this poorly the first time, but why don't we do:

``` py
prepared_request.headers.pop('Content-Length', None)
prepared_request.headers.pop('Content-Type', None)
```

Or if you really want to shave the yak

``` py
for part in ('Length', 'Type'):
    prepared_request.headers.pop('Content-' + part, None)
```

sigmavirus24: Just to be clear, I think the second suggestion is _very_ bad, but I realize some people might actually prefer that.

nateprewitt: I originally used `pop` over `del` when writing it, but reverted to keep with the surrounding code. If you're onboard with using `pop` instead, let's do it.

tests/test_requests.py
```diff
@@ -221,6 +221,23 @@ def test_http_303_doesnt_change_head_to_get(self, httpbin):
         assert r.history[0].status_code == 303
         assert r.history[0].is_redirect
 
+    def test_header_and_body_removal_on_redirect(self, httpbin):
+        purged_headers = ('Content-Length', 'Content-Type', 'Transfer-Encoding')
```
nateprewitt: This test doesn't actually cover 'Transfer-Encoding' since 'Content-Length' is present. I can add a smaller separate test for 'Transfer-Encoding' if it's deemed that we definitely want to remove it.

This does ensure that as long as `purged_headers` contains 'Transfer-Encoding' in `resolve_redirects`, that it will be removed if it occurs. It just won't fail for accidental removal from the `purged_headers` tuple.

Lukasa: Transfer-Encoding should be easily covered: a generator body will trigger it.

nateprewitt: Transfer-Encoding test code added.

tests/test_requests.py
```diff
@@ -221,6 +221,35 @@ def test_http_303_doesnt_change_head_to_get(self, httpbin):
...
+        resp_tran.headers['location'] = 'get'
+
+        # Run request through resolve_redirects
+        next_resp = next(ses.resolve_redirects(resp, prep))
+        next_resp_tran = next(ses.resolve_redirects(resp, prep))
+        assert next_resp.request.body is None
+        assert next_resp_tran.request.body is None
+        for header in purged_headers:
+            assert header not in next_resp.request.headers
+            assert header not in next_resp_tran.request.headers
```
Lukasa: Can we split this into two tests that look very similar, rather than have one really complex test?

nateprewitt: Sure thing. Tests split up now.

Lukasa:
Alright, fab, I'm happy with this. @sigmavirus24 you ok with us merging this for 2.11.1?

tests/test_requests.py
```diff
@@ -221,6 +221,41 @@ def test_http_303_doesnt_change_head_to_get(self, httpbin):
...
+        for header in purged_headers:
+            assert header not in next_resp.request.headers
+
+    def test_transfer_enc_removal_on_redirect(self, httpbin):
+        purged_headers = ('Transfer-Encoding', 'Content-Type')
+        ses = requests.Session()
+        req = requests.Request('POST', httpbin('post'), data=(b'x' for x in range(1)))
+        prep = ses.prepare_request(req)
+        assert 'Transfer-Encoding' in prep.headers
+        resp = ses.send(prep)
```
nateprewitt: I reduced the `range` call to size 1 because we were hitting [this bug](https://github.com/kevin1024/pytest-httpbin/issues/33) when sending more than one chunk. After some more testing, this only passes _most of the time_, which likely isn't something we want to introduce. 

The failures seem random and are tied specifically to the httpbin module. Calling the live http://httpbin.org/post endpoint won't produce the failure.

As I see it, two options are either calling the httpbin.org/post endpoint directly (probably not the best solution) or adding a try/except around `resp = ses.send(prep)` and manually creating the Response on failure.

Thoughts?

Lukasa: There's a question about whether we should just manually create the response unconditionally, to be honest. That might be the most sensible thing to do.

nateprewitt: Alright, I'll make a note and create the `Response` manually for this.

nateprewitt: Ok, test updated with manual `Response`. If it looks good I'll squash things down.

Lukasa:
Alright, all is well. @sigmavirus24, you want this in 2.11.1?

sigmavirus24:
👍

PR #3506, opened 2016-08-11
Prevent environment variables from overridding Session settings

pfista:
In some cases, `Session` settings are overridden by environment settings. Session should inherit from its own settings before trying to pull in environment settings. 

For example, say environment variable `REQUESTS_CA_BUNDLE` is set to a certificate bundle, a session object is created and `session.verify = False`, and then that session is mounted to an adapter. If subsequent requests made do not explicitly pass in `verify=False`, certificate verification will be enabled despite having set `session.verify = False`. This is due to `merge_setting` defaulting to use the `request_setting` when merging verify, which is currently incorrectly set to the environment variables certificate file before checking `session.verify`.

This will ensure that `Request` object settings take precedence, then `Session` settings, then global environment settings.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -167,3 +167,4 @@ Patches and Suggestions
 - piotrjurkiewicz
 - Jesse Shapiro <jesse@jesseshapiro.net> (`@haikuginger <https://github.com/haikuginger>`_)
 - Nate Prewitt <nate.prewitt@gmail.com> (`@nateprewitt <https://github.com/nateprewitt>`_)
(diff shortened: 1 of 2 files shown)


Lukasa:
Thanks! This is a known issue (see #2018), but fixing it is a breaking change that needs to go into 3.0.0. Can you please open a new pull request targeting the proposed/3.0.0 branch?

pfista:
It looks like this has already been fixed in proposed/3.0.0!

nateprewitt:
It may be worth noting that it's PR #2839 that fixes this.

PR #3510, opened 2016-08-12
adding passthrough in close() for non-urllib3-like Responses

nateprewitt:
This is minor, but when testing `resolve_redirects` I encountered an `AttributeError` when passing a non-urllib3 Response. This is because the method unconditionally calls [`close`](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L864) on the Response which in turn calls `release_conn` on the `Response.raw` attribute.

If the desire for requests is to maintain support for non-urllib3 objects in custom adapters, I think this would be a simple and useful addition. If urllib3 is going to be a de facto pattern for Responses though, this is probably just unnecessary code bloat.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -870,4 +870,6 @@ class Response(object):
         if not self._content_consumed:
             self.raw.close()
 
-        return self.raw.release_conn()
+        release_conn = getattr(self.raw, 'release_conn', None)
+        if release_conn is not None:
+            release_conn()
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1353,6 +1353,15 @@ class TestRequests:
         with pytest.raises(ValueError):
             r.json()
 
+    def test_response_without_release_conn(self):
+        """Test `close` call for non-urllib3-like raw objects.
+        Should work when `release_conn` attr doesn't exist on `response.raw`.
+        """
+        resp = requests.Response()
+        resp.raw = StringIO.StringIO('test')
+        assert not resp.raw.closed
+        resp.close()
+        assert resp.raw.closed
 
 class TestCaseInsensitiveDict:
 


tests/test_requests.py
```diff
@@ -1353,6 +1353,13 @@ def test_response_json_when_content_is_None(self, httpbin):
         with pytest.raises(ValueError):
             r.json()
 
+    def test_response_without_release_conn(self):
+        """Test `close` call for non-urllib3-like raw objects.
+        Should work when `release_conn` attr doesn't exist on `response.raw`.
+        """
+        resp = requests.Response()
+        resp.raw = StringIO.StringIO('test')
+        resp.close()
```
Lukasa, line 1363: Can we make some kind of positive assertion here? I don't want us to get to a place where we fix this by having `close` be a no-op without this test failing. =)

requests/models.py
```diff
@@ -869,5 +869,5 @@ def close(self):
         """
         if not self._content_consumed:
             self.raw.close()
-
-        return self.raw.release_conn()
+        release_conn = getattr(self.raw, 'release_conn', lambda *args: None)
```
Lukasa: I'd rather do an `if` here than create this temporary lambda every single time. =)

Lukasa:
Ok, I have put some small notes in. I don't want to merge this until after 2.11.1 gets out the door though.

nateprewitt:
Test updated and lambda logic changed to a conditional. No rush on this, just wanted to get it on the record before I forgot.

requests/models.py
```diff
@@ -870,4 +870,6 @@ def close(self):
         if not self._content_consumed:
             self.raw.close()
 
-        return self.raw.release_conn()
+        release_conn = getattr(self.raw, 'release_conn', None)
+        if release_conn is not None:
+            return release_conn()
```
sigmavirus24: Do we really need to return here?

nateprewitt: No we don't, I was just trying to change as little of the existing code as needed. I'm fine with removing it.

sigmavirus24:
Just reminding ourselves not to merge this until after 2.11.1 is released

nateprewitt:
Now that 2.11.1 is out, I don't think I have anything else on my checklist. Let me know if this needs any other final tweaks.

Lukasa:
Thanks @nateprewitt!

Issue #2228, opened 2014-09-16
Documentation examples.


mountainpaul:
Thanks for an excellent package. I believe the quick start examples should include a timeout value. This is the more common case, rather than the exception. 

In the more advanced documentation it maybe wise to mention that sockets.getdefaulttimeout() is not used when a value is not specified.

Lukasa:
Thanks for the suggestion! However, I disagree.

In many cases, users find it valuable to add a timeout to their calls. This is entirely reasonable. However, the quickstart code is not an example of 'general use', it's intended to be illustrative and educational. We should not obscure the pedagogical point in the pursuit of some form of pure correctness.

There is a [timeout section](http://docs.python-requests.org/en/latest/user/quickstart/#timeouts) in the quickstart docs, which we believe to be sufficient. Do you disagree?

As for mentioning the default socket timeout, I hadn't spotted that. That may actually be a bug. @sigmavirus24, what do you think: should we use the default socket timeout when no value is provided? If so, how do users specify they want no timeout at all? Am I overthinking the whole thing?

sigmavirus24:
@Lukasa I'm surprised that we don't use the default socket timeout. I'm more surprised that the "default" timeout is not automatically applied to a socket. That said, I don't think there's a good (or easy way) to handle adding this at this juncture. We could consider it for 3.0 but if we truly have no timeout set on sockets by default (which I'm prepared to believe), then we can't change it easily and it _should_ be documented for now.

The thing we should determine is whether we want that behaviour at all. The sad thing is that more often than not, not being able to put a timeout on `socket.getaddrinfo` has affected me more than a connection not timing out appropriately.

mountainpaul:
I disagree that it is sufficient as timeouts  are only mentioned at the
very bottom of the page.  Users often base their solutions off of examples
and don't read all the documentation, only what they need to get a result.
 While examples should be brief, they should not lead a user to writing
code that has a pretty strong potential to lock up.

The default socket timeout is used by urllib2 when a timeout is not
specified. This further compounds the above problem for people moving from
that library.  To have no timeout specified I can see either None or 0
being passed as the timeout value.  Explicit is better than implicit.

Lukasa:
> @Lukasa I'm surprised that we don't use the default socket timeout. I'm more surprised that the "default" timeout is not automatically applied to a socket.

[...]

PR #3513, opened 2016-08-14
doc updates

nateprewitt:
Various doc updates for clarity. Some of these may not match @kennethreitz's aesthetic sensibilities, which may require an inversion of the proposed change. I was mainly trying to make things consistent, input on preferences would be greatly appreciated.

Some notes:
* I think at least the faq.rst changes *need* to be merged because there is still live documentation stating that Requests supports 3.1 and 3.2. I can open a separate PR if we want those changes quicker.

* What was ScraperWiki, is now found at quickcode.io and appears to either be a paid or account-walled service. Wayback Machine is showing the site as a free service when the documentation was created, so I'm not sure if Kenneth still wants this included. At the very least, the URL should be updated.

* ~~Does Requests support pypy 2.2 officially? faq.rst says yes, but it's other versions were wrong.~~

* While I realize it's common to use `class.method` as a pattern when describing code, I felt like `r.json` is ambiguous when referenced next to `r.text`. Calling exactly `r.json` won't function as described in the provided `ValueError` example. Explicitly defining the call as `r.json()` reduces the chance of misinterpretation.

* api.rst has a pared down set of Exceptions and classes. I assume this is to keep extraneous information from bloating the page to a point of being unhelpful. I do think it may be worth including `BaseAdapter` for easy reference since anyone looking into a custom *non-HTTP* Transport Adapter will likely need to start there.

### DIFF
--- a/docs/dev/todo.rst
+++ b/docs/dev/todo.rst
@@ -41,7 +41,7 @@ Requests currently supports the following versions of Python:
 - Python 3.3
 - Python 3.4
 - Python 3.5
-- PyPy 1.9
+- PyPy
 
 Google AppEngine is not officially supported although support is available
 with the `Requests-Toolbelt`_.
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -87,11 +87,11 @@ See the :ref:`Session API Docs <sessionapi>` to learn more.
 Request and Response Objects
 ----------------------------
 
-Whenever a call is made to ``requests.get()`` and friends you are doing two
+Whenever a call is made to ``requests.get()`` and friends, you are doing two
 major things. First, you are constructing a ``Request`` object which will be
 sent off to a server to request or query some resource. Second, a ``Response``
-object is generated once ``requests`` gets a response back from the server.
-The Response object contains all of the information returned by the server and
+object is generated once Requests gets a response back from the server.
+The ``Response`` object contains all of the information returned by the server and
 also contains the ``Request`` object you created originally. Here is a simple
 request to get some very important information from Wikipedia's servers::
 
@@ -213,9 +213,7 @@ You can pass ``verify`` the path to a CA_BUNDLE file or directory with certifica
 
 This list of trusted CAs can also be specified through the ``REQUESTS_CA_BUNDLE`` environment variable.
 
-Requests can also ignore verifying the SSL certificate if you set ``verify`` to False.
-
-::
+Requests can also ignore verifying the SSL certificate if you set ``verify`` to False::
 
     >>> requests.get('https://kennethreitz.com', verify=False)
     <Response [200]>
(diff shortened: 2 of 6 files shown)


docs/dev/todo.rst
```diff
@@ -42,6 +42,7 @@ Requests currently supports the following versions of Python:
 - Python 3.4
 - Python 3.5
 - PyPy 1.9
+- PyPy 2.2
```
Lukasa: I think there is no value in being specific about PyPy versions: we just support PyPy.

docs/user/advanced.rst
```diff
@@ -940,7 +937,7 @@ If the remote server is very slow, you can tell Requests to wait forever for
 a response, by passing None as a timeout value and then retrieving a cup of
 coffee.
 
-.. code-block:: python
+::
```
Lukasa, line 941: Why was this removed?

nateprewitt: :: produces the same output and is used everywhere else in the documentation. I couldn't find any additional HTML markup generated by `.. code-block:: python` v.s. `::` and the [documentation](http://www.sphinx-doc.org/en/stable/markup/code.html) states it's only necessary when displaying a lanuage not in the default config (which is python for Requests). 

It looks like this all came from the same commit from Kevin though, so he may have a reason for this that I'm missing.

nateprewitt, line 941: :: produces the same output and is used everywhere else in the documentation. I couldn't find any additional HTML markup generated by `.. code-block:: python` v.s. `::` and the [documentation](http://www.sphinx-doc.org/en/stable/markup/code.html) states it's only necessary when displaying a language not in the default config (which is python for Requests). 

It looks like this all came from the same commit from Kevin though, so he may have a reason for this that I'm missing.

docs/dev/todo.rst
nateprewitt: Great, I'll update that.

docs/user/advanced.rst
nateprewitt, line 941: One more note on this. It looks like this does affect the GitHub highlighting but not Sphinx (Pygments).

The two previous code snippets above this one do not currently display with any syntax highlighting, but this one does. I don't have a strong opinion on which direction to go with this, but I do think it looks a little odd if highlighting isn't uniform across grouped examples.

sigmavirus24:
@nateprewitt is this related to #2228?

nateprewitt:
@sigmavirus24 It wasn't originally, I found a few minor hiccups making the `.json` change and then did a pass through the docs from there. I realize this is kind of a mess of unrelated doc changes so I can break it up into different commits or PRs if needed.

It looks like most of the discussion in #2228 is about timeouts documentation. Was there something low hanging you wanted to see integrated?

sigmavirus24:
> Was there something low hanging you wanted to see integrated?

No because the last person to comment on that thread wants to contribute and I wanted to make sure we gave them the first chance. If this is unrelated that's fine.

nateprewitt:
Sorry, I missed that. No, none of these changes touch the timeouts docs, so I think all is well.

kennethreitz:
These are all fanatastic changes, and ones I would have made myself, given an infinite amount of time :)

Thank you very much!

Issue #3518, opened 2016-08-17
ALL_PROXY is incorrectly preferred to HTTP(S)_PROXY


Lukasa:
Related to #3516.

We should aim to use HTTP(S)_PROXY in priority to ALL_PROXY. Generally speaking, more specific proxy settings should be preferred to less-specific ones.

Lukasa:
BTW, to any experienced contributor reading this: please can you leave this on the pile? This is a fairly easy change, so it'd be great to let someone who hasn't had the opportunity to get started in OSS take a swing at this if we possibly can.

Lukasa:
Resolved.

PR #3526, opened 2016-08-21
Fixed bug to give scheme proxy priority over "all"

shellhead:
Addresses bug called out in issue #3518.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -168,3 +168,4 @@ Patches and Suggestions
 - Jesse Shapiro <jesse@jesseshapiro.net> (`@haikuginger <https://github.com/haikuginger>`_)
 - Nate Prewitt <nate.prewitt@gmail.com> (`@nateprewitt <https://github.com/nateprewitt>`_)
 - Maik Himstedt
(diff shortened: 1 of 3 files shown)


sigmavirus24:
This seems reasonable to me. @Lukasa thoughts?

Lukasa:
This looks really good to me, I'm totally happy with it! Thanks so much @shellhead! :sparkles: :cake: :sparkles:

drpoggi:
Shouldn't this line also be swapped to close out this issue? https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L229

Lukasa:
Yup, looks like!

sentientcucumber:
Thanks for pointing this out.

I'm also thinking I missed this as well https://github.com/kennethreitz/requests/blob/master/requests/utils.py#L620

Is there any reason to not use `select_proxy` in sessions? It seems strange to have two different ways of selecting the proxy.

Lukasa:
`select_proxy` does something subtly different. In this case in the session we are trying to recompute the proxies from the environment, but otherwise not to do anything.

PR #3527, opened 2016-08-21
3518 Bug Fix and Response Header to Python equivalents

michelsch92:
I'm fairly new to open source contributions, so any feedback is appreciated. This PR
1) Addresses issue #3518 by iterating over ALL_PROXY last. 
2) Elaborates some of the most commonly used response headers into their corresponding Python equivalents for convenience, namely r.date() and r.last_modified() return datetime objects and r.content_length() and r.age() return ints when possible, None otherwise
3) Added corresponding tests for the above


Lukasa:
Thanks for this @michelsch92!

Some notes. Firstly, I'm afraid to say that #3518 was already resolved by #3527. I'm sorry about that, the timing was just a bit unlucky.

Secondly, the response helpers, I think we don't want to put them into the core requests library itself. Generally speaking we try to keep the surface area of the API down, because the more methods and properties that there are on the object the harder it gets to understand. However, @sigmavirus24 may be interested in having those properties in a headers helper object in the requests-toolbelt.

Unfortunately, that means this patch won't be accepted at this time. This has nothing to do with the quality of your work: altogether it was high quality work, and we'd have been pleased to accept it. Please do keep looking around for ways to contribute, because you're most welcome here!

Issue #3079, opened 2016-04-07
requests gets stuck on post redirect where data is readable


tzickel:
Checked on python 2.7. Unfortunately httpbin does not support 307 post redirects and thus it is not easy to test this issue in requests test suite currently.

Anyhow the issue seems to be with the stream being already over when the redirects tries to read it again. The response should be an exception if a stream is inputed and it's redirected (although this is known only in httplib).

example code:

``` python
from flask import Flask, request, redirect, url_for
import requests
import io

app = Flask('test')


@app.route('/redirect/', methods=['POST'])
def redirect():
  response = app.make_response('')
  response.status_code = 307
  response.headers['Location'] = 'http://localhost:5000/'
  return response


@app.route('/', methods=['POST'])
def index():
  return request.get_data()


if __name__ == "__main__":
  import threading
  def loop():
    app.run()
  t = threading.Thread(target=loop)
  t.daemon = True
  t.start()
  if requests.post('http://localhost:5000/redirect/', data='hey').text != 'hey':
    raise Exception()
  if requests.post('http://localhost:5000/redirect/', data=io.BytesIO('hey')).text != 'hey': # <-- This code gets stuck
    raise Exception()
```

Lukasa:
So this is a very real problem: when passing requests any file-like object it does not rewind the file pointer on redirect. That means that if the original POST consumed the data, the follow-on POST will not.

With normal files, that causes a small behavioural error, because we'll usually conclude that the file is zero-length and POST nothing. That's wrong, but program execution continues. With BytesIO we have a separate problem, which is that we use `len(BytesIO.getvalue())` to work out how much data we're going to send. `getvalue()` doesn't change with `read()`, which means that we tell the remote server that we're going to send 3 bytes of data but provide an empty body. This obviously causes the server to wait for the body that is never going to come.

Fixing this requires a degree of care: the PreparedRequest needs to make a note, for any object with a `tell()` method and a `read()` method, of what the current location of the file pointer is. It then needs to check, in `resolve_redirects`, whether it's keeping the request body, and if it is and has a saved location of the file pointer then it needs to restore that pointer.

This should be entirely do-able though. Are you interested in submitting a patch to fix this?

tzickel:
I don't like that BytesIO.getvalue() is called at super_len, because if i'm passing large data (let's say 100MB of data), the getvalue will allocate a new copy of it in memory, than super_len will get the length, than deallocate it...

[...]

PR #3535, opened 2016-08-24
avoid use of getvalues in super_len

nateprewitt:
This is a follow up on @jseabold's work in #3339. These last minor changes should fix the issues with return values of `seek` between Python 2 and Python 3.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -45,7 +45,7 @@ def dict_to_sequence(d):
 
 
 def super_len(o):
-    total_length = 0
+    total_length = None
     current_position = 0
 
     if hasattr(o, '__len__'):
@@ -53,10 +53,6 @@ def super_len(o):
 
     elif hasattr(o, 'len'):
         total_length = o.len
-
-    elif hasattr(o, 'getvalue'):
-        # e.g. BytesIO, cStringIO.StringIO
-        total_length = len(o.getvalue())
 
     elif hasattr(o, 'fileno'):
         try:
@@ -88,6 +84,13 @@ def super_len(o):
             # instance, we don't know what the length is, so set it to zero and
             # let requests chunk it instead.
             current_position = total_length
+
+    if hasattr(o, 'seek') and total_length is None:
+        # StringIO has a notimplemented fileno
+        # BytesIO has seek and not fileno
+        total_length = o.seek(0, 2)
+        # seek back to current position to support partially read file-like
+        o.seek(current_position)
 
     return max(0, total_length - current_position)
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -89,8 +89,14 @@ def super_len(o):
         # StringIO has a notimplemented fileno
         # BytesIO has seek and not fileno
         total_length = o.seek(0, 2)
+        # StringIO objects in py2 returns None
+        if total_length is None and hasattr(o, 'tell'):
+            total_length = o.tell()
+        elif total_length is None:
+            total_length = 0
+
         # seek back to current position to support partially read file-like
-        o.seek(current_position)
+        o.seek(current_position or 0)
 
     return max(0, total_length - current_position)
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -86,18 +86,16 @@ def super_len(o):
             if total_length is not None:
                 current_position = total_length
 
-    if hasattr(o, 'seek') and total_length is None:
-        # StringIO has a notimplemented fileno
-        # BytesIO has seek and not fileno
-        total_length = o.seek(0, 2)
-        # StringIO objects in py2 returns None
-        if total_length is None and hasattr(o, 'tell'):
-            total_length = o.tell()
-        elif total_length is None:
-            total_length = 0
+        if hasattr(o, 'seek') and total_length is None:
+            # StringIO and BytesIO have seek but no useable fileno
 
-        # seek back to current position to support partially read file-like
-        o.seek(current_position or 0)
+            # seek to end of file
+            o.seek(0, 2)
+            total_length = o.tell()
+
+            # seek back to current position to support
+            # partially read file-like objects
+            o.seek(current_position or 0)
 
     if total_length is None:
         total_length = 0
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -83,7 +83,8 @@ def super_len(o):
             # is actually a special file descriptor like stdin. In this
             # instance, we don't know what the length is, so set it to zero and
             # let requests chunk it instead.
-            current_position = total_length
+            if total_length is not None:
+                current_position = total_length
 
     if hasattr(o, 'seek') and total_length is None:
         # StringIO has a notimplemented fileno
@@ -97,6 +98,9 @@ def super_len(o):
 
         # seek back to current position to support partially read file-like
         o.seek(current_position or 0)
+
+    if total_length is None:
+        total_length = 0
 
     return max(0, total_length - current_position)
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -66,6 +66,34 @@ class TestSuperLen:
             assert super_len(fd) == 4
         assert len(recwarn) == warnings_num
 
+    def test_super_len_with__len__(self):
+        foo = [1,2,3,4]
+        len_foo = super_len(foo)
+        assert len_foo == 4
+
+    def test_super_len_with_no__len__(self):
+        class LenFile(object):
+            def __init__(self):
+                self.len = 5
+
+        assert super_len(LenFile()) == 5
+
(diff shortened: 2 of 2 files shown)


Lukasa:
It'd be really nice to have some more tests around this. Ideally tests that go into each of the branches. @nateprewitt, are you open to adding those?

nateprewitt:
Yep, I'll throw a few more together for `TestSuperLen`.

requests/utils.py
```diff
@@ -87,7 +83,23 @@ def super_len(o):
...
-            current_position = total_length
+            if total_length is not None:
+                current_position = total_length
+
+    if hasattr(o, 'seek') and total_length is None:
+        # StringIO and BytesIO have seek but no useable fileno
+
+        # seek to end of file
+        o.seek(0, 2)
+        if hasattr(o, 'tell'):
```
tzickel: why enter the enclosing if statement at all if the object has no 'seek' and 'tell' together ?

Lukasa: Yeah, good spot. This block does nothing sensible if there is no `tell`, so we need to check for both.

nateprewitt: Good catch!

requests/utils.py
```diff
@@ -87,7 +83,22 @@ def super_len(o):
             # is actually a special file descriptor like stdin. In this
             # instance, we don't know what the length is, so set it to zero and
             # let requests chunk it instead.
-            current_position = total_length
+            if total_length is not None:
+                current_position = total_length
+
+        if hasattr(o, 'seek') and total_length is None:
+            # StringIO and BytesIO have seek but no useable fileno
```
tzickel, line 90: StringIO.StringIO (not io.StringIO) has no fileno. (io.StringIO/io.BytesIO .seek does return the new position, while StringIO.StringIO does not).

Lukasa: io.StringIO also has no fileno.

nateprewitt: I believe that's what this comment conveys @tzickel, they all have seek and no _useable_ form of `fileno`. Was there something you were looking to see changed to make the comment clearer? 99364b6 was added to remove the use of seek to set `total_length` because of the inconsistency between the `io` and `StringIO` modules.

tests/test_utils.py
```diff
@@ -66,6 +66,32 @@ def test_file(self, tmpdir, mode, warnings_num, recwarn):
...
 
+    def test_super_len_with__len__(self):
+        foo = [1,2,3,4]
+        len_foo = super_len(foo)
+        assert len_foo == 4
+
+    def test_super_len_with_no__len__(self):
+        foo = BytesIO(b'1234')
+        assert getattr(foo, '__len__', None) is None
+        setattr(foo, 'len', 4)
```
nateprewitt: @Lukasa do you know of a class with `len` but no `__len__` this works but is a little hacky.

tests/test_utils.py
```diff
@@ -66,6 +66,32 @@ def test_file(self, tmpdir, mode, warnings_num, recwarn):
...
+        foo.read(2)
+        assert super_len(foo) == 3
+
+    def test_super_len_with_fileno(self):
+        f = open(__file__, 'rb')
+        length = super_len(f)
+        file_data = f.read()
+        f.close()
+        assert length == len(file_data)
+
```
nateprewitt, line 97: I was going to add a test for a file-like object that didn't have any of the attributes (`__len__`, `len`, `fileno`, or `tell`/`seek`) to make sure it defaults to 0. I can't delete `seek` or `tell` from any of the io classes, or default file object though. Any suggestions on how to simulate this?

tests/test_utils.py
Lukasa: Not off the top of my head, but you could write your own!

tests/test_utils.py
Lukasa: Once again, I think the easiest thing to do is write a custom class. =)

requests/utils.py
```diff
@@ -87,7 +83,22 @@ def super_len(o):
...
+
+            # seek to end of file
+            o.seek(0, 2)
+            total_length = o.tell()
+
+            # seek back to current position to support
+            # partially read file-like objects
+            o.seek(current_position or 0)
```
tzickel, line 98: This is 0 by default, so no need for the "or 0" here.

nateprewitt: So we set current_position on [line 80](https://github.com/nateprewitt/requests/blob/03535ce5baa0931cdce01b849ca37795229f3bc5/requests/utils.py#L80) to o.tell() which in most cases should be a valid number. If someone defines a custom tell() method that returns None under circumstance x, this will blow up with a `TypeError`.

I think this is a pretty rare edge case but I understand why @jseabold put it in place. Why leave a point of failure to avoid a quick comparison?

I'll defer to @Lukasa if he has an opinion, otherwise I think it's worth keeping.

nateprewitt:
@Lukasa, tests are updated with custom classes. I think things should be good to go with tests.

requests/utils.py
Lukasa: I'm happy to leave it. It costs relatively little.

tests/test_utils.py
```diff
@@ -66,6 +66,40 @@ def test_file(self, tmpdir, mode, warnings_num, recwarn):
...
+        assert super_len(foo) == 5
+
+    def test_super_len_with_tell(self):
+        foo = StringIO.StringIO('12345')
+        assert super_len(foo) == 5
+        foo.read(2)
+        assert super_len(foo) == 3
+
+    def test_super_len_with_fileno(self):
+        f = open(__file__, 'rb')
```
Lukasa: Can we use a context manager here?

tests/test_utils.py
```diff
@@ -66,6 +66,40 @@ def test_file(self, tmpdir, mode, warnings_num, recwarn):
...
+        f.close()
+        assert length == len(file_data)
+
+    def test_super_len_with_no_matches(self):
+        """Ensure that objects without any length methods default to 0"""
+        class NoMethodsFile(object):
+            pass
+
+        foo = NoMethodsFile()
+        assert super_len(foo) == 0
```
Lukasa: You can skip the extra class here and just do `assert super_len(object()) == 0`.

Lukasa:
Small notes about the tests, but this is looking really good!

tests/test_utils.py
nateprewitt: Definitely, I somehow got it in my head that context managers didn't work in 2.6 because you can't use them with assertRaises() in 2.6. Updated.

nateprewitt:
Great! Changes in last notes have been updated.

Lukasa:
Ok, cool, I like this. @sigmavirus24?

nateprewitt:
@sigmavirus24, wanted to ping in case this fell off the queue. No rush, just checking in :)

requests/utils.py
```diff
@@ -87,7 +83,22 @@ def super_len(o):
             # is actually a special file descriptor like stdin. In this
             # instance, we don't know what the length is, so set it to zero and
             # let requests chunk it instead.
-            current_position = total_length
+            if total_length is not None:
+                current_position = total_length
+
+        if hasattr(o, 'seek') and total_length is None:
```
sigmavirus24, line 89: So we know and agree that hasattr is bad, but I believe some of the `io` objects have a `seekable` method that will do what we want here. Thoughts?

requests/utils.py
sigmavirus24: I agree with @Lukasa

sigmavirus24:
I have one suggestion but it's optional.

requests/utils.py
nateprewitt: So I like that this as a simpler option to the two hasattr checks, but we'd still have to leave those in place for StringIO-like objects, so it would end up being a somewhat redundant portion of the conditional.

I do, however, think we're likely going to want to extract [this logic](https://github.com/nateprewitt/requests/blob/f5c59743e7a45b68eaa82c83baaaed5607348afe/requests/utils.py#L78-L98) into it's own function in order to solve #3079 so maybe it could fit in with that somehow? I haven't worked up a good solution for that yet though.

@Lukasa, do you have any opinions on using this?

Lukasa: I don't think that we can rely on `seekable`, sadly.

nateprewitt: Ok cool, we can probably table this then, at least until someone starts addressing #3079.

nateprewitt:
I don't think I've got anything else, so I think we're good to merge from my end. Let me know if you have anything else @lukasa.

We can probably start discussing #3545 too.

Lukasa:
Ok then, let's do this! =D

PR #3536, opened 2016-08-24
Test case for requests getting stuck on post redirect with seekable streams

tzickel:
https://github.com/kennethreitz/requests/issues/3079

This is a test case that depends on httpbin 0.5.0 to show the bug.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -154,6 +154,22 @@ class TestRequests:
         assert r.history[0].status_code == 302
         assert r.history[0].is_redirect
 
+    def test_HTTP_307_ALLOW_REDIRECT_POST(self, httpbin):
+        parts = urlparse(httpbin('post'))
+        url = "HTTP://" + parts.netloc + parts.path
+        r = requests.post(httpbin('redirect-to'), data='test', params={'url': url, 'status_code': 307})
+        assert r.status_code == 200
+        assert r.history[0].status_code == 307
+        assert r.history[0].is_redirect
+
+    def test_HTTP_307_ALLOW_REDIRECT_POST_WITH_SEEKABLE(self, httpbin):
+        parts = urlparse(httpbin('post'))
+        url = "HTTP://" + parts.netloc + parts.path
+        r = requests.post(httpbin('redirect-to'), data=io.BytesIO(b'test'), params={'url': url, 'status_code': 307})
+        assert r.status_code == 200
+        assert r.history[0].status_code == 307
+        assert r.history[0].is_redirect
+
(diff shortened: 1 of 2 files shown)


tests/test_requests.py
```diff
@@ -154,6 +154,22 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
...
+        parts = urlparse(httpbin('post'))
+        url = "HTTP://" + parts.netloc + parts.path
+        r = requests.post(httpbin('redirect-to'), data='test', params={'url': url, 'status_code': 307})
+        assert r.status_code == 200
+        assert r.history[0].status_code == 307
+        assert r.history[0].is_redirect
+
+    def test_HTTP_307_ALLOW_REDIRECT_POST_WITH_SEEKABLE(self, httpbin):
+        parts = urlparse(httpbin('post'))
+        url = "HTTP://" + parts.netloc + parts.path
```
Lukasa, line 167: Why do we need this?

Lukasa:
Thanks for these tests! They do, as expected, fail.

What we need now, I think, is for someone to step up and write a solution. Any volunteers?

tests/test_requests.py
tzickel: I thought to show a test where it works, and another where it doesn't, if it's redundant it can be removed.

Lukasa: Sorry, I wasn't clear enough. I mean literally why do we need this line that unpacks and rebuilds the URL?

tzickel:
I guess this issue can only be fixed inside urllib3 since it handles the actual redirects.

nateprewitt:
@tzickel At a quick glance, I don't think that urllib3 has any knowledge of whether our request is a redirect or not. We'd have to build extra params in both our `send` function and urllib3's `urlopen` to determine if it needs to rewind itself and to where. Otherwise, we could end up with unintentionally mangled body content. That would be extra overhead for *every* request, in order to solve what is a rarer edge case.

I'm not claiming #3655 is the best solution either though. Any input there or clarification on what I'm missing in a urllib3 implementation would be appreciated :)

nateprewitt:
@tzickel At a quick glance, I don't think that urllib3 has any knowledge of whether our request is a redirect or not. We pass each redirect call to urllib3 as if it were a single request.

#3655 addresses the problem posed in your second test, I'm not claiming it's the best solution though. Any input there or clarification on what I'm missing in a urllib3 implementation would be appreciated :)

nateprewitt:
@Lukasa tzickel's commit was merged into master with #3655, so this can probably be closed.

Lukasa:
Yup. Thanks for the work @tzickel!

PR #3539, opened 2016-08-26
Add persistent examples

RichieB2B:
Learning python and Requests it was not immediately clear that many arguments can also be made persistent using the Sessions class.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -208,6 +208,11 @@ You can pass ``verify`` the path to a CA_BUNDLE file or directory with certifica
 
     >>> requests.get('https://github.com', verify='/path/to/certfile')
 
+or persistent::
+
+    s = requests.Session()
+    s.verify = '/path/to/certfile'
+
 .. note:: If ``verify`` is set to a path to a directory, the directory must have been processed using
   the c_rehash utility supplied with OpenSSL.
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
@Lukasa I was certain this was described elsewhere, am I wrong?

Lukasa:
[The first sentence of the Session docs](http://docs.python-requests.org/en/master/user/advanced/#session-objects) reads:

> The Session object allows you to persist certain parameters across requests.

Is there a reason we need this change? Do we need to improve the Session docs directly?

RichieB2B:
While searching on how to use an SSL client certificate the [developer interface documentation](http://docs.python-requests.org/en/master/api/) did not help at all. SSL client certificates are only mentioned as a parameter for requests.request(). 

The [Session Object documentation](http://docs.python-requests.org/en/master/user/advanced/#session-objects) does have a [section](http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification) covering SSL client certificates but only as a parameter to requests.get(). Additional examples for using persistent parameters would have really helped me.

Lukasa:
Well, the scope of the changes here is pretty small, so I don't object to merging this for now. =)

Lukasa:
Thanks @RichieB2B!

PR #3541, opened 2016-08-26
Specify self.cert is used for SSL client certificates

RichieB2B:
In addition to #3539 clarify that self.cert is used for SSL client certificates


Lukasa:
Looks good to me, thanks!

PR #3545, opened 2016-08-29
removing redundant logic from prepare_content_length

nateprewitt:
Background
========
So while working on #3535 I noticed that `prepare_content_length` will be fairly redundant after the patch is merged. I was going to push the abbreviated version of `prepare_content_length` into #3535 but found some inconsistencies in how we're currently handling Content-Length.

So here's a quick rundown of what I've found, and what it seems like *should* happen.

* #957 addressed #223 and added Content-Length to *everything*.
* #1142 amended this to NOT send Content-Length with GET/HEAD requests (Issue #1051)
* #2329 addressed the issue of not being able to override Content-Length with a custom header (Issue #2329)

From this, I'd expect:

* Content-Length to be set on all requests that are not using a HEAD/GET method.
* If a Content-Length header exists, that it remains untouched when sending a request.

Next Steps
=========

Currently, setting a body on a request will override the custom Content-Length header. The test implemented in #2329 doesn't test the code path with a body. I've added that test and the required logic to make it succeed. However, the solution attached here is not my preferred approach, rather the closest to the old functionality of prepare_content_length.

*Ideally*, I think that any call to `prepare_content_length` will do just that, set the Content-Length. That way if someone updates the body on a PreparedRequest, they can use `prepare_content_length` without having to delete a header, or recreate the object. This approach would require moving `if self.headers.get('Content-Length') is None` up to `prepare_body` and `prepare_auth` to run the overwrite check there.

This has the issue of causing latent errors though, if someone decides to use `prepare_content_length` elsewhere without the existence check. So if we're concerned about that being a possibility, then I think `prepare_content_length` should instead have an `overwrite` param which will bypass the [check](https://github.com/kennethreitz/requests/compare/master...nateprewitt:new_prepare_content_length#diff-afd5aad80649cdfae687bee05242c8faR472) to see if the header exists.

==========
This PR is two tangentially related fixes (abbreviating prepare_content_length post #3535 and fixing inconsistencies in how we handle Content-Lengths). This SHOULD NOT be merged and probably doesn't need to even be addressed until after #3535 is merged.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -424,7 +424,6 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         # Nottin' on you.
         body = None
         content_type = None
-        length = None
 
         if not data and json is not None:
             # urllib3 requires a bytes-like body. Python 2's json.dumps
@@ -475,17 +474,16 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         self.body = body
 
     def prepare_content_length(self, body):
-        if hasattr(body, 'seek') and hasattr(body, 'tell'):
-            curr_pos = body.tell()
-            body.seek(0, 2)
-            end_pos = body.tell()
-            self.headers['Content-Length'] = builtin_str(max(0, end_pos - curr_pos))
-            body.seek(curr_pos, 0)
-        elif body is not None:
-            l = super_len(body)
-            if l:
-                self.headers['Content-Length'] = builtin_str(l)
-        elif (self.method not in ('GET', 'HEAD')) and (self.headers.get('Content-Length') is None):
+        """Prepare Content-Length header based on request method and body"""
+        if body is not None:
+            length = super_len(body)
+            if length:
+                # If length exists, set it. Otherwise, we fallback
+                # to Transfer-Encoding: chunked.
+                self.headers['Content-Length'] = builtin_str(length)
+        elif self.method not in ('GET', 'HEAD') and self.headers.get('Content-Length') is None:
+            # Set Content-Length to 0 for methods that can have a body
+            # but don't provide one. (i.e. not GET or HEAD)
             self.headers['Content-Length'] = '0'
 
     def prepare_auth(self, auth, url=''):
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -90,6 +90,16 @@ class TestRequests:
         req = requests.Request(method, httpbin(method.lower())).prepare()
         assert 'Content-Length' not in req.headers
 
+    @pytest.mark.parametrize('method', ('POST', 'PUT', 'PATCH', 'OPTIONS'))
+    def test_no_body_content_length(self, httpbin, method):
+        req = requests.Request(method, httpbin(method.lower())).prepare()
+        assert req.headers['Content-Length'] == '0'
+
+    @pytest.mark.parametrize('method', ('POST', 'PUT', 'PATCH', 'OPTIONS'))
+    def test_empty_content_length(self, httpbin, method):
+        req = requests.Request(method, httpbin(method.lower()), data='').prepare()
+        assert req.headers['Content-Length'] == '0'
+
     def test_override_content_length(self, httpbin):
         headers = {
             'Content-Length': 'not zero'


nateprewitt:
@Lukasa whenever you've got a chance, could I get your thoughts on this?

requests/models.py
```diff
@@ -470,18 +469,12 @@ def prepare_body(self, data, files, json=None):
...
-            if l:
-                self.headers['Content-Length'] = builtin_str(l)
-        elif (self.method not in ('GET', 'HEAD')) and (self.headers.get('Content-Length') is None):
-            self.headers['Content-Length'] = '0'
+        if self.headers.get('Content-Length') is None:
+            if body is not None:
+                length = super_len(body)
+                self.headers['Content-Length'] = builtin_str(length)
+            elif self.method not in ('GET', 'HEAD'):
+                self.headers['Content-Length'] = '0'
```
Lukasa: First note: we definitely need to comment this better to explain what we're doing. In this case, we're saying that we only want to override the `Content-Length` header if it hasn't already been set for any reason.

My question is: are we sure that that's the logic we want? Your description suggests that `prepare_content_length` should have an `overwrite` param, but this code doesn't add it.

Can you elaborate on what's going on here?

nateprewitt: Sorry, this probably should have been split into two PRs but I'll try to define the problems more clearly. 

This is the "solution" that was closest to the original but I would say not a complete or ideal one. I was looking for some more input on what direction you felt was best to proceed in. We definitely need to move `if self.headers.get('Content-Length') is None` outside of this method, or change it to `if self.headers.get('Content-Length') is None or overwrite` for this to work properly.

Currently, we overwrite 'Content-Length' in ALL cases where a body is attached. This will subtly override a 'Content-Length' header passed to a request. @kennethreitz stated in #2329 that he'd prefer users be able to do this and I feel like having to go back and change 'Content-Length' after already passing your custom value is awkward.

Some examples:

``` python
>>> import requests
>>> r = requests.post('http://httpbin.org/post', headers={'Content-Length': '5'}, data="thisisatest")
>>> r.request.headers
{'Connection': 'keep-alive', 'Content-Length': '11', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'User-Agent': 'python-requests/2.11.1'}
```

``` python
>>> import requests
>>> r = requests.Request('POST', 'http://httpbin.org/post', headers={'Content-Length': '5'}, data="thisisatest")
>>> r.headers
{'Content-Length': '5'}
>>> pr = r.prepare()
>>> pr.headers
{'Content-Length': '11'}
```

Having to do `pr.headers['Content-Length'] = '5'` after I've already passed in the custom headers I want is the issue I'm trying to address. If this is a feature we're going to support, does it need to subtly "fail" like this.

nateprewitt:
Ok, I'm going to do an abridged reframe for this PR because the first post is a bit wordy.

Now that we've updated `super_len` in #3535, `prepare_content_length` is mostly redundant and a space were we could accidentally end up with differing functionality in the future. I'm proposing we consolidate the method to the code below, which will have equivalent logic:

``` python
    def prepare_content_length(self, body):
        if body is not None:
            length = super_len(body)
            self.headers['Content-Length'] = builtin_str(length)
        elif self.method not in ('GET', 'HEAD') and self.headers.get('Content-Length') is None:
            self.headers['Content-Length'] = '0'
```

The second issue is the functionality over 'Content-Length' override, which I addressed [above](https://github.com/kennethreitz/requests/pull/3545#discussion_r79694617). They seemed somewhat related at first, which is why I raised it at a single PR.

requests/models.py
Lukasa: Allowing users to override `Content-Length` is a dangerous game. In particular, users who are _wrong_ about the header will encounter problems whereby either the server returns 400 Client Error, terminates the connection, or worse, the connection hangs while the server waits for more body. Your example is a great example of this case: the user should be _utterly forbidden_ from emitting that request. I have exactly no interest in allowing users to easily emit that request.

The case I _do_ care about is the possibility that the user may want to set Content-Length in one specific case: the case where we'd normally set `Transfer-Encoding: chunked`. In this instance we could take this as a signal to avoid chunked transfer encoding. However, in general, we haven't done this consistently and this PR is not the right place to address it.

Lukasa:
I think, given my comment above, that we should restrict ourselves to the redundancy removal for now. Overwriting Content-Length is, IMO, a severe misfeature, and I'm willing to make that case pretty forcefully to Kenneth if he needs convincing. ;)

nateprewitt:
Ok, that's good enough for me. For cataloging purposes, would you prefer I pare this PR down to just the redundancy fix, or close this and open a new one?

requests/models.py
nateprewitt: It's been a while since I was in this, so I've lost track of some of my thought process but it's slowly coming back. While I agree that it's pretty easy to accidentally shoot yourself in the foot with this approach, we specifically approved an odd subset of this functionality in #2329/#2332.

``` python
>>> import requests
>>> r = requests.Request('POST', 'http://httpbin.org/post', headers={'Content-Length': '5'})
>>> r.headers
{'Content-Length': '5'}
>>> pr = r.prepare()
>>> pr.headers
{'Content-Length': '5'}
```

Given the above reasoning, why is this acceptable here but not when a body is attached? I don't mean to waste your time for my own education. I just think there's enough ambiguity in the implementation that I want to make sure the situation is fully understood before we lay this to rest.

Lukasa:
Let's pare this one down. =)

requests/models.py
Lukasa: I think, frankly, that that decision was a mistake. Given that it was mine, I think we can hold ourselves accountable to that. The correct way to mess with content-length is in the PreparedRequest: nowhere else. =)

sigmavirus24:
I'd like to be included on this review prior to its merging. We (three) have made enough serious problems in the content-length portion of requests that it needs as many reviews as we can afford.

nateprewitt:
Ok, things are pared down. The patch is now pretty simple.

The logic for file-like objects (with seek/tell) is now in [`super_len`](https://github.com/kennethreitz/requests/blob/d7227fbb7e07af35f23a0d370ab3b01661af9e40/requests/utils.py#L78-L103). I'm removing [this chunk](https://github.com/kennethreitz/requests/blob/d7227fbb7e07af35f23a0d370ab3b01661af9e40/requests/models.py#L478-L483) since it is now redundant. I'm also adding two tests to verify that Content-Length is set to 0 appropriately.

@Lukasa, @sigmavirus24 feel free to take a peak when you have a moment.

sigmavirus24:
@Lukasa it looks as though @nateprewitt pushed new changes after your approval. Care to re-review?

sigmavirus24:
Woops, forgot to approve this. :)

PR #3554, opened 2016-09-05
ISO-8859-1 fallback for reason decoding

nateprewitt:
This implements the proposed solution to #3538 by falling back from Unicode to ISO-8859-1 for raw reason decoding.

This is a relatively trivial fix, and I wasn't sure if you wanted to waste bandwidth fixing, @mitsuhiko. I had a brief chat with @Lukasa who said to toss this up, but I'll gladly drop it in favor of #3538, if you had plans to update.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -848,6 +848,10 @@ class Response(object):
 
         http_error_msg = ''
         if isinstance(self.reason, bytes):
+            # We attempt to decode utf-8 first because some servers
+            # choose to localize their reason strings. If the string
+            # isn't utf-8, we fall back to iso-8859-1 for all other
+            # encodings. (See PR #3538)
             try:
                 reason = self.reason.decode('utf-8')
             except UnicodeDecodeError:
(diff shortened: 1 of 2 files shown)


requests/models.py
```diff
@@ -848,7 +848,10 @@ def raise_for_status(self):
 
         http_error_msg = ''
         if isinstance(self.reason, bytes):
-            reason = self.reason.decode('utf-8', 'ignore')
+            try:
+                reason = self.reason.decode('utf-8')
+            except UnicodeDecodeError:
+                reason = self.reason.decode('iso-8859-1')
```
Lukasa, line 858: Can we add comments to the start of this block explaining why we're doing this?

nateprewitt: Yep! Comment added.

Lukasa:
@sigmavirus24 are you happy with this?

sigmavirus24:
Yep!

Issue #2817, opened 2015-10-09
RFE: Be able to pass 'no_proxy' in proxies dictionary argument


JohnVillalovos:
It would be good if were able to pass the 'no_proxy' value in the proxies dictionary argument. Currently it appears have to set the environment variable 'no_proxy'.

 For example:

```
proxies = {'https_proxy': 'http://proxy.example.com:3128',
           'no_proxy': '10.0.0.0/16,example.com,.example.com'}
requests.get('http://www.example.com', proxies=proxies)
```

Lukasa:
This is reasonable request, but the use-case seems small, especially as we now support per-host proxies. The only risk is that you might get redirected off-host, to a host that you don't want to reach via a proxy, but in most cases I'd expect that to be relatively unlikely and covered by the per-host proxies functionality.

Is that incorrect?

JohnVillalovos:
We see this use case when using Ironic bare-metal provisioning in the OpenStack project to provision a system from scratch. So the underlying temporary installation OS isn't really configured with proxies in mind.

I have not heard of 'per-host proxies' before. So unsure if that would work or not.

Lukasa:
The per-host proxies feature is mentioned in our changelog for 2.8.0, but it allows you to set up a proxies dictionary like so:

``` python
proxies = {'http://my-host.com/': 'http://proxy.example.com:3128'}
```

This will only use the proxy for requests to `my-host.com`: other requests will not use the proxy.

JohnVillalovos:
Unfortunately we don't have a good idea which hosts we will be accessing. Thus want the more general method.

Lukasa:
So you don't know what hosts you _will_ access, just which ones you mustn't? It does feel like you should be setting the `NO_PROXY` environment variable, so that other tools obey these rules too.

JohnVillalovos:
Right. So this installation OS is thrown away when done. So doesn't have the variables set.  All the work is done by the Ironic Python Agent, which is a Python program.

Our current plan is to just use os.environ['no_proxy'] = '....'  but that seems a little hackish but will work. Thus why I filed this RFE.

I'd be happy to try to come up with a pull-request, given a few pointers on what to modify.

Lukasa:
I think we could do it easily enough, but I'm increasingly reluctant to add further features to the already quite complicated proxy logic. In the short term, your solution will work. Longer term, I need to think about where the line is with further proxy features.

Lukasa:
[...]

PR #3561, opened 2016-09-08
Allow use of 'no_proxy' in the proxies argument

JohnVillalovos:
Add the ability to add 'no_proxy' and a value to the 'proxies'
dictionary argument.

https://github.com/kennethreitz/requests/issues/2817
Closes gh-2817

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -218,13 +218,16 @@ class SessionRedirectMixin(object):
 
         :rtype: dict
         """
+        proxies = proxies if proxies is not None else {}
         headers = prepared_request.headers
         url = prepared_request.url
         scheme = urlparse(url).scheme
-        new_proxies = proxies.copy() if proxies is not None else {}
+        new_proxies = proxies.copy()
+        no_proxy = proxies.get('no_proxy')
 
-        if self.trust_env and not should_bypass_proxies(url):
-            environ_proxies = get_environ_proxies(url)
+        bypass_proxy = should_bypass_proxies(url, no_proxy=no_proxy)
+        if self.trust_env and not bypass_proxy:
+            environ_proxies = get_environ_proxies(url, no_proxy=no_proxy)
 
             proxy = environ_proxies.get(scheme, environ_proxies.get('all'))
 
@@ -638,7 +641,8 @@ class Session(SessionRedirectMixin):
         # Gather clues from the surrounding environment.
         if self.trust_env:
             # Set environment's proxies.
-            env_proxies = get_environ_proxies(url) or {}
+            no_proxy = proxies.get('no_proxy') if proxies is not None else None
+            env_proxies = get_environ_proxies(url, no_proxy=no_proxy)
             for (k, v) in env_proxies.items():
                 proxies.setdefault(k, v)
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -11,6 +11,7 @@ that are also useful for external consumption.
 import cgi
 import codecs
 import collections
+import contextlib
 import io
 import os
 import re
@@ -542,7 +543,29 @@ def is_valid_cidr(string_network):
     return True
 
 
-def should_bypass_proxies(url):
+@contextlib.contextmanager
+def set_environ(env_name, value):
+    """Set the environment variable 'env_name' to 'value'
+
+    Save previous value, yield, and then restore the previous value stored in
+    the environment variable 'env_name'.
+
+    If 'value' is None, do nothing"""
+    if value is not None:
+        old_value = os.environ.get(env_name)
+        os.environ[env_name] = value
+    try:
+        yield
+    finally:
+        if value is None:
+            return
+        if old_value is None:
+            del os.environ[env_name]
+        else:
+            os.environ[env_name] = old_value
+
+
+def should_bypass_proxies(url, no_proxy):
     """
     Returns whether we should bypass proxies or not.
 
@@ -552,7 +575,9 @@ def should_bypass_proxies(url):
 
     # First check whether no_proxy is defined. If it is, check that the URL
     # we're getting isn't in the no_proxy list.
-    no_proxy = get_proxy('no_proxy')
+    no_proxy_arg = no_proxy
+    if no_proxy is None:
+        no_proxy = get_proxy('no_proxy')
     netloc = urlparse(url).netloc
 
     if no_proxy:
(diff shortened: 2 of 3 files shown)


requests/sessions.py
```diff
@@ -222,9 +222,12 @@ def rebuild_proxies(self, prepared_request, proxies):
...
         scheme = urlparse(url).scheme
         new_proxies = proxies.copy() if proxies is not None else {}
+        no_proxy = proxies.get('no_proxy') if proxies is not None else None
 
-        if self.trust_env and not should_bypass_proxies(url):
-            environ_proxies = get_environ_proxies(url)
+        if self.trust_env and not should_bypass_proxies(
+                url, no_proxy=no_proxy):
+            environ_proxies = get_environ_proxies(
+                url, no_proxy=no_proxy)
```
Lukasa: I think this fits fine without needing a line break. =)

requests/sessions.py
```diff
@@ -222,9 +222,12 @@ def rebuild_proxies(self, prepared_request, proxies):
         url = prepared_request.url
         scheme = urlparse(url).scheme
         new_proxies = proxies.copy() if proxies is not None else {}
+        no_proxy = proxies.get('no_proxy') if proxies is not None else None
 
-        if self.trust_env and not should_bypass_proxies(url):
-            environ_proxies = get_environ_proxies(url)
+        if self.trust_env and not should_bypass_proxies(
+                url, no_proxy=no_proxy):
```
Lukasa: Rather than break this line, which is kind of ugly, let's just create a named variable (maybe `bypass`) that we assign the value of this to so we can keep the conditional clear.

requests/sessions.py
```diff
@@ -638,7 +641,9 @@ def merge_environment_settings(self, url, proxies, stream, verify, cert):
         # Gather clues from the surrounding environment.
         if self.trust_env:
             # Set environment's proxies.
-            env_proxies = get_environ_proxies(url) or {}
+            no_proxy = proxies.get('no_proxy') if proxies is not None else None
+            env_proxies = get_environ_proxies(
+                url, no_proxy=no_proxy)
```
Lukasa: Similarly, I don't think this needs a line break here.

requests/sessions.py
```diff
@@ -222,9 +222,12 @@ def rebuild_proxies(self, prepared_request, proxies):
         url = prepared_request.url
         scheme = urlparse(url).scheme
         new_proxies = proxies.copy() if proxies is not None else {}
+        no_proxy = proxies.get('no_proxy') if proxies is not None else None
```
Lukasa: At this point we're doing two `if proxies is not None` checks one after another. Given that the only places we use it in this function are these two lines, let's just add a line at the start that says `proxies = proxies if proxies is not None else {}`. That lets us remove the checks from these two lines.

requests/utils.py
```diff
@@ -585,24 +610,25 @@ def should_bypass_proxies(url):
...
     # legitimate problems.
-    try:
-        bypass = proxy_bypass(netloc)
-    except (TypeError, socket.gaierror):
-        bypass = False
+    with set_environ('no_proxy', no_proxy_arg):
+        try:
+            bypass = proxy_bypass(netloc)
+        except (TypeError, socket.gaierror):
+            bypass = False
```
Lukasa, line 617: Why are we doing this?

tests/test_utils.py
```diff
@@ -128,7 +128,32 @@ def test_bypass(self, url):
...
+            'http://192.168.0.1:5000/',
+            'http://192.168.0.1/',
+            'http://172.16.1.1/',
+            'http://172.16.1.1:5000/',
+            'http://localhost.localdomain:5000/v1.0/',
+        ))
+    def test_not_bypass_no_proxy_keyword(self, url, monkeypatch):
+        # This is testing that the 'no_proxy' argument overrides the
+        # environment variable 'no_proxy'
+        monkeypatch.setenv('http_proxy', 'http://proxy.example.com:3128/')
```
Lukasa, line 154: There is no environment variable `no_proxy` here, so this doesn't really test an override.

tests/test_utils.py
```diff
@@ -472,4 +497,24 @@ def test_should_bypass_proxies(url, expected, monkeypatch):
...
+            ('http://google.com:5000/v1.0/', False),
+    ))
+def test_should_bypass_proxies_no_proxy(
+        url, expected, monkeypatch):
+    """Tests for function should_bypass_proxies to check if proxy
+    can be bypassed or not using the 'no_proxy' argument
+    """
+    no_proxy = '192.168.0.0/24,127.0.0.1,localhost.localdomain,172.16.1.1'
+    # Test 'no_proxy' argument
+    assert should_bypass_proxies(url, no_proxy=no_proxy) == expected, url
```
Lukasa: There's no need for the comma, the test output will provide enough information for us.

Lukasa:
Alright, good start @JohnVillalovos! I've left some feedback in the diff that would be good to have addressed, thanks. =D

tests/test_utils.py
JohnVillalovos: Line 109 causes the 'no_proxy' environment variable to be set

https://github.com/kennethreitz/requests/pull/3561/files#diff-87010571bbeeee0b9a13279c493670f3R109

tests/test_utils.py
JohnVillalovos: Done

requests/utils.py
JohnVillalovos: This was so that the value we have for 'no_proxy' in 'proxies' gets set in the environment and then undone.

requests/sessions.py
JohnVillalovos: Done

requests/sessions.py
JohnVillalovos: Done

requests/sessions.py
JohnVillalovos: Done

requests/sessions.py
JohnVillalovos: Done

tests/test_utils.py
Lukasa: Ah, sorry, I missed that.

requests/utils.py
Lukasa: So I don't think we need it: see my comment in the issue for more.

JohnVillalovos: Thanks. I updated it to just return False if the passed in 'no_proxy' argument was set if it gets to this point in the code.

Lukasa: So, @sigmavirus24, I think we have two questions here. The first one is: is this an acceptable solution to this problem? The second one is: does this even work?

kennethreitz:

JohnVillalovos:
As far as I know the submitted patch is still valid, though I haven't dug into it.

I thought I had responded to all comments and latest version took them into account.

sigmavirus24:
@JohnVillalovos that's true but we cannot reopen this since your repository seems to have been deleted.

JohnVillalovos:
Okay. Looks like I have some merge conflicts. I will create another pull request after I fix it.

sigmavirus24:
Thanks @JohnVillalovos

JohnVillalovos:
Created new pull request: https://github.com/kennethreitz/requests/pull/3865

PR #3563, opened 2016-09-08
use enforce_content_length in Requests

nateprewitt:
This is a **DO NOT MERGE** until we bring urllib3 1.17 into the Requests `proposed/3.0.0` branch, which should likely happen before 3.0.0 gets out the door.

This is the final step for the work in shazow/urllib3#949 and addresses the issues originally raised in shazow/urllib3#723 and #2833.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -434,7 +434,8 @@ class HTTPAdapter(BaseAdapter):
                     preload_content=False,
                     decode_content=False,
                     retries=self.max_retries,
-                    timeout=timeout
+                    timeout=timeout,
+                    enforce_content_length=True
                 )
 
             # Send the request.
@@ -478,7 +479,9 @@ class HTTPAdapter(BaseAdapter):
                         pool=conn,
                         connection=low_conn,
                         preload_content=False,
-                        decode_content=False
+                        decode_content=False,
+                        enforce_content_length=True,
+                        request_method=request.method
                     )
                 except:
                     # If we hit any problems here, clean up the connection.
--- a/requests/models.py
+++ b/requests/models.py
@@ -28,7 +28,8 @@ from .packages.urllib3.fields import RequestField
 from .packages.urllib3.filepost import encode_multipart_formdata
 from .packages.urllib3.util import parse_url
 from .packages.urllib3.exceptions import (
-    DecodeError, ReadTimeoutError, ProtocolError, LocationParseError)
+    DecodeError, ReadTimeoutError, ProtocolError,
+    LocationParseError, ConnectionError)
 from .exceptions import (
     HTTPError, MissingScheme, InvalidURL, ChunkedEncodingError,
     ContentDecodingError, ConnectionError, StreamConsumedError)
@@ -701,7 +702,10 @@ class Response(object):
                     for chunk in self.raw.stream(chunk_size, decode_content=True):
                         yield chunk
                 except ProtocolError as e:
-                    raise ChunkedEncodingError(e)
+                    if self.headers.get('Transfer-Encoding') == 'chunked':
+                        raise ChunkedEncodingError(e)
+                    else:
+                        raise ConnectionError(e)
                 except DecodeError as e:
                     raise ContentDecodingError(e)
                 except ReadTimeoutError as e:
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -21,7 +21,7 @@ from ._internal_utils import to_native_string
 from .utils import to_key_val_list, default_headers
 from .exceptions import (
     TooManyRedirects, InvalidScheme, ChunkedEncodingError,
-    ContentDecodingError, InvalidHeader)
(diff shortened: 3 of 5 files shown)


requests/adapters.py
```diff
@@ -420,7 +420,9 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
                     preload_content=False,
                     decode_content=False,
                     retries=self.max_retries,
-                    timeout=timeout
+                    timeout=timeout,
+                    enforce_content_length=True,
+                    request_method=request.method
```
Lukasa: Do we need to explicitly provide this here?

requests/adapters.py
```diff
@@ -464,7 +466,9 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
                         pool=conn,
                         connection=low_conn,
                         preload_content=False,
-                        decode_content=False
+                        decode_content=False,
+                        enforce_content_length=True,
+                        request_method=request.method
```
Lukasa, line 484: Do we need to explicitly provide this here?

requests/adapters.py
nateprewitt: No, this was an oversight in which class we were calling `urlopen` on.

requests/adapters.py
nateprewitt: [see above](https://github.com/kennethreitz/requests/pull/3563#discussion_r78067817).

nateprewitt: Unfortunately yes, there's specific logic for a HEAD request in length_remaining and it's not accessible from the response object in urllib3. We're passing this information in manually in urllib3 in `urlopen` but not `from_httplib`.

~~I'd say this is probably an oversight I made in shazow/urllib3#949 though, I'd be happy to pass the method along there so we're not having to do this manually from outside of urllib3. I'll open a quick amendment patch in a second if that seems acceptable @Lukasa.~~

nateprewitt: Alright, the method is available inside of `from_httplib` but only as a private variable `_method` on the `httplib` object. I don't know if we want to use that inside of urllib3, so this would be the alternative option. `from_httplib` is a lower level method than `urlopen`, so we should probably be providing the method like this in the `response_kw` list, rather than trying to reimplement functionality that's already happening at a higher level.

nateprewitt:
@Lukasa, here's a bit of a hiccup I'd appreciate some input on. Currently all data reads in Requests end up using `itercontent`. 

Currently, regardless of whether or not the content is streamed, we end up initially running it through the `generate` function inside `itercontent`. This function assumes we're using Transfer-Encoding: chunked if it's called, and raises `ChunkedEncodingError` wrapping urllib3's `ProtocolError`. This is clearly incorrect because we're not chunked for responses with Content-Length but I'm not sure how best to catch this. 

We could parse the exception string for 'IncompleteRead', but that seems clunky to me. We could also add a conditional inside to raise `ProtocolError` on `stream=False` and `ChunkedEncodingError` on `stream=True` but that may be subtle enough to confuse since it's the same base error. The last option I can think of is removing `ChunkedEncodingError` completely and just raising the unmodified `ProtocolError` instead. Thoughts?

Lukasa:
...since when does `iter_content()` make any assumptions about the transfer-encoding of the response? Can you point to where in the code you believe this assumption is made?

Lukasa:
Oh, nevermind, I see! There's no need for all that, we can just check the response headers for a transfer encoding. =)

nateprewitt:
This is [the block](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L678-L679) I'm hitting now that we're raising the appropriate `ProtocolError` in urllib3. Perhaps this is simply a wording issue on the [exception](https://github.com/kennethreitz/requests/blob/master/requests/exceptions.py#L88-L89) though. When I read "chunked encoding", I immediately go to Transfer-Encoding, but perhaps that's me misreading things.

nateprewitt:
Ok, great! To clarify, are you suggesting a conditional in the `except ProtocolError` checking the headers? Is raising the `ProtocolError` otherwise appropriate?

Edit: The reason I ask is it would adversely affect `resolve_redirect` [here](https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L106-L107).

Lukasa:
We must raise a Requests exception.

nateprewitt:
Alright, I'll run with that then. Thanks!

nateprewitt:
@Lukasa we'll need to merge master into Proposed/3.0.0 to run the tests normally, but with 1.19 in Requests 2.12 this should be ready for a look.

Lukasa:
That merge is done. =)

nateprewitt:
Thanks @Lukasa! I'll merge it in now.

nateprewitt:
Branch updated and tests are running clean. I think this is ready for a review whenever you find a spare moment.

Lukasa:
Ok, I think this looks good! Thanks so much @nateprewitt! :sparkles:

PR #3574, opened 2016-09-15
Streaming responses require encoding to be set

shellhead:
Resolves the issues called out in #3359 and #3481 when `Response.iter_content(decode_unicode=True)` would return bytes instead of unicode. This is a breaking change as it raises an exception if `Response.encoding` is not set before invoking this function.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1042,6 +1042,22 @@ class TestRequests:
         chunks = r.iter_content(decode_unicode=True)
         assert all(isinstance(chunk, str) for chunk in chunks)
 
+        # raise an exception if encoding isn't set
+        r = requests.Response()
+        r.raw = io.BytesIO(b'the content')
+        r.encoding = None
+
+        with pytest.raises(TypeError):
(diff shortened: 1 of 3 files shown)


tests/test_requests.py
```diff
@@ -1042,6 +1042,22 @@ def test_response_decode_unicode(self):
...
+            chunks = r.iter_content(decode_unicode=True)
+
+        # raises an exception if the encoding is garbage
+        r = requests.Response()
+        r.raw = io.BytesIO(b'the content')
+        r.encoding = 'invalid encoding'
+
+        with pytest.raises(LookupError):
+            chunks = r.iter_content(decode_unicode=True)
+
```
Lukasa, line 1059: Is there any reason this isn't its own test? (Or ideally, two tests?)

sentientcucumber: No, I think I just added them to other without really thinking about it.

Not quite used to the new review process. If the new commit doesn't show up, please let me know.

PR #3575, opened 2016-09-15
added results of benchmark test about using session

vitaly-zdanevich:


Lukasa:
Thanks for this! I'm inclined to want to avoid having a specific benchmark, and instead to speak in general terms. Something general about approximate improvement would be better. Are you interested in that change?

vitaly-zdanevich:
When I first time read this documentation with words `significant performance` about session my thoughts was `eh maybe ten milliseconds`, I did not know that for every https call without session server must send certificate to client and get response and that this is seconds but not milliseconds, I was surprised when maked lazy test of it and get this numbers. Why are you against benchmarks? It is good illustration, I love benchmarks in any kind :)

Lukasa:
I'm not against benchmarks at all: I'm against benchmarks in _narrative documentation_. I'd be happy to have the benchmark in a `.py` file in the docs tree and then have the narrative docs just say what you said above with a link to that benchmark. It just doesn't need to be inline on the page, that's all.

sigmavirus24:
There's been no response to this pull request since feedback was provided. We can reopen this if interest in it is renewed at any point.

PR #3576, opened 2016-09-16
Fix encoding issue in test_response_reason_unicode_fallback

nateprewitt:
`test_response_reason_unicode_fallback` is currently failing in Python 2. The original proposed fix in #3557 unfortunately didn't address the underlying issue.

While this patch should fix the test now, it may be worth taking a brief moment to discuss how we envision the original PR #3554 working. Currently an error retrieved from an `except` clause can't be cast as `str`, `unicode`, written to a file, or easily decoded in Python 2. It prints to the console fine, but that's about it. This seems like a semi-serious issue for making this exception usable. I need to run a few more tests this evening or tomorrow, so we can sit on this for the weekend unless someone has some obvious insight I'm missing.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1026,13 +1026,13 @@ class TestRequests:
         # check raise_status falls back to ISO-8859-1
         r = requests.Response()
         r.url = 'some url'
-        reason = b'Komponenttia ei l\xf6ydy'
-        r.reason = reason
+        reason = u'Komponenttia ei löydy'
+        r.reason = reason.encode('latin-1')
         r.status_code = 500
         r.encoding = None
         with pytest.raises(requests.exceptions.HTTPError) as e:
             r.raise_for_status()
-        assert reason.decode('latin-1') in str(e)
+        assert reason in e.value.args[0]
 
     def test_response_chunk_size_type(self):
         """Ensure that chunk_size is passed as None or an integer, otherwise


tests/test_requests.py
```diff
@@ -1026,13 +1026,13 @@ def test_response_reason_unicode_fallback(self):
...
-        reason = b'Komponenttia ei l\xf6ydy'
-        r.reason = reason
+        reason = u'Komponenttia ei löydy'
+        r.reason = reason.encode('latin-1')
         r.status_code = 500
         r.encoding = None
         with pytest.raises(requests.exceptions.HTTPError) as e:
             r.raise_for_status()
-        assert reason.decode('latin-1') in str(e)
+        assert reason in e.value.args[0]
```
Lukasa, line 1035: This is surely not right: we're expecting the message to be _unicode_, not bytes.

nateprewitt: `e.value.args[0]` is the message, which is unicode. I believe this is part of the problem I'm discussing in the second paragraph.

``` python
import pytest
import requests

r = requests.Response()
r.url = 'some url'
reason = u'Komponenttia ei löydy'
r.reason = reason.encode('latin-1')
r.status_code = 500
r.encoding = None
with pytest.raises(requests.exceptions.HTTPError) as e:
        r.raise_for_status()
assert reason in e.value.args[0]
assert isinstance(e.value.args[0], unicode)
```

Note that `e` is pytest's `ExceptionInfo` class not an `HTTPError` in this case. That's why we're accessing `e.value.args[0]` instead of `e.args[0]`. `e.message` doesn't work in Python 3.

nateprewitt: Unless you were referencing `reason` v.s. `reason.decode('latin-1')` in which case, I reverted `reason` back to a unicode string so it was closer to the above unicode tests. Functionally, it's the same logic.

Lukasa: Aha, sorry, I misread this entire code base. I'm now 100% convinced that code reviews at conferences is a bad idea. ;) I have basically no problems.

nateprewitt:
Ok, I'm gonna call this good to go I guess. I feel like the behavior of the unicode string in an Exception is less than ideal but it's apparently one of the many long running unicode issues in Python 2.x.

For anyone reading this commit, trying to figure out how to do anything useful with the exception in Python 2, both of these should work:

``` python
unicode(e.message)  # unicode
unicode(e.message).decode('utf-8') # str
```

Lukasa:
@sigmavirus24 Would you like to take a look at this?

sigmavirus24:
@Lukasa I would like to pull it and run the tests locally. I suspect we might have a locale mismatch which is why this _appears_ to work for @nateprewitt and not for me. This is why I was using bytes explicitly previously.

nateprewitt:
Hey @sigmavirus24, while this could be a locale mismatch, there are 5 other tests in the same file using unicode characters in unicode strings like this (including \xf6). I _think_ I figured out why we were getting different results, and [detailed it](https://github.com/kennethreitz/requests/pull/3557#issuecomment-246908451), along with most of my other findings in #3557. I had @Lukasa confirm he could reproduce this issue before I opened the PR. Is this patch not working on your system?

sigmavirus24:
Quoting myself:

> I would like to pull it and run the tests locally.

nateprewitt:
> ...which is why **this** _appears_ to work for @nateprewitt and not for me.

Sorry @sigmavirus24, I interpreted "this" as meaning "this patch". I was just trying to clarify.

Lukasa:
Ok, so this test as written passes for me, as do all the tests. My locale is:

```
(env) cory@heimdall:requests/ % locale
LANG="en_GB.UTF-8"
LC_COLLATE="en_GB.UTF-8"
LC_CTYPE="en_GB.UTF-8"
LC_MESSAGES="en_GB.UTF-8"
LC_MONETARY="en_GB.UTF-8"
LC_NUMERIC="en_GB.UTF-8"
LC_TIME="en_GB.UTF-8"
LC_ALL=
```

sigmavirus24:
Thanks @nateprewitt. Care to update proposed/3.0.0 too then?

nateprewitt:
Sure thing, I'll move the changes over sometime today.

Issue #3579, opened 2016-09-17
Error when passing CookieJar to Session via PreparedRequest


agacek:
I'm using requests 2.11.1. The following code throws an exception:

```
from requests import Request, Session
from http.cookiejar import CookieJar

cj = CookieJar()
req = Request('GET', 'http://google.com', cookies=cj).prepare()
resp = Session().send(req)
```

The exception is:

```
Traceback (most recent call last):
  File "test.py", line 6, in <module>
    resp = Session().send(req)
  File "C:\Anaconda3\lib\site-packages\requests\sessions.py", line 617, in send
    history = [resp for resp in gen] if allow_redirects else []
  File "C:\Anaconda3\lib\site-packages\requests\sessions.py", line 617, in <listcomp>
    history = [resp for resp in gen] if allow_redirects else []
  File "C:\Anaconda3\lib\site-packages\requests\sessions.py", line 159, in resolve_redirects
    prepared_request._cookies.update(self.cookies)
AttributeError: 'CookieJar' object has no attribute 'update'
```

I'm able to avoid this by using `Session.prepare_request` or by passing the cookies via `Session.cookies`.

nateprewitt:
Thanks for opening this issue, @agacek. So the problem we're encountering here is a difference in how we "prepare" the CookieJar between a `PreparedRequest` and a `Session`.

`Session` will always [merge the cookies into a fresh `RequestsCookieJar`](https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L375-L376) when `prepare_request()` is called. `RequestsCookieJar` will have the update method, and is a subclass of `cookielib.CookieJar` so everything works fine.

`PreparedRequest` however doesn't perform this same merge, and won't modify the CookieJar if it's an instance of cookielib.CookieJar, so this one remains untouched. I'm not sure if this was an intentional feature of `PreparedRequest` to allow people to do what they wanted, or if it was an oversight. The code base is obviously expecting `cookies` to be `RequestsCookieJar`-like now though. If we're enforcing this in Session, we should probably also implement the same logic in `prepare_cookies` for the `PreparedRequest`. The user can always override the `cookies` param with their own CookieJar after if needed.

[...]

PR #3591, opened 2016-09-19
Fix RequestsCookieJar specific update call

nateprewitt:
This should solve the issue in #3579. `merge_cookies` performs the same `update` call inside of a try/except block which will fallback to the same logic that [`update`](https://github.com/kennethreitz/requests/blob/master/requests/cookies.py#L347-L353) itself implements for non-`RequestsCookieJar`-like objects.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -19,7 +19,8 @@ from requests.auth import HTTPDigestAuth, _basic_auth_str
 from requests.compat import (
     Morsel, cookielib, getproxies, str, urlparse,
     builtin_str, OrderedDict)
-from requests.cookies import cookiejar_from_dict, morsel_to_cookie
+from requests.cookies import (
+    cookiejar_from_dict, morsel_to_cookie, merge_cookies)
 from requests.exceptions import (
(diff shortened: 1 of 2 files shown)


tests/test_requests.py
```diff
@@ -344,6 +345,38 @@ def test_param_cookiejar_works(self, httpbin):
...
         assert r.json()['cookies']['foo'] == 'bar'
 
+    def test_cookielib_cookiejar_on_redirect(self, httpbin):
+        """Tests resolve_redirect doesn't fail when merging cookies
+        with non-RequestsCookieJar cookiejar.
+
+        See GH #3579
+        """
+        cj = cookielib.CookieJar()
+        cookiejar_from_dict({'foo': 'bar'}, cj)
```
Lukasa: While I know that `cookiejar_from_dict` mutates the cookiejar, not everyone will: can you change this to `cj = cookiejar_fro_dict({'foo': 'bar'}, cj)`? It's just slightly clearer.

nateprewitt: Sounds good, commit updated.

Lukasa:
Ok, per my note above I'm going ahead and merging this. Thanks @nateprewitt!

Issue #3597, opened 2016-09-25
TypeError in MockRequest when 'Host' header set to bytestring


bbamsch:
Issue is specific to Python3

An incorrect assumption is made in requests.cookies.MockRequest on Python3, [see here](https://github.com/kennethreitz/requests/blob/master/requests/cookies.py#L52):
```
class MockRequest(object):
    ...
    def get_full_url(self):
        # Only return the response's URL if the user hadn't set the Host
        # header
        if not self._r.headers.get('Host'):
            return self._r.url
        # If they did set it, retrieve it and reconstruct the expected domain
        host = self._r.headers['Host']
        parsed = urlparse(self._r.url)
        # Reconstruct the URL as we expect it
        return urlunparse([
            parsed.scheme, host, parsed.path, parsed.params, parsed.query,
            parsed.fragment
        ])
    ...
```

In Python2 this works fine where str and unicode objects are allowed interchangeably in urlunparse.

In Python3 this is an issue because ```host = self._r.headers['Host']``` may return a bytestring and ```        parsed = urlparse(self._r.url)``` returns a ParseResult with str objects. Combining these two causes urlunparse to throw a fit.

Steps to reproduce:
```
>>> import requests
>>> a = requests.Request('GET', 'http://www.google.com', headers={'Host': b'www.google.com'})
>>> b = requests.cookies.MockRequest(a)
>>> b.get_full_url()
```

Lukasa:
Yup, this looks wrong to me. It does look like we need to ensure that we have a match between the two types. In fact, we should probably ensure we're always returning the correct type for the platform.

bbamsch:
@Lukasa 
I've attempted a patch of this issue in PR #3598 
Would appreciate your input on the patch.

:smile:

bbamsch:
Will follow up with a fix for the dependency on httpbin.org. Please leave this open until then.

PR #3598, opened 2016-09-26
Issue #3597 - Avoid bytestring/str hodgepodge when resolving cookie from Request URL

bbamsch:
When resolving the request full URL, the Host cookie is used to override the URL's host, in Python 3 this causes a mix of str and bytestring objects when the Host cookie is set to a bytestring. Patch checks that ParsedResult is str type and host is non-str type before attempting to decode bytestring to str.

### DIFF
--- a/requests/_internal_utils.py
+++ b/requests/_internal_utils.py
@@ -1,0 +1,17 @@
+from .compat import is_py2, builtin_str
+
+
+def to_native_string(string, encoding='ascii'):
+    """Given a string object, regardless of type, returns a representation of
+    that string in the native string type, encoding and decoding where
+    necessary. This assumes ASCII unless told otherwise.
+    """
+    if isinstance(string, builtin_str):
+        out = string
+    else:
+        if is_py2:
+            out = string.encode(encoding)
+        else:
+            out = string.decode(encoding)
+
+    return out
--- a/requests/_internal_utils.py
+++ b/requests/_internal_utils.py
@@ -1,3 +1,13 @@
+# -*- coding: utf-8 -*-
+
+"""
+requests._internal_utils
+~~~~~~~~~~~~~~
+
+Provides utility functions that are consumed internally by Requests
+which depend on extremely few external helpers (such as compat)
+"""
+
 from .compat import is_py2, builtin_str
 
 
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -57,6 +57,12 @@ class MockRequest(object):
         # If they did set it, retrieve it and reconstruct the expected domain
         host = self._r.headers['Host']
         parsed = urlparse(self._r.url)
+
+        # If parsed url is str type, ensure that host is also str type
+        if isinstance(parsed.scheme, str) and not isinstance(host, str)\
+                and isinstance(host, bytes):
+            host = host.decode('ascii')
+
         # Reconstruct the URL as we expect it
         return urlunparse([
             parsed.scheme, host, parsed.path, parsed.params, parsed.query,
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -13,6 +13,8 @@ import copy
 import time
 import calendar
 import collections
+
+from . import utils
 from .compat import cookielib, urlparse, urlunparse, Morsel
 
 try:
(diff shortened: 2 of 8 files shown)


bbamsch:
Patch should address Issue #3597

requests/cookies.py
```diff
@@ -57,6 +57,12 @@ def get_full_url(self):
         # If they did set it, retrieve it and reconstruct the expected domain
         host = self._r.headers['Host']
         parsed = urlparse(self._r.url)
+
+        # If parsed url is str type, ensure that host is also str type
+        if isinstance(parsed.scheme, str) and not isinstance(host, str)\
+                and isinstance(host, bytes):
+            host = host.decode('ascii')
```
Lukasa: A quick dive into the CPython codebase shows that the current behaviour of the cookiejar module uses native strings (`str`). The same is true on Python 2.

That means that what we really want to do is just to call `to_native_str` on `host`. That is a Requests-specific function that lives in `utils`. That can skip a lot of this if statement.

While we're here, I should also note that `ascii` is probably a bad default encoding choice: we should at least consider defaulting to something a bit more sensible, like UTF-8.

bbamsch: Thanks for the tip. I missed the to_native_string utility function provided by requests.
I have made adjustments to this.

We may also want to consider making the import references in utils.py less explicit as to avoid issues with circular dependencies. utils.py imports some objects/functions from cookies explicitly with `from .cookies import RequestsCookieJar, cookiejar_from_dict`. These might be better achieved as a more generic import of the cookies module `from . import cookies`.

This would allow us to be more explicit in our imports within cookies.py, `from .utils import to_native_string` versus `from . import utils`, which is perhaps more beneficial?

requests/cookies.py
```diff
@@ -13,6 +13,8 @@
 import time
 import calendar
 import collections
+
+from . import utils
```
Lukasa: Yup, so this isn't going to work. This creates a circular import that breaks the code.

My recommendation is that we create an `_internal_utils.py` module that defines `to_native_string`. We then import from there in both this module and `utils.py` to expose the name in both places. How does that sound?

bbamsch: Odd, I didn't experience any issues during the test run. Perhaps it only appears if you import the cookies.py module directly.

Anyways, that sounds like a fine solution to me. I'll work on a patch later today to incorporate it into this pull request.

Thanks for the quick feedback.

bbamsch:
All appears to be in order 😃 

```
============================= test session starts =============================
platform win32 -- Python 3.5.2, pytest-2.8.7, py-1.4.31, pluggy-0.3.1
rootdir: C:\Users\Brian\PycharmProjects\requests, inifile:
plugins: httpbin-0.0.7, cov-2.2.1, mock-0.11.0
collected 381 items

tests\test_hooks.py ...
tests\test_lowlevel.py .........
tests\test_requests.py .............................................................................................................................................X.........X....................................................
tests\test_structures.py ....................
tests\test_testserver.py ...........
tests\test_utils.py ..s...................................................................................................................................

============== 378 passed, 1 skipped, 2 xpassed in 55.53 seconds ==============
```

requests/_internal_utils.py
```diff
@@ -0,0 +1,17 @@
+from .compat import is_py2, builtin_str
```
Lukasa, line 11: Let's add a few things to the top of this file. First, in the very first line, let's add a magic comment `# -*- coding: utf-8 -*-`, so that we can safely use unicode characters later on in the file.

Secondly, underneath that but before this import, let's add a docstring for the module that explains that this module contains internal utilities that need extremely few external helpers (nothing from any module other than `compat`).

requests/utils.py
```diff
@@ -20,10 +20,11 @@
 
 from . import __version__
 from . import certs
+# noinspection PyUnresolvedReferences
```
Lukasa: We shouldn't need this comment.

Lukasa: On the other hand, we _should_ put a comment above this import that says that we know that this is unused in this file, but we want it here for backward compat purposes.

bbamsch:
Went through your comments and reworked the module of internal references to to_native_string. 👍

tests/test_requests.py
```diff
@@ -858,6 +858,16 @@ def test_prepare_request_with_bytestring_url(self):
         prep = s.prepare_request(req)
         assert prep.url == "https://httpbin.org/"
 
+    def test_request_with_bytestring_host(self):
+        s = requests.Session()
+        resp = s.request(
+            'GET',
+            'http://httpbin.org/cookies/set?cookie=value',
```
nateprewitt, line 865: I know this is already merged, but we may want to amend this test to use the httpbin pytest fixture instead of calling httpbin.org directly. This test will fail if the user doesn't have an internet connection.

Lukasa: Oh yeah, I didn't even notice that. Can someone open a PR to fix that?

bbamsch: Wasn't aware of pytest-httpbin. I can follow up with a fix for this.

PR #3606, opened 2016-09-29
Issue #3597 - Rework Bytestring Host Test to use PyTest Httpbin

bbamsch:
Fixes dependency on internet connection for test added in PR #3598 for Issue #3597


Lukasa:
Looks good to me @bbamsch, thanks for fixing this up!

PR #3607, opened 2016-09-30
Remove error swallowing exception catching of AttributeError

frankier:
This exception catching is far too aggressive in my opinion. A lot of code could run in this and could throw AttributeError for any number of reasons. It's actually not clear to me what error it's trying to catch. It made it quite difficult for me to debug this problem: https://github.com/shazow/urllib3/pull/990

This PR removes it altogether since as far as I can see it is not needed in normal operation and can only serve to mask errors.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -747,18 +747,14 @@ class Response(object):
 
         if self._content is False:
             # Read the contents.
-            try:
-                if self._content_consumed:
-                    raise RuntimeError(
-                        'The content for this response was already consumed')
+            if self._content_consumed:
+                raise RuntimeError(
+                    'The content for this response was already consumed')
 
-                if self.status_code == 0:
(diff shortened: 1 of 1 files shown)


sigmavirus24:
I did some ~~dumpster~~ history diving and found https://github.com/kennethreitz/requests/commit/f7968b6797a50716a2f177ad23439a3c6bb9f683 and https://github.com/kennethreitz/requests/commit/296d8cc097b694ec8a9c4d8d7c00abed4755e5d5

The first is linked to https://github.com/kennethreitz/requests/issues/236 which indicates that at times `Response.raw` used to possibly be None. I don't think this could happen any longer (I think 98% of the time there's a valid urllib3 HTTPResponse) and we no longer access `self.raw` directly in this method, so I suspect this is probably safe.

PR #3620, opened 2016-10-14
Better support for internationalized domain names.

### LINKED ISSUE #3616: Consider vendoring the IDNA library.
As has been noted on the [Python Security SIG mailing list](https://mail.python.org/pipermail/security-sig/2016-October/000122.html), Python currently only includes the deprecated IDNA 2003 codec in the standard library. This is problematic, because IDNA 2003 is forbidden for some ccTLDs, such as `.de`: instead, the newer IDNA 2008 standard should be used. That standard is implemented in the PyPI package `idna`.

If we're going to support IDNA 2008, I would like it to not be optional: it leads to a fairly substantial and difficult to debug change, so it should be supported fully. To do that and keep in touch with Kenneth's wishes regarding dependencies, we'd have to vendor it.

I'd like contributors and packagers, particularly @eriol, @ralphbean, and @sigmavirus24 to weigh in with their thoughts here. What are your thoughts here? I'd also, if possible, like to hear from @kjd: while `idna` is licensed under a BSD-like license that should be broadly compatible with our own Apache 2.0 and so should present no legal blockers to vendoring, I'd still rather do it with the original author's permission than without it.

Lukasa:
Fixes #3616. This adds support for IDNA 2008 by vendoring the idna module, with the kind permission of @kjd.

For those keeping track, changes like this are another reason that Requests should stay out of the stdlib. ;) See also: [CPython issue 17305](https://bugs.python.org/issue17305).


vstinner:
"This adds support for IDNA 2008 by vendoring the idna module"

Why copying the whole module? Why not using a dependency? requests already uses requirements.txt.

Lukasa:
@haypo Without wanting to get too far into the weeds here, it remains the policy of the Requests project that we vendor all dependencies. If you'd like to discuss the amendment of that policy I recommend you email Kenneth, as all such decisions are his to make.

You'll also note that our requirements.txt are _development_ requirements, not our actual requirements for functioning.

vstinner:
"it remains the policy of the Requests project that we vendor all dependencies"

Oh ok. I misunderstood the requirements.txt file.

Lukasa:
@haypo No problem. =) This is an unusual policy, and it'll be interesting to see how much longer we keep it, but on a personal level I'd be totally happy to unvendor our deps.

requests/models.py
```diff
@@ -363,8 +364,8 @@ def prepare_url(self, url, params):
 
         # Only want to apply IDNA to the hostname
         try:
-            host = host.encode('idna').decode('utf-8')
-        except UnicodeError:
+            host = idna.encode(host).decode('utf-8')
```
jeremycline: Do we want to use `idna.encode(host, uts46=True)` here? It sounds like that option will enable a mapping between upper and lower case characters and other "safe" mappings, but I admit I just skimmed the document. Without it any uppercase letters here will result in an `InvalidURL`.

Lukasa: Are you sure? I just ran `idna.encode('GooGo.co')` and there were no problems there. Can you provide a sample hostname to encounter the issue?

jeremycline: I was testing with the sample hostname in the [idna docs](https://github.com/kjd/idna#compatibility-mapping-uts-46):

```
>>> import idna
>>> idna.encode(u'Königsgäßchen')
...
idna.core.InvalidCodepoint: Codepoint U+004B at position 1 of u'K\xf6nigsg\xe4\xdfchen' not allowed
>>> idna.encode(u'Königsgäßchen', uts46=True)
'xn--knigsgchen-b4a3dun'
```

Interestingly, it does work fine if there are not any characters that require encoding.

Lukasa: Aha, good spot. Yes, ok, that makes some sense. Let's enable the UTS46 mapping then, as there seems to be no reason not to.

tests/test_requests.py
```diff
@@ -1884,6 +1884,44 @@ def test_urllib3_pool_connection_closed(httpbin):
...
+            (u'http://ジェーピーニック.jp', u'http://xn--hckqz9bzb1cyrb.jp/'),
+            (
+                u'http://ジェーピーニック.jp'.encode('utf-8'),
+                u'http://xn--hckqz9bzb1cyrb.jp/'
+            ),
+            (u'http://straße.de/straße', u'http://xn--strae-oqa.de/stra%C3%9Fe'),
+            (
+                u'http://straße.de/straße'.encode('utf-8'),
+                u'http://xn--strae-oqa.de/stra%C3%9Fe'
+            ),
```
jeremycline, line 1910: It might be worth adding a test here to make sure uts46 is on.

`(u'http://Königsgäßchen.de/', u'http://xn--knigsgchen-b4a3dun/')` seems like a good candidate.

Lukasa:
Ok, addressed that issue.

jeremycline:
Cool, nothing else jumps out at me :shipit:

requests/packages/idna/codec.py
```diff
@@ -0,0 +1,118 @@
+from idna.core import encode, decode, alabel, ulabel, IDNAError
```
sigmavirus24: Will this work? I think we need to rewrite this import to be an absolute relative import.

Lukasa: Oh, good spot, it won't work.

Lukasa: Hrm, ideally this will be an automated rewrite. @kjd as a note for the future, relative imports might make life a bit easier here. ;)

In this case this should be simple enough I think: `sed "s/from idna/from /"` for each file in the `idna` directory should do the trick.

Lukasa: Ok, I've added the auto-rewrite to the `Makefile` and have rerun it to confirm that the imports are the only thing that get rewritten.

sigmavirus24: > In this case this should be simple enough I think: `sed "s/from idna/from /"` for each file in the idna directory should do the trick.

That will absolutely work for this sample. If @kjd, however, starts importing modules, e.g.,

``` py
from idna import codecs
```

Then that'll result in

``` py
from import codecs
```

So it won't quite work and we'll have to be extra careful with this vendor as we update it.

kjd:
I'll change the way imports are performed in the upstream.

sigmavirus24:
:cake:

PR #3625, opened 2016-10-17
fix issue when the file-like object raises an IOError with tell

mie00:


Lukasa:
Good spot!

Unfortunately, this fix has the effect of ensuring that files of this type pass transparently but send no data, as we've called `seek` on them to seek to the end. In this instance, probably the easiest fix is to move the `if` block to an `else` on the original `try`. That way, we know that the previous call to `tell()` didn't explode.

mie00:
Isn't is possible for seek itself to give an IOError?

for example:

``` python
import os
f = os.fdopen(0) # getting a file object for  the standard input
f.seek(10)
```

will give an IOError.

So I wonder if they can be handled with the same `except`, or they should be handled separately.

Lukasa:
@mie00 The question isn't "can `seek()` throw an `IOError`", it's "can `seek()` throw an `IOError` in a situation where `tell()` does not". I feel like the odds of that are pretty low. I'm inclined to want to worry about that when we know it's an issue.

mie00:
I tried to find a case, but I couldn't. The only thing that I was able to find is that `tell` giving `IOError` while `seek` gives `UnsupportedOperation` which is handled elsewhere (for example `requests.Response.raw`).

So, PR updated.

Lukasa:
Ok, rather than block on @sigmavirus24 having time to swing back to this, I'll merge this now. @sigmavirus24 can provide feedback whenever he gets time back. Thanks for the work @mie00!

PR #3627, opened 2016-10-18
remove RequestsCookieJar specific update call

nateprewitt:
This is a minor fix in the same vein as #3591. This function calls `update` on the CookieJar which only exists on `RequestsCookieJar` not the standard library `cookielib.CookieJar`. While this function will now be somewhat trivial, this will ensure it maintains backwards compatibility. It may be worth discussing removal in the future.

I have a test as well @Lukasa but it seems a bit overkill. I can include it if you'd like though.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -331,9 +331,7 @@ def add_dict_to_cookiejar(cj, cookie_dict):
     :rtype: CookieJar
     """
 
-    cj2 = cookiejar_from_dict(cookie_dict)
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hey @Lukasa, just pinging on this when you've got a moment. This should be the last problematic use of `update` on CookieJars in the codebase.

Lukasa:
Thanks @nateprewitt! :sparkles:

PR #3631, opened 2016-10-21
little modification  in consume_socket_content

lovedboy:


tests/testserver/server.py
```diff
@@ -8,17 +8,17 @@
 def consume_socket_content(sock, timeout=0.5):
     chunks = 65536
     content = b''
-    more_to_read = select.select([sock], [], [], timeout)[0]
 
-    while more_to_read:
-        new_content = sock.recv(chunks)
+    while 1:
```
Lukasa: Let's avoid `while 1`: can we have this be `while True` instead?

ls0f: @Lukasa I will update.

ls0f: @Lukasa updated.

PR #3632, opened 2016-10-21
default dispose_func, less condition check

lovedboy:


Lukasa:
Thanks for this! However, urllib3 is a separate project that Requests includes, in its entirety, with no patches. You'll need to provide this patch against urllib3 directly.

PR #3634, opened 2016-10-23
Keep ``verify`` setting when no CA_BUNDLE variable exists

jeremycline:
If the ``trust_env`` flag is set on a session and ``verify`` is ``True``
or ``None``, the environment is checked for ``CURL_CA_BUNDLE`` and
``REQUESTS_CA_BUNDLE``. Before this patch, if neither existed,
``verify`` would always be set to ``None`` rather than ``True`` when it
was originally ``True``.

I found this while working on a patch for https://github.com/kennethreitz/requests/issues/3633.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -674,7 +674,8 @@ class Session(SessionRedirectMixin):
             # with cURL.
             if verify is True or verify is None:
                 verify = (os.environ.get('REQUESTS_CA_BUNDLE') or
-                          os.environ.get('CURL_CA_BUNDLE'))
+                          os.environ.get('CURL_CA_BUNDLE') or
+                          verify)
 
         # Now we handle proxies.
         # Proxies need to be built up backwards. This is because None values
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1466,6 +1466,21 @@ class TestRequests:
         with pytest.raises(KeyError):
             proxies['http']
 
+    @pytest.fixture(autouse=True)
+    def test_merge_environment_settings_verify(self, monkeypatch):
+        """Assert CA environment settings are merged as expected when missing"""
+        session = requests.Session()
+        monkeypatch.delenv('CURL_CA_BUNDLE', raising=False)
+        monkeypatch.delenv('REQUESTS_CA_BUNDLE', raising=False)
+
+        assert session.trust_env is True
+        assert session.verify is True
+        assert 'REQUESTS_CA_BUNDLE' not in os.environ
+        assert 'CURL_CA_BUNDLE' not in os.environ
+        merged_settings = session.merge_environment_settings(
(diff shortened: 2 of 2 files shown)


tests/test_requests.py
```diff
@@ -1466,6 +1466,18 @@ def test_environment_comes_after_session(self, httpbin):
         with pytest.raises(KeyError):
             proxies['http']
 
+    def test_merge_environment_settings_verify(self):
+        """Assert CA environment settings are merged as expected when missing"""
+        session = requests.Session()
+        assert session.trust_env is True
+        assert session.verify is True
+        assert 'REQUESTS_CA_BUNDLE' not in os.environ
+        assert 'CURL_CA_BUNDLE' not in os.environ
```
Lukasa, line 1479: This test needs to be adjusted slightly to patch the environment to make these things true. Otherwise this test will fail on machines where the user has set these environment variables on purpose.

jeremycline:
@Lukasa I've adjusted the test.

tests/test_requests.py
```diff
@@ -1466,6 +1466,31 @@ def test_environment_comes_after_session(self, httpbin):
...
+    def test_merge_environment_settings_verify(self):
+        """Assert CA environment settings are merged as expected when missing"""
+        session = requests.Session()
+
+        old_curl_bundle = os.environ.get('CURL_CA_BUNDLE')
+        old_requests_bundle = os.environ.get('REQUESTS_CA_BUNDLE')
+        if old_curl_bundle:
+            del os.environ['CURL_CA_BUNDLE']
+        if old_requests_bundle:
+            del os.environ['REQUESTS_CA_BUNDLE']
```
Lukasa: Let's save ourself a bunch of time and code maintenance and just use [py.test's monkeypatch fixture](http://doc.pytest.org/en/latest/monkeypatch.html#method-reference-of-the-monkeypatch-fixture).

requests/sessions.py
```diff
@@ -674,7 +674,7 @@ def merge_environment_settings(self, url, proxies, stream, verify, cert):
             # with cURL.
             if verify is True or verify is None:
                 verify = (os.environ.get('REQUESTS_CA_BUNDLE') or
-                          os.environ.get('CURL_CA_BUNDLE'))
+                          os.environ.get('CURL_CA_BUNDLE') or verify)
```
Lukasa: While I'm being nitpicky...let's put `verify` on its own line? =D

jeremycline:
@Lukasa no worries. I haven't really worked with pytest and didn't see examples of patching in the existing tests so I got lazy :frowning_face:.

Lukasa:
Hurrah! No need to worry @jeremycline, this is what code review is for! I'm delighted to say that I'm now happy with this patch, and ready to merge. Thanks so much! :sparkles: :cake: :sparkles:

Issue #3633, opened 2016-10-22
HTTPS requests through proxies in proposed/3.0.0 aren't configured correctly


jeremycline:
In current master:

```
>>> import requests
>>> requests.__version__
'2.11.1'
>>> session = requests.Session()
>>> r = session.get('https://www.jcline.org/', verify=True, proxies={'http': 'http://vagrant:vagrant@localhost:3128', 'https': 'http://vagrant:vagrant@localhost:3128'})
>>> 
```

In current proposed/3.0.0:

```
>>> import requests
>>> requests.__version__
'3.0.0'
>>> session = requests.Session()
>>> r = session.get('https://www.jcline.org/', verify=True, proxies={'http': 'http://vagrant:vagrant@localhost:3128', 'https': 'http://vagrant:vagrant@localhost:3128'})
requests/packages/urllib3/connectionpool.py:838: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/security.html
  InsecureRequestWarning)
>>> 
```

This is a problem I introduced in https://github.com/kennethreitz/requests/pull/3109 :disappointed:. What happens right now is if a request is _not_ through a proxy and it's HTTPS, the urllib3 pool manager's `connection_pool_kw` are updated before requesting a new connection using [requests.adapters.HTTPAdapter._update_poolmanager_ssl_kw](https://github.com/kennethreitz/requests/blob/proposed/3.0.0/requests/adapters.py#L204). If it _is_ through a proxy, the keywords aren't updated and the request is made with the default settings for urllib3.

To me, the most appealing way to fix this is to add a keyword argument, `connection_kwargs` or something, to all the `urllib3.poolmanager.PoolManager.connection_from_*` methods that is either merged into `connection_pool_kw` or overrides them. That way `urllib3` can handle getting the connection pool with the new kwargs in a thread-safe manner. Currently, `requests` has to manage updating the keys and getting the new connection pool with a lock. It seems like that would be better in `urllib3`.

The other option is to patch up what's currently in `HTTPAdapter` so it handles updating the proxy manager or plain pool manager based on whether proxies are in use.

What do people think?

Lukasa:
I am open to having a urllib3 patch for this, at least in principle. What would it look like?

jeremycline:
This current patch I've got should give you the general idea, although I'm still playing with it a bit:

```
[...]

PR #3636, opened 2016-10-25
Update SSL options on proxy pool manager.

jortel:
In a previous PR, the SSL options are updated on the main pool manager in `get_connection()` but not when the selected pool manager is a proxy pool manager.  This PR fixes that.  

In addition, I renamed the `_pool_kw_lock` because it's used in  `get_connection()` to create a _critical section_ for both updating the selected pool manager _connection_pool_kw_  and calling `connection_from_url()` without being preempted.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -172,3 +172,4 @@ Patches and Suggestions
 - Maik Himstedt
 - Michael Hunsinger
 - Jeremy Cline <jcline@redhat.com> (`@jeremycline <https://github.com/jeremycline>`_)
+- Jeff Ortel <jortel@redhat.com> (`@jortel <https://github.com/jortel>`_)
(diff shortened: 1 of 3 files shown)


Lukasa:
Thanks for this @jortel!

I think that @jeremycline is planning a related PR that focuses on changes in urllib3 first. @jeremycline, how does this relate to your plan and what are your thoughts about this approach versus yours?

jeremycline:
Hey @jortel, I filed #3633 and attached a few proposed patches. Let me know what you think. I need to take the time to write a few tests before I submit a PR, but I will probably have time for that tomorrow night. 

I like making a change to urllib3 first because that way the issue of thread safety is all bundled into urllib3 rather than the responsibility of each user of urllib3.

jortel:
@jeremycline I like what you've proposed in #3633 and will close this PR in favor of that effort pending the outcome.

Issue #3649, opened 2016-10-26
The data= parameter barfs on all dictionary-like objects that don't subclass dict itself


gruns:
Example:

``` python
import requests
from collections import MutableMapping

class mydict(MutableMapping):
  def __init__(self, *args, **kwargs):
    self.d = dict(*args, **kwargs)
  def __iter__(self):
    return iter(self.d)
  def __getitem__(self, key):
    return self.d.get(key)
  def __setitem__(self, key, value):
    self.d[key] = value
  def __delitem__(self, key):
    del self.d[key]
  def __len__(self):
    return len(self.d)

d = {'key':'value'}
myd = mydict(d)

# Works.
print requests.post('http://httpbin.org/post', data=d).json().get('form')

# Should also work but doesn't.
print requests.post('http://httpbin.org/post', data=myd).json().get('form')
```

In models.py, the `is_stream` check

``` python
is_stream = all([
    hasattr(data, '__iter__'),
    not isinstance(data, (basestring, list, tuple, dict))
])
```

explicitly checks if the `data` parameter is an instance of `dict`, so only
instances of `dict` can be passed through `data=`. All other dictionary-like
objects later bomb with

``` python
TypeError: must be convertible to a buffer, not mydict
```

Is it intentional that all dictionary-like objects not explicitly `dict`
subclasses are rejected? I can't fathom why.

Lukasa:
Nope, probably we should check for `Mapping` instead. Would you like to provide a PR to fix this up?

sigmavirus24:
We had this at some point. Why did we get rid of this?

Lukasa:
I'm honestly not sure. =(

lejkos:
Hello,
I am a seasoned Python developer and I want to make my first steps 
as an open source contributor. Can you assign this bug to me?

Lukasa:
@gardiac2002 Unfortunately, GitHub doesn't allow assigning issues to anyone other than contributors or the person who opened the issue. However, no-one on the team is currently working on this, so you are welcome to work on it and open a pull request.

lejkos:
@Lukasa thank you, I am going to take a look at the issue :)

lejkos:
I created a pull request for the issue: https://github.com/kennethreitz/requests/pull/3652
All tests ran successful from Python2.6 to Python3.5

nateprewitt:
It looks like this is now resolved thanks to @gardiac2002's work in #3652.

PR #3652, opened 2016-10-27
The data= parameter barfs on all dictionary-like objects that don't subclass dict itself #3649

gardiac2002:
requests.post does not accept custom mappings as
data parameters. I changed the type check to Mapping.

Original bug report can be found at #3649:
The data= parameter barfs on all dictionary-like objects that don't subclass dict itself #3649


kennethreitz:
✨🍰✨

PR #3655, opened 2016-10-29
Rewind readable body before POST redirect

nateprewitt:
This addresses the issue (#3079) of Requests hanging when a POST with a file-like body encounters a temporary redirect. This currently has the unmerged tests from #3536, so I'll need to remove those if we reach a point where we're ready to merge this.

This approach has a couple of things that I'm not especially happy with but couldn't come up with a nicer alternative. I've added a few outstanding questions below.

1.) I feel like `body_position` should probably be `_body_position`, but we need to use it in the `Session` module. Making it a private var would require an additional getter method for `PreparedRequest` to expose the data to `Session`, but I wanted to run that by before adding more  code. It just seems problematic to provide any encouragement to manually set `body_position` since it will lead to fairly unintuitive behaviour.

2.) As I noted in the comment [here](https://github.com/nateprewitt/requests/commit/b84d71461e238b123a39ce91a48a99e3eb03717c#commitcomment-19619276), this still allows for the possibility that a request with a file-like object, that has tell but no seek, to hang. I feel like we should raise an `IOError` if we can't rewind, at least allowing the user to manually recreate the body and POST it to the new URL. Does that seem reasonable?

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -50,6 +50,7 @@ if is_py2:
     str = unicode
     basestring = basestring
     numeric_types = (int, long, float)
+    integer_types = (int, long)
 
 elif is_py3:
     from urllib.parse import urlparse, urlunparse, urljoin, urlsplit, urlencode, quote, unquote, quote_plus, unquote_plus, urldefrag
@@ -64,3 +65,4 @@ elif is_py3:
     bytes = bytes
     basestring = (str, bytes)
     numeric_types = (int, float)
+    integer_types = (int,)
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -100,6 +100,8 @@ class StreamConsumedError(RequestException, TypeError):
 class RetryError(RequestException):
     """Custom retries logic failed"""
 
+class UnrewindableBodyError(RequestException):
+    """Requests encountered an error when trying to rewind a body"""
 
 # Warnings
 
--- a/requests/models.py
+++ b/requests/models.py
@@ -291,6 +291,8 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         self.body = None
         #: dictionary of callback hooks, for internal usage.
         self.hooks = default_hooks()
+        #: integer denoting starting position of a readable file-like body.
+        self._body_position = None
 
     def prepare(self, method=None, url=None, headers=None, files=None,
         data=None, params=None, auth=None, cookies=None, hooks=None, json=None):
@@ -320,6 +322,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         p._cookies = _copy_cookie_jar(self._cookies)
         p.body = self.body
         p.hooks = self.hooks
+        p._body_position = self._body_position
         return p
 
     def prepare_method(self, method):
@@ -446,6 +449,17 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
 
         if is_stream:
             body = data
+
+            if getattr(body, 'tell', None) is not None:
+                # Record the current file position before reading.
+                # This will allow us to rewind a file in the event
+                # of a redirect.
+                try:
+                    self._body_position = body.tell()
+                except (IOError, OSError):
+                    # This differentiates from None, allowing us to catch
+                    # a failed `tell()` later when trying to rewind the body
+                    self._body_position = object()
 
             if files:
                 raise NotImplementedError('Streamed bodies and files are mutually exclusive.')
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -28,7 +28,7 @@ from .adapters import HTTPAdapter
 
 from .utils import (
     requote_uri, get_environ_proxies, get_netrc_auth, should_bypass_proxies,
-    get_auth_from_url
+    get_auth_from_url, rewind_body
 )
 
 from .status_codes import codes
@@ -163,6 +163,18 @@ class SessionRedirectMixin(object):
             # Rebuild auth and proxy information.
             proxies = self.rebuild_proxies(prepared_request, proxies)
             self.rebuild_auth(prepared_request, resp)
+
+            # A failed tell() sets `_body_position` to `object()`. This non-None
+            # value ensures `rewindable` will be True, allowing us to raise an
+            # UnrewindableBodyError, instead of hanging the connection.
+            rewindable = (
+                prepared_request._body_position is not None and
+                ('Content-Length' in headers or 'Transfer-Encoding' in headers)
+            )
+
+            # Attempt to rewind consumed file-like object.
+            if rewindable:
+                rewind_body(prepared_request)
 
             # Override the original request.
             req = prepared_request
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -165,6 +165,22 @@ class TestRequests:
         assert r.history[0].status_code == 302
         assert r.history[0].is_redirect
 
+    def test_HTTP_307_ALLOW_REDIRECT_POST(self, httpbin):
+        parts = urlparse(httpbin('post'))
+        url = "HTTP://" + parts.netloc + parts.path
+        r = requests.post(httpbin('redirect-to'), data='test', params={'url': url, 'status_code': 307})
+        assert r.status_code == 200
+        assert r.history[0].status_code == 307
+        assert r.history[0].is_redirect
+
+    def test_HTTP_307_ALLOW_REDIRECT_POST_WITH_SEEKABLE(self, httpbin):
+        parts = urlparse(httpbin('post'))
+        url = "HTTP://" + parts.netloc + parts.path
+        r = requests.post(httpbin('redirect-to'), data=io.BytesIO(b'test'), params={'url': url, 'status_code': 307})
+        assert r.status_code == 200
+        assert r.history[0].status_code == 307
+        assert r.history[0].is_redirect
+
     def test_HTTP_302_TOO_MANY_REDIRECTS(self, httpbin):
         try:
             requests.get(httpbin('relative-redirect', '50'))
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -24,7 +24,7 @@ from requests.cookies import (
 from requests.exceptions import (
     ConnectionError, ConnectTimeout, InvalidSchema, InvalidURL,
     MissingSchema, ReadTimeout, Timeout, RetryError, TooManyRedirects,
-    ProxyError, InvalidHeader)
+    ProxyError, InvalidHeader, UnrewindableBodyError)
 from requests.models import PreparedRequest
 from requests.structures import CaseInsensitiveDict
 from requests.sessions import SessionRedirectMixin
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1386,6 +1386,107 @@ class TestRequests:
         r3 = next(rg)
         assert not r3.is_redirect
 
+    def test_prepare_body_position_non_stream(self):
+        data = b'the data'
+        s = requests.Session()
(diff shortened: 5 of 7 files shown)


requests/models.py
```diff
@@ -291,6 +291,9 @@ def __init__(self):
         self.body = None
         #: dictionary of callback hooks, for internal usage.
         self.hooks = default_hooks()
+        #: integer denoting position of readable file-like bodies
+        self.body_position = None
```
Lukasa: Feel free to make this private: Session can reach in and grab it if it wants.

requests/sessions.py
```diff
@@ -164,6 +164,14 @@ def resolve_redirects(self, resp, req, stream=False, timeout=None,
             proxies = self.rebuild_proxies(prepared_request, proxies)
             self.rebuild_auth(prepared_request, resp)
 
+            # Attempt to rewind consumed file-like object
+            rewindable = (req.body_position is not None and prepared_request.body is not None)
+            if rewindable and getattr(prepared_request.body, 'seek', None) is not None:
+                try:
+                    prepared_request.body.seek(req.body_position)
+                except (IOError, OSError):
+                    pass # Probably raise an exception here so we don't hang on send?
```
Lukasa: Let's begin by splitting this out to a method. We have a tendency in this project to shove too many things into an inline function. Let's fight that.

Lukasa: Yeah, I think we should throw an exception here.

requests/models.py
```diff
@@ -447,6 +451,15 @@ def prepare_body(self, data, files, json=None):
...
             body = data
 
+            if getattr(body, 'tell', None) is not None:
+                # Record the current file position before reading.
+                # This will allow us to rewind a file in the event
+                # of a redirect.
+                try:
+                    self._body_position = body.tell()
+                except (IOError, OSError):
+                    pass
```
nateprewitt: @Lukasa, thoughts on setting `_body_position` to a sentinel object here? We currently have a gap where `_body_position` will remain `None` when `tell` fails and won't trigger the exception in `rewind_body`. It seems hasty to raise the exception here since we aren't sure a redirect will ever occur.

requests/exceptions.py
```diff
@@ -100,6 +100,8 @@ class StreamConsumedError(RequestException, TypeError):
 class RetryError(RequestException):
     """Custom retries logic failed"""
 
+class RedirectError(RequestException):
```
nateprewitt: `RedirectError` seems a little nebulous, the current occurrences are more of an `IOError`, but I'm not sure what we'd want to call a Requests version of that. Open to suggestions.

Lukasa: Eh, let's just be specific and call it an "UnrewindableBodyError". Specificity never hurt anyone.

requests/models.py
Lukasa: I think a sentinel object is a good idea.

tests/test_requests.py
```diff
@@ -1386,6 +1401,64 @@ def test_manual_redirect_with_partial_body_read(self, httpbin):
         r3 = next(rg)
         assert not r3.is_redirect
 
+    def test_rewind_body(self):
+        data = io.BytesIO(b'the data')
+        s = requests.Session()
+        prep = requests.Request('GET', 'http://example.com', data=data).prepare()
+        prep.body.read()
```
Lukasa: Can we assert the content of the read here too?

tests/test_requests.py
```diff
@@ -1386,6 +1401,64 @@ def test_manual_redirect_with_partial_body_read(self, httpbin):
...
+        # rewind it back
+        s.rewind_body(prep)
+        assert prep.body.read() == b'the data'
+
+    def test_rewind_partially_read_body(self):
+        data = io.BytesIO(b'the data')
+        s = requests.Session()
+        data.read(4)  # read some data
+        prep = requests.Request('GET', 'http://example.com', data=data).prepare()
+        prep.body.read()
```
Lukasa: Same deal, let's assert the content of the read.

tests/test_requests.py
```diff
@@ -1386,6 +1401,64 @@ def test_manual_redirect_with_partial_body_read(self, httpbin):
...
+
+        data = BadFileObj('the data')
+        s = requests.Session()
+        prep = requests.Request('GET', 'http://example.com', data=data).prepare()
+
+        with pytest.raises(RedirectError) as e:
+            s.rewind_body(prep)
+
+        assert 'error occured when rewinding request body' in str(e)
+
```
Lukasa, line 1483: Presumably we should have a test here for an object with a `seek()` method but no working `tell()` method?

nateprewitt:
Alright @Lukasa, I think everything in that last round should be addressed. A couple of notes on some decisions I made in these tweaks.

Python 2.7 decided it needs to return a long from `tell` instead of an int like every other version. I've adjusted for that by casting it to int because we can't test for long in Py3.

I didn't declare the "sentinel object" as a global because I couldn't come up with a scenario where it was equivocal to anything in the sessions module. Simply setting it to object accomplishes what we were looking for but perhaps there's a caveat I'm missing here?

~~**Tests.** I went a little crazy. I have a few "lifecycle" tests, but found a snag when adding `test_rewind_body_failed_tell`. I realized things were failing in different ways (`_body_position` was never being set with the custom test objects) but the error messages produced were still "correct". In order to prevent accidental regressions, I add specific tests for setting `_body_position` and failure conditions in `rewind_body` by manually setting `_body_position`. Let me know if you want these collapsed back.~~

Edit: I went ahead and expanded the test objects to simplify the testing.

requests/models.py
```diff
@@ -447,6 +451,18 @@ def prepare_body(self, data, files, json=None):
         if is_stream:
             body = data
 
+            if getattr(body, 'tell', None) is not None:
+                # Record the current file position before reading.
+                # This will allow us to rewind a file in the event
+                # of a redirect.
+                try:
+                    # cast value to int for Python 2.7 which returns a long
+                    self._body_position = int(body.tell())
```
Lukasa: The fact that Python 2.7 returns a long is, I suspect, meaningful. In particular, Python's 2.7's `int` is a 32/64-bit number, whereas the long is unbounded (it's a bignum).

For this reason, we should probably drop the cast and instead define an `integer_types` field in `compat.py` that is `(int, long)` on Python 2 and `int` on Python 3.

requests/sessions.py
```diff
@@ -267,6 +277,19 @@ def rebuild_method(self, prepared_request, response):
...
+        """
+        body_seek = getattr(prepared_request.body, 'seek', None)
+        if body_seek is not None and isinstance(prepared_request._body_position, int):
+            try:
+                body_seek(prepared_request._body_position)
+            except (IOError, OSError):
+                raise UnrewindableBodyError("An error occured when rewinding request "
+                                            "body for redirect.")
+        else:
+            raise UnrewindableBodyError("Unable to rewind request body for redirect.")
```
Lukasa: The variable `self` is unused in this method, which is usually a good sign that it's a function masquerading as a method. I suggest moving this function to utils. =)

requests/sessions.py
```diff
@@ -267,6 +277,19 @@ def rebuild_method(self, prepared_request, response):
 
         prepared_request.method = method
 
+    def rewind_body(self, prepared_request):
+        """Move file pointer back to it's recorded starting position
```
Lukasa: Nit: "its".

tests/test_requests.py
```diff
@@ -165,6 +165,21 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
         assert r.history[0].status_code == 302
         assert r.history[0].is_redirect
 
+    def test_HTTP_307_ALLOW_REDIRECT_POST(self, httpbin):
+        r = requests.post(httpbin('redirect-to'), data='test', params={'url': 'post', 'status_code': 307})
+        assert r.status_code == 200
+        assert r.history[0].status_code == 307
+        assert r.history[0].is_redirect
```
Lukasa, line 172: This test should probably confirm that httpbin has seen the correct data.

tests/test_requests.py
```diff
@@ -165,6 +165,21 @@ def test_HTTP_302_ALLOW_REDIRECT_GET(self, httpbin):
...
+
+    def test_HTTP_307_ALLOW_REDIRECT_POST_WITH_SEEKABLE(self, httpbin):
+        byte_str = b'test'
+        r = requests.post(httpbin('redirect-to'), data=io.BytesIO(byte_str), params={'url': 'post', 'status_code': 307})
+        content = json.loads(r.content.decode('utf-8'))
```
Lukasa: Better to use `r.json()`.

tests/test_requests.py
```diff
@@ -1386,6 +1401,107 @@ def test_manual_redirect_with_partial_body_read(self, httpbin):
         r3 = next(rg)
         assert not r3.is_redirect
 
+    def test_prepare_body_position_non_stream(self):
+        data = b'the data'
+        s = requests.Session()
+        prep = requests.Request('GET', 'http://example.com', data=data).prepare()
+        assert prep._body_position == None
```
Lukasa: `is`

nateprewitt:
Here's another round of changes @Lukasa.

I moved the `rewind_body` to a util function but when I started looking at moving tests over to `test_utils` they're going to bring a lot of import overhead with them. I can chop them up into individual `_body_position` and `rewind_body` tests and divide them up accordingly. Otherwise, we can leave them where they are.

requests/models.py
```diff
@@ -447,6 +451,17 @@ def prepare_body(self, data, files, json=None):
         if is_stream:
             body = data
 
+            if getattr(body, 'tell', None) is not None:
+                # Record the current file position before reading.
+                # This will allow us to rewind a file in the event
+                # of a redirect.
+                try:
+                    self._body_position = body.tell()
+                except (IOError, OSError, ValueError):
```
nateprewitt: We don't need this ValueError anymore, I'll yank that on the next pass.

requests/compat.py
```diff
@@ -64,3 +65,4 @@
     bytes = bytes
     basestring = (str, bytes)
     numeric_types = (int, float)
+    integer_types = (int)
```
Lukasa: This is a common tricky area, but `(int)` doesn't actually make a tuple. In this case, you'd want `(int,)`.

requests/sessions.py
```diff
@@ -164,6 +164,14 @@ def resolve_redirects(self, resp, req, stream=False, timeout=None,
             proxies = self.rebuild_proxies(prepared_request, proxies)
             self.rebuild_auth(prepared_request, resp)
 
+            # Attempt to rewind consumed file-like object
```
Lukasa: I think we need to expressly extend this comment to talk about why we want to allow the `object()` from `models.py` to mark the body `rewindable`.

nateprewitt:
Alright, ValueError removed, tuple properly instantiated, and comment added. I also squashed most of the commits down. There's an unnecessary space I need to pull from beneath the `_body_position` declaration but I'll do that when I remove bc199e7 before merging. That is unless we have another round :)

Lukasa:
Want to do your final tidying up now, @nateprewitt? I don't think I'm going to have any further feedback at this time.

nateprewitt:
Ok, I think things are good to go from my end.

nateprewitt:
Just a reminder, we'll need to merge #3536 before this is merged.

Lukasa:
Ah, sorry, I didn't realise that's what was going on there. Can you merge the branch from #3536 into this branch and then I'll merge them both as one? I'd rather not have a weird interim master state where we have deliberately failing tests on the master branch.

nateprewitt:
Ok, everything is merged in and this should be ready to go. It will now supersede #3536.

Lukasa:
Thanks for this @nateprewitt!

PR #3660, opened 2016-10-31
[httpAdapter] allow empty password in proxy credentials

### LINKED ISSUE #3659: URL proxy auth with empty passwords doesn't emit auth header.
I'm using a proxy that requires authentication to send request that receives 302 response with Location header. I would like python.requests to follow this redirect and make request via proxy with specified credentials. But it seems like this doesn't happen, if I provide credentials in HTTPProxyAuth they will work ok for 200 responses but will fail for 302. See below code sample:

```python

import requests
from requests.auth import HTTPProxyAuth

sess = requests.Session()
url1 = 'http://httpbin.org/'
url2 = 'http://httpbin.org/redirect/2'
auth = HTTPProxyAuth('frank', 'hunter2')
proxies = {
    "http": "http://localhost:9000"
}
response1 = sess.get(url1, proxies=proxies, auth=auth)
response1.raise_for_status()
response2 = sess.get(url2, proxies=proxies, auth=auth)
response2.raise_for_status()
```
Now launch MITM proxy on localhost

```
> mitmproxy -p 9000 --singleuser=frank:hunter2
```

This fails with 407 for me, and proxy logs only two requests

```
    response2.raise_for_status()
  File "----------", line 862, in raise_for_status
requests.exceptions.HTTPError: 407 Client Error: Proxy Authentication Required for url: http://httpbin.org/relative-redirect/1
```

```
>> GET http://httpbin.org/
       ← 200 text/html 11.87kB 3.57MB/s
   GET http://httpbin.org/redirect/2
       ← 302 text/html 247B 76.59kB/s

```
it does not log request to `Location`. 

I see that putting credentials in proxies dictionary somehow fixes this issue when I use MITM proxy but it doesn't fix it for my production proxy (can't share code or proxy details here, need to check closer why it doesn't work for my proxy). I guess some details in setup of proxies might vary.

Is this a bug? I see some issues for proxy auth but they are mostly about HTTPS, not sure if someone reported this thing I describe here. Should this be fixed?

EDIT:

It looks like this always fails if proxy password is empty string.

change auth to 

```python
auth = HTTPProxyAuth('frank', '')

proxies = {
    "http": "http://frank:@localhost:9000"
}
```

will now always fail on redirect.

```python
auth = HTTPProxyAuth('frank', 'hunter2')
proxies = {
    "http": "http://frank:hunter2@localhost:9000"
}
```
works fine on redirects, but seems somewhat duplicated.

I noticed this on Ubuntu 14.04, requests 2.11.1, python 2.7.6, mitmproxy 0.10.1

pawelmhm:
fixes #3659

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -366,7 +366,7 @@ class HTTPAdapter(BaseAdapter):
         headers = {}
         username, password = get_auth_from_url(proxy)
 
-        if username and password:
+        if username:
             headers['Proxy-Authorization'] = _basic_auth_str(username,
                                                              password)
 
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1472,6 +1472,16 @@ class TestRequests:
         proxies['one'].clear.assert_called_once_with()
         proxies['two'].clear.assert_called_once_with()
 
+    def test_proxy_auth(self, httpbin):
+        adapter = HTTPAdapter()
+        headers = adapter.proxy_headers("http://user:pass@httpbin.org")
+        assert headers == {'Proxy-Authorization': 'Basic dXNlcjpwYXNz'}
+
+    def test_proxy_auth_empty_pass(self, httpbin):
+        adapter = HTTPAdapter()
+        headers = adapter.proxy_headers("http://user:@httpbin.org")
+        assert headers == {'Proxy-Authorization': 'Basic dXNlcjo='}
+
     def test_response_json_when_content_is_None(self, httpbin):
         r = requests.get(httpbin('/status/204'))
         # Make sure r.content is None


pawelmhm:
Some things that could be improved in future:
- add some unit test using real proxy, e.g. MITM-proxy, something along the lines of Scrapy: https://github.com/scrapy/scrapy/blob/129421c7e31b89b9b0f9c5f7d8ae59e47df36091/tests/test_proxy_connect.py#L35 where there is fake proxy started and all tests for proxy requests go via this fake proxy. This would be nice to have to test somethings with real proxies (instead of doing unit test of some specific adapter method).

Lukasa:
Cool, so this looks really good. You'll note we have some tests with proxies in `test_lowlevel.py`, so if you feel like adding more testing in this area I'll happily merge it. However, for this specific issue, I think this test is sufficient.

Thanks @pawelmhm! :sparkles: :cake: :sparkles:

PR #3663, opened 2016-11-02
fix path for cookiejar_from_dict in api.rst

nateprewitt:
Right now the `cookiejar_from_dict` path provided to autofunction in api.rst is incorrect. It happens to work though because Sphinx is smart enough to correct the source url for us. This will make the display string consistent with the correct path.


Lukasa:

Issue #2979, opened 2016-01-24
JSON content type for PUT method


Neraste:
According to the [doc](http://docs.python-requests.org/en/latest/api/), you can specify to send a request with JSON content type for POST method. But there is nothing like this for PUT or PATCH methods.

Is there a reason? Can this be implemented?

sigmavirus24:
If you look at the documentation for [`requests.put`](http://docs.python-requests.org/en/latest/api/#requests.put) and [`requests.patch`](http://docs.python-requests.org/en/latest/api/#requests.patch) they take `**kwargs` that are passed to [`requests.request`](http://docs.python-requests.org/en/latest/api/#requests.request).

This means that `json=` is already a valid parameter for those methods.

PR #3664, opened 2016-11-02
json parameter added to patch and put methods

hootnot:
referring to #2979 ...

this is answered that this can be done with **kwargs, which is true. But wouldn't that would go for **post** too ?
It would make consistent use of parameters possible if **put** and **patch** would support the _json_ parameter also, so then there is no need for additional logic to distinguish between the different methods, take for example:

```python
   ...
   self.client = requests.Session()  
   ...
   request_args = {}
   if method == 'get':
        request_args['params'] = params
   elif hasattr(endpoint, "data") and endpoint.data:
       request_args['json'] = endpoint.data
   ....
   func = getattr(self.client, method)
   response = func(url, stream=stream, headers=headers, **request_args)
   ...

```
post, put and patch can all be handled the same this way.

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -110,30 +110,32 @@ def post(url, data=None, json=None, **kwargs):
     return request('post', url, data=data, json=json, **kwargs)
 
 
-def put(url, data=None, **kwargs):
+def put(url, data=None, json=None, **kwargs):
     """Sends a PUT request.
 
     :param url: URL for the new :class:`Request` object.
     :param data: (optional) Dictionary, bytes, or file-like object to send in the body of the :class:`Request`.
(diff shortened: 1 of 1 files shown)


sigmavirus24:
>  if put and patch would support the json parameter also

They already do.

hootnot:
Yup, you are right. 

Somehow reading the docs one gets a direction because the **post** method explicit names  the **json** parameter in it's documentation. While the **put** and **post** don't, they indirectly refer to the params of request.

Lukasa:
@hootnot I think in this case this should be considered a documentation issue, rather than explicitly naming the parameters. Mind opening a docs PR?

PR #3667, opened 2016-11-08
Upgrade urllib3 to version 1.19

frankier:


Lukasa:
Thanks for opening this, but this is usually done by maintainers as part of the release and maintenance process. In particular, it requires some care around handling the packaging side of things. Sorry, but that means we're not going to merge this.

Thanks for the work though!

PR #3673, opened 2016-11-12
Enable unicode in basic http auth

klimenko:
Fixed the issue with unicode characters in basic http auth

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -28,14 +28,14 @@ CONTENT_TYPE_MULTI_PART = 'multipart/form-data'
 def _basic_auth_str(username, password):
     """Returns a Basic Auth string."""
     
-    if isinstance(username, bytes):
-        username = username.decode('latin1')
+    if isinstance(username, str):
+        username = username.encode('latin1')
 
-    if isinstance(password, bytes):
-        password = password.decode('latin1')
+    if isinstance(password, str):
+        password = password.encode('latin1')
 
     authstr = 'Basic ' + to_native_string(
-        b64encode(('%s:%s' % (username, password)).encode('latin1')).strip()
+        b64encode(b':'.join((username, password))).strip()
     )
 
     return authstr
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -15,7 +15,7 @@ import threading
 
 from base64 import b64encode
 
-from .compat import urlparse, str
+from .compat import urlparse, str, bytes
 from .cookies import extract_cookies_to_jar
 from ._internal_utils import to_native_string
 from .utils import parse_dict_header
@@ -27,6 +27,12 @@ CONTENT_TYPE_MULTI_PART = 'multipart/form-data'
 
 def _basic_auth_str(username, password):
     """Returns a Basic Auth string."""
+    
+    if isinstance(username, bytes):
+        username = username.decode('latin1')
+
+    if isinstance(password, bytes):
+        password = password.decode('latin1')
 
     authstr = 'Basic ' + to_native_string(
         b64encode(('%s:%s' % (username, password)).encode('latin1')).strip()
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -15,7 +15,7 @@ import threading
 
 from base64 import b64encode
 
-from .compat import urlparse, str, bytes
+from .compat import urlparse, str
 from .cookies import extract_cookies_to_jar
 from ._internal_utils import to_native_string
 from .utils import parse_dict_header
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -465,14 +465,14 @@ class TestRequests:
             ('user', 'pass'),
             (u'имя'.encode('utf-8'), u'пароль'.encode('utf-8')),
         ))
-    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin, username, password, auth_str):
+    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin, username, password):
         auth = (username, password)
         url = httpbin('get')
-        
-        r = Request('GET', url, auth=auth)
+
+        r = requests.Request('GET', url, auth=auth)
         p = r.prepare()
 
-        assert p['Authorization'] == _basic_auth_str(username, password)
+        assert p.headers['Authorization'] == _basic_auth_str(username, password)
         
         r = requests.get(url, auth=auth)
         assert r.status_code == 200
@@ -1582,7 +1582,7 @@ class TestRequests:
         'username, password, auth_str', (
             ('test', 'test', 'Basic dGVzdDp0ZXN0'),
             (u'имя'.encode('utf-8'), u'пароль'.encode('utf-8'), 'Basic 0LjQvNGPOtC/0LDRgNC+0LvRjA=='),
-        ))    
+        ))
     def test_basic_auth_str_is_always_native(self, username, password, auth_str):
         s = _basic_auth_str(username, password)
         assert isinstance(s, builtin_str)
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -460,12 +460,27 @@ class TestRequests:
         r = requests.put(httpbin('put'))
         assert r.status_code == 200
 
+    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin):
+        auth = ('user', 'pass')
+        url = httpbin('basic-auth', 'user', 'pass')
+
+        r = requests.get(url, auth=auth)
+        assert r.status_code == 200
+
+        r = requests.get(url)
+        assert r.status_code == 401
+
+        s = requests.session()
+        s.auth = auth
+        r = s.get(url)
+        assert r.status_code == 200
+
     @pytest.mark.parametrize(
         'username, password', (
             ('user', 'pass'),
             (u'имя'.encode('utf-8'), u'пароль'.encode('utf-8')),
         ))
-    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin, username, password):
+    def test_set_basicauth(self, httpbin, username, password):
         auth = (username, password)
         url = httpbin('get')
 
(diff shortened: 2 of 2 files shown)


requests/auth.py
```diff
@@ -27,6 +27,12 @@
 
 def _basic_auth_str(username, password):
     """Returns a Basic Auth string."""
+    
+    if isinstance(username, bytes):
+        username = username.decode('latin1')
+
+    if isinstance(password, bytes):
+        password = password.decode('latin1')
```
Lukasa: Let's not do it this way around. Given that we need to re-encode immediately (`b64encode` only works on `bytes` objects), we should instead check for _unicode_ strings and encode them using "latin1", then change the `authstr` line to do `b"%s:%s" % (username, password)` where `username` and `password` are both `bytes`.

ghost: Sorry, I don't sure, I understood you correctly. Do you talk about this?

``` python
def _basic_auth_str(username, password):
    """Returns a Basic Auth string."""

    if isinstance(username, str):
        username = username.encode('latin1')

    if isinstance(password, str):
        password = password.encode('latin1')

    authstr = 'Basic ' + to_native_string(
        b64encode(b'%s:%s' % (username, password)).strip()
    )

    return authstr
```

Lukasa: Yeah, I did mean that, though I note that we will have to do a `b':'.join` instead of string formatting, so we'll have to do this:

``` python
def _basic_auth_str(username, password):
    """Returns a Basic Auth string."""

    if isinstance(username, str):
        username = username.encode('latin1')

    if isinstance(password, str):
        password = password.encode('latin1')

    authstr = 'Basic ' + to_native_string(
        b64encode(b':'.join((username, password))).strip()
    )

    return authstr
```

ghost: I just committed this and one test. Please review.

tests/test_requests.py
```diff
@@ -460,9 +460,14 @@ def test_HTTP_200_OK_PUT(self, httpbin):
         r = requests.put(httpbin('put'))
         assert r.status_code == 200
 
-    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin):
-        auth = ('user', 'pass')
-        url = httpbin('basic-auth', 'user', 'pass')
+    @pytest.mark.parametrize(
+        'username, password', (
+            ('user', 'pass'),
+            ('имя'.encode('utf-8'), 'пароль'.encode('utf-8')),
```
Lukasa: It should be safe to use `u''` strings here, which saves the encode call.

ghost: Oh, sorry, I forgot about this.
Fixed.

tests/test_requests.py
```diff
@@ -460,9 +460,14 @@ def test_HTTP_200_OK_PUT(self, httpbin):
...
-        auth = ('user', 'pass')
-        url = httpbin('basic-auth', 'user', 'pass')
+    @pytest.mark.parametrize(
+        'username, password', (
+            ('user', 'pass'),
+            (u'имя'.encode('utf-8'), u'пароль'.encode('utf-8')),
+        ))
+    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin, username, password):
+        auth = (username, password)
+        url = httpbin('basic-auth', username, password)
```
Lukasa: Ah, sorry, this test fails because of the way the URL has been constructed here. The httpbin url fixture doesn't support bytes here.

Instead of doing the test this way, let's remove httpbin entirely. We should be able to check what the correct value is by using the prepared request flow:

``` python
r = Request('GET', httpbin('get'), auth=(username, password))`
p = r.prepare()

assert p['Authorization'] == "whatever the auth string should be"
```

Want to change the test to that form?

ghost: Yeah, sure. I just fixed this.
I think, that hardcoding is the more right way in this case (vs using _basic_auth_str()).

ghost: Pls, wait, I want to add one more test )

ghost: Ok, I removed hardcode from this test (from previous commit) and added test for _basic_auth_str with bytes. Review, please.

tests/test_requests.py
```diff
@@ -460,10 +460,20 @@ def test_HTTP_200_OK_PUT(self, httpbin):
...
+    @pytest.mark.parametrize(
+        'username, password', (
+            ('user', 'pass'),
+            (u'имя'.encode('utf-8'), u'пароль'.encode('utf-8')),
+        ))
+    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin, username, password, auth_str):
+        auth = (username, password)
+        url = httpbin('get')
+        
+        r = Request('GET', url, auth=auth)
```
Lukasa: Sorry, this needs to be `requests.Request`.

ghost: Fixed.
I found https://github.com/kennethreitz/requests/blob/master/docs/dev/todo.rst#development-dependencies :)

And now I can see this error :(

```
py.test tests/test_requests.py
================================================= test session starts ==================================================
platform darwin -- Python 3.5.2, pytest-2.8.7, py-1.4.31, pluggy-0.3.1
rootdir: requests, inifile:
plugins: cov-2.2.1, httpbin-0.2.0, mock-0.11.0
collected 231 items

tests/test_requests.py ..............................................................FF....................................................................................X................127.0.0.1 - - [18/Nov/2016 15:23:05] "GET /stream/4 HTTP/1.1" 200 1080
X.................................................................

======================================================= FAILURES =======================================================
_____________________________ TestRequests.test_BASICAUTH_TUPLE_HTTP_200_OK_GET[user-pass] _____________________________

self = <tests.test_requests.TestRequests object at 0x104b15ef0>
httpbin = <function prepare_url.<locals>.inner at 0x104445ea0>, username = 'user', password = 'pass'

    @pytest.mark.parametrize(
        'username, password', (
            ('user', 'pass'),
            (u'имя'.encode('utf-8'), u'пароль'.encode('utf-8')),
        ))
    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin, username, password):
        auth = (username, password)
        url = httpbin('get')

        r = requests.Request('GET', url, auth=auth)
        p = r.prepare()

        assert p.headers['Authorization'] == _basic_auth_str(username, password)

        assert r.status_code == 200

        r = requests.get(url)
>       assert r.status_code == 401
E       assert 200 == 401
E        +  where 200 = <Response [200]>.status_code

tests/test_requests.py:481: AssertionError
------------------------------------------------- Captured stderr call -------------------------------------------------
127.0.0.1 - - [18/Nov/2016 15:23:03] "GET /get HTTP/1.1" 200 337
127.0.0.1 - - [18/Nov/2016 15:23:03] "GET /get HTTP/1.1" 200 294
 TestRequests.test_BASICAUTH_TUPLE_HTTP_200_OK_GET[\xd0\xb8\xd0\xbc\xd1\x8f-\xd0\xbf\xd0\xb0\xd1\x80\xd0\xbe\xd0\xbb\xd1\x8c]

self = <tests.test_requests.TestRequests object at 0x102240240>
httpbin = <function prepare_url.<locals>.inner at 0x103f73730>, username = b'\xd0\xb8\xd0\xbc\xd1\x8f'
password = b'\xd0\xbf\xd0\xb0\xd1\x80\xd0\xbe\xd0\xbb\xd1\x8c'

    @pytest.mark.parametrize(
        'username, password', (
            ('user', 'pass'),
            (u'имя'.encode('utf-8'), u'пароль'.encode('utf-8')),
        ))
    def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin, username, password):
        auth = (username, password)
        url = httpbin('get')

        r = requests.Request('GET', url, auth=auth)
        p = r.prepare()

        assert p.headers['Authorization'] == _basic_auth_str(username, password)

        assert r.status_code == 200

        r = requests.get(url)
>       assert r.status_code == 401
E       assert 200 == 401
E        +  where 200 = <Response [200]>.status_code

tests/test_requests.py:481: AssertionError
------------------------------------------------- Captured stderr call -------------------------------------------------
127.0.0.1 - - [18/Nov/2016 15:23:03] "GET /get HTTP/1.1" 200 353
127.0.0.1 - - [18/Nov/2016 15:23:03] "GET /get HTTP/1.1" 200 294
=================================== 2 failed, 227 passed, 2 xpassed in 24.46 seconds ===================================
```

ghost: Now is:

```
$ py.test tests/test_requests.py
================================================= test session starts ==================================================
platform darwin -- Python 3.5.2, pytest-2.8.7, py-1.4.31, pluggy-0.3.1
rootdir: requests, inifile:
plugins: cov-2.2.1, httpbin-0.2.0, mock-0.11.0
collected 231 items

tests/test_requests.py ....................................................................................................................................................X................X.................................................................

======================================== 229 passed, 2 xpassed in 24.31 seconds ========================================
```

ghost: Heh, but test don't pass in python2. I look this right now.

Lukasa: I recommend just skipping the actual request: simply remove those lines entirely. We only care about the auth headers anyway in this case.

ghost: It looks that the problem on httpbin side. We request same url in py2 and py3, and we set same authorization header in both cases.

May be skip the test with unicode auth in py2?

Lukasa: Don't skip the test, just skip the actual `requests.get()` part of the test. Nothing after the assertion of the `Authorization` header is needed.

ghost: Ok, I missed your message. Github doesn't autoupdate the conversation (:

ghost: I just committed. Please review.

nateprewitt: @Lukasa @klimenko this test was originally here for specifically testing that the auth header would send and receive a 200. It may be worth creating a separate test for verifying `prepare` works correctly, rather than circumventing the original functionality of this test.

Lukasa: Yeah, that's probably the best idea tbh. @klimenko want to keep your new test _and_ restore the original one, but only testing the values it originally used?

ghost: Yes, of course.

ghost: Please review.

Issue #3675, opened 2016-11-12
Streaming Requests documentation error for Python 3


nateprewitt:
The [documentation](http://docs.python-requests.org/en/master/user/advanced/#streaming-requests) for streaming requests supplies a code snippet (below) that hits #3359 which is also experienced in `iter_lines` due to its use of `iter_content`. 

```python
r = requests.get('http://httpbin.org/stream/20', stream=True)

for line in r.iter_lines():
    if line:
        print(json.loads(line))
```

Since `iter_lines` default is `decode_unicode=False`, it returns bytes causing Python 3 to barf on `json.loads(lines)`. Unfortunately, `decode_unicode=True` doesn't work either because httpbin.org doesn't provide a charset in the Content-Type header.

As for fixing the example, we can manually decode `line` before passing it into json.loads. This exposes some of our dirty laundry though, and may give the user the wrong idea that they need to manually decode their strings in all cases.

Alternatively, we update httpbin to pass a charset with the Content-Type and change the example to use `decode_unicode=True`.

Edit: Moving semi-related comment to appropriate issue. (#3359)

Lukasa:
I view this as strictly a Python 2 vs Python 3 problem, rather than a problem with `iter_lines`. `iter_lines` returns bytes by default, that's a good sensible default, so we should just decode those bytes ourselves.

nateprewitt:
Alright, I'll update the example accordingly.

nateprewitt:
Resolved with #3678.

PR #3678, opened 2016-11-14
decoding fix for Streaming Requests docs

nateprewitt:
This will address #3675 with 71b8941. I've added an additional commit (d81ad06) noting a suggested best practice of providing an encoding fallback when using the `decode_unicode` parameter. This is an optional improvement that I can strip if we only want the bare bones fix.


Lukasa:
LGTM, let's do this! Thanks @nateprewitt! :sparkles: :cake: :sparkles:

Issue #3070, opened 2016-03-29
Consider making Timeout option required or have a default


mlissner:
I have a feeling I'm about to get a swift education on this topic, but I've been thinking about the pros/cons of changing `requests` so that somehow there is a timeout value configured for every request.

I think there are two ways to do this:
1. Provide a default value. I know browsers have a default, so that may be a simple place to begin.
2. Make every user configure this in every request -- bigger API breakage. Probably not the way to go.

The reason I'm thinking about this is because I've used requests for a few years now and until now I didn't realize the importance of providing a timeout. It took one of my programs hanging forever for me to realize that the default here isn't really very good for my purposes. (I'm in the process of updating all my code...)

I also see that a lot of people want `Session` objects to have a timeout parameter, and this might be a way to do that as well. 

If a large default were provided to all requests and all sessions, what negative impact would that have? The only thing I can think of is that some programs will get timeout exceptions where they previously hung, which seems like an improvement to me.
## Caveat, added May 13, 2016:

Please don't use this issue to discuss adding a timeout attribute to requests. There are a number of discussions about this elsewhere (search closed issues), and we don't want that conversation to muddy this issue too. Thanks.

Lukasa:
This is an entirely reasonable suggestion.

Honestly, I'm not averse to doing it: there are definitely worse things to do than this. I'd be open to providing a default timeout. However, I don't think we can do it until 3.0.0.

Of course @kennethreitz as keeper of the spirit of requests has got the final say on this, and should definitely weigh in.

sigmavirus24:
Changing the default would definitely need to wait until 3.0.0.

I'm not against having a well-reasoned (and therefore, also well-documented) default timeout parameter.

I'm still pretty strongly against sessions having a timeout attribute though and I think that shouldn't muddy the waters of this discussion because it will sidetrack things and prevent an otherwise productive discussion.

mlissner:
@sigmavirus24 I don't know the argument pro/con sessions having a timeout, but I'm happy to defer to other's judgement and experience on that, keeping these waters clean, like you say. 

I'm just happy that I haven't gotten a swift education about timeouts.

justanr:
I'm definitely in favor of this in light of dealing with a package that doesn't provide a way to set a timeout, but does accept a fully configured session. My current fix is a specialized `HTTPAdapter` that sets a default timeout if none is provided:

``` python
class TimeoutHTTPAdapter(HTTPAdapter):
[...]

PR #3679, opened 2016-11-14
Amplifies the timeout warning in the docs.

mlissner:
Partially addresses #3070 until a fix can come in the next major version.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -484,7 +484,9 @@ Timeouts
 --------
 
 You can tell Requests to stop waiting for a response after a given number of
-seconds with the ``timeout`` parameter::
+seconds with the ``timeout`` parameter. Nearly all production code should use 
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks @mlissner! :sparkles:

nateprewitt:
Sorry, I don't mean to be a squeaky wheel. Isn't this already stated two sentences later? We may want to delete one so we're not saying the same thing twice.

> If no timeout is specified explicitly, requests do not time out.

Lukasa:
I don't think I mind the duplication here: they say the same thing in different language, which is potentially useful in a case like this.

mlissner:
I'm fine with removing the second mention if folks want, but I agree with @Lukasa too. I didn't realize for ages that a request not timing out meant my program would hang until I killed it. It's a pretty critical issue in my opinion, so being super-duper, repetitively explicit doesn't seem like a bad thing to me.

PR #3684, opened 2016-11-15
Add method to url tests

nateprewitt:
Looks like we forgot methods in the tests for #3620 😉

Proposed/3.0.0 is having a tantrum because we started checking for a non-None method there. We should probably get ahead of that and just patch it on master now though.


Lukasa:
Thanks @nateprewitt!

PR #3690, opened 2016-11-16
Clarify allow_redirects

galgeek:
Make consistent with quickstart doc.

_If_ true, this would be better:

:param allow_redirects: (optional) Boolean. Whether redirects will be followed. Defaults to False for HEAD requests, True for all other methods.


Lukasa:
Thanks @galgeek!

Lukasa:
:sparkles: :cake: :sparkles:

PR #3693, opened 2016-11-16
Change setuptools extra for secure extra.

### LINKED ISSUE #3689: security extras need to be updated
Starting with `urllib3` 1.17, `ndg-httpsclient` and `pyasn1` are no longer required to inject `pyopenssl` into `urllib3`. `cryptography` and `idna` are required instead. The result of this is if an older version of `pyopenssl` is installed and you run `pip install requests[security]`, you will see import errors when trying to inject `pyopenssl`.

Lukasa:
Resolves #3689.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,8 @@ setup(
     cmdclass={'test': PyTest},
     tests_require=test_requirements,
     extras_require={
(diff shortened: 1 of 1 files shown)


Lukasa:
I took this version count from urllib3, which seems like a sensible lead to follow here.

Issue #3687, opened 2016-11-15
because of idna2008 enforcement some real urls that work in the browser are now broken


nlevitt:
Because of idna2008 enforcement 2.12.0 some real urls that work in the browser are now broken.

For example:
http://☃.net/
http://xn--n3h.net/

My suggestion would be to try idna2008 first, catch exception, then try idna2003.

```
>>> requests.get('http://xn--n3h.net/')
Traceback (most recent call last):
  File "/Users/nlevitt/workspace/warcprox/warcprox-ve35/lib/python3.5/site-packages/requests/models.py", line 370, in prepare_url
    host = idna.encode(host, uts46=True).decode('utf-8')
  File "/Users/nlevitt/workspace/warcprox/warcprox-ve35/lib/python3.5/site-packages/requests/packages/idna/core.py", line 355, in encode
    result.append(alabel(label))
  File "/Users/nlevitt/workspace/warcprox/warcprox-ve35/lib/python3.5/site-packages/requests/packages/idna/core.py", line 276, in alabel
    check_label(label)
  File "/Users/nlevitt/workspace/warcprox/warcprox-ve35/lib/python3.5/site-packages/requests/packages/idna/core.py", line 253, in check_label
    raise InvalidCodepoint('Codepoint {0} at position {1} of {2} not allowed'.format(_unot(cp_value), pos+1, repr(label)))
requests.packages.idna.core.InvalidCodepoint: Codepoint U+0027 at position 2 of "b'xn--n3h'" not allowed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nlevitt/workspace/warcprox/warcprox-ve35/lib/python3.5/site-packages/requests/api.py", line 70, in get
    return request('get', url, params=params, **kwargs)
  File "/Users/nlevitt/workspace/warcprox/warcprox-ve35/lib/python3.5/site-packages/requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/nlevitt/workspace/warcprox/warcprox-ve35/lib/python3.5/site-packages/requests/sessions.py", line 474, in request
    prep = self.prepare_request(req)
  File "/Users/nlevitt/workspace/warcprox/warcprox-ve35/lib/python3.5/site-packages/requests/sessions.py", line 407, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
[...]

Issue #3683, opened 2016-11-15
URL has an invalid label.


MikeVL:
requests v2.12.0
```python
>>> import requests
>>> requests.get('http://vluki_develop_624_solr:8983/solr/vluki')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 70, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 474, in request
    prep = self.prepare_request(req)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 407, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
  File "/usr/local/lib/python2.7/dist-packages/requests/models.py", line 302, in prepare
    self.prepare_url(url, params)
  File "/usr/local/lib/python2.7/dist-packages/requests/models.py", line 372, in prepare_url
    raise InvalidURL('URL has an invalid label.')
requests.exceptions.InvalidURL: URL has an invalid label.
```

Lukasa:
So this is arguably an increase in strictness thanks to our new idna library, but underscores are _not_ allowed in hostnames. See kjd/idna#17 for more. While this previously worked, I suspect that really it shouldn't have. Underscores are perfectly allowed in _DNS_, which is why this never encountered a problem prior to this, but for hostnames they're entirely forbidden.

I'm inclined to say that this is within the bounds of things that are allowed to stop working in a minor release, as it's simply outside the realm of what RFCs allow.

Lukasa:
To be clear, your suitable workarounds are either to downgrade Requests or to change your hostnames to use hyphens instead of underscores. I'm proposing not to consider this a Requests bug, so you'd either be stuck on 2.11.1 for the rest of eternity or you'll need to change your hostnames to use hyphens.

MikeVL:
Ok, thanks! I will change my urls.

Lukasa:
Thanks for the report! :sparkles:

Lothiraldan:
[...]

PR #3695, opened 2016-11-16
idna bypass

nateprewitt:
This addresses #3687 allowing URLs that are already idna encoded to pass through while everything else is still properly encoded.

### DIFF
--- a/requests/_internal_utils.py
+++ b/requests/_internal_utils.py
@@ -8,7 +8,7 @@ Provides utility functions that are consumed internally by Requests
 which depend on extremely few external helpers (such as compat)
 """
 
-from .compat import is_py2, builtin_str
+from .compat import is_py2, builtin_str, str
 
 
 def to_native_string(string, encoding='ascii'):
@@ -25,3 +25,18 @@ def to_native_string(string, encoding='ascii'):
             out = string.decode(encoding)
 
     return out
+
+
+def unicode_is_ascii(u_string):
+    """Determine if unicode string only contains ASCII characters.
+
+    :param str u_string: unicode string to check. Must be unicode
+        and not Python 2 `str`.
+    :rtype: bool
+    """
+    assert isinstance(u_string, str)
+    try:
+        u_string.encode('ascii')
+        return True
+    except UnicodeEncodeError:
+        return False
--- a/requests/models.py
+++ b/requests/models.py
@@ -30,7 +30,7 @@ from .packages.urllib3.exceptions import (
 from .exceptions import (
     HTTPError, MissingSchema, InvalidURL, ChunkedEncodingError,
     ContentDecodingError, ConnectionError, StreamConsumedError)
-from ._internal_utils import to_native_string
+from ._internal_utils import to_native_string, unicode_is_ascii
 from .utils import (
     guess_filename, get_auth_from_url, requote_uri,
     stream_decode_response_unicode, to_key_val_list, parse_header_links,
@@ -365,11 +365,17 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         if not host:
             raise InvalidURL("Invalid URL %r: No host supplied" % url)
 
-        # Only want to apply IDNA to the hostname
+        # In general, we want to try IDNA encoding every hostname, as that
+        # allows users to automatically get the correct behaviour. However,
+        # we’re quite strict about IDNA encoding, so certain valid hostnames
+        # may fail to encode. On failure, we verify the hostname meets a
+        # minimum standard of only containing ASCII characters, and not starting
+        # with a wildcard (*), before allowing the unencoded hostname through.
         try:
             host = idna.encode(host, uts46=True).decode('utf-8')
         except (UnicodeError, idna.IDNAError):
-            raise InvalidURL('URL has an invalid label.')
+            if not unicode_is_ascii(host) or host.startswith(u'*'):
+                raise InvalidURL('URL has an invalid label.')
 
         # Carefully reconstruct the network location
         netloc = auth or ''
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -2093,6 +2093,7 @@ class TestPreparingURLs(object):
         (
             ('http://google.com', 'http://google.com/'),
             (u'http://ジェーピーニック.jp', u'http://xn--hckqz9bzb1cyrb.jp/'),
+            (u'http://xn--n3h.net/', u'http://xn--n3h.net/'),
             (
                 u'http://ジェーピーニック.jp'.encode('utf-8'),
                 u'http://xn--hckqz9bzb1cyrb.jp/'
@@ -2113,6 +2114,18 @@ class TestPreparingURLs(object):
                 u'http://Königsgäßchen.de/straße'.encode('utf-8'),
                 u'http://xn--knigsgchen-b4a3dun.de/stra%C3%9Fe'
             ),
+            (
+                b'http://xn--n3h.net/',
+                u'http://xn--n3h.net/'
+            ),
+            (
+                b'http://[1200:0000:ab00:1234:0000:2552:7777:1313]:12345/',
+                u'http://[1200:0000:ab00:1234:0000:2552:7777:1313]:12345/'
+            ),
+            (
+                u'http://[1200:0000:ab00:1234:0000:2552:7777:1313]:12345/',
+                u'http://[1200:0000:ab00:1234:0000:2552:7777:1313]:12345/'
+            )
         )
     )
     def test_preparing_url(self, url, expected):
@@ -2127,6 +2140,7 @@ class TestPreparingURLs(object):
             b"http://*",
             u"http://*.google.com",
             u"http://*",
+            u"http://☃.net/"
         )
     )
     def test_preparing_bad_url(self, url):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -17,6 +17,7 @@ from requests.utils import (
     to_key_val_list, to_native_string,
     unquote_header_value, unquote_unreserved,
     urldefragauth, add_dict_to_cookiejar)
+from requests._internal_utils import unicode_is_ascii
 
 from .compat import StringIO, cStringIO
 
@@ -515,6 +516,7 @@ def test_should_bypass_proxies(url, expected, monkeypatch):
     monkeypatch.setenv('NO_PROXY', '192.168.0.0/24,127.0.0.1,localhost.localdomain,172.16.1.1')
     assert should_bypass_proxies(url) == expected
 
+
 @pytest.mark.parametrize(
     'cookiejar', (
         compat.cookielib.CookieJar(),
@@ -529,3 +531,14 @@ def test_add_dict_to_cookiejar(cookiejar):
     cj = add_dict_to_cookiejar(cookiejar, cookiedict)
     cookies = dict((cookie.name, cookie.value) for cookie in cj)
     assert cookiedict == cookies
+
+
+@pytest.mark.parametrize(
+    'value, expected', (
+                (u'test', True),
+                (u'æíöû', False),
+                (u'ジェーピーニック', False),
+    )
+)
+def test_unicode_is_ascii(value, expected):
+    assert unicode_is_ascii(value) is expected


requests/utils.py
```diff
@@ -825,3 +825,19 @@ def rewind_body(prepared_request):
                                         "body for redirect.")
     else:
         raise UnrewindableBodyError("Unable to rewind request body for redirect.")
+
+def is_ascii(data):
```
nateprewitt: We could also rework this as:

``` python
    if isinstance(data, str):
        coding_func = getattr(data, 'encode')
    elif isinstance(data, bytes):
        coding_func = getattr(data, 'decode')
    else:
        return False

    # Try to encode/decode data into ascii
    try:
        coding_func('ascii')
        return True
    except (UnicodeError):
        return False
```

Either way it's kind of an ugly function but pulls that out of the method and hopefully can provide value elsewhere later.

requests/models.py
```diff
@@ -365,11 +365,12 @@ def prepare_url(self, url, params):
         if not host:
             raise InvalidURL("Invalid URL %r: No host supplied" % url)
 
-        # Only want to apply IDNA to the hostname
-        try:
-            host = idna.encode(host, uts46=True).decode('utf-8')
-        except (UnicodeError, idna.IDNAError):
-            raise InvalidURL('URL has an invalid label.')
+        if not host.startswith('xn--') or not is_ascii(host):
```
Lukasa: We need a comment to justify why we're leaving some URLs alone: specifically, because we think they've _already_ been IDNA-encoded and don't want to stomp on that logic.

We think this if they begin with the IDNA tag (`xn--`) and if all the bytes within them are ASCII.

requests/utils.py
```diff
@@ -825,3 +825,19 @@ def rewind_body(prepared_request):
...
+    """Determine is data can be encoded/decoded as ascii"""
```
Lukasa: Nit: `if`.

requests/utils.py
Lukasa: I think I'd like to condense the code in this form.

tests/test_requests.py
```diff
@@ -2113,6 +2113,11 @@ class TestPreparingURLs(object):
                 u'http://Königsgäßchen.de/straße'.encode('utf-8'),
                 u'http://xn--knigsgchen-b4a3dun.de/stra%C3%9Fe'
             ),
+            (
+                u'http://xn--n3h.net/'.encode('utf-8'),
+                u'http://xn--n3h.net/'
+            ),
+            ('http://xn--n3h.net/', 'http://xn--n3h.net/')
```
Lukasa: Let's also confirm this works for things explicitly marked as byte strings.

nateprewitt:
Alright, all systems are nominal. 🚀

requests/models.py
```diff
@@ -365,11 +365,14 @@ def prepare_url(self, url, params):
         if not host:
             raise InvalidURL("Invalid URL %r: No host supplied" % url)
 
-        # Only want to apply IDNA to the hostname
-        try:
-            host = idna.encode(host, uts46=True).decode('utf-8')
-        except (UnicodeError, idna.IDNAError):
-            raise InvalidURL('URL has an invalid label.')
+        # If the host doesn't start with 'xn--' or contains non-ascii
+        # characters, it may not be idna encoded, so we'll do it here.
```
Lukasa: I think this comment needs to be reframed: almost all URIs aren't IDNA encoded, so we're just aiming to catch the edge case.

nateprewitt:
Comment updated, hopefully that's a bit closer.

requests/models.py
```diff
@@ -365,11 +365,19 @@ def prepare_url(self, url, params):
...
-        except (UnicodeError, idna.IDNAError):
-            raise InvalidURL('URL has an invalid label.')
+        # In general, we want to IDNA encode everything, as that allows users
+        # to automatically get correct behaviour in the majority of cases.
+        # However, we’re quite strict about IDNA encoding, so if a user has
+        # already IDNA encoded their domain name, we shouldn’t bother validating
+        # their work. We determine something *looks* IDNA encoded by checking
+        # for the IDNA-encoding tag (xn--) at the start of the hostname, and
+        # confirming that the hostname is entirely ASCII characters.
+        if not host.startswith('xn--') or not is_ascii(host):
```
Lukasa: Let's try to avoid problems by setting this to `u'xn--'`. On top of this, given that we know that the string will always be unicode here, we can safely rewrite `is_ascii` to `unicode_is_ascii` which will just work for unicode strings for now.

Save ourselves the unneeded complexity.

nateprewitt: While this is definitely doable/simpler, I have a couple outstanding question:

1. The new function will only be the single try/except block, so does it really warrant being a util function at that point?

2. While we are now explicitly stating that this is only for unicode, if you supply the function a bytestring without the current checks, it will evaluate to `True` in Python 2 and `False` in Python 3. If this ever gets reused in a place were people are *pretty* sure it will always be unicode but may not, that's a tricky place for errors to crop up. I feel like having the more robust, but well tested, function is probably the safer of these two options. That is if we're even leaving this as a standalone function.

nateprewitt: @Lukasa, sorry I deleted this the first time, thought I'd made a mistake but I've arrived back at the same conclusion. I'm onboard with `u'xn--'`. 

While `unicode_is_ascii` is definitely doable/simpler, I do have a couple outstanding questions:
1. The new function will only be the single try/except block, so does it really warrant being a util function at that point?
2. While we are now explicitly stating that this is only for unicode, if you supply the function a bytestring without the current checks, it will evaluate to `True` in Python 2 and `False` in Python 3. If this ever gets reused in a place were people are _pretty_ sure it will always be unicode but may not, that's a tricky place for errors to crop up.

The repro is simply `b'test'.encode('ascii')`. Python 2 will let you do this all day, Python 3 raises an `AttributeError`. We could alternatively wrap the try/except in a conditional block to check if it's a string, but that has the inverse problem of string in the format `'my string'` being `False` in Python 2 and `True` in Python 3.

Lukasa: A single try...except block is still a totally valuable thing to have in one function, particularly when we're trying to use it in a conditional. So yeah, I have no problem with that.

Ultimately, I'm not concerned about the failure modes. If you're really worried, add `assert isinstance(argument, unicode)` where `unicode` is imported from `compat.py`. That should avoid the problem with unicode by adding a check that can be optimised out.

nateprewitt: So `unicode` won't work here because it's not defined for Python 3 in compat. We bind `str` to `unicode` for Python 2. However, strings that are declared like `'my string'`, as I noted above, are not `unicode` and will evaluate to `False` in `isinstance(string, str)` for Python 2.

I can implement the standalone try/except block, I just wanted to make sure the subtle failure points were understood before making the changes. Bytestrings (including Python 2's default str) will inconsistently evaluate between versions, the conditional only toggles which value is returned respectively.

nateprewitt:
Alright, ASCII check updated 😃

requests/models.py
nateprewitt: Just another passing thought, perhaps this should be moved to _internal_utils due to the caveats? It just seems like this has enough nuance to not be a utility people are actively encouraged to use externally.

requests/utils.py
```diff
@@ -825,3 +825,16 @@ def rewind_body(prepared_request):
...
     else:
         raise UnrewindableBodyError("Unable to rewind request body for redirect.")
+
+def unicode_is_ascii(u_string):
+    """Determine if unicode string only contains ASCII characters.
+
+    :param str u_string: unicode string to check. Must be unicode
+        and not Python 2 `str`.
+    :rtype: bool
+    """
```
Lukasa: Ok, let's move this to `_internal_utils`. Also, I apologise about the confusion of `unicode`: let's import `str` from `compat.py` and use it here for an assertion.

nateprewitt:
OK, we're all moved over to _internal_utils.

nateprewitt:
> An alternative solution to this is to say that if the idna module fails to encode we'll just go ahead and try to encode as ASCII. If that works then we shrug our shoulders and say everything is probably ok, and if it fails we catch that and throw InvalidURL. @sigmavirus24, how does that sound?

([comment](/kennethreitz/requests/issues/3683#issuecomment-261240279))

Do we want to hold here until a decision is made on this so we're not having to back things out?

Lukasa:
It's certainly worth us holding off until we can resolve that ambiguity.

sigmavirus24:
That method sounds good to me. I'm not as familiar with the idna standard though, so while it sounds perfectly practical, I don't know if it fits in with the rest of the work around the standard.

Lukasa:
Ok then, let's go ahead with that and see how it goes.

tests/test_requests.py
```diff
@@ -2127,6 +2133,7 @@ def test_preparing_url(self, url, expected):
             b"http://*",
             u"http://*.google.com",
             u"http://*",
+            u"http://☃.net/"
         )
     )
     def test_preparing_bad_url(self, url):
```
nateprewitt, line 2146: Hrmm, so I'm assuming the urls in this `test_preparing_bad_url` are to show we prevent wildcards? If we fallback to simply checking for ASCII on failed IDNA encoding, these all pass now. Is there a discrete subset of ASCII URLs that we still want/need to catch?

nateprewitt:
I guess we should probably clarify what we're attempting to do here. The above comment was in reference to the the inability to use IPv6 with Requests 2.12+. Personally, I think that definitely needs to be addressed, the other complaints I have no opinion on.

The fallback approach though essentially gives carte blanche to anything that can be ASCII encoded, which seems to defeat some benefits of doing IDNA encoding. They've special cased `.` delimiters in the IDNA library, I'm assuming for things like subdomains, which allows IPv4 addresses to pass through (unintentionally?). However, they don't appear to have taken into account IPv6 style addresses with `:` delimiters or wrapping brackets.

From a cursory reading of [RFC5891](https://tools.ietf.org/html/rfc5891), it seems like IDNA is only intended for domain names. That would suggest to me we shouldn't be passing IP addresses into this function to begin with.

If we're ONLY attempting to handle IPv6, we can check to skip `host` values that are entirely numeric or delimiters. Otherwise, if we're entertaining allowing things like underscores then maybe the passthrough is the only viable option to avoid constantly programming around use cases.

**Edit**: Sorry @Lukasa I just realized I essentially reiterated your entire thought process from the other thread. Vendoring in a dependency is likely not what we want to do. `any([c.isalpha() for c in host])` is probably kludgy enough to keep out of the main path for Requests. So I guess that leaves the bypass.

tests/test_requests.py
Lukasa: Yeah, this is about preventing wildcards.

To find this kind of thing we really may need to add a custom validator that pulls the hostname apart and checks for wildcards. It's probably sufficient to confirm the URI doesn't start with `u'*.'`.

Lukasa:
Yup, that's why I landed on the bypass: it just seems like the simplest approach that is also in line with what most browsers do (i.e. try to do the right thing but allow the wrong thing if it's unambiguous). In this case I'd like to filter out certain obviously bad URLs that IDNA was saving us from (those with leading wildcards, for example), but otherwise if they're the kind of garden-variety bad that issue #3683 is discussing (or if they aren't hostnames at all!) then we can probably just pass them through and let the rest of the codebase handle them.

nateprewitt:
Ok, I removed the initial check for `xn--` because we end up performing fairly redundant logic both before the try...except block and again in the except block itself.

This should work for our given test cases, as well as the other concerns about unix socket files and hostnames containing underscores, provided they're both only ASCII. Let me know if I'm leaving anything else out.

tests/test_utils.py
```diff
@@ -529,3 +531,13 @@ def test_add_dict_to_cookiejar(cookiejar):
...
     cookies = dict((cookie.name, cookie.value) for cookie in cj)
     assert cookiedict == cookies
+
+
+@pytest.mark.parametrize(
+    'value, expected', (
+                (u'test', True),
+                (u'ジェーピーニック', False),
+    )
+)
```
Lukasa, line 542: I'd like to see a test here that uses something that's present in Latin1, as well, just to assuage concerns.

tests/test_requests.py
```diff
@@ -2113,6 +2115,14 @@ class TestPreparingURLs(object):
...
                 u'http://xn--knigsgchen-b4a3dun.de/stra%C3%9Fe'
             ),
+            (
+                u'http://xn--n3h.net/'.encode('utf-8'),
+                u'http://xn--n3h.net/'
+            ),
+            (
+                'http://[1200:0000:ab00:1234:0000:2552:7777:1313]:12345/',
+                'http://[1200:0000:ab00:1234:0000:2552:7777:1313]:12345/'
+            )
```
Lukasa, line 2128: Can we test this in unicode as well please?

requests/_internal_utils.py
```diff
@@ -25,3 +25,18 @@ def to_native_string(string, encoding='ascii'):
...
+
+    :param str u_string: unicode string to check. Must be unicode
+        and not Python 2 `str`.
+    :rtype: bool
+    """
+    assert isinstance(u_string, str)
+    try:
+        u_string.encode('ascii')
+        return True
+    except UnicodeError:
```
Lukasa: It should only be possible to see a `UnicodeEncodeError` here now, as we're sure that the instance is `str`. Let's tighten the exception handling so we don't accidentally mask weird bugs.

nateprewitt:
Alright, tests extended and exception tightened.

tests/test_requests.py
```diff
@@ -2113,6 +2114,18 @@ class TestPreparingURLs(object):
                 u'http://Königsgäßchen.de/straße'.encode('utf-8'),
                 u'http://xn--knigsgchen-b4a3dun.de/stra%C3%9Fe'
             ),
+            (
+                u'http://xn--n3h.net/'.encode('utf-8'),
```
Lukasa: This can just be a literal byte string (as in `b''`), no need for a `.encode`.

tests/test_requests.py
```diff
@@ -2113,6 +2114,18 @@ class TestPreparingURLs(object):
...
+                u'http://xn--n3h.net/'
+            ),
+            (
+                b'http://[1200:0000:ab00:1234:0000:2552:7777:1313]:12345/',
+                'http://[1200:0000:ab00:1234:0000:2552:7777:1313]:12345/'
```
Lukasa: This should presumably be a unicode string.

nateprewitt:
Yep, those were oversights. Should be updated now.

Lukasa:
Alright, I'm going to merge this now. @sigmavirus24 feel free to give feedback regardless, we can always revisit this. =) Thanks @nateprewitt!

olgert:
Guys, do you have any idea of when these changes could be released?

Lukasa:
I am tentatively considering next week for a possible release. No promises though.

olgert:
Thanks for prompt response!

Issue #3696, opened 2016-11-16
A whitespace at the start of Url causes error on request


hussaintamboli:
In [1]: import requests

    In [2]: requests.get("http://stackoverflow.com/")
    Out[2]: <Response [200]>

    In [3]: requests.get(" http://stackoverflow.com/")
    # ... stacktrace skipped ...
    InvalidSchema: No connection adapters were found for ' http://stackoverflow.com/'

Url should have been stripped. But I may be missing something here.

Lukasa:
I'm open to a PR that will cause us to strip URLs, that seems reasonable enough to me.

hussaintamboli:
Sure.

hussaintamboli:
I have created a pull request #3697. Please have a look.

hussaintamboli:
PR is merged. Closing it as it's fixed.

PR #3697, opened 2016-11-16
Fixed issue #3696. Added a test for it as well.

hussaintamboli:


requests/models.py
```diff
@@ -343,6 +343,9 @@ def prepare_url(self, url, params):
         else:
             url = unicode(url) if is_py2 else str(url)
 
+        # Remove leading and trailing whitespaces from url
+        url = url.strip()
```
Lukasa: So while I agree that we want to do this, I think we should be cautious. Stripping trailing whitespace will be a behavioural change because right now trailing whitespace _does_ work (albeit in a way that is a bit unexpected to people).

So, let's just do `lstrip()` here: we only want to remove leading whitespace.

hussaintamboli: I agree. Fixed it.

PR #3700, opened 2016-11-17
Adding 'import requests' to readme code snippet.

rmhasan:
The interpreter snippet at the top of the readme.rst is the first snippet of code that users will see for using requests. I think an import statement should be shown at the top of the code snippet, so a novice or even intermediate python user will know to import the module first before attempting to use it. It makes the snippet more complete. I've seen other github projects include the import statement in snippets like the one in the readme. I changed the doc/index.rst file for consistency, since they are the same code snippet.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ constantly reading documentation, depression, headaches, or even death.
 Behold, the power of Requests:
 
 .. code-block:: python
-
+    >>> import requests
     >>> r = requests.get('https://api.github.com/user', auth=('user', 'pass'))
     >>> r.status_code
     200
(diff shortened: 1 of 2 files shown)


rmhasan:
For instance [tensorflow](https://github.com/tensorflow/tensorflow)'s readme has a similar code snippet and it has the import statement at the top of the code snippet.

Lukasa:
Thanks for this!

kennethreitz:
undo

rmhasan:
Hi @kennethreitz , I was wondering why my pull request got reverted.

kennethreitz:
I prefer it the way it is, and it was made that way intentionally. 

I understand why you made this suggestion, and I appreciate it — but this is the way I want to present the project, for aesthetic reasons.

PR #3704, opened 2016-11-17
clarify that the `chunk_size` is optional when streaming to a file

afeld:
Tripped me up briefly...figured others might run into the same thing. Thanks!

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -184,7 +184,7 @@ streamed to a file::
 Using ``Response.iter_content`` will handle a lot of what you would otherwise
 have to handle when using ``Response.raw`` directly. When streaming a
 download, the above is the preferred and recommended way to retrieve the
-content. Note that `chunk_size` is optional.
+content. Note that ``chunk_size`` is optional.
 
 
 Custom Headers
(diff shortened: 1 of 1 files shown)


docs/user/quickstart.rst
```diff
@@ -184,7 +184,7 @@ streamed to a file::
 Using ``Response.iter_content`` will handle a lot of what you would otherwise
 have to handle when using ``Response.raw`` directly. When streaming a
 download, the above is the preferred and recommended way to retrieve the
-content.
+content. Note that `chunk_size` is optional.
```
Lukasa: In restructuredtext you need to use two backticks, as elsewhere in the paragraph. Can you amend to do that please?

nateprewitt:
It's probably important to note that you _should_ provide a chunk_size here despite it being optional. The default of 1 is likely not what the user wants for any non-trivial file.

afeld:
@nateprewitt Yeah, thought about that...might make sense to give a value of `128` or something reasonable in the example, and just say that it can be changed. I'm no expert on this, so whatever y'all think!

nateprewitt:
Great! How does changing the example to:

```python
with open(filename, 'wb') as fd:
    for chunk in r.iter_content(chunk_size=128):
        fd.write(chunk)
```
and then that line saying you can adjust the chunk_size to a number that fits your requirements sound? This will probably better serve the user than pointing them towards the default behaviour of iter_content. That's just one guys thoughts though :)

Lukasa:
I'd be happy to merge that docs change.

nateprewitt:
@afeld if you think this change is agreeable and have the inclination, feel free to throw the PR together. Otherwise, I'll try to open something next week.

PR #3709, opened 2016-11-20
Added new function that returns json output in a pretty printed string form.

rmhasan:
The function sjson in the Response class will return a json object in string form. Unless provided arguments, it will return a pretty printed string, when printed is human readable. You can look at the docs for the json module's dumps object to know which arguments to send to the sjson function for custom output.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -173,3 +173,4 @@ Patches and Suggestions
 - Om Prakash Kumar <omprakash070@gmail.com> (`@iamprakashom <https://github.com/iamprakashom>`_)
 - Philipp Konrad <gardiac2002@gmail.com> (`@gardiac2002 <https://github.com/gardiac2002>`_)
 - Hussain Tamboli <hussaintamboli18@gmail.com> (`@hussaintamboli <https://github.com/hussaintamboli>`_)
+- Rakib Hasan
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -156,9 +156,26 @@ It should be noted that the success of the call to ``r.json()`` does **not**
 indicate the success of the response. Some servers may return a JSON object in a
 failed response (e.g. error details with HTTP 500). Such JSON will be decoded
 and returned. To check that a request is successful, use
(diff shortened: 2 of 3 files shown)


rmhasan:
out
This is what the output should look like.

drpoggi:
I believe requests is in an API freeze at this point, especially when user code could do the same thing with `r.content` and their preferred indentation level and separators. Contributions are always welcome, but API changes are the type of thing that would be better served by raising an issue and getting feedback on the interface, before you spend your time writing the code. 

EDIT: Of course I'm not a requests maintainer, I'm just trying to give feedback based on what I've seen. Some interface of this type may well get accepted.

rmhasan:
@drpoggi You can provide custom indentation level and separators as arguments sjson. I didn't know about the freeze. Ill take a look at r.content. Thanks for the feedback.

rmhasan:
@drpoggi Can you explain how users can create the same output with r.content.  The output from r.content is different from r.json().  Also how do you provide preferred indentation level and separators. As far as I know it doesn't take any arguments. You can provide preferred indentation level and separators to r.sjson through key value arguments. You can look at the documentation for the dumps function in the json module to give custom arguments.  The whole point of creating the function was so that users can get the json object in a human readable form with simple readable code. Also how do you know there is a freeze in requests. Was there a post somewhere, or is it in the documentation somewhere?

EDIT: Also thanks for the feedback.

nateprewitt:
@rmhasan, thanks for your interest in contributing to Requests. You can find information about the freeze and other contributing topics in Requests [Contributor's Guide](http://docs.python-requests.org/en/master/dev/contributing). The specific information you're looking for is under [feature requests](http://docs.python-requests.org/en/master/dev/contributing/#feature-requests).

rmhasan:
@nateprewitt Thanks.

Lukasa:
Thanks for this, but ultimately this duplicates the logic of the `pprint` module in the standard library. Because of that, and because the potential user base is fairly limited, I don't think this feature warrants an exception to the API freeze.

Thanks for your contribution though!

PR #3713, opened 2016-11-21
Restrict URL preparation to HTTP/HTTPS

tiran:
Requests treats all URLs starting with the string 'http' as HTTP URLs.
Preparation with IDNA breaks non-standard URIs like http+unix. Requests
now prepares only URLs with prefix http:// and https://.

Signed-off-by: Christian Heimes <christian@python.org>

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -347,9 +347,9 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         url = url.lstrip()
 
         # Don't do any URL preparation for non-HTTP schemes like `mailto`,
-        # `data` etc to work around exceptions from `url_parse`, which
-        # handles RFC 3986 only.
-        if ':' in url and not url.lower().startswith('http'):
+        # `data`, `http+unix` etc to work around exceptions from `url_parse`,
+        # which handles RFC 3986 only.
+        if ':' in url and not url.lower().startswith(('http://', 'https://')):
             self.url = url
             return
 
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -2138,3 +2138,20 @@ class TestPreparingURLs(object):
         r = requests.Request('GET', url=url)
         with pytest.raises(requests.exceptions.InvalidURL):
             r.prepare()
+
+    @pytest.mark.parametrize(
+        'protocol, url',
+        (
+            ("http+unix://", b"http+unix://%2Fvar%2Frun%2Fsocket/path"),
+            ("http+unix://", u"http+unix://%2Fvar%2Frun%2Fsocket/path"),
+            ("mailto", b"mailto:user@example.org"),
+            ("mailto", u"mailto:user@example.org"),
+            ("data", b"data:SSDimaUgUHl0aG9uIQ=="),
+        )
+    )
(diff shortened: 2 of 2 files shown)


tiran:
http+unix is used in Docker, Custodia and SSSD, e.g. https://github.com/latchset/custodia/blob/master/custodia/client.py#L60 with URI like ```http+unix://%2Fvar%2Frun%2Fcustodia.socket/secrets``` for HTTP over Unix sockets on socket file ```/var/run/custodia.socket```.

Lukasa:
This looks entirely reasonable to me, thanks @tiran!

Lukasa:
One note though: can we get a test for this?

requests/models.py
```diff
@@ -347,9 +347,9 @@ def prepare_url(self, url, params):
         url = url.lstrip()
 
         # Don't do any URL preparation for non-HTTP schemes like `mailto`,
-        # `data` etc to work around exceptions from `url_parse`, which
-        # handles RFC 3986 only.
-        if ':' in url and not url.lower().startswith('http'):
+        # `data`, `http+unix` etc to work around exceptions from `url_parse`,
+        # which handles RFC 3986 only.
+        if ':' in url and not url.lower().startswith(('http://', 'https://')):
```
sigmavirus24, line 352: Didn't we have this previously? Can we check the history and see if we can find why we switched to what we have today?

sigmavirus24: Nope. This is written as it was introduced in #1717, so I'm clearly thinking of something else.

nateprewitt: The original commit (b149be5) was to solve issues like @tiran's but may have been implemented a bit naively. It doesn't look like this would negatively affect the original intent of the conditional. I do think we can remove the `if ':' in url` portion of the conditional now though since that's covered in the new `startswith` tuple.

tests/test_requests.py
```diff
@@ -2138,3 +2138,18 @@ def test_preparing_bad_url(self, url):
...
+
+    @pytest.mark.parametrize(
+        'url',
+        (
+            b"http+unix://%2Fvar%2Frun%2Fsocket/path",
+            u"http+unix://%2Fvar%2Frun%2Fsocket/path",
+        )
+    )
+
+    def test_prepare_http_unix(self, url):
```
Lukasa: So this is good, let's just make this a bit more tempting for people to extend in future by changing the name: `test_url_passthrough`.

requests/models.py
nateprewitt, line 352: I think we can remove the `if ':' in url` portion of the conditional now since that's covered in the new `startswith` tuple.

tiran: I tried it but got a test failure in the test case ```MissingSchema, 'hiwpefhipowhefopw'```.

Lukasa: Yup, it's needed: it exists to allow for things that don't look to us like URLs at all but still might be handle-able by something down in the stack.

nateprewitt: Yep, oversight on my part.

Lukasa:
Cool, thanks @tiran! All tests pass, so it seems like it's good to go. Thanks so much! :sparkles: :cake: :sparkles:

PR #3717, opened 2016-11-23
streaming doc clarification

nateprewitt:
Minor example update and rewording in response to the continued discussion after merging #3704. @afeld, if you're inclined to comment, does this change still address your original concerns?


afeld:
:+1:

Lukasa:
Thanks @afeld and @nateprewitt!

PR #3338, opened 2016-06-16
Refactor prepare body

davidsoncasey:
PR for proposed/3.0 with refactoring of `prepare_body` method. For discussion see #3184.

PR #3754, opened 2016-12-08
test auth doesn't generate incorrect http headers

nateprewitt:
These are a few tests from #3338 which is slated for Proposed/3.0.0. #3066 is the issue #3338 was originally opened to fix, but has since been resolved on master with #3082 and #3535. Adding these tests will allow us to confirm #3066 is indeed resolved and help prevent regressions until 3.0.0 is released.


Lukasa:
Is the plan for this to land in 3.0.0 or master? Your summary isn't entirely clear.

Lukasa:
Based on some of the other issues you linked I *think* you did want these tests in master.

nateprewitt:
Yep, master was the goal here. Thanks!

Issue #2590, opened 2015-05-04
PATCH get response 301(moved permanently) is redirected with empty body


sainttail:
I don't know this is normal behaviour or not but any reason to set body to empty with patch?

Lukasa:
What kind of body are you sending?

sainttail:
i sent something like this
`r = requests.patch(url, data={"data": "test"})`

Lukasa:
I think we're a bit over-zealous with stripping bodies on redirects. In #1084 we allowed 307 and 308 to keep their bodies, but all other redirects lose theirs. Arguably, a 301 for a non-POST should also behave the same way. Assigning this to 3.0.0, because it'd be a breaking change.

texttheater:
Same for response 302.

texttheater:
<blockquote>I think we're a bit over-zealous with stripping bodies on redirects.</blockquote>

Out of interest, why would you _ever_ strip bodies?

sigmavirus24:
On a redirect? Consider being redirected to a different domain that you did not want to send that body to... That is why we tend to be overzealous

texttheater:
If you don't want to send the body to the new location, presumably, you don't want to send the request at all. Either preserving the body or raising an exception would be sensible behavior. Sending an altogether _different_ request to the new location makes no sense at all, IMHO. Certainly not as default behavior.

-------- Original message --------

Lukasa:
> Out of interest, why would you ever strip bodies?

Well that's easy, really. A very common pattern is to POST to some endpoint, receive a 302, and GET the location from the 302. You obviously want to strip the body in that case.

> Sending an altogether _different_ request to the new location makes no sense at all, IMHO. Certainly not as default behavior.

Again, sure it does. That's exactly what many redirects _are_. Status codes 301, 302 are standardised to turn POST into GET (which is definitionally an "altogether different request"), and status code 303 is standardised to turn _any_ verb into HEAD or GET (again, an "altogether different request"). If your web server really wants to keep everything the same it should issue 307 (302 with no verb change) or 308 (301 with no verb change).

texttheater:
Now I'm really confused.

<blockquote>A very common pattern is to POST to some endpoint, receive a 302, and GET the location from the 302. You obviously want to strip the body in that case.</blockquote>

What is the purpose of this pattern?

From the perspective of the client, requests is doing something completely different from what it was asked to do, without giving any notice about this.

[...]

PR #3757, opened 2016-12-08
301/302 redirects for non-POST with body

nateprewitt:
This patch is an attempt to address #2590 by relaxing the way Requests handles redirects. 

Taking a look at how we’re handling stripping bodies, I modified the conditional to avoid an overly complicated check. The only requests we want to strip bodies off of are ones that have been modified from the redirect. That means the method will have already been changed to GET in `rebuild_method` and we can perform a uniform stripping for only GET requests. Personally, I'd like to have `rebuild_method` return a value, either a boolean or the new method, and use that as the parameter for the check. Then move the code inside the conditional into it's own function. This will help a bit with the overall size of `resolve_redirects`.

I also removed the 302 catch-all which was put in place in #1704 when RFC 2616 was still the standard. The previous mandates for 302 aren’t included in [RFC 7231 6.4.3](https://tools.ietf.org/html/rfc7231#section-6.4.3) and don’t appear to be reproducible in Chrome or Firefox with any tools supporting PUT, PATCH, or DELETE. (Note: if there’s a repro for this that I’m missing please let me know and I’ll gladly revert this. I’m just going off of the information available.)

The "test all methods" tests may be a bit overboard, I added them primarily for my own assurance to make sure I wasn't missing edge cases. I can remove those if they're deemed unnecessary.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -157,13 +157,11 @@ class SessionRedirectMixin(object):
             if response.is_permanent_redirect and request.url != prepared_request.url:
                 self.redirect_cache[request.url] = prepared_request.url
 
-            old_method = prepared_request.method
-            self.rebuild_method(prepared_request, response)
-            new_method = prepared_request.method
+            method_changed = self.rebuild_method(prepared_request, response)
 
             # https://github.com/kennethreitz/requests/issues/2590
             # If method is changed to GET we need to remove body and associated headers.
-            if old_method != new_method and new_method == 'GET':
+            if method_changed and prepared_request.method == 'GET':
                 # https://github.com/kennethreitz/requests/issues/3490
                 purged_headers = ('Content-Length', 'Content-Type', 'Transfer-Encoding')
                 for header in purged_headers:
@@ -284,8 +282,11 @@ class SessionRedirectMixin(object):
     def rebuild_method(self, prepared_request, response):
         """When being redirected we may want to change the method of the request
         based on certain specs or browser behavior.
+
+        :rtype bool:
+        :return: boolean expressing if the method changed during rebuild.
         """
-        method = prepared_request.method
+        method = original_method = prepared_request.method
 
         # http://tools.ietf.org/html/rfc7231#section-6.4.4
         if response.status_code == codes.see_other and method != 'HEAD':
@@ -300,6 +301,7 @@ class SessionRedirectMixin(object):
             method = 'GET'
 
         prepared_request.method = method
+        return method != original_method
 
 
 class Session(SessionRedirectMixin):
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -157,11 +157,13 @@ class SessionRedirectMixin(object):
             if response.is_permanent_redirect and request.url != prepared_request.url:
                 self.redirect_cache[request.url] = prepared_request.url
 
+            old_method = prepared_request.method
             self.rebuild_method(prepared_request, response)
+            new_method = prepared_request.method
 
-            # https://github.com/kennethreitz/requests/issues/1084
-            if response.status_code not in (codes.temporary_redirect,
-                    codes.permanent_redirect):
+            # https://github.com/kennethreitz/requests/issues/2590
+            # If method is changed to GET we need to remove body and associated headers.
+            if old_method != new_method and new_method == 'GET':
                 # https://github.com/kennethreitz/requests/issues/3490
                 purged_headers = ('Content-Length', 'Content-Type', 'Transfer-Encoding')
                 for header in purged_headers:
@@ -289,14 +291,12 @@ class SessionRedirectMixin(object):
         if response.status_code == codes.see_other and method != 'HEAD':
             method = 'GET'
 
-        # Do what the browsers do, despite standards...
-        # First, turn 302s into GETs.
-        if response.status_code == codes.found and method != 'HEAD':
-            method = 'GET'
-
-        # Second, if a POST is responded to with a 301, turn it into a GET.
-        # This bizarre behaviour is explained in Issue 1704.
-        if response.status_code == codes.moved and method == 'POST':
+        # If a POST is responded to with a 301 or 302, turn it into a GET. This has
+        # become a common pattern in browsers and was introduced into later versions
+        # of HTTP RFCs. While some browsers transform other methods to GET, little of
+        # that has been standardized. For that reason, we're using curl as a model
+        # which only supports POST->GET.
+        if response.status_code in (codes.found, codes.moved) and method == 'POST':
             method = 'GET'
 
         prepared_request.method = method
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -234,11 +234,51 @@ class TestRequests:
 
     def test_http_301_doesnt_change_head_to_get(self, httpbin):
         r = requests.head(httpbin('status', '301'), allow_redirects=True)
-        print(r.content)
         assert r.status_code == 200
         assert r.request.method == 'HEAD'
         assert r.history[0].status_code == 301
         assert r.history[0].is_redirect
+
+    def test_http_301_doesnt_change_non_post_to_get(self, httpbin):
+        r = requests.patch(httpbin('redirect-to'),
+                           data='test body',
+                           params={'url': 'patch', 'status_code': '301'})
+        assert r.status_code == 200
+        assert r.request.method == 'PATCH'
+        assert r.history[0].status_code == 301
+        assert r.history[0].is_redirect
+        assert r.request.body == 'test body'
+        assert r.json()['data'] == 'test body'
+
+    @pytest.mark.parametrize(
+        'method, body, expected', (
+            ('GET', None, 'GET'),
+            ('HEAD', None, 'HEAD'),
+            ('POST', 'test', 'GET'),
+            ('PUT', 'put test', 'PUT'),
+            ('PATCH', 'patch test', 'PATCH'),
+            ('DELETE', '', 'DELETE')
+        )
+    )
+    def test_http_301_for_redirectable_methods(self, httpbin, method, body, expected):
+        """Tests all methods except OPTIONS for expected redirect behaviour.
+
+        OPTIONS responses can behave differently depending on the server, so
+        we don't have anything uniform to test except how httpbin responds
+        to them. For that reason they aren't included here.
+        """
+        params = {'url': '/%s' % expected.lower(), 'status_code': '301'}
+        r = requests.request(method, httpbin('redirect-to'), data=body, params=params)
+
+        assert r.request.url == httpbin(expected.lower())
+        assert r.request.method == expected
+        assert r.history[0].status_code == 301
+        assert r.history[0].is_redirect
+
+        if expected in ('GET', 'HEAD'):
+            assert r.request.body is None
+        else:
+            assert r.json()['data'] == body
 
     def test_http_302_changes_post_to_get(self, httpbin):
         r = requests.post(httpbin('status', '302'))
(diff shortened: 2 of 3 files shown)


tests/test_requests.py
```diff
@@ -229,12 +229,59 @@ def test_http_301_changes_post_to_get(self, httpbin):
...
+            ('POST', 'test', 'GET'),
+            ('PUT', 'put test', 'PUT'),
+            ('PATCH', 'patch test', 'PATCH'),
+            ('DELETE', '', 'DELETE')
+        )
+    )
+    def test_http_301_for_redirectable_methods(self, httpbin, method, body, expected):
+        """Tests all methods except OPTIONS for expected redirect behaviour.
+
+        OPTIONS isn't a redirectable method and will always return values for
```
nateprewitt: This was a hasty comment. There isn't anything in the RFC that states this, and it seems servers from major platforms treat the request differently (Facebook treats it as a GET??). I think the problem here is that it's unpredictable how a server may respond. I'll clarify the comment in the next review pass.

Lukasa: I think there is no reason to assume that OPTIONS cannot be redirected, and nothing in the Requests codebase forbids it.

tests/test_requests.py
```diff
@@ -229,12 +229,59 @@ def test_http_301_changes_post_to_get(self, httpbin):
...
+        assert r.history[0].is_redirect
+        assert r.request.body == 'test body'
+        assert r.json()['data'] == 'test body'
+
+    def test_http_301_for_OPTIONS(self, httpbin):
+        r = requests.options(httpbin('status', '301'))
+        assert r.status_code == 200
+        assert r.history == []
+        assert r.request.method == 'OPTIONS'
+        assert 'Allow' in r.headers
```
Lukasa: This is really only a test of httpbin's particular behaviour.

tests/test_requests.py
```diff
@@ -249,6 +296,36 @@ def test_http_302_doesnt_change_head_to_get(self, httpbin):
...
+            ('PUT', 'put test', 'PUT'),
+            ('PATCH', 'patch test', 'PATCH'),
+            ('DELETE', '', 'DELETE')
+        )
+    )
+    def test_http_302_for_redirectable_methods(self, httpbin, method, body, expected):
+        """Tests all methods except OPTIONS for expected redirect behaviour.
+
+        OPTIONS isn't a redirectable method and will always return values for
+        the original URI.
```
Lukasa: Same note re OPTIONS here.

tests/test_requests.py
```diff
@@ -263,6 +340,33 @@ def test_http_303_doesnt_change_head_to_get(self, httpbin):
...
+            ('PUT', 'put test', 'GET'),
+            ('PATCH', 'patch test', 'GET'),
+            ('DELETE', '', 'GET')
+        )
+    )
+    def test_http_303_for_redirectable_methods(self, httpbin, method, body, expected):
+        """Tests all methods except OPTIONS for expected redirect behaviour.
+
+        OPTIONS isn't a redirectable method and will always return values for
+        the original URI.
```
Lukasa: Same note re OPTIONS.

requests/sessions.py
```diff
@@ -290,13 +291,8 @@ def rebuild_method(self, prepared_request, response):
...
         # Do what the browsers do, despite standards...
-        # First, turn 302s into GETs.
-        if response.status_code == codes.found and method != 'HEAD':
-            method = 'GET'
-
-        # Second, if a POST is responded to with a 301, turn it into a GET.
-        # This bizarre behaviour is explained in Issue 1704.
-        if response.status_code == codes.moved and method == 'POST':
+        # If a POST is responded to with a 301 or 302, turn it into a GET.
+        if response.status_code in (codes.found, codes.moved) and method == 'POST':
```
Lukasa, line 300: We may want to flesh this comment out a bit here to point out that there is a lack of consistency over what exactly happens here. Some browsers have historically changed other methods (I can find some citations of Chrome changing DELETE to GET), but ultimately most browsers rarely ever issue methods outside of POST and GET.

Our best option, then, is to mostly emulate what curl does, and curl restricts this behaviour to POST only. That seems like the most consistent position to take. But we should aim to cite our sources. ;)

nateprewitt:
Alright, comments updated.

> Personally, I'd like to have rebuild_method return a value, either a boolean or the new method, and use that as the parameter for the check. Then move the code inside the conditional into it's own function. This will help a bit with the overall size of resolve_redirects.

Thoughts on mucking with this? If not, I think we're good to go.

tests/test_requests.py
```diff
@@ -229,12 +229,53 @@ def test_http_301_changes_post_to_get(self, httpbin):
...
+    )
+    def test_http_301_for_redirectable_methods(self, httpbin, method, body, expected):
+        """Tests all methods except OPTIONS for expected redirect behaviour.
+
+        OPTIONS responses can behave differently depending on the server, so
+        we don't have anything uniform to test except how httpbin responds
+        to them. For that reason they aren't included here.
+        """
+        s = requests.Session()
+        params = {'url': '/%s' % method.lower(), 'status_code': '301'}
```
Lukasa: Should the URL parameter not be `expected`?

tests/test_requests.py
```diff
@@ -229,12 +229,53 @@ def test_http_301_changes_post_to_get(self, httpbin):
...
+        req = requests.Request(method, httpbin('redirect-to'), data=body, params=params)
+        r = s.send(req.prepare())
```
Lukasa: Why are we using the `PreparedRequest` flow?

tests/test_requests.py
```diff
@@ -249,6 +290,37 @@ def test_http_302_doesnt_change_head_to_get(self, httpbin):
...
+        """Tests all methods except OPTIONS for expected redirect behaviour.
+
+        OPTIONS responses can behave differently depending on the server, so
+        we don't have anything uniform to test except how httpbin responds
+        to them. For that reason they aren't included here.
+        """
+        s = requests.Session()
+        params = {'url': '/%s' % method.lower()}
+        req = requests.Request(method, httpbin('redirect-to'), data=body, params=params)
+        r = s.send(req.prepare())
```
Lukasa: Same notes as above.

tests/test_requests.py
```diff
@@ -263,6 +335,34 @@ def test_http_303_doesnt_change_head_to_get(self, httpbin):
...
+        """Tests all methods except OPTIONS for expected redirect behaviour.
+
+        OPTIONS responses can behave differently depending on the server, so
+        we don't have anything uniform to test except how httpbin responds
+        to them. For that reason they aren't included here.
+        """
+        s = requests.Session()
+        params = {'url': '/%s' % method.lower(), 'status_code': '303'}
+        req = requests.Request(method, httpbin('redirect-to'), data=body, params=params)
+        r = s.send(req.prepare())
```
Lukasa: Same notes here.

tests/test_requests.py
nateprewitt: So I went with the `PreparedRequest` flow because I felt it was easier to understand. We could use the API mappings by writing the test as:
```python
params = {'url': '/%s' % method.lower(), 'status_code': '301'}
req = getattr(requests, method)
req(httpbin('redirect-to'), data=body, params=params)
```
but that's probably not a paradigm to encourage and makes the test kind of odd for the sake of saving 2 lines.

Lukasa: Is it not easier just to use `requests.request`?

nateprewitt: First method in the api and I don't know if I knew that existed before now. That's definitely simpler, I'll get that updated.

requests/sessions.py
```diff
@@ -157,13 +157,11 @@ def resolve_redirects(self, response, stream=False, timeout=None,
             if response.is_permanent_redirect and request.url != prepared_request.url:
                 self.redirect_cache[request.url] = prepared_request.url
 
-            old_method = prepared_request.method
-            self.rebuild_method(prepared_request, response)
-            new_method = prepared_request.method
+            method_changed = self.rebuild_method(prepared_request, response)
```
nateprewitt, line 160: Eh, so I guess I'm indifferent on this now that I'm really looking at it. Pulling the logic from the conditional into it's own function adds code without significantly reducing the `resolve_redirects` footprint.

The return value for `rebuild_method` provides an op vs no-op check but isn't particularly useful. I threw it in there as a slight readability improvement, but we can ditch it if you aren't feeling it.

requests/sessions.py
```diff
@@ -157,12 +157,11 @@ def resolve_redirects(self, response, stream=False, timeout=None,
...
                 self.redirect_cache[request.url] = prepared_request.url
 
-            self.rebuild_method(prepared_request, response)
+            method_changed = self.rebuild_method(prepared_request, response)
 
-            # https://github.com/kennethreitz/requests/issues/1084
-            if response.status_code not in (codes.temporary_redirect,
-                    codes.permanent_redirect):
-                # https://github.com/kennethreitz/requests/issues/3490
+            # https://github.com/kennethreitz/requests/issues/2590
```
sigmavirus24, line 162: We should probably keep old issue references if only for historical context.

requests/sessions.py
```diff
@@ -282,24 +281,27 @@ def rebuild_proxies(self, prepared_request, proxies):
     def rebuild_method(self, prepared_request, response):
         """When being redirected we may want to change the method of the request
         based on certain specs or browser behavior.
+
+        :rtype bool:
+        :return: boolean expressing if the method changed during rebuild.
         """
-        method = prepared_request.method
+        original_method = prepared_request.method
```
sigmavirus24: Nit: This might be less confusing if we did 

```py
method = original_method = prepared_request.method
```

nateprewitt: Agreed, I like that better too.

requests/sessions.py
nateprewitt: I think #3490 still makes sense to have, but #1084 is relating to code that is being removed here. We can keep it in if you think it's still providing value, but I feel it will end up as more of a red herring. Maybe making a note in #2590 with the link would be more useful if it's needed?

nateprewitt:
@Lukasa, I think this is ready for a peek whenever you've got a moment :)

tests/test_requests.py
```diff
@@ -234,12 +234,52 @@ def test_http_301_changes_post_to_get(self, httpbin):
...
+    def test_http_301_doesnt_change_non_post_to_get(self, httpbin):
+        r = requests.patch(httpbin('redirect-to'),
+                           data='test body',
+                           params={'url': 'patch', 'status_code': '301'})
+        assert r.status_code == 200
+        assert r.request.method == 'PATCH'
+        assert r.history[0].status_code == 301
+        assert r.history[0].is_redirect
+        assert r.request.body == 'test body'
+        assert r.json()['data'] == 'test body'
```
Lukasa: This test looks pretty close to a duplicate of the one below. Any reason we wouldn't just move the check on `status_code` into the test below?

nateprewitt: You're right, this one was written before `test_http_301_for_redirectable_methods` was, following the styling above. We can actually collapse the three tests above into the one below, but I left them to ensure we were always testing for those specific cases. The thought being if `test_http_301_for_redirectable_methods` is later modified it would be nice for those to still fail, but I understand if you'd like to see them squashed together.

As for the status code, I left that out of the test below because /head doesn't exist in httpbin (because it's non-sensical). The test logic becomes more complex trying to code around the HEAD case and the status code isn't needed confirm we got what we wanted from the test, only that httpbin is functioning properly.

Lukasa: I think there's no reason to keep them separate. =)

nateprewitt: Ok, I'll squash them all down then.

nateprewitt:
Ok, 301-303 individual tests removed in favor of the big ones.

Issue #3753, opened 2016-12-08
Basic auth with integer username broken in 2.12.2


oschwald:
Previously, it was possible to set `auth` to something like `(42,'password')`. This has worked fine since before requests hit 1.0. 2.12.2, a patch release, broke this, causing failures such as the ones seen in this [Travis build](https://travis-ci.org/maxmind/GeoIP2-python/jobs/181427000).

I am not saying the new behavior is incorrect, but this is a breaking change.

This is similar but not the same as #3750,  #3736, and #3751.

StewPoll:
This is different to the other three issues. Those three issues are all passing `None` type to `auth()` - You're passing an `int` object instead.

It would appear that they were all caused by 329a5cf though. 

What decisions those that maintain requests want to make is up to them though, just thought I'd help debug this.

StewPoll:
A possible solution:
`codecs.encode("{0}:{1}".format(username, password).encode(), "base-64").strip()`
Not sure how versatile this is though

nateprewitt:
I've got a feeling this may fall under Requests' policy of "All header values must be a string, bytestring, or unicode.", since the Authorization field generated is a header, and the `auth` keyword is even discussed in the [Custom Headers](http://docs.python-requests.org/en/master/user/quickstart/#custom-headers) docs.

@Lukasa may have different feelings on this, but trying to permit non-string values here seems to have the same ambiguity issues we've tried to avoid elsewhere. Just like how `None` hasn't been permitted because `None` is cast to a string, when the user much more likely wants `""`, which will generate different auth values.

Lukasa:
This gets to be another entry on my list of problems with semantic versioning.

We never intended non-string values to work here. I didn't care about the removal of support for `None`, because it was a colossally stupid value to allow that would basically never do what people expected. But at this point we begin to go off the rails again. And semantic versioning, read strictly, doesn't allow tightening of any check at all in a minor release.

I think the best thing to do is to re-add support for non-string values by checking for them and casting them to str explicitly. I don't want to hide it: I want us to put a big-ol' comment on top of it that says that this behaviour is dumb but we need to preserve it because people are relying on it. I then think we should catch-up master into proposed/3.0.0 and then merge a PR that *removes* that explicit cast.

Lukasa:
Ok, right now master and proposed/3.0.0 are caught up. So if someone wants to make this patch, I can walk them through the two PRs we'll need.

oschwald:
[...]

PR #3758, opened 2016-12-08
adding string casting for non-bytes values

nateprewitt:
This addresses #3753, I'll open a PR for 3.0.0 removing this when we merge master.

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -15,7 +15,7 @@ import threading
 
 from base64 import b64encode
 
-from .compat import urlparse, str
+from .compat import urlparse, str, basestring
 from .cookies import extract_cookies_to_jar
 from ._internal_utils import to_native_string
 from .utils import parse_dict_header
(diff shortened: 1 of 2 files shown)


kennethreitz:
i support this

kennethreitz:
it's stupid, but no reason not to

Lukasa:
LGTM. Thanks for this @nateprewitt! :sparkles:

Lukasa:
And I've just caught the 3.0.0 branch up, so you can go ahead and open the second PR now.

requests/auth.py
```diff
@@ -27,7 +27,21 @@
...
-    
+
+    # "I want us to put a big-ol' comment on top of it that
+    # says that this behaviour is dumb but we need to preserve
+    # it because people are relying on it."
+    #    - Lukasa
+    #
+    # These are here solely to maintain backwards compatibility
+    # for things like ints. This will be removed in 3.0.0.
+    if not isinstance(username, basestring):
```
sigmavirus24, line 38: We should also have included a deprecation warning in here too to warn people this was happening in 3.0

tests/test_requests.py
```diff
@@ -484,6 +484,7 @@ def test_BASICAUTH_TUPLE_HTTP_200_OK_GET(self, httpbin):
         'username, password', (
             ('user', 'pass'),
             (u'имя'.encode('utf-8'), u'пароль'.encode('utf-8')),
+            (42, 42)
```
sigmavirus24, line 487: This should also have had a test with `(None, None)` just to cover all of our bases.

requests/auth.py
Lukasa: Honestly, a whole lot of stuff got removed that doesn't have deprecation warnings. I kinda think that ship has sailed.

sigmavirus24: The difference being that we're adding this *back* with the intent to remove it. If we're doing that, we have enough time to add a deprecation warning

Lukasa: I'm certainly happy to accept a patch that adds the deprecation warning. =) I have no objection to adding it.

nateprewitt:
@sigmavirus24 you're suggesting using `warnings.warn` to actively notify the user about the deprecation, correct? I can add that and the test here in a moment if you're not already working on it.

PR #3760, opened 2016-12-09
Add deprecation warnings for 3.0

sigmavirus24:
Add extra test parameter for basic auth encoding

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -12,6 +12,7 @@ import re
 import time
 import hashlib
 import threading
+import warnings
 
 from base64 import b64encode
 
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Don't merge yet, I think I swapped the warning message and class

sigmavirus24:
r4r @Lukasa @nateprewitt

nateprewitt:
@Lukasa when you get another second to merge master into Proposed/3.0.0, I'll throw up the other patch.

Lukasa:
Already done. ;)

kennethreitz:
hell yeah!

requests/auth.py
```diff
@@ -36,9 +37,23 @@ def _basic_auth_str(username, password):
     # These are here solely to maintain backwards compatibility
     # for things like ints. This will be removed in 3.0.0.
     if not isinstance(username, basestring):
+        warnings.warn(
+            "Non-string usernames will no longer be supported in Requests "
+            "3.0.0. Please convert the object you've passed in ({!r}) to "
```
sigmavirus24, line 42: Worth noting - People have complained in the past when a warning has variable content in it because it can defeat the warnings filters. At the same time as many (if not more) people complain when there's a warning and they have no indication of what *caused* the warning. So I've erred on the side of using the dynamic content because a filter will handle the following situations:

```py
AUTH = (None, None)

requests.get(url, auth=AUTH)
requests.postt(url, auth=AUTH)

s = requests.Session()
s.auth = AUTH
s.get(url)
```

They'll get exactly one warning for the None username and one for the None password. But if they change the `AUTH` to have `(42, None)` they'll get a third warning for passing `42` but not for passing `None` as the password. This should ideally work fairly well.

nateprewitt:
Looks like we might be hitting a problem in 2.6 with zero length values in the string formatter (None specifically).

PR #3763, opened 2016-12-09
Revert "use enforce_content_length in Requests"

nateprewitt:
This is more of a TODO reminder. Given the upcoming work for urllib3 2.0, which I'm assuming will be used in a Requests 3.0.0, this won't work. We don't need to remove this immediately, but I'll leave this here for when we're sure 2.0 is happening.


Lukasa:
Heh, we won't forget: we'll run immediately into errors because I removed the kwarg entirely. Let's not queue up a whole bunch of PRs: let's focus on doing 2.0 first. :wink:

PR #3765, opened 2016-12-11
Make code match behavior as defined in RFC 2818 and RFC 6125 and documented behavior

noxxi:
The CN should be checked if no dNSName exist in subjectAltNames. The code treated a IP address as DNS name by appending it to dnsnames. This way CN was not checked in a certificate which contained the hostname as CN but only IP addresses as SAN.

See also http://stackoverflow.com/questions/41089539/authentication-issue-with-ssl-certificate-using-python-requests-lib

### DIFF
--- a/requests/packages/urllib3/packages/ssl_match_hostname/_implementation.py
+++ b/requests/packages/urllib3/packages/ssl_match_hostname/_implementation.py
@@ -132,7 +132,6 @@ def match_hostname(cert, hostname):
         elif key == 'IP Address':
             if host_ip is not None and _ipaddress_match(value, host_ip):
                 return
-            dnsnames.append(value)
     if not dnsnames:
         # The subject is only checked when there is no dNSName entry
         # in subjectAltName


Lukasa:
Thanks for this! For the moment, I'm not going to merge this, for the following reasons.

1. This patch affects urllib3, which we vendor without changes. It would need to be proposed to that project. 
2. The inclusion of match_hostname is currently intended to behave the same way on all platforms. Thus, if we edit it, we'd also have to stop preferring the stdlib until they matched our behaviour. This patch is therefore incomplete.
3. Your assertion about the incorrectness of our logic is not as clear-cut as it seems. Our behaviour currently matches curl, Firefox, and Chrome, which is some good company to be in (see curl/curl#1065).
4. It seems the Python crypto experts [aren't convinced of this either, and ordinarily I'd default to respecting their logic](http://bugs.python.org/issue28938).
5. There is basically no good reason to have a cert that fails in this manner. CN has been deprecated for years, SAN has been around forever. Just put the hostname in the SAN field as well.

Issue #1979, opened 2014-03-28
Authentication Handlers lost on redirect.


costinb7:
I'am trying to use the requests library by making a redirection with  Digest authentication method, but the response is 401. I mention that it works with basic authentication. I've captured the packets with wireshark, and noticed that the first HTTP request is without the Authorization header, the 401 unauthorized answered is received, and after that the traffic continues as it should be, the Authorization header is added, the 302 answer is received, and after that with the https cyphers exchange. I don't know why the requests.send method returns 401.

Lukasa:
Wait, I'm confused. Your description of the wireshark trace suggests that everything is happening as expected, but you're getting an unexpected status code?

costinb7:
Yes, but in fact, I don't know if it is happening _exactlly_ as it should be. I tried to make a redirection from http to https. When I try with a browser it works well, and also it works well when using requests with Basic authentication. When using with Digest, what happens is as I described above (+ the https message exchange: server hello, server hello done, client key exchange, change cipher spec etc), but the send() method responds with 401.

Lukasa:
Can I see some sample code?

costinb7:
The sample code is very simple:

url1 = 'http://192.168.40.248/param.cgi?action=list&group=root.AutoTracking'
res = s.get(url1, auth=requests.auth.HTTPDigestAuth('user', 'pass'), verify=False)
print res.status_code

The url1 address is automatic redirecting to : "https://192.168.40.248/..."(the same url but with https).
So, the above code prints: "401"

If I change the authentication class to HTTPBasicAuth, and change the server to also use Basic authentication, everything works well (200 is printed).

Ofcourse, you don't have access to the DUT that I use. But I suppose you can try with any URL that automatic redirects to use https and use digest auth method.

Lukasa:
What version of requests are you using, and where did you install it from?

costinb7:
I am using requests-2.2.1, I've downloaded it from here: https://pypi.python.org/pypi/requests (requests-2.2.1.tar.gz).

Lukasa:
Aha, ok.

I don't think auth handlers can handle an auth challenge that occurs _after_ a redirect: I think we lose track of the auth handler. @sigmavirus24 does this match your reading of the code?

If that's the case, do we consider that behaviour a bug?

Lukasa:
Note: I'm renaming this issue to more accurately reflect the bug.

costinb7:
[...]

PR #3766, opened 2016-12-12
testing HTTPDigestAuth hooks

nateprewitt:
#1979 is still opened as "unresolved" but it looks like a patch (#2253) was merged to address it in 2014. There wasn't a testing framework at the time, so no tests were included to verify things actually worked. I haven't been able to concoct a end-to-end test with httpbin for this, but the individual pieces work as intended. I can work on devising a socket test if needed.

### DIFF
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -5,7 +5,7 @@ import pytest
 import threading
 import requests
 
-from tests.testserver.server import Server
+from tests.testserver.server import Server, consume_socket_content
 
 from .utils import override_environ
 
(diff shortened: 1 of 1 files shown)


Lukasa:
I think I'd be much happier seeing a socket test than the httpbin ones: they just end up testing implementation details more than actual behaviour.

nateprewitt:
I don't think either of these are testing anything specific to httpbin's implementation, they should work the same regardless of which server they hit as long as the status codes are correct.

I added a somewhat monstrous test to simulate the full lifecycle of 401->302-401->200. I'm cutting a few corners with the actual digest piece because of the thread state, but the scenario we're trying to prove is correct. I've tested this both against 2.2.1 and 2.12.3 without the `handle_redirect` hook mounted. It's failing as expected in those scenarios.

tests/test_lowlevel.py
```diff
@@ -25,6 +25,71 @@ def test_chunked_upload():
...
+    """Ensure we correctly reset num_401_calls after a successful digest auth,
+    followed by a 302 redirect to another digest auth prompt.
+
+    See https://github.com/kennethreitz/requests/issues/1979.
+    """
+    text_401 = ('HTTP/1.1 401 UNAUTHORIZED\r\n'
+                'Connection:keep-alive\r\n'
+                'Content-Length: 0\r\n'
+                'Set-Cookie:fake=fake_value\r\n'
+                'WWW-Authenticate:Digest nonce="6bf5d6e4da1ce66918800195d6b9130d"'
```
Lukasa: Why so stingy with spaces after colons?

tests/test_lowlevel.py
```diff
@@ -25,6 +25,71 @@ def test_chunked_upload():
...
+
+    expected_digest = ('Authorization: Digest username="user", realm="me@kennethreitz.com", '
+                       'nonce="6bf5d6e4da1ce66918800195d6b9130d", uri="/"')
+
+    auth = requests.auth.HTTPDigestAuth('user', 'pass')
+
+    def digest_response_handler(sock):
+        # Respond to initial get with a challenge.
+        request_content = consume_socket_content(sock, timeout=0.5)
+        assert b"GET / HTTP/1.1" in request_content
```
Lukasa: We specifically expect this to *begin* with this string.

tests/test_lowlevel.py
```diff
@@ -25,6 +25,71 @@ def test_chunked_upload():
...
+        sock.send(text_401.encode('utf-8'))
+
+        # Verify we receive an Authorization header in response and redirect.
+        request_content = consume_socket_content(sock, timeout=0.5)
+        assert expected_digest.encode('utf-8') in request_content
+        sock.send(text_302.encode('utf-8'))
```
Lukasa: Any reason not to just store these as bytestrings to begin with?

tests/test_lowlevel.py
```diff
@@ -25,6 +25,71 @@ def test_chunked_upload():
...
+    close_server = threading.Event()
+    server = Server(digest_response_handler)
+
+    with server as (host, port):
+        url = 'http://{0}:{1}/'.format(host, port)
+        r = requests.get(url, auth=auth)
+        # Verify server succeded in authenticating
+        assert r.status_code == 200
+        # Verify redirect happened as we expected
+        assert r.history != []
```
Lukasa: Can we be a bit clearer about how many redirects we want?

tests/test_lowlevel.py
nateprewitt: Nope, I was just following the trend from the other test. I can instantiate them as bytes to avoid the overhead.

tests/test_lowlevel.py
nateprewitt: Yep, that should be a 302 check.

tests/test_lowlevel.py
nateprewitt: That's a good question, I copied these directly from an httpbin response to Requests. I'll separate them out.

nateprewitt:
> I remain unconvinced about whether checking num_401_calls is a good idea or whether we should consider it an implementation detail.

Sorry, could you clarify this a little? Are you referring to this in relation to the tests, or the actual implementation? I'm not sure how else we'd test that this is appropriately reset since that's the single flag that determines whether the 401 challenge is responded to or not.

Lukasa:
I mean it about the tests. Generally speaking if we can we should restrict ourselves to testing observable behaviour, by which I mean bytes output or public return values.

nateprewitt:
Ok, I'll remove those assert statements and rely on having expected header and status_code values.

nateprewitt:
Alright, assertions about the state of `num_401_calls` should be gone now and we're only relying upon observable behaviour.

tests/test_lowlevel.py
```diff
@@ -25,6 +25,73 @@ def test_chunked_upload():
...
+def test_digestauth_401_count_reset_on_redirect():
+    """Ensure we correctly reset num_401_calls after a successful digest auth,
+    followed by a 302 redirect to another digest auth prompt.
+
+    See https://github.com/kennethreitz/requests/issues/1979.
+    """
+    text_401 = (b'HTTP/1.1 401 UNAUTHORIZED\r\n'
+                b'Connection: keep-alive\r\n'
+                b'Content-Length: 0\r\n'
+                b'Set-Cookie: fake=fake_value\r\n'
```
Lukasa: May as well get rid of this header, we don't care about the cookie.

tests/test_lowlevel.py
```diff
@@ -25,6 +25,73 @@ def test_chunked_upload():
...
 
 
+def test_digestauth_401_count_reset_on_redirect():
+    """Ensure we correctly reset num_401_calls after a successful digest auth,
+    followed by a 302 redirect to another digest auth prompt.
+
+    See https://github.com/kennethreitz/requests/issues/1979.
+    """
+    text_401 = (b'HTTP/1.1 401 UNAUTHORIZED\r\n'
+                b'Connection: keep-alive\r\n'
```
Lukasa: May as well remove this, it's assumed for HTTP/1.1.

tests/test_requests.py
```diff
@@ -623,6 +623,21 @@ def test_DIGESTAUTH_QUOTES_QOP_VALUE(self, httpbin):
...
+    def test_DIGESTAUTH_DOESNT_RESEND(self, httpbin):
+        """If HTTPDigestAuth fails a 401 challenge, it won't try again"""
+        s = requests.Session()
+        auth = requests.auth.HTTPDigestAuth('user', 'pass')
+
+        r = requests.Request('GET', httpbin('digest-auth/auth/user/pass'), auth=auth)
+        p = r.prepare()
+
+        # simulate previously failed 401 so auth fails
+        auth._thread_local.num_401_calls = 2
```
Lukasa: So, this doesn't assert about the state of `num_401_calls`, but it does still *tweak* the state of `num_401_calls`. That is, if we removed `num_401_calls`, this test would fail. I'm still not sure that's a good idea.

nateprewitt: So this was originally written for specifically testing `handle_401` as a method, now it's morphed into leaning towards testing a behaviour. I'll rewrite this as a socket test to make sure HTTPDigestAuth functions correctly rather than focusing on the hook.

I do think removing this will make it a bit harder to pinpoint the problem in a future failure though. I understand the hesitance to not test a variable that may be ephemeral, but only having the end-to-end tests kind of hides the smoking gun. There have been previous instances of HTTPDigestAuth failing specifically because `num_401_calls` was modified incorrectly and there were no tests to catch it.

`num_401_calls` is a well defined attribute that's been the core of how HTTPDigestAuth functions over the wire for 4 years.  We know that it's an unconventional boolean that will be either 1 and 2 at any point after it's been invoked. Unless we're planning to rewrite HTTPDigestAuth in the near future, wouldn't we want more granularity in the tests where there have been previous failures?

Sorry if I'm coming across as pig headed over such an minor issue, I'm just trying to understand how you'd solve this.

nateprewitt:
Alright, we're on only socket tests now testing behaviour rather than implementation details.

tests/test_lowlevel.py
```diff
@@ -25,6 +25,116 @@ def test_chunked_upload():
...
+        sock.send(text_401)
+
+        # Verify Authorization is sent correctly again, and return 200 OK.
+        request_content = consume_socket_content(sock, timeout=0.5)
+        assert expected_digest in request_content
+        sock.send(text_200)
+
+        return request_content
+
+    close_server = threading.Event()
```
Lukasa, line 77: These events don't seem to be being used by anything: should they be?

nateprewitt: Yep, Server needs to be initialized with `wait_to_close_event=close_server`. Pushing the changes now.

nateprewitt:
@Lukasa I rescinded my last comment on `num_401_calls` because while what I was testing worked, it's not reproducible in the wild. I still think there may be value in testing the hooks individually but I don't think it's a blocker here. These tests cover the use cases we're concerned about in #1979.

Lukasa:
Cool, tests are good. Thanks for the work @nateprewitt!

PR #3771, opened 2016-12-15
Reuse all connections

kianxineki:
When testing, every time I make a request, the message appears:

`` `Starting a new HTTP connection (1): www. *****. With

I have seen one session per application. He tried to reuse the "session" for all requests and seems to work properly, even with connections that have been closed.

What implications would this change have?

``` python
In [9]: requests.get("https://github.com")
Starting new HTTPS connection (1): github.com
Out[9]: <Response [200]>

In [10]: requests.get("https://github.com")
Out[10]: <Response [200]>

In [11]: requests.get("https://stackoverflow.com")
Starting new HTTPS connection (1): stackoverflow.com
Out[11]: <Response [200]>

In [12]: requests.get("https://stackoverflow.com")
Out[12]: <Response [200]>

In [13]: requests.get("https://github.com")
Out[13]: <Response [200]>

In [14]: requests.get("https://github.com")
Out[14]: <Response [200]>
```

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -12,6 +12,7 @@ This module implements the Requests API.
 
 from . import sessions
 
+session = sessions.Session()
 
 def request(method, url, **kwargs):
     """Constructs and sends a :class:`Request <Request>`.
@@ -49,11 +50,7 @@ def request(method, url, **kwargs):
       <Response [200]>
     """
 
-    # By using the 'with' statement we are sure the session is closed, thus we
-    # avoid leaving sockets open which can trigger a ResourceWarning in some
-    # cases, and look like a memory leak in others.
-    with sessions.Session() as session:
-        return session.request(method=method, url=url, **kwargs)
+    return session.request(method=method, url=url, **kwargs)
 
 
 def get(url, params=None, **kwargs):


nateprewitt:
Hey @kianxineki, the api methods are intended to be single connection requests. If you'd like to utilize connection pooling in Requests, you'll need to use the `Session` object as shown below.

```python
s = requests.Session()
req_git = requests.Request('GET', 'https://github.com')
req_stack = requests.Request('GET', 'https://stackoverflow.com')

resp_git = s.send(req_git.prepare())
resp_stack = s.send(req_stack.prepare())
```

You can find more information on how to use `Session` [here](http://docs.python-requests.org/en/master/user/advanced/#session-objects).

nateprewitt:
Hey @kianxineki, I skimmed through your question and responded a bit too quickly. The implications of this change is we would now use a single `Session` object across ALL connections with `requests.request` which may not be what the user wants. The api methods are intended to be single connections that are cleaned up after completion.

As I think you've already noted above, you can accomplish the change by simply adding an extra line  in your code declaring your own Session instance. This will allow you to utilize connection pooling and avoid the new connections.

```python
s = requests.Session()
resp_git = s.get('https://github.com')
resp_stack = s.get('https://stackoverflow.com')
```

You can find more information on how to use `Session` [here](http://docs.python-requests.org/en/master/user/advanced/#session-objects).

agalera:
In my current code I do this to not have to use s.get() and to be able to use the methods that facilitates me requests.

``` python
import requests


session = requests.sessions.Session()


def request(method, url, **kwargs):
    return session.request(method=method, url=url, **kwargs)


requests.api.request = request

requests.get("https://google.com")
```

sigmavirus24:
@kianxineki Thanks for sending us a pull request! It looks to be your first to the project. Welcome!

It seems as though you're trying to ask why we do not use a global session object in `requests.api`. To answer you, let me explain a little bit of history first.

We used to use a global session object and our Session objects used to be thoroughly thread-safe. At some point, something changed about our Session objects that no one has yet been able to identify. This means that they're no longer properly threadsafe. As such, people were having significant problems sharing a session between threads which also applied to using the functional API. Further, having a session that sits around until the process finishes was holding onto sockets that were never being reused. 

The solution to the thread-safety issue and the socket leak was to stop using a global session.

Currently, if you want to eliminate those lines in your logs and utilize connection pooling, you *must* use a Session. There's no way to provide connection pooling to you via our functional API without actively harming other users.

In the future, please direct all of your questions to StackOverflow, which is the proper Question & Answer forum for Requests.

Again, thank you for your pull request, but we can't accept it.

PR #3781, opened 2016-12-21
updating idna.uts46data from upstream project

mplonka:


mplonka:
This is meant to include https://github.com/kjd/idna/pull/33 which will fix https://github.com/kennethreitz/requests/issues/3711.

The patch was done by running `make idna`. Also patching Makefile itself, removing `sed` invocation, which isn't necessary anymore.

Lukasa:
Thanks for this patch @mplonka. However, we would generally prefer to update the bundled packages ourselves at the time of release: that way, we have some confidence that we have done the sensible thing. Can you pare this patch down to just the makefile change, please?

mplonka:
Thanks for reviewing this PR, @Lukasa .

As per your request, here's the revert commit.

Lukasa:
Thanks! For the sake of clarity, do you mind squashing these into one commit?

mplonka:
I actually wanted to suggest squashing.

LordGaav:
@Lukasa it seems that idna hasn't been updated yet, but a new version has been released in the mean time. Will idna be updated to fix the Jython issue?

Lukasa:
@LordGaav Per [our documentation](http://docs.python-requests.org/en/master/community/release-process/#hotfix-releases), we do not update vendored dependencies in patch releases. Please wait for v2.13, which will contain an update.

LordGaav:
That was quick. I'll use `requests!=2.12.*` in my requirements.txt in the mean time then. Thanks.

PR #3782, opened 2016-12-21
Closes ##3780 High memory usage due to idma import

moin18:
Fix related to bug: https://github.com/kennethreitz/requests/issues/3780

`idma` package will now be imported within the function


Lukasa:
Thanks for this!

However, this patch does not meaningfully reduce the memory usage. Any use of Requests will encounter this branch immediately and unconditionally, meaning that in practice all that we have done is delay the loading of the module. I think we need a better solution than this one.

moin18:
How about doing selective import of the needed objects?

Raised another PR: https://github.com/kennethreitz/requests/pull/3783

PR #3783, opened 2016-12-21
Closes #3780 High memory usage due to idma import

### LINKED ISSUE #3780: idna package increases memory usage by ~20MB
I am running an Ubuntu 16.04.1 LTS machine with Python 3.5.2 which has about 40 Python scripts loaded, each one of which uses Requests. I recently updated Requests and suddenly found I was running out of memory.

Upon further investigation I discovered that as of version 2.12.0, the `idna` package has been included in requests and this has caused the memory consumption for each script to increase by about 20MB.

This can be observed with the following steps:

1. Install Requests version 2.11.1
2. Run `python3 -c "import requests; import time; time.sleep(60)" &`.
3. While the previous command is running, run `top` or `htop` and observe the resident memory usage of the command.
4. Install Requests version 2.12.0.
5. Repeat steps 2 and 3.

Is there any way that Requests could lazily load the idna library so that this extra 20MB of memory is only used when actually working with IDNA domains?

moin18:


Lukasa:
Because of how Python imports work, this also will not have any effect. Specifically, `from x import y` is approximately the same as `import x; y = x.y`. I'm afraid we need a better solution than even this. 

Sorry!

moin18:
By better solution do you mean the changes in: `idna/core.py` that is imported by `__init__.py`?

Lukasa:
Ideally we'd only bring in IDNA when it was actually needed: that is, when we had a domain we had to internationalise.

moin18:
What are your thoughts on this?

```
@staticmethod:
def _get_idna_encoded_host(host):
      from . package import idna
      try:
         host = idna.encode(host, uts46=True).decode('utf-8')
     except idna.IDNAError:
         raise UnicodeError

.....

if not unicode_is_ascii(host):
    try:
        host =self._get_idna_encoded_host(host)
    except UnicodeError:
         raise InvalidURL('URL has an invalid label.')
elif host.startswith(u'*'):
    raise InvalidURL('URL has an invalid label.')
```

DanielGibbsNZ:
@moin18 The problem I ran up against was that `idna` is part of `packages` so whenever `packages` is loaded (e.g. in `models.py`) `idna` will be loaded at that point. So the fix would be for `idna` not to be loaded at that point but somehow still be loaded when needed :/

PR #3787, opened 2016-12-22
Lazy load idna to free memory

moin18:
Fix for issue: https://github.com/kennethreitz/requests/issues/3780

Lazily loading `idna`. This CR includes utilities to:
* lazily load idna package
* And, delete module from sys.modules after use

Changes looks to work fine for me. Now, None of the reference to `idna` is getting used directly. GC will release the memory when required.


moin18:
I think it will be a better idea to create a separate package to lazy load module and add that package as the part of both `urllib3` and `requests` package

Lukasa:
urllib3 does not vendor idna, so such a package cannot really be easily shared.

Lukasa:
Additionally, Requests does not carry patches to urllib3: it vendors an unedited copy. Any urllib3 patch you need to make must be proposed to that project.

moin18:
@Lukasa : In that case I will be creating a separate patch for urllib3 discarding the urllib3 related changes from this PR. Apart from that, does the `requests` library related changes looks good to you?

Lukasa:
I think it looks really over complex. Repeatedly loading and unloading the module is solving a non-existent problem. All we should do is defer loading as late in the process as possible.

PR #3789, opened 2016-12-23
3780: Lazy load idna library

moin18:
Changes based on comment from previous Pull Request: https://github.com/kennethreitz/requests/pull/3787

Fix for issue: https://github.com/kennethreitz/requests/issues/3780

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -9,6 +9,7 @@ This module contains the primary objects that power Requests.
 
 import collections
 import datetime
+import sys
 
 # Import encoding now, to avoid implicit import later.
 # Implicit import within threads may cause LookupError when standard library is in a ZIP,
@@ -21,7 +22,6 @@ from .structures import CaseInsensitiveDict
 
 from .auth import HTTPBasicAuth
 from .cookies import cookiejar_from_dict, get_cookie_header, _copy_cookie_jar
-from .packages import idna
 from .packages.urllib3.fields import RequestField
 from .packages.urllib3.filepost import encode_multipart_formdata
 from .packages.urllib3.util import parse_url
@@ -331,6 +331,22 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         if self.method is not None:
             self.method = to_native_string(self.method.upper())
 
+    @staticmethod
+    def _get_idna_encoded_host(host):
+        try:
+            from .packages import idna
+        except ImportError:
+            # tolerate the possibility of downstream repackagers unvendoring `requests`
+            # For more information, read: packages/__init__.py
+            import idna
+            sys.modules['requests.packages.idna'] = idna
+
+        try:
+            host = idna.encode(host, uts46=True).decode('utf-8')
+        except idna.IDNAError:
+            raise UnicodeError
+        return host
+
     def prepare_url(self, url, params):
         """Prepares the given HTTP URL."""
         #: Accept objects that have string representations.
@@ -368,17 +384,17 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         if not host:
             raise InvalidURL("Invalid URL %r: No host supplied" % url)
 
-        # In general, we want to try IDNA encoding every hostname, as that
-        # allows users to automatically get the correct behaviour. However,
-        # we’re quite strict about IDNA encoding, so certain valid hostnames
-        # may fail to encode. On failure, we verify the hostname meets a
-        # minimum standard of only containing ASCII characters, and not starting
-        # with a wildcard (*), before allowing the unencoded hostname through.
-        try:
-            host = idna.encode(host, uts46=True).decode('utf-8')
-        except (UnicodeError, idna.IDNAError):
-            if not unicode_is_ascii(host) or host.startswith(u'*'):
+        # In general, we want to try IDNA encoding the hostname if the string contains
+        # non-ASCII characters. This allows users to automatically get the correct IDNA
+        # behaviour. For strings containing only ASCII characters, we need to also verify
+        # it doesn't start with a wildcard (*), before allowing the unencoded hostname.
+        if not unicode_is_ascii(host):
+            try:
+                host = self._get_idna_encoded_host(host)
+            except UnicodeError:
                 raise InvalidURL('URL has an invalid label.')
+        elif host.startswith(u'*'):
+            raise InvalidURL('URL has an invalid label.')
 
         # Carefully reconstruct the network location
         netloc = auth or ''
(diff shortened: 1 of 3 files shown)


moin18:
@Lukasa Please let me know your thought on this. I think it should address your concern in my previous Pull Request.

requests/packages/__init__.py
```diff
@@ -34,9 +34,3 @@
 except ImportError:
     import chardet
     sys.modules['%s.chardet' % __name__] = chardet
-
-try:
-    from . import idna
-except ImportError:
-    import idna
-    sys.modules['%s.idna' % __name__] = idna
```
Lukasa, line 42: So, here's the crux of the issue: I'm not sure we can remove this.

@sigmavirus24, we added these blocks to help tolerate the possibility of downstream repackagers unvendoring us. Do we still believe we need this?

moin18: @Lukasa : In order to prevent the same behavior, I think we should still be using the logic i.e. `import` with `try..except` for `idna`. But remove this code block from `__init__.py` to the place where we are importing `idna` i.e. within `_get_idna_encoded_host()` of my PR in `models.py`. What are your thoughts on this?

sigmavirus24: @Lukasa we do still need this. That said, I think we can hide it in a function performs the import though.

moin18: @Lukasa @sigmavirus24 Updated the pull request.

requests/models.py
```diff
@@ -331,6 +330,22 @@ def prepare_method(self, method):
...
 
+    @staticmethod
+    def _get_idna_encoded_host(host):
+        try:
+            from .packages import idna
+        except ImportError:
+            # tolerate the possibility of downstream repackagers unvendoring `requests`
+            # For more information, read: packages/__init__.py
+            import idna
+            sys.modules['%s.idna' % __name__] = idna
```
sigmavirus24: This does not work. `__name__` here is `requests.models`, not `requests.packages` which is what we *want*.

The function to import `idna` belongs in `requests/packages/__init__.py`, not here.

moin18: @sigmavirus24 Fixed

requests/models.py
```diff
@@ -368,17 +384,19 @@ def prepare_url(self, url, params):
         if not host:
             raise InvalidURL("Invalid URL %r: No host supplied" % url)
 
-        # In general, we want to try IDNA encoding every hostname, as that
-        # allows users to automatically get the correct behaviour. However,
-        # we’re quite strict about IDNA encoding, so certain valid hostnames
+        # In general, we want to try IDNA encoding every hostname if unicode string is
```
nateprewitt: We may want to consider rewording this comment a bit since it no longer matches the logic below.

Perhaps something like,

>In general, we want to try IDNA encoding the hostname if the string contains
  non-ASCII characters. This allows users to automatically get the correct IDNA 
  behaviour. For strings containing only ASCII characters, we need to also verify
  it doesn't start with a wildcard (*), before allowing the unencoded hostname
  through.

moin18: Updated comment

moin18:
@sigmavirus24  Not sure why it is showing the status as "Changes requested" even though they are fixed in commit: https://github.com/kennethreitz/requests/pull/3789/commits/965eb3d1525e3d384623c0267654c2f07acc4aea

May be it due to additional import of `sys` (needed but was missing in previous PR)

nateprewitt:
@moin18 that will be in place until sigmavirus24 has a chance to review your changes again. Today is still a "holiday" for a lot of people working in the US, so sigmavirus24 likely won't respond immediately. Things look pretty cleaned up now, so I think you can leave these changes and he'll respond when he has a spare moment :)

requests/models.py
moin18: @sigmavirus24 Does this looks ok now?

moin18:
@sigmavirus24  I have fixed the comments you mentioned. Please can you review them once more? :)

sigmavirus24:
@moin18 would you be willing to rebase this rather than merging master into this branch? Alternatively, @Lukasa how do you feel about squash merging this?

Lukasa:
I have no objection to squash merging.

moin18:
@sigmavirus24 @Lukasa cleaned up the commit history. Merged all commits into single commit.

requests/models.py
```diff
@@ -368,17 +384,17 @@ def prepare_url(self, url, params):
...
+        # non-ASCII characters. This allows users to automatically get the correct IDNA
+        # behaviour. For strings containing only ASCII characters, we need to also verify
+        # it doesn't start with a wildcard (*), before allowing the unencoded hostname.
+        if not unicode_is_ascii(host):
+            try:
+                host = self._get_idna_encoded_host(host)
+            except UnicodeError:
                 raise InvalidURL('URL has an invalid label.')
+        elif host.startswith(u'*'):
+            raise InvalidURL('URL has an invalid label.')
```
Lukasa, line 397: Ok, so I think this change represents a possibly-breaking behaviour. I am not *certain* of that fact, but it seems to me that we should be willing to move this into a semver-minor release to allow for the possibility that this reordering breaks some people's code. Do you agree @sigmavirus24?

sigmavirus24: I agree with putting it in 2.13 or 2.14. I'm not sure I see where this could break things but that said, I'm also not super familiar with the way IDNA works and how we can determine if something is valid here.

Lukasa: Ok, this suggests to me that sometime over the next few weeks I should push out a urllib3 release and then we should look at a new requests minor release.

It'll be a pretty minor minor release (there are very few changes in Requests), but that's probably a good thing: would be nice to have a quiet minor release for once!

How does that sound to you @sigmavirus24?

sigmavirus24: 👍 Perhaps we should push out a bug fix requests release before that just to be safe. If people have problems with 2.13.0 they have a recent bug-fix release to fall back onto in that case. I'm not sure if we have had that many bug fixes merged since our last release though, so it may also be entirely unnecessary. ;)

Lukasa: We've had basically only one bugfix that affects code: #3791. Everything else is either documentation or test code.

Lukasa: Still, version numbers are free, so we could ship a 2.12.5 with that bugfix in it and then move on from there.

sigmavirus24: Yeah that would be good. Means that anyone who needs #3791 but runs into problems with this would have something to fall back on. =)

moin18: Is there anything which is needed to be taken care of from my end?

PR #3795, opened 2016-12-30
Require pytest-mock for the tests

AdamWill:
test_requests.py `test_session_close_proxy_clear` uses the
`mocker` fixture, which is provided by pytest-mock.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ packages = [
 ]
 
 requires = []
-test_requirements = ['pytest>=2.8.0', 'pytest-httpbin==0.0.7', 'pytest-cov']
+test_requirements = ['pytest>=2.8.0', 'pytest-httpbin==0.0.7', 'pytest-cov', 'pytest-mock']
 
 with open('requests/__init__.py', 'r') as fd:
     version = re.search(r'^__version__\s*=\s*[\'"]([^\'"]*)[\'"]',


AdamWill:
Now I look at it, the pytest-httpbin required version also differs between setup.py and requirements.txt (as a downstream distributor I also much prefer `>=` to `==`, but...:>)

nateprewitt:
Perhaps it would be worth reducing test_requirements to a non-versioned list like [urllib3 does](https://github.com/shazow/urllib3/blob/master/setup.py#L48-L54)? That allows a minimum starting point for someone running tests through setup.py, while avoiding having to remember updating dependencies in multiple places. It looks like the chosen versions in test_requirements were simply the latest ones available when test_requirements was added (6c2942b).

AdamWill:
FWIW, we distributors like `install_requires` and `test_requires` over the 'everything together' format of `requirements.txt` because we have the same split at the distro level: we need to have all the test requirements present when *building* the package (assuming we're going to run the test suite) - so as `BuildRequires`, in RPM terms - but we only need the generated packages to require the stuff from `install_requires` at install time (so as `Requires` in RPM terms). Since `requirements.txt` draws no distinction it's hard for us to use.

Lukasa:
We have no `requirements.txt` requirement: all the setup.py stuff should work. It's just that requirements files can be handy for things like tox, which can try to be clever about tracking changes on it. 

Regardless, we should being setup.py into line with requirements.txt for testing purposes. I see no reason to remove the version pins at this time.

PR #3796, opened 2017-01-05
updating https demo urls

nateprewitt:
The doc examples for SSL Cert Verification are pointing at an older url that no longer serves a response which is causing the example to hang indefinitely. This updates the urls to the current, correct domain.

Also removing a swap file that I accidentally snuck in a couple months ago 😅


kennethreitz:
✨🍰✨

PR #3802, opened 2017-01-09
Add **kwargs to cookiejar_from_dict

Atterratio:
It's can be useful for set not default cookies attributes values, for example "domain" or "discard".

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -500,7 +500,7 @@ def morsel_to_cookie(morsel):
     )
 
 
-def cookiejar_from_dict(cookie_dict, cookiejar=None, overwrite=True):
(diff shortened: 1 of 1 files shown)


Lukasa:
So, while we can do this, are you not likely to be better served by creating a CookieJar yourself at this point? `cookiejar_from_dict` as a utility function is mostly intended for internal use to support the usage of dict in our APIs: we kind of expect that users who want to get more granular control over cookie settings will simply directly use a CookieJar.

sigmavirus24:
I agree with @Lukasa. While I know a significant number of people use this function, it's not a public API that we support and even so, this is the one of the least optimal ways of accomplishing what you want.

PR #3804, opened 2017-01-10
Improve discoverability of OAuth 2 support

ncoghlan:
The previous summary gave the impression that requests-oauthlib only supports
OAuth 1.

This updates makes it clear that it also supports OAuth 2, and links directly to the
use case specific authentication flow guides.

### DIFF
--- a/docs/user/authentication.rst
+++ b/docs/user/authentication.rst
@@ -60,7 +60,7 @@ OAuth 1 Authentication
 ----------------------
 
 A common form of authentication for several web APIs is OAuth. The ``requests-oauthlib``
-library allows Requests users to easily make OAuth authenticated requests::
+library allows Requests users to easily make OAuth 1 authenticated requests::
 
     >>> import requests
     >>> from requests_oauthlib import OAuth1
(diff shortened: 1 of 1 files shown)


ncoghlan:
I'm currently working on adding OpenID Connect support to a pre-existing web application (https://github.com/fedora-infra/anitya/issues/329) and found it very hard to find good documentation on writing a full semi-automated integration test where the only manual step is the user authorising a temporary access token.

Part of the problem was hitting this page in the docs and thinking that requests didn't have native OAuth 2 support - it was only after clicking through to the full requests-oauthlib documentation that I realised that that initial impression was wrong.

ncoghlan:
Thanks for having the code available to find!

You may be amused by the horrible hack I wrote to automate obtaining a refresh token for Anitya's Fedora Account System integration tests: https://github.com/ncoghlan/anitya/blob/authenticated-api/request_oidc_credentials.py :)

glensc:
permalink for above broken link:
https://github.com/release-monitoring/anitya/blob/c369715f5ae7873066ca92b7ae3c412f01711e16/request_oidc_credentials.py

PR #3809, opened 2017-01-11
Remove unused module from tests

JungWinter:
merge_cookies and cookiejar_from_dict functions are already tested by other test functions.


Lukasa:
Thanks for this @JungWinter! :sparkles: :cake: :sparkles:

PR #3814, opened 2017-01-13
Add ability to load ca bundle from data

Kentzo:
SSLContext.load_verify_locations allows to specify cadata with 2.7.9+, 3.4+

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -175,3 +175,4 @@ Patches and Suggestions
 - Hussain Tamboli <hussaintamboli18@gmail.com> (`@hussaintamboli <https://github.com/hussaintamboli>`_)
 - Casey Davidson (`@davidsoncasey <https://github.com/davidsoncasey>`_)
 - Andrii Soldatenko (`@a_soldatenko <https://github.com/andriisoldatenko>`_)
+- Ilya Kulakov (`@kentzo <https://github.com/kentzo>`_)
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -221,10 +221,15 @@ class HTTPAdapter(BaseAdapter):
 
             conn.cert_reqs = 'CERT_REQUIRED'
 
-            if not os.path.isdir(cert_loc):
-                conn.ca_certs = cert_loc
-            else:
-                conn.ca_cert_dir = cert_loc
+            try:
+                if os.path.isfile(cert_loc):
+                    conn.ca_certs = cert_loc
+                elif os.path.isdir(cert_loc):
+                    conn.ca_cert_dir = cert_loc
+                else:
+                    conn.ca_certs_data = cert_loc
+            except:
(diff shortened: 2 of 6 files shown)


Lukasa:
Thanks for this! Some notes:

- Firstly, this patch contains changes to the urllib3 project. This is a sub-project vendored in its entirety unchanged into Requests.  That means any changes to urllib3 need to be made on that project, and they will then make their way into Requests over time.
- Secondly, I am not sure how important this patch is. In the current Requests release it is possible to pass a SSLContext in to a transport adapter, as discussed in a few other issues (e.g. [here](https://github.com/kennethreitz/requests/issues/3774#issuecomment-267871876)). This means it's already possible to use cadata, which drastically reduces the urgency of this patch.
- Thirdly, I suspect in v3 we'll want to refactor this so that verify can accept an `SSLContext` directly, rather than continuing to overload it with more stringly-typed inputs.

Because of the first reason I cannot merge this patch directly: you'd need to open a PR on urllib3 first. However, you may want to consider whether it's worth pursuing the "pass an SSLContext" approach instead.

Kentzo:
Agree, passing custom SSLContext solves the issue.

Could you update documentation to better reflect that, with an example of passing custom SSLContext that loads its certificates from memory?

Kentzo:
@Lukasa One question: won't implementation of `ssl_wrap_socket` override custom CA certificates by calling load_verify_locations inside?

Kentzo:
@Lukasa That's how I did it so far:

```python


class CustomSSLContextAdapter(HTTPAdapter):
    def __init__(self, *args, ssl_context=None, **kwargs):
        super().__init__(*args, **kwargs)
        self._ssl_context = ssl_context

        kwargs['ssl_context'] = self._ssl_context
```

Lukasa:
It's certainly possible. You may need to make further adapter overrides so you can set `ca_certs` and `ca_cert_dir` to `None`.

Kentzo:
@Lukasa Something like this:

```python
class CustomSSLContextAdapter(HTTPAdapter):
    """
    HTTP adapter with custom SSLContext.

    >>> from requests.sessions import Session
    >>>
    >>> with Session() as session:
    >>>     session.mount('https://', CustomSSLContextAdapter(ssl_context=...))
    >>>     session(method=..., url=..., ...)
    """
    def __init__(self, *args, ssl_context=None, **kwargs):
        """
        @param ssl_context: Custom SSL context.
        @type ssl_context: ssl.SSLContext
        """
        self._ssl_context = ssl_context
        super().__init__(*args, **kwargs)

        kwargs['ssl_context'] = self._ssl_context

    def cert_verify(self, conn, *args, **kwargs):
        super().cert_verify(conn, *args, **kwargs)
        conn.ca_certs = None
        conn.ca_cert_dir = None
```
?

Lukasa:
I think that should be fine.

Issue #3213, opened 2016-05-21
Importing requests is fairly slow


cournape:
We are using requests in our projects, and it is working great. Unfortunately, for our CLI tools, using requests is an issue because it is slow to import. E.g. on my 2014 macbook:

```
python -c "import requests"
```

Takes close to 90 ms.

Is optimizing import time worth of consideration for the project ?

Lukasa:
This is almost certainly the result of CFFI. Can you print what libraries you have installed in your environment (`pip freeze`)?

cournape:
This is not linked to cffi (I don't have it installed). I did the test w/ a quasi empty virtualenv:

```
$ pip list
pip (8.1.2)
requests (2.10.0)
setuptools (21.1.0)
wheel (0.24.0)
$ time python -c "import requests"

real    0m0.090s
user    0m0.061s
sys 0m0.026s
```

sigmavirus24:
I have a 2012 Macbook pro:, using a slightly different way of measuring the import time:

```
$ pip list
pip (8.0.2)
requests (2.10.0)
setuptools (19.6.2)
wheel (0.26.0)
$ python -m timeit -n 10000000 'import requests'
10000000 loops, best of 3: 0.571 usec per loop
```

I also upgraded pip and setuptools to match your environment:

```
$ pip list
pip (8.1.2)
requests (2.10.0)
setuptools (21.1.0)
wheel (0.26.0)
$ python -m timeit -n 10000000 'import requests'
10000000 loops, best of 3: 0.59 usec per loop
```

Further

```
$ time python -c 'import requests'
python -c 'import requests'  0.07s user 0.05s system 61% cpu 0.200 total
```

Which is still not ideal, but is probably higher than the simple import of requests due to Python's initialization.

sigmavirus24:
That said, if you have information about specific things in our `__init__.py` (and elsewhere) that are causing slowness, I think we can fix them so long as they don't break backwards compatibility or our functional API.

cournape:
`timeit` is not informative for imports, you have to create a new process every time because of python import caching.

The main cost is importing urllib3, so I suspect that should be addressed there.

nicktimko:
@sigmavirus24 as @cournape mentions you can't time it like that; try using `time` on the command line. On my machine (ran each ~6-10 times to try to get a reliable average) for user+sys times (including sys because if there's any strange calls made into the kernel as a result of the import, that should be counted):
[...]

PR #3817, opened 2017-01-13
Only load .packages.urllib3.contrib.pyopenssl if we have an old version of OpenSSL.

dsully:
See issue #3213

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -56,7 +56,7 @@ __copyright__ = 'Copyright 2016 Kenneth Reitz'
 # See https://github.com/kennethreitz/requests/issues/3213 as well.
 import ssl
 
-if ssl.OPENSSL_VERSION_INFO < (1, 0, 1):
+if hasattr(ssl, 'OPENSSL_VERSION_INFO') and getattr(ssl, 'OPENSSL_VERSION_INFO') < (1, 0, 1):
     try:
         from .packages.urllib3.contrib import pyopenssl
         pyopenssl.inject_into_urllib3()
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -48,11 +48,20 @@ __license__ = 'Apache 2.0'
 __copyright__ = 'Copyright 2016 Kenneth Reitz'
 
 # Attempt to enable urllib3's SNI support, if possible
-try:
-    from .packages.urllib3.contrib import pyopenssl
-    pyopenssl.inject_into_urllib3()
-except ImportError:
-    pass
+# If the user has a recent version of OpenSSL, this
+# isn't needed as Python's core ssl module handles SNI, etc correctly.
+# If the requirements of urllib3.contrib.pyopenssl change,
+# this version should be modified appropraitely.
+#
+# See https://github.com/kennethreitz/requests/issues/3213 as well.
+import ssl
+
+if ssl.OPENSSL_VERSION_INFO < (1, 0, 1):
+    try:
+        from .packages.urllib3.contrib import pyopenssl
+        pyopenssl.inject_into_urllib3()
+    except ImportError:
+        pass
 
 import warnings
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -175,3 +175,4 @@ Patches and Suggestions
 - Hussain Tamboli <hussaintamboli18@gmail.com> (`@hussaintamboli <https://github.com/hussaintamboli>`_)
 - Casey Davidson (`@davidsoncasey <https://github.com/davidsoncasey>`_)
 - Andrii Soldatenko (`@a_soldatenko <https://github.com/andriisoldatenko>`_)
+- Dan Sully (`@dsully <https://github.com/dsully>`_)


Lukasa:
Ok, @alex and @reaperhulk really need to sign off on this before I consider merging it. ;)

requests/__init__.py
```diff
@@ -48,11 +48,20 @@
...
-    pass
+# If the user has a recent version of OpenSSL, this
+# isn't needed as Python's core ssl module handles SNI, etc correctly.
+# If the requirements of urllib3.contrib.pyopenssl change,
+# this version should be modified appropraitely.
+#
+# See https://github.com/kennethreitz/requests/issues/3213 as well.
+import ssl
+
+if ssl.OPENSSL_VERSION_INFO < (1, 0, 1):
```
reaperhulk: pyopenssl uses cryptography, which can (and frequently is) linked against a different version of OpenSSL from the stdlib. This check would prevent it from loading in those cases.

dsully: Quite true. Looks like I'll need to move this loader into a different module in order to opt-out then. Do you agree?

reaperhulk: What do you perceive as the benefit of doing a version check here? To verify what version of cryptography you've got you need to attempt to import it. If it imports successfully and the version is new enough then you would attempt to import pyOpenSSL, but we already effectively do that. I suppose a version check could potentially catch a mismatch between cryptography version and the minimum required by pyOpenSSL (since pip doesn't do much in the way of version resolution this can occur in some scenarios), but that check would more properly be part of pyOpenSSL.

The check present in the currently vendored version of urllib3 is insufficiently robust to catch the scenario where pyOpenSSL < 0.14 was present, but that's been updated (see: https://github.com/shazow/urllib3/pull/1063) and will be in a future requests.

dsully: Sorry - let's unwind the stack a bit here. I have an environment where pyOpenSSL is present, but I don't want to use this automatically imported module, because I'm guaranteed to have a modern Python and a recent version of OpenSSL. I want to reduce imports and import time. Loading `contrib.pyopenssl` brings in a bunch of code I don't want, including `pkg_resources` via `cryptography`.  The OpenSSL version check was a potential work-around proposed by @sigmavirus24. I just want a mechanism to opt-out of trying to do this import at all.

dsully: @reaperhulk To be clear, I could hack it in my code like this:
```
sys.modules['requests.packages.urllib3.contrib'] = object()
```

But would prefer a first class way of opting out that isn't a hammer like the above. :)

reaperhulk: Ah, okay I understand. I'm not sure I think a version check is the appropriate mechanism here, although ultimately that'd be the maintainers' call. My reservation about it is scenarios like this:

* User has Python linked against 1.0.1 or 1.0.2 (or even 1.1.0!)
* User has pyOpenSSL and cryptography (linked against OpenSSL 1.1.1 because they got it via a wheel and we linked the latest version)

In the above a version check would not import pyOpenSSL, but pyOpenSSL would actually provide TLS 1.3, which is a desirable feature. An explicit opt-out seems like a better path, although I admit I am not entirely sure how one would accomplish that short of `os.getenv('NO_PYOPENSSL', None)`

dsully: @Lukasa already nixed the environment variable option. So far my best bet is the `sys.modules` hack above.

requests/__init__.py
```diff
@@ -56,7 +56,7 @@
 # See https://github.com/kennethreitz/requests/issues/3213 as well.
 import ssl
 
-if ssl.OPENSSL_VERSION_INFO < (1, 0, 1):
+if hasattr(ssl, 'OPENSSL_VERSION_INFO') and getattr(ssl, 'OPENSSL_VERSION_INFO') < (1, 0, 1):
```
drpoggi, line 59: So this disables all the SSL niceties Requests and cryptography provide for someone on 2.6 that has installed `requests[security]`. I'm not sure how much the Requests maintainers care about that particular environment config, but as it stands it's currently supported (albeit with a strong warning from cryptography to update your Python version, so I'm not sure how much longer this will remain to be true.)

Lukasa:
Ok, all commenters, *please* read some of the discussion on #3213 before making further comments here. You'll find it helps you avoid retreading some ground. ;)

@reaperhulk, my main concern here is that we are fundamentally looking at adding this patch because "importing PyOpenSSL is slow". My question is: do you consider that a bug?

reaperhulk:
It would help to quantify slow. In the long long ago (pre-cryptography 1.0) importing pyOpenSSL was slow (1+ seconds), but on my 4 year old laptop I just tested and it took about 50ms to import pyOpenSSL 16.2 + cryptography 1.7.  `import requests` with no pyOpenSSL installed takes ~63ms and ~108ms for requests+pyOpenSSL. That doesn't seem particularly problematic.

reaperhulk:
It'd also help if I took @Lukasa's advice and actually read the whole issue that spawned this.

I don't consider the current import performance to be a bug, but if someone wants to try to improve it on cryptography (which is probably the dominant source of import time for pyopenssl) I'm happy to review. The biggest single source of time is likely when it loops over the lib object to build a new conditional lib object. That could be optimized if cffi supported a means of conditional binding, but we're likely talking only ~10ms?

Lukasa:
So step 1 is: are we talking about the same level of "slow" here?

sigmavirus24:
@Lukasa @reaperhulk so it's worth noting that `slow` will change based on how many things are installed + how many possible entry-points exist. `pkg_resources`, if I remember correctly, will scan all of `site_packages` for entry-points. So if you're testing `pyOpenSSL` in a fresh virtual environment, you're import speed (given that cryptography apparently scans entry-points at import) is going to be faster than @dsully since they seem to be installing a lot of things in one `site_packages` directory.

Granted, this is a fundamental flaw of how `pkg_resources` works, but I think it's still a legitimate problem. I haven't looked at what cryptography uses `pkg_resources` to find at import, but there would only be "import time" benefits to avoiding that scan rather than any real performance benefit to not doing it when cryptography is imported.

I'd also like a better understanding of `slow` from @dsully + maybe a better description of how much is being installed into their site-packages directory.

sigmavirus24:
Also, to set everyone's expectations appropriately, this appears to be a problem @dsully is encountering at work, so I completely expect them to *not* respond until Monday during work hours.

requests/__init__.py
drpoggi: @Lukasa I read that issue, and maybe I'm just dense, but I don't see where dropping `requests[security]` support for 2.6 is on the table, or maybe my understanding is wrong and that was never supported. The `OPENSSL_VERSION_INFO` attribute was added in 2.7.

requests/__init__.py
```diff
@@ -48,11 +48,20 @@
...
-    pass
+# If the user has a recent version of OpenSSL, this
+# isn't needed as Python's core ssl module handles SNI, etc correctly.
+# If the requirements of urllib3.contrib.pyopenssl change,
+# this version should be modified appropraitely.
+#
+# See https://github.com/kennethreitz/requests/issues/3213 as well.
+import ssl
+
+if hasattr(ssl, 'OPENSSL_VERSION_INFO') and getattr(ssl, 'OPENSSL_VERSION_INFO') < (1, 0, 1):
```
sigmavirus24, line 59: Why not more simply:

```py
_OPENSSL_VERSION = getattr(ssl, 'OPENSSL_VERSION_INFO', None)
if _OPENSSL_VERSION is None or _OPENSSL_VERSION < (1, 0, 1):
    try:
        # ...
    except ImportError:
        pass
```

This then will import pyOpenSSL on 2.6 *or* in situations where the built-in OpenSSL isn't quite so great.

requests/__init__.py
sigmavirus24: @drpoggi so, a few things:

1. No one is calling you dense
2. No one is talking about dropping support for `requests[security]` on Python 2.6
3. This is a provisional pull request. More investigation is necessary prior to merging it.

   Specifically, we need to know what "slow" means to @dsully and figure out how to reproduce that slowness to see if we can fix it.

Finally, I don't think you intended it* but your tone in your comments is a bit hostile. Please take a step back and come back with a less adversarial approach. Your contributions have been valuable in the past, and I have reason to believe that with a more constructive and collaborative tone they'll continue to be now and in the future. 🎉 

*intentions aren't magical - they don't fix anything.

drpoggi: @sigmavirus24 I didn't mean to come off as hostile at all, apologies for that. I didn't think anyone else was calling me dense, I thought that was myself just _being_ dense, which I would like to retain the right to insult myself (only) 😉 . Thank you for your feedback, still trying to get a handle on my tone when contributing to FOSS, will adjust.

requests/__init__.py
piotr-dobrogost: > Loading contrib.pyopenssl brings in a bunch of code I don't want, including pkg_resources via cryptography.

I didn't look at the code but it sounds like some package in the chain of imports (cryptography?) does too much work (probably finding/registering plugins) in the init module which in turn requires using pkg_resources. If that's the case then we have classic problem of unwanted side effects during import. The solution would be to either make this work explicit instead of implicit by requiring user of the library to request this functionality or to postpone work until it's really needed by introducing some laziness. The former makes the usage of library marginally harder and the latter is not always possible.

requests/__init__.py
sigmavirus24: No worries @drpoggi. I think we're all striving to be better at this F/OSS collaboration thing.

reaperhulk:
Okay, this is getting a bit more interesting. `pkg_resources` is invoked to determine what backends are available to cryptography. This is done in `cryptography.hazmat.backends.__init__`. requests doesn't need that check since it directly imports the backend it requires. However, if you look in that you can see that we **are** doing it lazily. To summarize a wasted hour here, `import pkg_resources` is actually where we pay the speed penalty. I haven't looked into what's going on inside `pkg_resources`, but it seems likely that a side effect of importing is an immediate scan of the packages?

Anyway, this points us to a potential significant performance improvement: just `import pkg_resources` inside the function call so it is lazily imported as well. On my machine with a test virtualenv containing 100 packages this drops the import time by 150ms.

sigmavirus24:
@reaperhulk If the cryptography team is okay with moving the import inside the function that seems suitable and should eliminate the need for this PR based on your findings.

dsully:
@reaperhulk Yes - removing the pkg_resources import until it's needed will certainly help. I'll create a PR over there to move the import.

@sigmavirus24 nails it - we have a lot installed in site-packages, and are extremely sensitive about CLI tools start up time. Every millisecond counts in the eyes of our users. I do agree this is a fundamental flaw in pkg_resources. (My weekend was pretty packed, so I wasn't able to reply until now).

I do feel that requests automatically trying to use pyopenssl if it's installed without any way to opt-out is surprise functionality. I do have a work around for now, so this PR can be dropped.

Lukasa:
> I do feel that requests automatically trying to use pyopenssl if it's installed without any way to opt-out is surprise functionality.

I mean, you can opt-out *after* import with no difficulty at all. ;)

PR #3819, opened 2017-01-14
Fix #3818. avoid doing an import inside a method

### LINKED ISSUE #3818: Import error with idna version
Hello,

I recently upgraded to 2.12 and I started to see this import error:

> (Pdb) request('get', url, params=params, **kwargs)
> *** ImportError: No module named uts46data

Looking in the changes I saw that it was recently added idna. 
My projects are using import hooks and inside of it the hook use requests.

The code of idna is doing an import inside the method uts46_remap, which is breaking my code.
Moving the import uts46_remap outside fixes this issue.

llazzaro:


Lukasa:
Thanks for this!

The idna module is vendored without changes from [here](https://github.com/kjd/idna). We do not carry patches to vendored modules in our own source tree. Please propose this pull request upstream, and we will pull in the changes when a new release is cut.

llazzaro:
@Lukasa thanks for the commet! I already propose the changes on that repository.

PR #3827, opened 2017-01-24
Release v2.13.0

sigmavirus24:
This includes updates to:

- when we load idna
- urllib3 (updated to version 1.20)
- idna (updated to version 2.2)
- release publishing machinery


nateprewitt:
It looks like `init` in the Makefile may need `pipenv shell` run to initialize the virtualenv?

Lukasa:
Maybe? pipenv 0.1.14 worked, 0.2.9 doesn't. The changelog is not immediately forthcoming as to what the specific problem is.

Lukasa:
For now we'll just pin it so the builds keep working: @kennethreitz should feel free to remove the pin and investigate when he has some time.

sigmavirus24:
Just tested with pipenv 0.2.7 locally and that seems to work, so I'm going to say it's just 0.2.8 that's busted. We can blacklist it after this is merged and 2.13.0 is released

Lukasa:
BAM.

PR #3828, opened 2017-01-24
Blacklist known bad version of pipenv

sigmavirus24:
We pinned pipenv to release v2.13.0 but in reality, we could have just
blacklisted the known bad version. For future us, we now blacklist it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: docs
 
 init:
-	pip install 'pipenv===0.2.0'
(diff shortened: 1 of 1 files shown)


nateprewitt:
I think it's more than just 2.8.0, I can't get a working build on travis after 2.5.0. It looks like something happened there that's breaking the build, checking it out now.

sigmavirus24:
Odd, 0.2.7 worked just fine locally

sigmavirus24:
We can blacklist everything between 0.2.5 and 0.2.8 though

PR #3830, opened 2017-01-24
Fixing Pipfile source

nateprewitt:
This fixes the issue that was encountered this morning during the 2.13.0 release.

The source specified in the original Pipfile was for an incorrect pypi endpoint. This didn't matter until pipenv 2.6 was released, specifically [cb22a12](https://github.com/kennethreitz/pipenv/commit/cb22a129eae8c8e800e603c38bf1fe04d420fbde), which started using the provided `source` value.

@kennethreitz I used `pipenv lock` with pipenv 3.0.0 to generate the new lock file which I'm assuming is what we want here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: docs
 
 init:
-	pip install 'pipenv>=0.1.6,!=0.2.6,!=0.2.7,!=0.2.8'
+	pip install 'pipenv>=0.1.6'
 	pipenv install --dev
 
 test:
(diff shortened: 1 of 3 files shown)


nateprewitt:
While we're here, it looks like this value is still used in pipenv [here](https://github.com/kennethreitz/pipenv/blob/3fa204b734f67522d7858ab47b1d10263a76dc90/pipenv/_pipfile/api.py#L73) , but I'm not sure if that code is actively being used. I can open a quick PR updating it over there too if needed.

kennethreitz:
ah yes, this is the issue.

PR #3832, opened 2017-01-25
fixing codecov with pipenv

nateprewitt:
codecov broke with the introduction of pipenv because the dependency isn't installed outside of the virtualenv. Running it inside of pipenv fixes this.


Lukasa:
LGTM, thanks @nateprewitt!

Issue #3772, opened 2016-12-16
Digest Auth will auth regardless of status code


Lukasa:
Discovered on IRC.

The digest auth handler in Requests' codebase doesn't ever actually check that it is responding to a 401: it just looks for an Authorization header. That's pretty dumb, so we should add a check to only actually do the execution for 401s.

This is a contributor friendly change.

nateprewitt:
Are we sure that tightening this check is what we want to do? It looks like [RFC 7235](https://tools.ietf.org/html/rfc7235#section-4.1) suggests that it's acceptable to send a WWW-Authenticate challenge with a non-401 status code. I'm not sure `handle_401` is the most apt name for the hook, but the functionality seems correct.

>A server generating a 401 (Unauthorized) response MUST send a
   WWW-Authenticate header field containing at least one challenge.  A
   server MAY generate a WWW-Authenticate header field in other response
   messages to indicate that supplying credentials (or different
   credentials) might affect the response.

sigmavirus24:
@nateprewitt So we have two role models to look at here:

1. Browsers - If you read the logs for our IRC channel, you'll see that Browsers are doing what we do (although probably not as loosely as we do it)

2. Curl/Wget - They disregard the WWW-Authenticate headers on anything that's not a 401. In the case of the user on IRC they were receiving a 403 with WWW-Authenticate headers and we were authenticating while Curl and Wget were not.

Frankly, I'm not sold on restricting this to 401s only, but I do think we should restrict it to 4xx codes. Having these headers sent back on a 30x, for example, could lead to interesting behaviour. On the one hand, we have to answer the challenge, right? On the other we're being redirected, potentially to a totally different domain. I think the only right answer in non 4xx cases is to not authenticate. Unfortunately, we will answer any challenge we receive at the moment and that's problematic.

sigmavirus24:
Also @nateprewitt, please let someone else pick this up.

nateprewitt:
Yeah, I was planning on leaving it on the stack. I just did work on testing this functionality last week which is why I felt the need to comment.

I just implemented tests confirming that we won't send Authorization on a 3xx request unless challenged after the redirect, so I don't think that's an issue we need to worry about. 

I don't think that restricting this to the 4XX range is unreasonable but I do think the RFC is pretty permissive.  I'm just trying to provide information that I couldn't find in the IRC messages that seems relevant. Hopefully helping avoid similar problems that we just had with tightening basic auth.

[...]

PR #3835, opened 2017-01-26
Only send HTTPDigestAuth on 4xx challenges

mmedal:
Resolves: #3772

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -227,6 +227,12 @@ class HTTPDigestAuth(AuthBase):
         :rtype: requests.Response
         """
 
+        # If response is not 4xx, do not auth
+        # See https://github.com/kennethreitz/requests/issues/3772
+        if not 400 <= r.status_code < 500:
+            self._thread_local.num_401_calls = 1
+            return r
+
         if self._thread_local.pos is not None:
             # Rewind the file position indicator of the body to where
             # it was to resend the request.
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -136,6 +136,43 @@ def test_digestauth_401_only_sent_once():
         close_server.set()
 
 
+def test_digestauth_only_on_4xx():
+    """Ensure we only send digestauth on 4xx challenges.
+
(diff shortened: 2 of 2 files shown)


nateprewitt:
@Lukasa I believe there were issues with ONLY doing 401 as noted in #3772. The related RFCs allow for the WWW-Authenticate header in a non-401 response which the client SHOULD respond to if received. I think the conclusion we came to is to support 4XX response codes.

mmedal:
I believe @nateprewitt is correct. It seemed to me that the conclusion of the discussion was to support 4XX response codes.

requests/auth.py
```diff
@@ -227,6 +227,11 @@ def handle_401(self, r, **kwargs):
         :rtype: requests.Response
         """
 
+        # If response is not 4xx, do not auth
```
Lukasa, line 230: Given that I am clearly inclined to forget why we did this, let's add a link to the relevant issue here too. 😁

mmedal:
Done!

mmedal:
Might I also suggest renaming the hook to something like handle_4xx_auth_challenge?

sigmavirus24:
> Might I also suggest renaming the hook to something like handle_4xx_auth_challenge?

I think I'd rather not. Most auth plugins use the same method name for consistency. But I'm not firmly set in that opinion.

Lukasa:
Hrm. @mmedal, can you rebase on top of the current master? We seem to be having some build problems and I want to check whether they're ones we've seen before and fixed, or new ones.

mmedal:
I triggered a new build (my branch is up to date), but it looks like you're still having build problems.

nateprewitt:
@Lukasa this is a bug in pipenv (kennethreitz/pipenv#90) that may not have an immediate solution. It may be best to pin Requests' dependency for a bit. `pipenv==3.1.9` should work for now.

Lukasa:
@nateprewitt I'll merge a PR that does that. 😁

nateprewitt:
On it!

mmedal:
Retriggering build.

PR #3836, opened 2017-01-27
pin pipenv until kennethreitz/pipenv#90 is resolved

nateprewitt:
Fixing build for now until we can properly parse all dependencies in pipenv.


Lukasa:
Thanks @nateprewitt!

Issue #3837, opened 2017-01-28
Proposal: potential way to follow bad url shorteners 


jvanasco:
I've ran into an issue that runs tangent to issue #441.

We have a content indexer that is powered by `requests`.  Some of the larger URL shorteners exhibit some weird behaviors depending on the user agent string.

* (variant of #441) depending on the user-agent header, a particular shortener will either send a proper 301 response OR a HTTP 200 containing a meta-refresh value of 0.

         <head><meta name="referrer" content="always"><noscript><META http-equiv="refresh" content="0;URL={URL}"></noscript><title>{URL}</title></head><script>window.opener = null; location.replace("{URL}")</script>

* (new?) a handful of url shorteners may send a HTTP-200 response with a `location` header to the redirect.  (yes I know, it breaks spec and makes no sense.

To handle both of these (and several other scenarios) I have a novel suggestion.  The callback hooks could be used to allow developers to catch and follow these types of redirect oddities within a single configured request.

A developer would be able to handle edge cases like the above by with a hook:

      def bad_shortener_callback(r):
          r_location = r.headers.get('location')
          if r.status_code == 200 and r_location!= r.url:
                r.is_redirect = True
                r.redirect_location_override = r_location
          return r

In order to make this work, a slight change would be needed: 

The redirect url is pulled via this line:

https://github.com/kennethreitz/requests/blob/f72684e13c5074a671506d29c1b5638156680ea7/requests/sessions.py#L116

    url = resp.headers['location']

I propose this change to sessions.py

    -url = resp.headers['location']
    +url = resp.redirect_location

 and then extending models.py accordingly

    + redirect_location_override = None

    @property
    def redirect_location(self):
          if self.redirect_location_override is not None:
             return self.redirect_location_override
         return self.headers['location']

if this is acceptable, I would be happy to issue a PR that includes tests.

Lukasa:
Thanks for the suggestion!

So, I think I don't fully understand why this needs to be a Requests patch. What advantages would you or other users get from having this integrated into Requests?

jvanasco:
I'm not sure how to interpret your question, so I'm going to answer it a few ways:

And I'll refer to this URL flow, where the 301s are a proper redirect and the 200 is standing in for a malformed response (meta-refresh or 200+location).

	A -> [301] -> B -> [200] -> C -> [301] -> D

1. What is the utility in general?

[...]

PR #3846, opened 2017-02-02
initial attempt at `get_redirect_target`

jvanasco:
This is the first attempt at a PR to address my proposal #3837 to support pluggable redirect handling.  Feedback is welcome.

This does not break existing tests and adds a new test to for handling malformed 200+location responses using a custom session mixin.

I've been testing it for a day and it seems to be fine.

Thank you for simply considering this, and even more thanks for all the suggestions and advice in the related thread.  It has been a long week of little sleep trying to pin down some issues.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,15 @@
 Release History
 ---------------
 
+**Unreleased**
++++++++++++++++++++
+
+- The behavior of ``SessionRedirectMixin`` was slightly altered.
+  ``resolve_redirects`` will now detect a redirect by calling
+  ``get_redirect_target(response)`` instead of directly
+  querying ``Response.is_redirect`` and ``Response.headers['location']``.
+  Advanced users will be able to process malformed redirects more easily.
+
 2.13.0 (2017-01-24)
 +++++++++++++++++++
 
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -86,34 +86,38 @@ def merge_hooks(request_hooks, session_hooks, dict_class=OrderedDict):
 
 
 class SessionRedirectMixin(object):
+
+    def get_redirect_target(self, resp):
+        """Receives a Response. Returns a redirect URI or ``None``"""
+        if resp.is_redirect:
+            return resp.headers['location']
+        return None
+
     def resolve_redirects(self, resp, req, stream=False, timeout=None,
                           verify=True, cert=None, proxies=None, **adapter_kwargs):
         """Receives a Response. Returns a generator of Responses."""
 
-        i = 0
         hist = [] # keep track of history
 
-        while resp.is_redirect:
+        url = self.get_redirect_target(resp)
+        while url:
             prepared_request = req.copy()
 
-            if i > 0:
-                # Update history and keep track of redirects.
-                hist.append(resp)
-                new_hist = list(hist)
-                resp.history = new_hist
+            # Update history and keep track of redirects.
+            # resp.history must ignore the original request in this loop
+            hist.append(resp)
+            resp.history = hist[1:]
 
             try:
                 resp.content  # Consume socket so it can be released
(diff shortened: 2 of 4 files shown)


HISTORY.rst
```diff
@@ -3,6 +3,15 @@
...
 ---------------
 
+**Unreleased**
++++++++++++++++++++
+
+- The behavior of `SessionRedirectMixin` was slightly altered.
+  `resolve_redirects` will now detect a redirect by calling
+  `get_redirect_target(response)` instead of directly
+  querying `Response.is_redirect` and `Response.headers['location'].  Advanced
+  users will be able to process malformed redirects more easily.
```
Lukasa: reStructuredText uses double-backticks to force monospace font, not single-backticks.

requests/sessions.py
```diff
@@ -86,14 +86,22 @@ def merge_hooks(request_hooks, session_hooks, dict_class=OrderedDict):
 
 
 class SessionRedirectMixin(object):
+
+    def get_redirect_target(self, resp):
+        """Receives a Response. Returns a redirect URI or `None`"""
```
Lukasa: Double-backtick.

tests/test_requests.py
```diff
@@ -1740,6 +1740,50 @@ def test_chunked_upload_does_not_set_content_length_header(self, httpbin):
         assert 'Transfer-Encoding' in prepared_request.headers
         assert 'Content-Length' not in prepared_request.headers
 
+    def test_custom_redirect_mixin(self, httpbin):
+        """This tests using a custom mixin to overwrite `get_redirect_target`.
```
Lukasa: Double-backticks

sigmavirus24: We don't really auto-doc our test functions. I'm less concerned about it here.

jvanasco: Yeah, I noticed that -- but the last few tests added to `TestRequests` had docs, and this one can take a minute of staring at to figure out.  I figured it would be best to submit docs and edit as needed.

jvanasco:
I originally hoped to change the while-loop.  I found it a difficult to decipher what is going on, as it's doing a lot of copying/resetting of vars and prepping for the next iteration. I admittedly missed some obvious behaviors in it.

In terms of maintenance and readability, I agree a generator or closure would likely be better.    Tracking the iteration with `i` could be replaced by just referencing `len(history)`too.

kennethreitz:
💫✨🍰✨💫

kennethreitz:
this is great.

jvanasco:
here's a quick edit to get rid of that `i`:

https://github.com/kennethreitz/requests/commit/6c40707b93f98fac9a7982cc5f50521079b155b8

IMHO this is a bit cleaner to understand.  it not only removes the `i` (i found it annoying), but instead of having to handle the weird iteration conditions and copying the history via`list`, just uses an array slice to create a new list excluding the first element.

Lukasa:
Yup, that does seem like a positive step.

jvanasco:
Should I merge that change into this PR?

Lukasa:
Let's give it a try and see how it looks.

kennethreitz:
needs a rebase!

jvanasco:
Rebased, but a conflict on the `AUTHORS.txt` left an ugly merge commit :(

I pulled in the clarification to the while-loop too.

I *really dislike* how the while-loop handles the history, but I'm still trying to clean it up.

Lukasa:
So FWIW, it may be easier to cherry-pick the commits you need to a new branch and force-push it up to this PR. You dragged a whole bunch of other history into the PR which makes reviewing it really hard.

jvanasco:
Actually, I'm going to manually patch and then force-push.  That'll give you one commit. Sorry about this.  Git is often not my friend.

jvanasco:
Cleaned up.  Sorry for the mess.

kennethreitz:
💫✨🍰✨💫

PR #3853, opened 2017-02-06
Do not convert /o\ into /O\

StyXman:
There is code that already check that a guy raising (at least) his right arm does not get big-headed (a.k.a uppercased). Just add the same for the other, poor guy complaining about an error.

### DIFF
--- a/requests/status_codes.py
+++ b/requests/status_codes.py
@@ -87,5 +87,5 @@ codes = LookupDict(name='status_codes')
 for code, titles in _codes.items():
     for title in titles:
         setattr(codes, title, code)
-        if not if not title.startswith(('\\', '/')):
+        if not title.startswith(('\\', '/')):
             setattr(codes, title.upper(), code)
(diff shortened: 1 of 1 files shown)


requests/status_codes.py
```diff
@@ -87,5 +87,5 @@
 for code, titles in _codes.items():
     for title in titles:
         setattr(codes, title, code)
-        if not title.startswith('\\'):
+        if not title.startswith('\\') and not title.strartswith('/'):
```
Lukasa: I don't think `strartswith` is the name of this method.

Lukasa: Additionally, you can write this as `if not title.startswith(('\\', '/'))`, which may be clearer.

StyXman:
Yes, sorry, I just (w)hacked it in 3 minutes without proper testing. Shame on me. Let me see...

requests/status_codes.py
```diff
@@ -87,5 +87,5 @@
 for code, titles in _codes.items():
     for title in titles:
         setattr(codes, title, code)
-        if not title.startswith('\\'):
+        if not if not title.startswith(('\\', '/')):
```
Lukasa: You have duplicated the `if not` here.

Lukasa:
Cool, looks good now, thanks! :sparkles:

StyXman:
\o/

PR #3864, opened 2017-02-10
Proposed/3.0.0

kennethreitz:
# v3.0.0 of Requests

This is a long-polling pull request for the next breaking-change version of Requests — version 3.0.0.

Any feature requests should be added as comments here — this is the time to get anything into Requests, if there's any time at at all. 

**Expected landing date**: *this year*. 

## Upcoming changes

- Remove the HTTPProxyAuth class in favor of supporting proxy auth via the proxies parameter.
- Relax how Requests strips bodies from redirects. 3.0.0 only supports body removal on 301/302 POST redirects and all 303 redirects.
- Remove support for non-string/bytes parameters in _basic_auth_str.
- Prevent Session.merge_environment from erroneously setting the verify parameter to None instead of True.
- Streaming responses with Response.iter_lines or Response.iter_content now requires an encoding to be set if one isn't provided by the server.
- Raise exception if multiple locations are returned during a redirect.
- Update ConnectionPool connections when TLS/SSL settings change.
- Remove simplejson import and only use standard json module.
- Strip surrounding whitespace from urls.
- MissingSchema and InvalidSchema renamed to MissingScheme and InvalidScheme respectively.
- Change merge order for environment settings to avoid excluding Session-level settings.
- Encode redirect URIs as latin-1 before performing redirects in Python 3 to avoid mangling during the requoting process.
- Remove the __bool__ and __nonzero__ methods from a Response object.
- This has been a planned feature for over a year. The behaviour is surprising to most people and breaks most of the assumptions that people have about Response objects. This resolves issue #2002
- Skip over empty chunks in iterators. Empty chunks could prematurely signal the end of a request body's transmission, skipping them allows all of the data through. See #2631 for more details.
- Remove the req argument from Session.resolve_redirects method.
- Rename the resp argument from Session.resolve_redirects to response.
- New PreparedRequest.send method. Now, you can Request().prepare().send().
- All porcelain API functions (e.g. requests.get, etc) now accept an optional session parameter. If provided, the session given will be used for the request, in place of one being created for you.
- URLs are now automatically stripped of leading/trailing whitespace.

## What's left

- Pluggable redirect handler (#3846)
- Anything Kenneth thinks of. 
- Urllib3 v2 update


kennethreitz:
Needs a rebase!

kennethreitz:
Needs a rebase!

AraHaan:
ok, so can someone help me with solving an issue with aiohttp that requests.post can work with so that way I can close the issue with my PR for it? The issue is with the files param and well my PR semi adds it in however I am confused on the implementation for it into aiohttp. So, can someone here help me with that PR?

Issue: https://github.com/KeepSafe/aiohttp/issues/1651
PR: https://github.com/KeepSafe/aiohttp/pull/1653
thanks in advance.

Lukasa:
@AraHaan The relevant code is [here](https://github.com/kennethreitz/requests/blob/master/requests/models.py#L109): this is the code path that Requests uses if `files` has been provided. You'll find it's not remarkably helpful to you because it uses a lot of custom code in both Requests and urllib3, but essentially it encodes both the `files` and `data` arguments as multipart form data.

obskyr:
>All porcelain API functions (e.g. requests.get, etc) now accept an optional session parameter. If provided, the session given will be used for the request, in place of one being created for you.

What's the use case for this? Is there a scenario where passing a session as an argument is comfortable, but calling a method on the session isn't? Is it perhaps to alleviate boilerplate for user-defined functions that take an optional session argument (e.g. `def cool_function(thing, another_thing, session=None):`)?

>URLs are now automatically stripped of leading/trailing whitespace.

I'm also wondering about this one... Is this really desirable? I suppose I can't really come up with a case where you'd want to turn away URLs with leading/trailing whitespace, hahah, but still wondering about the real rationale behind it.

Lukasa:
> What's the use case for this?

It's quite common to have APIs that wrap Requests that pass `**kwargs` to the library. This allows the use of a custom `Session` transparently by those APIs.

ml31415:
Additionally to providing a session parameter (or maybe instead of it), wouldn't it make sense, to create a default session in the `api` module, and keep using this session throughout all porcelain functions? Right now a session gets created on the fly for every request, also throwing away stored cookies etc. This might be quite unintuitive when doing requests consequently in a browser-like fashion in simple scripts.

obskyr:
@ml31415 I don't think that's reasonable. If you want to save state, you should explicitly use a session. Making the top-level functions share a session would make "stateless" requests really cumbersome - you'd have to manually create a new session between each request.

sigmavirus24:
@ml31415 

> Right now a session gets created on the fly for every request, also throwing away stored cookies etc. 

We found that there was actually significantly more confusion about why cookies were saved across porcelain request calls. @obskyr is exactly correct.

ml31415:
Agreed. Though, this stateful behavior might be more what someone expects, when he's looking for a simple library, automating what he's used from using his browser every day. Another option would be to simply initialize a default session, so that the stateful behavior would be more readily accessible:
```
# without state
import requests
requests.get()
# vs with state
from requests import session
session.get()
```
Would save a `session = Session()` line for each simple script out there in system administration and so on. This might be more helpful than having `session()` defined in `sessions` as a kind of synonym for `Session()`, except ignoring all the arguments. The gain in convenience is of course not huge, just saving this one line, but then again, this is what gets discussed in terms of usability e.g. here: https://gist.github.com/kennethreitz/973705

Or if this API change might be too intrusive, how about this instead:

```
from requests import stateful
stateful.get()
```

sigmavirus24:
@ml31415 Please don't try to conflate not having to copy and paste around several lines to get a minimum viable HTTP request to having to include 1 line to have consistent state. Besides the fact that having a module with a global session that can keep state and cause confusion. There's also the problem where having that Session has caused memory leaks in the past. So we're just not going to do it.

PR #3866, opened 2017-02-10
More extensive while-loop altering

jvanasco:
This is offered as an alternate/addition to my other PR, https://github.com/kennethreitz/requests/pull/3846

Specifically the code is in this commit: https://github.com/jvanasco/requests/commit/5e9371854176164db9ce5fee5ceb39b58f4035b2

This passes current tests, but it's a bit of a change so I'm offering it separately.

The existing redirect code was incredibly confusing, because objects were being overwritten mid-loop, select items had to be removed, and then everything needed to be switched back again.

This approach does a major reorganization of the logic, but I think for the better:

# SessionRedirectMixin.resolve_redirects

The `hist` tracker makes more sense.

* `hist` is started off with the original response. 
* after the redirect response is created, it is updated with the "current" history
* `hist` is then updated with the new response

# Session.send

Because the `Request.history` was carefully constructed within the loop context, the history had to then be shuffled around and response updated.  The above change makes that unnecessary, and instead `r ` is just popped off and used as-is.

I think this makes the code easier to read and maintain, and it removes that weird shuffle at the end.  It's a bit of a change though, and might break the API in a way that is currently untested.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -177,3 +177,5 @@ Patches and Suggestions
 - Andrii Soldatenko (`@a_soldatenko <https://github.com/andriisoldatenko>`_)
 - Moinuddin Quadri <moin18@gmail.com> (`@moin18 <https://github.com/moin18>`_)
 - Matt Kohl (`@mattkohl <https://github.com/mattkohl>`_)
+- Jonathan Vanasco (`@jvanasco <https://github.com/jvanasco>`_)
+
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,15 @@
 Release History
 ---------------
 
+**Unreleased**
++++++++++++++++++++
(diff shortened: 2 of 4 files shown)


kennethreitz:
closing in favor of other PR #3846

jvanasco:
Thanks!

It is worth resubmitting the clarified loop as a new PR for discussion?  

Rebased against the current master, it looks like this:  https://github.com/jvanasco/requests/commit/f1cbdfbc7a18e1ac41f8cdcec107a634674ae5ae

and gets rid of some of the janky looping code.

Lukasa:
Certainly I'd be happy to see it. =) Can't speak for Kenneth, of course.

sigmavirus24:
@jvanasco yes but please make that commit message easier on the eyes :)

PR #3868, opened 2017-02-11
Make Response.raise_for_status() return the response object if the response is successful

vmalloc:
This allows for chaining method calls in cases where we want to raise for bad codes but use the response otherwise,
e.g. `requests.get(URL).raise_for_status().json()['value']`

### DIFF
--- a/3.0-HISTORY.rst
+++ b/3.0-HISTORY.rst
@@ -58,5 +58,7 @@
 
 - URLs are now automatically stripped of leading/trailing whitespace.
 
+- ``Response.raise_for_status()`` now returns the response object for good responses
+
 .. _#2002: https://github.com/kennethreitz/requests/issues/2002
 .. _#2631: https://github.com/kennethreitz/requests/issues/2631
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -354,9 +354,13 @@ But, since our ``status_code`` for ``r`` was ``200``, when we call
 ``raise_for_status()`` we get::
 
     >>> r.raise_for_status()
-    None
+    <Response [200]>
 
 All is well.
+
+.. note:: ``raise_for_status`` returns the response object for a successful response. This eases chaining in trivial cases, where we want bad codes to raise an exception, but use the response otherwise:
+
+    >>> value = requests.get('http://httpbin.org/ip').raise_for_status().json()['origin']
 
 
 Response Headers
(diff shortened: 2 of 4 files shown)


requests/models.py
```diff
@@ -902,7 +902,7 @@ def links(self):
         return l
 
     def raise_for_status(self):
-        """Raises stored :class:`HTTPError`, if one occurred."""
+        """Raises stored :class:`HTTPError`, if one occurred. Returns the response object (self) otherwise."""
```
Lukasa: Can you shorten this line please?

vmalloc: Committed and pushed again. Let me know if it's ok now

vmalloc:
@Lukasa added a changelog entry to 3.0-HISTORY. Also the Travis build fails, but it doesn't appear like it's because of my change... not sure how to investigate...

Lukasa:
Yeah, I don't think it is: the 3.0.0 build has a breakage on it that was caused by one of the recent merges. I'll chase it on Monday. In the meantime we'll let this sit, if that's ok?

vmalloc:
@Lukasa of course. No rush. Thanks a lot!

Lukasa:
Ok, this branch should be fixed.

vmalloc:
@Lukasa will rebase and push again.

3.0-HISTORY.rst
```diff
@@ -58,5 +58,8 @@
 
 - URLs are now automatically stripped of leading/trailing whitespace.
 
+- ``Response.raise_for_status()`` now returns the response object if the
+  response status is not erroneous.
```
Lukasa: Heh, I know this is really nitpicking now, but can we rephrase this? It should be enough to say "now returns the response object if no exception is raised."

docs/user/quickstart.rst
```diff
@@ -354,10 +354,14 @@ But, since our ``status_code`` for ``r`` was ``200``, when we call
 ``raise_for_status()`` we get::
 
     >>> r.raise_for_status()
-    None
+    <Response [200]>
 
 All is well.
 
+.. note:: ``raise_for_status`` returns the response object for a successful response. This eases chaining in trivial cases:, where we want bad codes to raise an exception, but use the response otherwise:
```
Lukasa: Nit: colon followed by comma is unnecessary. Remove the colon, as having two colons in one sentence is unwieldy.

vmalloc: Fixed. Sorry for that - was a typo

3.0-HISTORY.rst
vmalloc: Fixed. How is it now?

PR #3869, opened 2017-02-12
revert removal of request param from resolve_redirects

nateprewitt:
This will fix the build issue on the proposed/3.0.0 branch.

When the `resp` and `req` parameters were condensed into `response`, it had some unintentional effects on responses modified by hooks. The `req` parameter was the request prior to being sent and hooks being run. Since `handle_401` resends the request again with the Authorization header, using `response.request` continues to pass it on with every redirect.

Reverting this change is the simplest approach since it's mostly cosmetic. We could alternatively look into a more explicit removal of Authorization headers on redirects.

### DIFF
--- a/3.0-HISTORY.rst
+++ b/3.0-HISTORY.rst
@@ -1,4 +1,4 @@
-3.0.0 (2016-xx-xx)
+3.0.0 (2017-xx-xx)
 ++++++++++++++++++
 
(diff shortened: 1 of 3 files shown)


Lukasa:
This is an annoyance, but looks right. Thanks @nateprewitt!

PR #3871, opened 2017-02-13
altered internal loops of how ``SessionRedirectMixin.resolve_redirect…

jvanasco:
This is an idea for updating the `resolve_redirects` loop.  

The existing code was hard to untangle as variables get overwritten midway through the loop (`resp` switches between the previous and the next, which can be easy to miss).  It manages the `hist` by omitting the first request, appending the previous request, then shuffling some stuff around in a cleanup during the `send` method.

In this approach, populating the `hist` is moved down the loop so the current request is topmost on the stack.   The cleanup function is then simplified to just popping it off.

This passes all tests, but causes a slight undocumented API change -- the value of `request.history` changes, requiring a slightly different cleanup.  The only code that calls `resolve_redirects` is within `Session.send`, but I don't know if this would break any hooks.

anyways, I think this makes the block more straightforward as the "cleanup" function in `send` is simplified and the context of `resp` when appending the history doesn't change as much.  another approach I thought of was using inline comments or variable name changes to clearly note which is the "previous_request" in that loop.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -97,16 +97,11 @@ class SessionRedirectMixin(object):
                           verify=True, cert=None, proxies=None, **adapter_kwargs):
         """Receives a Response. Returns a generator of Responses."""
 
-        hist = [] # keep track of history
+        hist = [resp, ] # keep track of history; seed it with the original response
 
         url = self.get_redirect_target(resp)
         while url:
             prepared_request = req.copy()
-
-            # Update history and keep track of redirects.
-            # resp.history must ignore the original request in this loop
-            hist.append(resp)
-            resp.history = hist[1:]
 
             try:
                 resp.content  # Consume socket so it can be released
@@ -193,6 +188,10 @@ class SessionRedirectMixin(object):
                 allow_redirects=False,
                 **adapter_kwargs
             )
+            # copy our history tracker into the response
+            resp.history = hist[:]
+            # append the new response to the history tracker for the next iteration
+            hist.append(resp)
 
             extract_cookies_to_jar(self.cookies, prepared_request, resp.raw)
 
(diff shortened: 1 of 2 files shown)


requests/sessions.py
```diff
@@ -97,17 +97,12 @@ def resolve_redirects(self, resp, req, stream=False, timeout=None,
                           verify=True, cert=None, proxies=None, **adapter_kwargs):
         """Receives a Response. Returns a generator of Responses."""
 
-        hist = [] # keep track of history
+        hist = [resp, ] # keep track of history; seed it with the original response
```
Lukasa: No need for the trailing comma in the list.

nateprewitt:
Whether this gets moved to 3.0.0 or not, it should probably note that this addresses #1953.

For what it's worth, I'm +1 on 3.0.0. The seeding does seem like an unexpected semantics change. `response.history[0]` will now be a different value than it was before.

jvanasco:
3.0 sounds good to me.  I can remove the trailing comma.  flake8 yells at me when i leave it out ;)

sigmavirus24:
> flake8 yells at me when i leave it out ;)

What flake8 plugins are you using? That is *not* default behaviour. (Also, side note, Kenneth hates Flake8 with a passion ;))

jvanasco:
updated, however I should probably do a different pr against the 3.0 branch instead -- right?

> What flake8 plugins are you using? That is not default behaviour.

`flake8-commas`, but installed after years of enforcing it as a style guide, after years of having it enforced on me as a style guide.  I have a love-hate relationship with flake8, my exceptions list is long, and there are a lot of situations that pyflakes doesn't cover... but it's become really useful as a sanity-check before running unit tests.

Lukasa:
Yeah, you'd have to do a different PR against the 3.0 branch.

jvanasco:
closed in favor of #3886 for the 3.0 branch

PR #3874, opened 2017-02-14
variable name consistency

nateprewitt:
Just need to clean up a few variable names.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -87,13 +87,13 @@ def merge_hooks(request_hooks, session_hooks, dict_class=OrderedDict):
 
 
 class SessionRedirectMixin(object):
-    def get_redirect_target(self, resp):
+    def get_redirect_target(self, response):
         """Receives a Response. Returns a redirect URI or ``None``"""
-        if resp.is_redirect:
+        if response.is_redirect:
             if not is_valid_location(response):
                 raise InvalidHeader('Response contains multiple Location headers. '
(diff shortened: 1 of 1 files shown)


requests/sessions.py
```diff
@@ -104,7 +104,7 @@ def resolve_redirects(self, response, request, stream=False, timeout=None,
         """
 
         redirect_count = 0
-        history = [] # keep track of history
+        hist = [] # keep track of history
```
Lukasa: Let's call the other instances `history`, rather than call this `hist`.

nateprewitt: :+1:

nateprewitt:
All set.

Lukasa:
So to be clear, the clearer names are on the 3.0 branch already: I just accidentally busted them with names from the master branch like a dummy.

nateprewitt:
Yep @sigmavirus24, `response` and `history` were added to 3.0.0 in commits by Kenneth.

Edit: 66eedec to be specific.

jvanasco:
Is there an ETA of this getting merged in?  I'd like to refile #3871 against the 3.0 branch, but will need to schedule the time to do that after this is merged.

nateprewitt:
I think it's good to go @jvanasco, we were just waiting for confirmation from @kennethreitz. @Lukasa we can probably just merge though since this isn't actually a change, yeah?

kennethreitz:
+1

PR #3876, opened 2017-02-16
added docstring for response.ok property

mislavcimpersak:
Nothing much here, just added a docstring for `response.ok` property.

Took the docstring straight from the other methods to stay consistent.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -672,6 +672,7 @@ class Response(object):
 
     @property
     def ok(self):
+        """Returns true if :attr:`status_code` is 'OK'."""
         try:
             self.raise_for_status()
         except HTTPError:


Lukasa:
LGTM, thanks! :sparkles: :cake: :sparkles:

requests/models.py
```diff
@@ -672,6 +672,7 @@ def __iter__(self):
 
     @property
     def ok(self):
+        """Returns true if :attr:`status_code` is 'OK'."""
```
sigmavirus24, line 675: That's actually not true. This returns `True` if `status_code < 400`. So this will return True for `2xx`, `3xx` responses. This isn't a `status_code == 200` check.

mislavcimpersak: Ouh. My bad then. I can change the docstring to:

`Returns true if :attr:status_code is 'OK' or 'redirect'.`

Would something like that be ok?

mislavcimpersak: Ironically, this is the best proof that the docstring there was really necessary 😁

sigmavirus24: =D I already wrote up a PR for that. I'm also fixing up `__bool__` and `__nonzero__` at @Lukasa's request. Thanks for spurring all of this. =) :cake:

mislavcimpersak: A happy ending for all 🌈 😄

PR #3877, opened 2017-02-16
Correct docstring for Response.ok

sigmavirus24:
We added a docstring just a few hours ago but it wasn't entirely accurate.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -672,7 +672,13 @@ class Response(object):
 
     @property
     def ok(self):
-        """Returns true if :attr:`status_code` is 'OK'."""
+        """Returns True if :attr:`status_code` is less than 400.
+
+        This attribute checks if the status code of the response is between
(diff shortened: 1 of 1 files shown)


Lukasa:
@sigmavirus24 That PR was merged on the grounds that it was consistent with the docstrings that relied on that method. If we're changing one, we should change them all. =)

sigmavirus24:
@Lukasa which docstrings are those? I was under the impression that the docstring format was what was borrowed from, not the content of other docstrings.

Lukasa:
`__bool__` and `__nonzero__`, immediately above.

sigmavirus24:
I'll happily fix those too. In my opinion, misleading docs are worse than no docs. Those methods have been subject to criticism for years at this point and are targetted for correction in 3.0. May as well correct their documentation until 3.0 is out.

sigmavirus24:
Updated `__bool__` and `__nonzero__`.

PR #3884, opened 2017-02-22
Removed response hook hardcode.

duriantang:


Lukasa:
Thanks for this! I think this is absolutely worth doing but there's an element of style consideration here, so I'm going to let @kennethreitz make the call on this one.

requests/hooks.py
```diff
@@ -11,12 +11,16 @@
 ``response``:
     The response generated from a Request.
 """
-HOOKS = ['response']
+
+RESPONSE_HOOK = 'response'
+
+HOOKS = [RESPONSE_HOOK, ]
```
sigmavirus24: There's no need for `, ` here. It should just be `[RESPONSE_HOOK]` (if we even do this)

Issue #2988, opened 2016-02-01
Proxy bypass detection could add delay of ~4.5 seconds to each request 


schlamar:
Requests uses `urlib.proxy_bypass` to check if the proxy should be bypassed. This uses `socket.gethostbyaddr` to check if the provided IP matches a hostname entered in the proxy bypass list. In some cases, `gethostbyaddr` takes up to 4.5 seconds on Windows (e.g. when the IP is found but the primary DNS doesn't recognize it).

Currently, the proxy bypass detection is done for _every_ request, making it pretty unusable when the conditions are met where `gethostbyaddr` is this slow.

A reasonable solution might be caching the results from `urlib.proxy_bypass`.

Lukasa:
Yeah, 4.5 seconds is probably too long.

The concern with caching the result is that the resolver doesn't give enough information for us to invalidate that cache appropriately, but we will from time to time want to invalidate it (so that long running processes don't end up so broken that they need a restart).

A simple approach here might be to use an `@memoize` decorator with a cache that invalidates entries after a certain amount of time. Given that [`functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) provides a `cache_clear` function, we may be able to provide a very simple wrapper function that handles this logic.

lazywei:
May I try this one :-)?

My plan for implementation:
1. wrap the `proxy_bypass` by our function like `cached_proxy_bypass`
2. cache our wrapper function

``` python
 @functools.lru_cache(maxsize=32)
  def cached_proxy_bypass(netloc):
      return proxy_bypass(netloc)
```

This can cache all the function call to `proxy_bypass` with the same `netloc` as long as they are in the latest 32 calls. However, if we want to invalidate the same `netloc` from time to time, we might maintain a counter, such that we define the wrapper function like this

``` python
@functools.lru_cache(maxsize=32)
def cached_proxy_bypass(netloc, _):
    return proxy_bypass(netloc)
```

and call it like

``` python
cached_proxy_bypass(netloc, int(counter / 5)); counter += 1;
```

5 is a magic number here, which can be any other number, such that lru will invalidate the result every 5 calls with the same `netloc` (because for every 5 calls, the passing arguments are the same, and thus lru will cache them).

How do you think?

Thanks a lot.

Lukasa:
[...]

PR #3885, opened 2017-02-22
Cache proxy bypass

davidfontenot:
Added simple cache over `urllib.proxy_bypass` to remedy #2988. Holds on to entries for a minute. Once it reaches its max size, will evict the first entry that is the oldest.

Added tests over the structure, did not add a test that `should_bypass_proxies` invokes the cache.

This is my first pull request; please let me know if I've overlooked something.

Additionally, when I check out master, run `make coverage` in pipenv, the test named 'TestRequests.test_proxy_error' fails. This is still the only failing test when I use my branch. The error is `ConnectionError: ('Connection aborted.', BadStatusLine("''",))`. Is there some additional setup I need to do to run the test suite?

### DIFF
--- a/requests/structures.py
+++ b/requests/structures.py
@@ -8,8 +8,11 @@ Data structures that power Requests.
 """
 
 import collections
+import time
 
 from .compat import OrderedDict
+
+current_time = getattr(time, 'monotonic', time.time)
 
 
 class CaseInsensitiveDict(collections.MutableMapping):
@@ -103,3 +106,89 @@ class LookupDict(dict):
 
     def get(self, key, default=None):
         return self.__dict__.get(key, default)
+
+
+class TimedCacheManaged(object):
+    """
+    Wrap a function call in a timed cache
+    """
+    def __init__(self, fnc):
+        self.fnc = fnc
+        self.cache = TimedCache()
+
+    def __call__(self, *args, **kwargs):
+        key = args[0]
+        found = None
+        try:
+            found = self.cache[key]
+        except KeyError:
+            found = self.fnc(key, **kwargs)
+            self.cache[key] = found
+
+        return found
+
+
+class TimedCache(collections.MutableMapping):
+    """
+    Evicts entries after expiration_secs. If none are expired and maxlen is hit,
+    will evict the oldest cached entry
+    """
+    def __init__(self, maxlen=32, expiration_secs=60):
+        """
+        :param maxlen: most number of entries to hold on to
+        :param expiration_secs: the number of seconds to hold on
+        to entries
+        """
+        self.maxlen = maxlen
+        self.expiration_secs = expiration_secs
+        self._dict = OrderedDict()
+
+    def __repr__(self):
+        return '<TimedCache maxlen:%d len:%d expiration_secs:%d>' % \
+            (self.maxlen, len(self._dict), self.expiration_secs)
+
+    def __iter__(self):
+        return map(lambda kv: (kv[0], kv[1][1]), self._dict.items()).__iter__()
+
+    def __delitem__(self, item):
+        return self._dict.__delitem__(item)
+
+    def __getitem__(self, key):
+        """
+        Look up an item in the cache. If the item
+        has already expired, it will be invalidated and not returned
+
+        :param key: which entry to look up
+        :return: the value in the cache, or None
+        """
+        occurred, value = self._dict[key]
+        now = int(current_time())
+
+        if now - occurred > self.expiration_secs:
+            del self._dict[key]
+            raise KeyError
+        else:
+            return value
+
+    def __setitem__(self, key, value):
+        """
+        Locates the value at lookup key, if cache is full, will evict the
+        oldest entry
+
+        :param key: the key to search the cache for
+        :param value: the value to be added to the cache
+        """
+        now = int(current_time())
+
+        while len(self._dict) >= self.maxlen:
+            self._dict.popitem(last=False)
+
+        return self._dict.__setitem__(key, (now, value))
+
+    def __len__(self):
+        """:return: the length of the cache"""
+        return len(self._dict)
+
+    def clear(self):
+        """Clears the cache"""
+        return self._dict.clear()
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -28,7 +28,7 @@ from .compat import (
     quote, urlparse, bytes, str, OrderedDict, unquote, getproxies,
     proxy_bypass, urlunparse, basestring, integer_types)
 from .cookies import RequestsCookieJar, cookiejar_from_dict
-from .structures import CaseInsensitiveDict
+from .structures import CaseInsensitiveDict, TimedCache, TimedCacheManaged
 from .exceptions import (
     InvalidURL, InvalidHeader, FileModeWarning, UnrewindableBodyError)
 
@@ -579,6 +579,16 @@ def set_environ(env_name, value):
             os.environ[env_name] = old_value
 
 
+@TimedCacheManaged
+def _proxy_bypass_cached(netloc):
+    """
+    Looks for netloc in the cache, if not found, will call proxy_bypass
+    for the netloc and store its result in the cache
+
+    :rtype: bool
+    """
+    return proxy_bypass(netloc)
+
 def should_bypass_proxies(url, no_proxy):
     """
     Returns whether we should bypass proxies or not.
@@ -626,7 +636,7 @@ def should_bypass_proxies(url, no_proxy):
     # legitimate problems.
     with set_environ('no_proxy', no_proxy_arg):
         try:
-            bypass = proxy_bypass(netloc)
+            bypass = _proxy_bypass_cached(netloc)
         except (TypeError, socket.gaierror):
             bypass = False
 
--- a/tests/test_structures.py
+++ b/tests/test_structures.py
@@ -2,7 +2,7 @@
 
 import pytest
 
-from requests.structures import CaseInsensitiveDict, LookupDict
+from requests.structures import CaseInsensitiveDict, LookupDict, TimedCache, TimedCacheManaged
 
 
 class TestCaseInsensitiveDict:
@@ -74,3 +74,75 @@ class TestLookupDict:
     @get_item_parameters
     def test_get(self, key, value):
         assert self.lookup_dict.get(key) == value
+
+
+class TestTimedCache(object):
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        self.any_value = 'some value'
+        self.expiration_secs = 60
+        self.cache = TimedCache(expiration_secs=self.expiration_secs)
+        yield
+        self.cache.clear()
+
+    def test_get(self):
+        self.cache['a'] = self.any_value
+        assert self.cache['a'] is self.any_value
+
+    def test_repr(self):
+        repr = str(self.cache)
+        assert repr == '<TimedCache maxlen:32 len:0 expiration_secs:60>'
+
+    def test_get_expired_item(self, mocker):
+        self.cache = TimedCache(maxlen=1, expiration_secs=self.expiration_secs)
+
+        mocker.patch('requests.structures.current_time', lambda: 0)
+        self.cache['a'] = self.any_value
+        mocker.patch('requests.structures.current_time', lambda: self.expiration_secs + 1)
+        assert self.cache.get('a') is None
+
+    def test_evict_first_entry_when_full(self, mocker):
+        self.cache = TimedCache(maxlen=2, expiration_secs=2)
+        mocker.patch('requests.structures.current_time', lambda: 0)
+        self.cache['a'] = self.any_value
+        mocker.patch('requests.structures.current_time', lambda: 1)
+        self.cache['b'] = self.any_value
+        mocker.patch('requests.structures.current_time', lambda: 3)
+        self.cache['c'] = self.any_value
+        assert len(self.cache) is 2
+        with pytest.raises(KeyError, message='Expected key not found'):
+            self.cache['a']
+        assert self.cache['b'] is self.any_value
+        assert self.cache['c'] is self.any_value
+
+    def test_delete_item_removes_item(self):
+        self.cache['a'] = self.any_value
+        del self.cache['a']
+        with pytest.raises(KeyError, message='Expected key not found'):
+            self.cache['a']
+
+    def test_iterating_hides_timestamps(self):
+        self.cache['a'] = 1
+        self.cache['b'] = 2
+        expected = [('a', 1), ('b', 2)]
+        actual = [(key, val) for key, val in self.cache]
+        assert expected == actual
+
+
+class TestTimedCacheManagedDecorator(object):
+    def test_caches_repeated_calls(self, mocker):
+        mocker.patch('requests.structures.current_time', lambda: 0)
+
+        nonlocals = {'value': 0}
+
+        @TimedCacheManaged
+        def some_method(x):
+            nonlocals['value'] = nonlocals['value'] + x
+            return nonlocals['value']
+
+        first_result = some_method(1)
+        assert first_result is 1
+        second_result = some_method(1)
+        assert second_result is 1
+        third_result = some_method(2)
+        assert third_result is 3
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -178,4 +178,5 @@ Patches and Suggestions
 - Moinuddin Quadri <moin18@gmail.com> (`@moin18 <https://github.com/moin18>`_)
 - Matt Kohl (`@mattkohl <https://github.com/mattkohl>`_)
 - Jonathan Vanasco (`@jvanasco <https://github.com/jvanasco>`_)
+- David Fontenot (`@davidfontenot <https://github.com/davidfontenot>`_)
 


requests/structures.py
```diff
@@ -103,3 +104,94 @@ def __getitem__(self, key):
 
     def get(self, key, default=None):
         return self.__dict__.get(key, default)
+
+
+class TimedCache(object):
```
Lukasa: So, this is basically a `collections.MutableMapping`! We should make it *actually* be one, because that gives us a number of handy-dandy convenience methods.

requests/structures.py
```diff
@@ -103,3 +104,94 @@ def __getitem__(self, key):
...
+
+    def __getitem__(self, key):
+        """
+        Look up an item in the cache. If the item
+        has already expired, it will be invalidated and not returned
+        :param key: which entry to look up
+        :return: the value in the cache, or None
+        """
+
+        return self.get(key, None)
```
Lukasa: So, generally speaking it is better to implement `get` in terms of `__getitem__` than the other way around. `__getitem__` should almost always maintain dict-like behaviour of raising `KeyError` if the entry is not present, rather than returning `None`. The core concern here is that it needs to be possible to distinguish between "key not present in dictionary" and "key present with a value of `None`".

requests/structures.py
```diff
@@ -103,3 +104,94 @@ def __getitem__(self, key):
...
+        """
+        Look up an item in the cache. If the item
+        has already expired, it will be invalidated and not returned
+        :param key: which entry to look up
+        :return: the value in the cache, or None
+        """
+
+        found = self._dict.get(key, default)
+        if not found:
+            return default
```
Lukasa: This logic is wrong. Happily, it doesn't matter: if you make this a subclass of `MutableMapping` you don't actually need to implement this method at all, we get it for free.

requests/structures.py
```diff
@@ -103,3 +104,94 @@ def __getitem__(self, key):
...
+
+        occurred, value = found
+        now = int(time.time())
```
Lukasa: So, this is a problem.

*If possible* we'd like to use `time.monotonic` instead, as `time.time` can go backwards. So I think in this module you'll need to feature detect on import to see whether `time.monotonic` is present and to prefer it if it is.

requests/structures.py
```diff
@@ -103,3 +104,94 @@ def __getitem__(self, key):
...
+        """
+        Locates the value at lookup key,
+        if cache is full, will evict the first
+        found expired entry, otherwise will evict
+        the oldest non-expired item
+        :param key: the key to search the cache for
+        :param value: the value to be added to the cache
+        """
+
+        now = int(time.time())
```
Lukasa: Same note about monotonic time.

requests/structures.py
```diff
@@ -103,3 +104,94 @@ def __getitem__(self, key):
...
+            return default
+        else:
+            return value
+
+    def __setitem__(self, key, value):
+        """
+        Locates the value at lookup key,
+        if cache is full, will evict the first
+        found expired entry, otherwise will evict
+        the oldest non-expired item
```
Lukasa: You're wrapping this paragraph *very* short. Feel free to go all the way out to 89 columns!

Lukasa: While we're here, can you add a blank line separating the paragraph blurb from the `:param:` directives?

requests/structures.py
```diff
@@ -103,3 +104,94 @@ def __getitem__(self, key):
...
+        :param key: the key to search the cache for
+        :param value: the value to be added to the cache
+        """
+
+        now = int(time.time())
+
+        if len(self._dict) >= self.maxlen:
+            replacement_candidate_key = None
+            max_time_delta = 0
+            for cur_key, (occurred, _) in self._dict.items():
```
Lukasa: This complete search can probably be avoided by using an `OrderedDict`, can't it? urllib3 includes one that works on all the Python versions we support, so you can import that and use it as the base implementation to save ourselves this iterative walk.

requests/utils.py
```diff
@@ -579,6 +581,25 @@ def set_environ(env_name, value):
...
+    """
+
+    cached_result = PROXY_BYPASS_CACHE[netloc]
+
+    if not cached_result:
+        result = proxy_bypass(netloc)
+        PROXY_BYPASS_CACHE[netloc] = result
+        return result
+    else:
+        return cached_result
```
Lukasa: My suggestion is that this be reimplemented as a generic decorator, rather than written to specifically handle `proxy_bypass`. That way we get a much easier time of testing the timed cache and all the relevant code. =)

tests/test_structures.py
```diff
@@ -74,3 +74,56 @@ def test_getitem(self, key, value):
     @get_item_parameters
     def test_get(self, key, value):
         assert self.lookup_dict.get(key) == value
+
+
+class TestTimedCache:
```
Lukasa: Let's avoid making this an old-style class by putting `(object)` as well.

requests/structures.py
dfontenot: Thanks. To clarify, you're saying if `time.monotonic` is not present, it's ok to fall back to `time.time`?

requests/structures.py
dfontenot: Will update.

dfontenot: Will update.

requests/structures.py
Lukasa: Yes.

tests/test_structures.py
dfontenot: Will update.

requests/structures.py
dfontenot: Will update to have TimedCache inherit `collections.MutableMapping`

requests/structures.py
dfontenot: Will update.

requests/structures.py
dfontenot: Will replace `self._dict` with an `OrderedDict`.

dfontenot:
@Lukasa I see a line like this already in the code base

```
current_time = getattr(time, 'monotonic', time.time)
```

as a top level definition. I've attempted to mock those, but my attempts so far have had some code smells. I've added a new commit that attempts to mock those in the least smelly way, but it's not ideal. Can you help me iron these out so they aren't so strange?

My approaches so far:
1. I set `current_time` in the class and mock it using `mocker.patch.object`. If I set it as a staticmethod or a classmethod, I'm not able to replace the function with a lambda w/o any args or with one anonymous arg. I'd rather not set it as an instance method of the class as that's a bit strange.
1. I use `mocker.patch` and set `current_time` at the global level of requests.structures. The only problem then is I have to import requests.structures (or just requests.structures.current_time), and pyflakes complains about an unused import.

Would appreciate any suggestions.

Lukasa:
Your mocking will get a lot easier if you make the `_current_time` callable a module-level object rather than putting it at class scope. =)

dfontenot:
I think I must have had a typo, `mocker.patch` with the fully-qualified package name for `current_time` works without doing any extra non-used imports.

requests/structures.py
```diff
@@ -103,3 +106,110 @@ def __getitem__(self, key):
...
+            self.cache[key] = found
+
+        return found
+
+
+class TimedCache(collections.MutableMapping):
+    """
+    Evicts entries after expiration_secs.
+    If none are expired and maxlen is hit,
+    will evict the oldest cached entry
```
Lukasa: You can have a slightly wider docstring than this! :grin:

requests/structures.py
```diff
@@ -103,3 +106,110 @@ def __getitem__(self, key):
...
+        :param expiration_secs: the number of seconds to hold on
+        to entries
+        """
+
+        self.maxlen = maxlen
+        self.expiration_secs = expiration_secs
+        self._dict = OrderedDict()
+
+    def __repr__(self):
+        return '<timed-cache maxlen:%d len:%d>' % \
```
Lukasa: This should be `TimedCache`, and it should also provide the `expiration_secs`.

requests/structures.py
```diff
@@ -103,3 +106,110 @@ def __getitem__(self, key):
...
+        be invalidated and not returned
+
+        :param key: which entry to look up
+        :return: the value in the cache, or None
+        """
+
+        try:
+            return self.__getitem__(key)
+        except KeyError:
+            return default
```
Lukasa: You don't need to implement this: `collections.MutableMapping` will do it for you.

requests/structures.py
```diff
@@ -103,3 +106,110 @@ def __getitem__(self, key):
...
+
+        :param key: the key to search the cache for
+        :param value: the value to be added to the cache
+        """
+
+        now = int(current_time())
+
+        if len(self._dict) >= self.maxlen:
+            for _ in range(len(self._dict) - self.maxlen + 1):
+                self._dict.popitem(last=False)
```
Lukasa: This can be more cleanly expressed as:

```python
while len(self._dict) >= self.maxlen:
    self._dict.popitem(last=False)
```

requests/structures.py
```diff
@@ -103,3 +106,110 @@ def __getitem__(self, key):
...
+
+        return self._dict.__setitem__(key, (now, value))
+
+    def __len__(self):
+        """:return: the length of the cache"""
+        return len(self._dict)
+
+    def clear(self):
+        """Clears the cache"""
+        return self._dict.clear()
```
Lukasa, line 194: This one we also technically get for free from `MutableMapping`, but in this case we can do it much faster than the generic implementation so I'm happy to have this override.

tests/test_structures.py
```diff
@@ -74,3 +74,97 @@ def test_getitem(self, key, value):
...
+        assert fnc_stub.call_count is 1
+
+        assert len(cache) is 1
+        some_method(1)
+        assert getitem_spy.call_count is 2
+        assert setitem_spy.call_count is 1
+        assert fnc_stub.call_count is 1
+        assert len(cache) is 1
+
+        assert invocation_spy.call_count is 2
```
Lukasa: Heh, so this is a good test, but it mostly tests implementation details. For this, it's enough to just mock out the time and then write a small function that constantly returns an incrementing value, then wrap it in the decorator and use the timer to demonstrate the behaviour.

Generally speaking, avoiding mocks is a good thing wherever we can. =)

requests/structures.py
dfontenot: Fixed

requests/structures.py
dfontenot: Fixed

requests/structures.py
dfontenot: Fixed

requests/structures.py
dfontenot: Agreed. Updated.

requests/structures.py
dfontenot: Good to know. I've left the override intact.

tests/test_structures.py
dfontenot: Ok. I've re-made the test more along these lines. Let me know what you think.

requests/structures.py
```diff
@@ -103,3 +106,94 @@ def __getitem__(self, key):
...
+    will evict the oldest cached entry
+    """
+
+    def __init__(self, maxlen=32, expiration_secs=60):
+        """
+        :param maxlen: most number of entries to hold on to
+        :param expiration_secs: the number of seconds to hold on
+        to entries
+        """
+
```
Lukasa: Really minor nit, we don't need this blank line.

requests/utils.py
```diff
@@ -579,6 +579,19 @@ def set_environ(env_name, value):
...
+def _proxy_bypass_cached(netloc):
+    """
+    Looks for netloc in the cache,
+    if not found, will call proxy_bypass
+    for the netloc and store its result
+    in the cache
+
+    :rtype: bool
+    """
+
```
Lukasa: Same again, don't need this blank line.

requests/utils.py
```diff
@@ -579,6 +579,19 @@ def set_environ(env_name, value):
             os.environ[env_name] = old_value
 
 
+@TimedCacheManaged
+def _proxy_bypass_cached(netloc):
+    """
+    Looks for netloc in the cache,
+    if not found, will call proxy_bypass
+    for the netloc and store its result
+    in the cache
```
Lukasa: This again can be flowed out to column 79 as well.

tests/test_structures.py
```diff
@@ -74,3 +74,76 @@ def test_getitem(self, key, value):
...
+
+        @TimedCacheManaged
+        def some_method(x):
+            nonlocals['value'] = nonlocals['value'] + x
+            return nonlocals['value']
+
+        first_result = some_method(1)
+        assert first_result is 1
+        second_result = some_method(1)
+        assert second_result is 1
```
Lukasa, line 146: Should also check that a different argument doesn't return a cached value.

dfontenot:
@Lukasa thanks for your help with this. As for squashing up all of the commits this PR has accumulated--who should do that? Do the maintainers of this lib do that? Or I can squash locally and push to a new PR (or just force push).

requests/structures.py
dfontenot: Removed

requests/utils.py
dfontenot: Fixed

requests/utils.py
dfontenot: Removed

Lukasa:
If you squash locally and force push that's fine, GitHub will work it all out. 👍

requests/structures.py
```diff
@@ -103,3 +106,93 @@ def __getitem__(self, key):
 
     def get(self, key, default=None):
         return self.__dict__.get(key, default)
+
+
+class TimedCacheManaged(object):
+    """
+    Wrap a function call in a timed cache
+    """
+
```
Lukasa: This one.

requests/structures.py
```diff
@@ -103,3 +106,93 @@ def __getitem__(self, key):
...
+        return found
+
+
+class TimedCache(collections.MutableMapping):
+    """
+    Evicts entries after expiration_secs. If none are expired and maxlen is hit,
+    will evict the oldest cached entry
+    """
+
```
Lukasa: This one

requests/structures.py
```diff
@@ -103,3 +106,93 @@ def __getitem__(self, key):
...
+    def __getitem__(self, key):
+        """
+        Look up an item in the cache. If the item
+        has already expired, it will be invalidated and not returned
+
+        :param key: which entry to look up
+        :return: the value in the cache, or None
+        """
+
```
Lukasa: This one

requests/structures.py
```diff
@@ -103,3 +106,93 @@ def __getitem__(self, key):
...
+    def __setitem__(self, key, value):
+        """
+        Locates the value at lookup key, if cache is full, will evict the
+        oldest entry
+
+        :param key: the key to search the cache for
+        :param value: the value to be added to the cache
+        """
+
```
Lukasa: This one

tests/test_structures.py
```diff
@@ -74,3 +74,78 @@ def test_getitem(self, key, value):
     @get_item_parameters
     def test_get(self, key, value):
         assert self.lookup_dict.get(key) == value
+
+
+class TestTimedCache(object):
+
```
Lukasa: This one

tests/test_structures.py
```diff
@@ -74,3 +74,78 @@ def test_getitem(self, key, value):
...
+    def test_iterating_hides_timestamps(self):
+        self.cache['a'] = 1
+        self.cache['b'] = 2
+        expected = [('a', 1), ('b', 2)]
+        actual = [(key, val) for key, val in self.cache]
+        assert expected == actual
+
+
+class TestTimedCacheManagedDecorator(object):
+
```
Lukasa: This one

tests/test_structures.py
```diff
@@ -74,3 +74,78 @@ def test_getitem(self, key, value):
...
+    def test_caches_repeated_calls(self, mocker):
+
```
Lukasa: This one.

dfontenot:
Force pushed commit addresses the whitespace issues

requests/structures.py
```diff
@@ -103,3 +106,89 @@ def __getitem__(self, key):
...
+        self.maxlen = maxlen
+        self.expiration_secs = expiration_secs
+        self._dict = OrderedDict()
+
+    def __repr__(self):
+        return '<TimedCache maxlen:%d len:%d expiration_secs:%d>' % \
+            (self.maxlen, len(self._dict), self.expiration_secs)
+
+    def __iter__(self):
+        return map(lambda kv: (kv[0], kv[1][1]), self._dict.items()).__iter__()
```
sigmavirus24, line 151: I'm curious why we don't just build a generator here.

1. `map()` is slow and `lambda` s are slow. This intentional on Guido's part.
2. Calling `__iter__` is really poor form.
3. This would be far simpler and easier to read as

```py
return ((key, value[1]) for key, value in self._dict.items())
```

requests/structures.py
```diff
@@ -103,3 +106,89 @@ def __getitem__(self, key):
...
+
+    def __delitem__(self, item):
+        return self._dict.__delitem__(item)
```
sigmavirus24, line 154: Why not write this as:

```py
del self._dict[item]
```

`__delitem__` doesn't return anything so there's no need to worry about returning something here. And it would avoid calling the dunder method directly, which I'll repeat, is poor form.

requests/structures.py
```diff
@@ -103,3 +106,89 @@ def __getitem__(self, key):
...
+
+        :param key: which entry to look up
+        :return: the value in the cache, or None
+        """
+        occurred, value = self._dict[key]
+        now = int(current_time())
+
+        if now - occurred > self.expiration_secs:
+            del self._dict[key]
+            raise KeyError
```
sigmavirus24, line 169: This should raise a `KeyError` with the key as an argument, e.g., `KeyError(key)`.

requests/structures.py
```diff
@@ -103,3 +106,89 @@ def __getitem__(self, key):
...
+
+        :param key: the key to search the cache for
+        :param value: the value to be added to the cache
+        """
+        now = int(current_time())
+
+        while len(self._dict) >= self.maxlen:
+            self._dict.popitem(last=False)
+
+        return self._dict.__setitem__(key, (now, value))
```
sigmavirus24, line 186: Again, `__setitem__` doesn't return anything. This should just be

```py
self._dict[key] = (now, value)
```

There's no need to obfuscate any of this.

PR #3886, opened 2017-02-22
proposed 3.0 - altered internal loops of resolve_redirect

jvanasco:
This is a resubmit of #3871 against the 3.0.0 branch.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -12,6 +12,11 @@ Release History
   querying ``Response.is_redirect`` and ``Response.headers['location']``.
   Advanced users will be able to process malformed redirects more easily.
 
+- Altered how ``SessionRedirectMixin.resolve_redirects`` and ``Session.send``
+  process redirect history.  Developers who subclass ``resolve_redirects`` will
+  find a different ``.history`` attribute - the first element now contains the
+  original response, and the last element now contains the active response.
+
 2.13.0 (2017-01-24)
 +++++++++++++++++++
 
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -103,18 +103,12 @@ class SessionRedirectMixin(object):
         reached.
         """
 
-        redirect_count = 0
-        history = [] # keep track of history
+        history = [response] # keep track of history; seed it with the original response
 
         url = self.get_redirect_target(response)
 
         while url:
             prepared_request = request.copy()
-
-            # Update history and keep track of redirects.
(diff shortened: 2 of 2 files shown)


sigmavirus24:
> The fact that this counts as regressing code review is stupid but there we are.

How do you mean, @Lukasa?

This looks generally fairly good, but I'm not sure I'm entirely clear. We're changing the ordering of the response history? Is there a good reason why? Has it been wrong all this time?

jvanasco:
> This looks generally fairly good, but I'm not sure I'm entirely clear. We're changing the ordering of the response history? Is there a good reason why? Has it been wrong all this time?

The original approach managed the history *before* a new request was made and ignored the original request.  So it was "wrong" as it missed the original request.  The logic on the loop's construction is somewhat painful to follow, and requires a lot of cleanup in `send` to build the correct history.

Assuming the URI `a` is requested and `g` is the final URI, the original code sets up history like this:

    history = [b, c, d, e, f, g, ]

the send routine then cleans it up as such:

    history.insert(0, a)
    r = history.pop()
    r.history = history

So the result is:

    r = g
    r.history = [a, b, c, d, e, f, ]

The new setup maintains the actual history:

    history = [a, b, c, d, e, f, g, ]

So the cleanup in `send` is more straightforward:

    r = history.pop()

and we end up with:

    r = g
    r.history = [a, b, c, d, e, f, ]

This patch makes the `history` less-wrong and removes the weird juggling that was required by the loop.  The first element is now correct (the original `a`), however the last element of `history` is still the most-recent request/response and needs to be popped off to replace `r`.   in the `send` method, `r.history` and `history` are distinct lists.  `r.history` contains the actual history for that response (thanks to the fix in the above loop), however `history` is more like a list of the yielded/generated `responses` (that each contain a history).

sigmavirus24:
Ah, I see. So this makes it better for folks handling their own redirect logic to see the *complete* history. 👍 Sounds good to me.

jvanasco:
> On Feb 25, 2017, at 10:05 AM, Ian Cordasco <notifications@github.com> wrote:
> 
> Ah, I see. So this makes it better for folks handling their own redirect logic to see the complete history. 👍 Sounds good to me.
> 
More importantly, It will make things much more straightforward for people subclassing "send" and future edits to the core redirect logic will be easier (I lost a few hours deciphering what was going on, because the variables swap around midway ).

PR #3892, opened 2017-02-26
Fixed some typos

vpfautz:


requests/packages/urllib3/util/request.py
```diff
@@ -108,7 +108,7 @@ def rewind_body(body, body_pos):
         try:
             body_seek(body_pos)
         except (IOError, OSError):
-            raise UnrewindableBodyError("An error occured when rewinding request "
+            raise UnrewindableBodyError("An error occurred when rewinding request "
```
Lukasa: Can you remove this patch please? urllib3 is a separate project that we vendor in, and we don't carry patches to it in tree. If you'd like to fix this, please make this patch upstream.

nateprewitt:
This looks good now, I think. @vpfautz if you feel like it, this can be patched in [urllib3](https://github.com/shazow/urllib3) too.

Also thanks for catching this! I shouldn't be typing without spell check in my IDE :)

PR #3897, opened 2017-02-28
Refactor Prepare Body

nateprewitt:
This PR supersedes #3338 with the branch brought up to date, as well as a few minor changes, including logic tweaks based on updates to `super_len` since the original PR.

This will forcibly prevent the transmission of headers that include both Content-Length and Transfer-Encoding, as well as simplify redundant code we use for both `prepare_content_length` and `prepare_body`. All the credit here goes to @davidsoncasey, the original creator of this PR.

### DIFF
--- a/3.0-HISTORY.rst
+++ b/3.0-HISTORY.rst
@@ -1,6 +1,10 @@
 3.0.0 (2017-xx-xx)
 ++++++++++++++++++
 
+- Simplified logic for determining Content-Length and Transfer-Encoding.
+  Requests will now avoid setting both headers on the same request, and
+  raise an exception if this is done manually by a user.
+
 - Remove the HTTPProxyAuth class in favor of supporting proxy auth via
   the proxies parameter.
 
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -107,6 +107,10 @@ class RetryError(RequestException):
 class UnrewindableBodyError(RequestException):
     """Requests encountered an error when trying to rewind a body"""
 
+class ConflictingHeaderError(RequestException):
+    """Mutually exclusive request headers set"""
+
+
 # Warnings
 
 
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -104,11 +104,14 @@ class StreamConsumedError(RequestException, TypeError):
 class RetryError(RequestException):
     """Custom retries logic failed"""
 
+
 class UnrewindableBodyError(RequestException):
     """Requests encountered an error when trying to rewind a body"""
 
+
 class ConflictingHeaderError(RequestException):
     """Mutually exclusive request headers set"""
+
 
 class InvalidBodyError(RequestException, ValueError):
     """An invalid request body was specified"""
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -110,6 +110,8 @@ class UnrewindableBodyError(RequestException):
 class ConflictingHeaderError(RequestException):
     """Mutually exclusive request headers set"""
 
+class UnreachableCodeError(RequestException, RuntimeError):
+    """Unreachable code block reached"""
 
 # Warnings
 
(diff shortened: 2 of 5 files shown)


requests/exceptions.py
```diff
@@ -104,9 +104,18 @@ class StreamConsumedError(RequestException, TypeError):
...
+
 class UnrewindableBodyError(RequestException):
     """Requests encountered an error when trying to rewind a body"""
 
+
+class ConflictingHeaderError(RequestException):
+    """Mutually exclusive request headers set"""
+
+
+class InvalidBodyError(RequestException, ValueError):
```
nateprewitt: Only minor note I have here is we've added an `InvalidHeader` exception since this PR was created which could likely be used instead of this. I didn't want to tweak the original work too much in the initial pass though. Thoughts?

requests/exceptions.py
```diff
@@ -104,9 +104,18 @@ class StreamConsumedError(RequestException, TypeError):
 class RetryError(RequestException):
     """Custom retries logic failed"""
 
+
 class UnrewindableBodyError(RequestException):
     """Requests encountered an error when trying to rewind a body"""
 
+
+class ConflictingHeaderError(RequestException):
```
nateprewitt: Only minor note I have here is we've added an `InvalidHeader` exception since this PR was created which could likely be used instead of this. I didn't want to tweak the original work too much in the initial pass though. Thoughts?

Lukasa: Yeah, might be better to use that new exception, or possibly to make this a child exception of `InvalidHeader`.

requests/models.py
```diff
@@ -509,27 +497,37 @@ def prepare_body(self, data, files, json=None):
...
             length = super_len(body)
+
             if length:
-                # If length exists, set it. Otherwise, we fallback
-                # to Transfer-Encoding: chunked.
                 self.headers['Content-Length'] = builtin_str(length)
+            elif is_stream(body):
+                self.headers['Transfer-Encoding'] = 'chunked'
+            else:
+                raise InvalidBodyError('Non-null body must have length or be streamable.')
```
Lukasa, line 524: I don't think this is ever imported, is it?

requests/models.py
```diff
@@ -509,27 +497,37 @@ def prepare_body(self, data, files, json=None):
...
         elif self.method not in ('GET', 'HEAD') and self.headers.get('Content-Length') is None:
-            # Set Content-Length to 0 for methods that can have a body
-            # but don't provide one. (i.e. not GET or HEAD)
```
Lukasa: Did we need to get rid of this comment?

nateprewitt: Nope, I think that was probably a bad merge conflict resolution.

requests/models.py
nateprewitt: It wasn't but  it should be added now. Thoughts on adding a test to ensure hitting this block?

Lukasa: I think we should do that, yeah. Ideally we'd cover all of this logic if we can.

nateprewitt: Alright, the other tests from #3338 cover the other two cases according to pytest. We now have 100% coverage over `prepare_content_length` with the new test. This should be set after the build completes, unless you have further notes.

davidsoncasey:
@nateprewitt nice work on this, thanks for getting this over the finish line

nateprewitt:
@davidsoncasey, glad we're able to get all of your work merged, thanks again!

@Lukasa, 3.0-HISTORY is updated :)

3.0-HISTORY.rst
```diff
@@ -1,6 +1,13 @@
 3.0.0 (2017-xx-xx)
 ++++++++++++++++++
 
+- Simplified logic for determining Content-Length and Transfer-Encoding.
+  Requests will now avoid setting both headers on the same request, and
+  raise an exception if this is done manually by a user.
+
+  It will now generally be unnecessary (and unadvised) to manually set
+  these headers.
```
Lukasa: We probably don't need this paragraph. 😎

nateprewitt: A little too thorough I guess 😉

Updated.

nateprewitt:
It's the little things you forget setting up a new machine 😅 Time to add a global .gitignore.

PR #3901, opened 2017-03-01
Smaller logo

nedbat:
The logo as-is is too large, it won't even fit on my screen!

I made a smaller version for use in the README.rst, because GitHub has neutered some useful features and doesn't care: https://github.com/github/markup/issues/295

NOTE: the image path is to my fork. Once merged, the path should be changed to refer to the main repo.


README.rst
```diff
@@ -50,7 +50,7 @@ Behold, the power of Requests:
 
 See `the similar code, sans Requests <https://gist.github.com/973705>`_.
 
-.. image:: http://docs.python-requests.org/en/master/_static/requests-sidebar.png
+.. image:: https://raw.githubusercontent.com/nedbat/requests/master/docs/_static/requests-logo-small.png
     :target: http://docs.python-requests.org/
```
Lukasa: This should probably be pointed into our copy of the repo, not yours. 😉

nedbat: I know, i mentioned that on the pull request. :)  If I made it point to kenneth's now, I wouldn't be able to see it in my repo, and neither would you :)

nedbat: I can change the URL now in the PR if you like, or fix it afterwards.

Lukasa: Let's fix it now. I don't like adding extra commits if I can avoid it. 😆

nedbat: Done

Lukasa:
Nice first commit, thanks @nedbat!

PR #3903, opened 2017-03-01
fix #3902 Add proposed documentation change for post data

### LINKED ISSUE #3902: Propose additional documentation for post data
## Problem statement

Current documentation only specifies the most common case of post data - a dictionary. The code in fact supports other forms of post data which is not documented. Propose to enhance the quick start section that deals with post data to include examples of other uses.

Section of text in the current documentation: [More complicated POST requests](http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests)

## Questions in the wild indicating the need for more documentation in this regard
On stockoverflow:

* [how to post multiple value with same key in python requests?](http://stackoverflow.com/questions/23384230/how-to-post-multiple-value-with-same-key-in-python-requests)
* [handling duplicate keys in HTTP post in order to specify multiple values](
http://stackoverflow.com/questions/27116424/handling-duplicate-keys-in-http-post-in-order-to-specify-multiple-values)
* [How to send multiple of the same key/parameter via POST?](
http://stackoverflow.com/questions/19004403/how-to-send-multiple-of-the-same-key-parameter-via-post)
* [How to send an array using requests.post (Python)? “Value Error: Too many values to unpack”](
http://stackoverflow.com/questions/31168819/how-to-send-an-array-using-requests-post-python-value-error-too-many-values/35534695)

And there is this issue #1641: Cannot send Single Key, Multiple value data in post because of the use of pythons dict.  

## Proposed modification:

```rst
You can also pass a list of tuples to the ``data`` argument. This is particularly
useful when the form has multiple elements that use the same key::

    >>> payload = (('key1', 'value1'), ('key1', 'value2'))
    >>> r = requests.post('http://httpbin.org/post', data=payload)
    >>> print(r.text)
    {
      ...
      "form": {
        "key1": [
          "value1",
          "value2"
        ]
      },
      ...
    }

Value of each payload element can be a scalar or an iterable::

    >>> payload = (('key1', ('value1', 1, (1, 2))), {'key2', ('value2', 2)})
    >>> r = requests.post('http://httpbin.org/post', data=payload)
    >>> print(r.text)
    {
      ...
      "form": {
        "key1": [
          "value1",
          "1",
          "1",
          "2"
        ],
        "key2": [
          "value2",
          "2"
        ]
      },
      ...
    }

```

garywu:


docs/user/quickstart.rst
```diff
@@ -234,6 +234,45 @@ dictionary of data will automatically be form-encoded when the request is made::
...
+      "form": {
+        "key1": [
+          "value1",
+          "value2"
+        ]
+      },
+      ...
+    }
+
+Value of each payload element can be a scalar or an iterable::
```
Lukasa, line 254: This should begin "The value". Also, let's avoid the word "scalar": let's say "single element".

docs/user/quickstart.rst
```diff
@@ -234,6 +234,45 @@ dictionary of data will automatically be form-encoded when the request is made::
...
+
+    >>> payload = (('key1', ('value1', 1, (1, 2))), {'key2', ('value2', 2)})
```
Lukasa, line 256: We try not to document that we accept non-string elements here, in part because we don't promise to support them.

docs/user/quickstart.rst
garywu: Thanks. Sounds good.

docs/user/quickstart.rst
garywu: In that case,  this section goes away. I can just delete this section so only tuples section above remains. Does that sound good?

Lukasa: Works for me. 👍

docs/user/quickstart.rst
kennethreitz: Agreed.

docs/user/quickstart.rst
kennethreitz: :sparkles: :cake: :sparkes:

kennethreitz:
Can you do a rebase?

kennethreitz:
If you need help with that, let me know.

kennethreitz:
I'm going to merge this by hand.

PR #3909, opened 2017-03-04
Update implementation of TimedCache object

sigmavirus24:
There were some odd decisions made about the implementation of some of
the required methods for MuttableMapping in the TimedCache object. This
cleans those up and makes the implementation, ever so slightly, easier
to read.

See also #3885


sigmavirus24:
:cake:

PR #3913, opened 2017-03-07
Add mention to list od tuples as data param.

StyXman:
`requests.request` accepts list of tuples in the `data` param, but thereś no mention of it in the docs.

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -19,7 +19,7 @@ def request(method, url, **kwargs):
     :param method: method for the new :class:`Request` object.
     :param url: URL for the new :class:`Request` object.
     :param params: (optional) Dictionary or bytes to be sent in the query string for the :class:`Request`.
-    :param data: (optional) Dictionary or list of tuples ``[(key, value)]`` (will be form-encoded), bytes, or file-like object to send in the body of the :class:`Request`. The list of tuple version is to allow duplication of fields.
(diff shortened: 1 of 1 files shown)


requests/api.py
```diff
@@ -19,7 +19,7 @@ def request(method, url, **kwargs):
     :param method: method for the new :class:`Request` object.
     :param url: URL for the new :class:`Request` object.
     :param params: (optional) Dictionary or bytes to be sent in the query string for the :class:`Request`.
-    :param data: (optional) Dictionary (will be form-encoded), bytes, or file-like object to send in the body of the :class:`Request`.
+    :param data: (optional) Dictionary or list of tuples ``[(key, value)]`` (will be form-encoded), bytes, or file-like object to send in the body of the :class:`Request`. The list of tuple version is to allow duplication of fields.
```
Lukasa: I think we'd rather avoid the extra sentence at the end: we don't really need to justify the reasoning for the data types here because it starts drastically increasing the length of the property documentation.

If we want to add stuff about why we allow different formats we should add a new section to the prose documentation about the fact that almost everywhere that accepts a dictionary also accepts an iterable of tuples.

StyXman:
Changed.

Lukasa:
Cool, this looks good. I'll wait for the build to go green and merge. Thanks so much @StyXman! :sparkles:

PR #3918, opened 2017-03-08
DefaultSelector fallback to less efficient methods

programatix:
Test the selector before making it the default selector. This is due to issue reported by myself at https://github.com/kennethreitz/requests/issues/3906


Lukasa:
Thanks for this! However, this patch is entirely contained in [urllib3](https://github.com/shazow/urllib3), which we vendor in unchanged. You'll need to propose this patch on that repository.

programatix:
Ok, but it's weird to have another module within requests itself, which made me thought that urllib3 is part of requests. The only other python project I usually got involved in is flexget which never place other external module inside itself.

Anyway, your comment clears things up for me. Thanks.

PR #3745, opened 2016-12-02
Rebase #2431 against proposed/3.0.0

alexwlchan:
This is really quite unexciting: #2431 had a tentative +1 from @Lukasa, pending a rebase and retargeting against 3.0.0, but it’s been sat unloved for nearly eight months. 😢 

This patch does the required rebase and targets the 3.0.0 branch.

PR #3923, opened 2017-03-14
Rebase #3745 and add some tests

vbarbaresi:
2 years after the original #2431, and after #3745
I encountered the strange behavior of `r.iter_lines()`, found this and tried to complete it.
Rebased once more and added more tests using the breakdown of @ianepperson https://github.com/kennethreitz/requests/pull/2431#issuecomment-72333964

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -780,7 +780,10 @@ class Response(object):
 
         for chunk in self.iter_content(chunk_size=chunk_size,
                                        decode_unicode=decode_unicode):
-            # Skip any null responses
+            # Skip any null responses: if there is pending data it is necessarily an
+            # incomplete chunk, so if we don't have more data we don't want to bother
+            # trying to get it. Unconsumed pending data will be yielded anyway in the
+            # end of the loop if the stream ends.
             if not chunk:
                 continue
 
--- a/requests/models.py
+++ b/requests/models.py
@@ -813,7 +813,7 @@ class Response(object):
             #
             # If we're using `splitlines()`, we only do this if the chunk
             # ended midway through a line.
-            incomplete_line = lines[-1] and lines[-1].endswith(chunk[-1])
+            incomplete_line = lines[-1] and lines[-1][-1] == chunk[-1]
             if delimiter or incomplete_line:
                 pending = lines.pop()
 
--- a/requests/models.py
+++ b/requests/models.py
@@ -776,23 +776,36 @@ class Response(object):
 
         .. note:: This method is not reentrant safe.
         """
-
         pending = None
 
-        for chunk in self.iter_content(chunk_size=chunk_size, decode_unicode=decode_unicode):
+        for chunk in self.iter_content(chunk_size=chunk_size,
+                                       decode_unicode=decode_unicode):
 
+            # Skip any null responses
+            if not chunk:
+                continue
+
+            # Consume any pending data
             if pending is not None:
                 chunk = pending + chunk
+                pending = None
 
+            # Either split on a line, or split on a specified delimiter
             if delimiter:
                 lines = chunk.split(delimiter)
             else:
                 lines = chunk.splitlines()
 
-            if lines and lines[-1] and chunk and lines[-1][-1] == chunk[-1]:
+            # The split(delimiter) will always end with whatever remains past
+            # the delimiter ('' if nothing more).  However splitlines() will
+            # not end with a '' if the final text is a line delimiter.
+
+            # Therefore, if we're in delimiter mode, always pop the final
+            # item to prepend to the next chunk. However, only do this for
+            # non-delimiter mode if the chunk does not match the end of the
+            # last line.
+            if delimiter or (lines[-1] and lines[-1][-1] == chunk[-1]):
                 pending = lines.pop()
-            else:
-                pending = None
 
             for line in lines:
                 yield line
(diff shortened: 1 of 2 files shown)


requests/models.py
```diff
@@ -776,23 +776,43 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
...
         """
-
         pending = None
 
-        for chunk in self.iter_content(chunk_size=chunk_size, decode_unicode=decode_unicode):
+        for chunk in self.iter_content(chunk_size=chunk_size,
+                                       decode_unicode=decode_unicode):
+            # Skip any null responses
+            if not chunk:
+                continue
```
Lukasa: Should this check move after the pending data logic?

vbarbaresi: You're right
Fixed and added tests for this behavior :ok_hand:

I don't know if it's licit to receive empty chunks in the middle of a stream, but I guess it's possible in theory. (and it would work)

vbarbaresi: Actually I replied too fast: if we want to return the same result on
`['line\r\n']` and  `['line\r\n', '']` (which seems right),
we need to keep this check on top

Otherwise we consume `pending`  and discard it

I think it's safe to discard empty chunks early, `pending` will be carried to the next non-empty chunk

Lukasa: > I don't know if it's licit to receive empty chunks in the middle of a stream, but I guess it's possible in theory. (and it would work)

urllib3 won't give them to us, but people like putting things that aren't urllib3 underneath us, and they might. So we should tolerate it.

As to your note about wanting to return the same result, I think that makes sense. But we should add a comment to say that "pending is necessarily an incomplete chunk, and so if we don't have more data we don't want to bother trying to get it".

vbarbaresi: Done 👌

vbarbaresi:
I'll take a look

vbarbaresi:
I attempted a fix; though I'm not sure it's correct
Maybe use `self.encoding` or `self.apparent_encoding`?

When `decode_unicode` is False we don't deal with the encoding, so I'm not sure it makes sense to assume utf-8 and to encode only the last byte

requests/models.py
```diff
@@ -813,7 +813,10 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             #
             # If we're using `splitlines()`, we only do this if the chunk
             # ended midway through a line.
-            incomplete_line = lines[-1] and lines[-1].endswith(chunk[-1])
+            if isinstance(chunk[-1], basestring):  # Decoded string (decode_unicode=True or Py2)
+                incomplete_line = lines[-1] and lines[-1].endswith(chunk[-1])
+            else:  # Bytestring (decode_unicode=False and Py3)
+                incomplete_line = lines[-1] and lines[-1].endswith(chr(chunk[-1]).encode('utf-8'))
```
Lukasa: So, to be clear, the error in this fix was assuming that we need to tolerate `str` here. The tests were in error, not the code: `iter_content` is required to always yield `bytes`. You'll find if you go through the tests and use `bytes` everywhere the problem goes away. :grin: So we don't need this.

vbarbaresi: I'm confused: I checked in a normal usage (outside of the tests) and could reproduce the issue:
with Python2 I get `str`, and with Python3 I get `bytes` (and if I use `iter_lines(decode_unicode=True)`
I get `str` (Py3) or `unicode` (Py2)

The previous code wasn't accepting bytes (and was working only with Python 2)

This failed on Python3 before my last fix:
```
r = requests.get("https://github.com", stream=True)
list(r.iter_lines())
```
TypeError: endswith first arg must be bytes or a tuple of bytes, not int

That's why I don't think the tests were in error (a test that I didn't write was failing, `test_response_iter_lines`)

nateprewitt: @vbarbaresi The issue here seems to be stemming from 46e3fdd in #3745. The conditional on master, and prior to that commit, is done by comparing the last index of each string/bytestring. When retrieving a single index from a bytestring in Python 3 it returns an integer representation of the byte. This works in the case of `lines[-1] and lines[-1][-1] == chunk[-1]` but not `lines[-1] and lines[-1].endswith(chunk[-1])`.

I'd suggest we simply move back to using the original double indexing, rather than introducing more logic to handle the seemingly unneeded `endswith`.

tests/test_requests.py
```diff
@@ -1741,10 +1816,12 @@ def test_json_param_post_should_not_override_data_param(self, httpbin):
         prep = r.prepare()
         assert 'stuff=elixr' == prep.body
 
-    def test_response_iter_lines(self, httpbin):
+    @pytest.mark.parametrize('decode_unicode', (True, False))
+    def test_response_iter_lines(self, httpbin, decode_unicode):
```
nateprewitt, line 1820: Also, I'm not sure if you're going to back this commit out, but if not, `decode_unicode` doesn't seem to be used for anything here in this tests current configuration.

requests/models.py
vbarbaresi: Thanks, I missed that... much neater. It works fine with a simple indexing in all cases.

tests/test_requests.py
vbarbaresi: Nice catch
I overwrote the commit, and fixed the unused `decode_unicode` parameter :ok_hand:

Issue #3926, opened 2017-03-17
SSLError raised when path to CA_BUNDLE is not valid could be more meaningful


bagerard:
Hi there,
First of all thanks for the requests library, it's just awesome!
In one of my project I was pulling 'verify' from a config file and passing it along every requests as an argument. Instead of receiving a boolean, it turned out that I was actually passing verify='true' to the requests methods, which makes requests think that I was providing a path to a ca bundle. 
Although I was obviously misusing the api, it took me some time to debug it and I believe that the error message could have been more meaningful.

In fact the following:
```
requests.get('https://github.com/', verify='true')
```
raises
```
/.../requests/requests/adapters.py in send(self, request, stream, timeout, verify, cert, proxies)
    495         except (_SSLError, _HTTPError) as e:
    496             if isinstance(e, _SSLError):
--> 497                 raise SSLError(e, request=request)
    498             elif isinstance(e, ReadTimeoutError):
    499                 raise ReadTimeout(e, request=request)

SSLError: [Errno 2] No such file or directory
```
I'll do a PR and make it raise an IOError (with the path being printed) in case the `verify` param does not point to a valid file/dir. As minor as it is, I hope you'll find it valuable.

Lukasa:
So, before you leap into this (which is a good idea), be aware that you need to be able to distinguish this case from when either of the paths given to the `cert` parameter are bad.

bagerard:
Thanks for the quick feedback, I'll take that into account

PR #3927, opened 2017-03-17
Raise IOError when providing an invalid path to a CA bundle or …

bagerard:
Use IOError whenever invalid paths are used for ssl certificates (#3926)

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -223,10 +223,10 @@ class HTTPAdapter(BaseAdapter):
 
             conn.cert_reqs = 'CERT_REQUIRED'
 
-            if os.path.isdir(cert_loc):
-                conn.ca_cert_dir = cert_loc
-            else:
+            if not os.path.isdir(cert_loc):
                 conn.ca_certs = cert_loc
+            else:
+                conn.ca_cert_dir = cert_loc
         else:
             conn.cert_reqs = 'CERT_NONE'
             conn.ca_certs = None
@@ -234,7 +234,8 @@ class HTTPAdapter(BaseAdapter):
 
         if cert:
             if not isinstance(cert, basestring):
-                conn.cert_file, conn.key_file = cert
+                conn.cert_file = cert[0]
+                conn.key_file = cert[1]
             else:
                 conn.cert_file = cert
             if conn.cert_file and not os.path.exists(conn.cert_file):
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -202,7 +202,8 @@ class HTTPAdapter(BaseAdapter):
 
         :param conn: The urllib3 connection object associated with the cert.
         :param url: The requested URL.
-        :param verify: Whether we should actually verify the certificate.
+        :param verify: Whether we should actually verify the certificate
+            or it could be a path to a ca bundle
         :param cert: The SSL certificate to verify.
         """
         if url.lower().startswith('https') and verify:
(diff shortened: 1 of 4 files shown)


requests/adapters.py
```diff
@@ -216,26 +217,32 @@ def cert_verify(self, conn, url, verify, cert):
...
 
             conn.cert_reqs = 'CERT_REQUIRED'
 
-            if not os.path.isdir(cert_loc):
-                conn.ca_certs = cert_loc
-            else:
+            if os.path.isdir(cert_loc):
                 conn.ca_cert_dir = cert_loc
+            else:
+                conn.ca_certs = cert_loc
```
Lukasa: Why did this conditional get inverted?

requests/adapters.py
```diff
@@ -216,26 +217,32 @@ def cert_verify(self, conn, url, verify, cert):
...
         else:
             conn.cert_reqs = 'CERT_NONE'
             conn.ca_certs = None
             conn.ca_cert_dir = None
 
         if cert:
             if not isinstance(cert, basestring):
-                conn.cert_file = cert[0]
-                conn.key_file = cert[1]
+                conn.cert_file, conn.key_file = cert
```
Lukasa: You can't make this change in a backwards compatible way: the current code allows for tuples longer than length two.

requests/adapters.py
```diff
@@ -202,7 +202,8 @@ def cert_verify(self, conn, url, verify, cert):
 
         :param conn: The urllib3 connection object associated with the cert.
         :param url: The requested URL.
-        :param verify: Whether we should actually verify the certificate.
+        :param verify: Whether we should actually verify the certificate
+            or it could be a path to a ca bundle
```
Lukasa: So, this docstring needs to be reworded. Let's have it be: `Either a boolean, in which case it controls whether we verify the server's TLS certificate, or a string, in which case it must be a path to a CA bundle to use`.

requests/adapters.py
```diff
@@ -236,6 +238,12 @@ def cert_verify(self, conn, url, verify, cert):
                 conn.key_file = cert[1]
             else:
                 conn.cert_file = cert
+            if conn.cert_file and not os.path.exists(conn.cert_file):
+                raise IOError("Could not find the SSL certificate file, "
+                              "invalid path: {0}".format(conn.cert_file))
+            if conn.key_file and not os.path.exists(conn.key_file):
+                raise IOError("Could not find the SSL key file, "
+                              "invalid path: {0}".format(conn.key_file))
```
Lukasa: For new exception strings, can we use the acronym `TLS` instead of the acronym `SSL`? I don't want to change the string in the block above because people may be checking the actual string message, but these ones should use the proper name for the protocol.

tests/test_requests.py
```diff
@@ -769,6 +769,22 @@ def test_unicode_header_name(self, httpbin):
...
+        assert str(e.value) == 'Could not find a suitable SSL CA certificate bundle, invalid path: {0}'.format(INVALID_PATH)
+
+    def test_invalid_ssl_certificate_files(self):
+        INVALID_PATH = '/garbage'
+        with pytest.raises(IOError) as e:
+            requests.get('http://github.com', cert=INVALID_PATH)
+        assert str(e.value) == 'Could not find the SSL certificate file, invalid path: {0}'.format(INVALID_PATH)
+
+        with pytest.raises(IOError) as e:
+            requests.get('http://github.com', cert=('.', INVALID_PATH))
```
Lukasa: For these tests, can you use the `httpbin_secure` fixture like the tests above? We want to avoid making live web requests in our tests.

bagerard:
Thanks for the review. In addition I also ported the definitions you provided for the `verify` parameter to the other methods that were getting this argument for consistency

bagerard:
I'm happy to help :) Just by curiosity, is it a good practice to squash the commits into one?

Lukasa:
It depends very much on the project. For my part, I generally allow people to work the way they feel most comfortable unless the number of commits vastly outweighs the complexity of the change. Other projects may ask you to do so: it's polite to do what they want. But if no-one asks, then you're totally safe. :smile:

bagerard:
Ok good to know, thanks

PR #3937, opened 2017-03-23
[RFC] Add debugging submodule for bug reporters

sigmavirus24:
The suggested usage in a bug report would be

    python -c 'from requests import _bug; _bug.print_information()'

This should generate most of the information we tend to ask for
repeatedly from bug reporters.

----

**Note** This is merely a POC and an idea I had this morning. I've done similar things in other projects with varying success. Most of the success will ride on us adding a `.github` folder with an issue template. This just makes the bug report process easier for users.


requests/_bug.py
```diff
@@ -0,0 +1,99 @@
...
+    pyopenssl_info = {
+        'version': None,
+        'openssl_version': '',
+    }
+    if OpenSSL:
+        pyopenssl_info = {
+            'version': OpenSSL.__version__,
+            'openssl_version': '%x' % OpenSSL.SSL.OPENSSL_VERSION_NUMBER,
+        }
+    cryptography_info = {
```
Lukasa, line 78: This presumably needs to be guarded by `if cryptography`.

requests/_bug.py
```diff
@@ -0,0 +1,99 @@
...
+        'version': getattr(cryptography, '__version__', ''),
+    }
+
+    return {
+        'platform': platform_info,
+        'implementation': implementation_info,
+        'system_ssl': {
+            'version': '%x' % ssl.OPENSSL_VERSION_NUMBER,
+        },
+        'using_pyopenssl': pyopenssl is not None,
```
Lukasa, line 88: Rather than this boolean, can we just optionally include the pyOpenSSL and cryptography keys?

requests/_bug.py
sigmavirus24: It doesn't. That's the beauty of getattr with a fallback. There wasn't something as simple for the pyOpenSSL `OPENSSL_VERSION_NUMBER` access.

Lukasa: Oh sweet, of course. Ignore me. :D

requests/_bug.py
sigmavirus24: I'd rather not optionally include any part of this, to be fair. This is why I'm including everything. Ostensibly, we could (in the future) start scraping this out of issues with a bot and doing some analysis on the data. This would allow us to filter on that instead of filtering on null values of something else.

Lukasa: Ok, works for me.

sigmavirus24:
This also (obviously) needs tests. I just want Kenneth's opinion on it before progressing. I'm also completely open to renaming the public interfaces to the function =P I am not emotionally attached to them here.

requests/_bug.py
```diff
@@ -0,0 +1,99 @@
+"""Module containing bug report helper(s)."""
+from __future__ import print_function
```
kennethreitz, line 2: I'm not a huge fan of the `_` used here, for the module name. 

I'd rather see `python -m requests.report` or something maybe.

kennethreitz:
Love the concept. I'd just tweak the module name a bit. I like `requests.report` or `requests.bugtrap` or something.

requests/_bug.py
```diff
@@ -0,0 +1,99 @@
...
+        'cryptography': cryptography_info,
+        'requests': {
+            'version': requests_version,
+        },
+    }
+
+
+def print_information():
+    """Pretty-print the bug information as JSON."""
+    print(json.dumps(information(), sort_keys=True, indent=2))
```
kennethreitz, line 99: can we give this an `if __name__ == '__main__'` please?

requests/_bug.py
kennethreitz: `python -m requests.qa`

kennethreitz:
merged this into the "help" branch, with some changes

PR #3943, opened 2017-03-29
HTTPAdapter.send() timeout defaults

leo-b:
This PR allows to use urllib3 ``Timeout`` objects as ``timeout`` argument in ``HTTPAdapter.send()``.

Besides it omits the ``timeout`` argument when calling ``conn.urlopen()`` if ``HTTPAdapter.send()`` is called without a timeout. This allows setting a default timeout at connection pool level:
`adapter.poolmanager.connection_pool_kw['timeout'] = urllib3.Timeout(...)`

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -420,14 +420,11 @@ class HTTPAdapter(BaseAdapter):
                 raise ValueError(err)
         elif isinstance(timeout, TimeoutSauce):
             pass
-        elif timeout is not None:
+        else:
             timeout = TimeoutSauce(connect=timeout, read=timeout)
 
         try:
             if not chunked:
-                kwargs = {}
-                if timeout is not None:
-                    kwargs['timeout'] = timeout
                 resp = conn.urlopen(
                     method=request.method,
                     url=url,
@@ -438,7 +435,7 @@ class HTTPAdapter(BaseAdapter):
                     preload_content=False,
                     decode_content=False,
                     retries=self.max_retries,
-                    **kwargs
+                    timeout=timeout
                 )
 
             # Send the request.
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -391,7 +391,7 @@ class HTTPAdapter(BaseAdapter):
         :param timeout: (optional) How long to wait for the server to send
             data before giving up, as a float, or a :ref:`(connect timeout,
             read timeout) <timeouts>` tuple.
-        :type timeout: float or tuple
+        :type timeout: float or tuple or urllib3 Timeout object
         :param verify: (optional) Either a boolean, in which case it controls whether
             we verify the server's TLS certificate, or a string, in which case it
             must be a path to a CA bundle to use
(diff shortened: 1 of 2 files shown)


requests/adapters.py
```diff
@@ -408,7 +408,10 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
 
         chunked = not (request.body is None or 'Content-Length' in request.headers)
 
-        if isinstance(timeout, tuple):
+        
+        if isinstance(timeout, TimeoutSauce):
+            pass
```
Lukasa: Can this be moved lower down the list? A timeout object isn't a tuple, so the original `isinstance(tuple)` check won't catch it, so let's try to avoid rearranging the code too much.

requests/adapters.py
```diff
@@ -418,11 +421,14 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
                        "timeout tuple, or a single float to set "
                        "both timeouts to the same value".format(timeout))
                 raise ValueError(err)
-        else:
+        elif timeout is not None:
             timeout = TimeoutSauce(connect=timeout, read=timeout)
```
Lukasa: Unfortunately, this change is breaking. You'll either need to remove the default timeout change, or move this PR to the 3.0 branch.

You're also welcome to split this PR into two parts, the first bit making just the non-breaking changes against the master branch and the second bit making the breaking ones against 3.0.

leo-b: Hmm. Is it really breaking?
Looking at urllib3, leaving out the timeout argument will result in calling ``connectionpool.HTTPConnectionPool._get_timeout(_Default)``, which will use ``Timeout.DEFAULT_TIMEOUT`` if the ``HTTPConnectionPool`` had been initialized without a timeout. And ``Timeout.from_float(DEFAULT_TIMEOUT)`` yields the same result as if it had been called with ``None``: https://github.com/shazow/urllib3/blob/master/urllib3/util/timeout.py#L117

So the results should be the same...

Lukasa: Nope, they aren't the same. They both validate to the identity, but that identity is then used as the actual timeout. Note: https://github.com/shazow/urllib3/blob/master/urllib3/util/timeout.py#L33

leo-b: Ah - ok. Your are right.
I removed the change for ``timeout`` is ``None``.

leo-b:
I've added copies of the existing timeout tests using Timeout() objects instead of timeout tuples.

tests/test_requests.py
```diff
@@ -33,6 +33,7 @@
 
 from .compat import StringIO, u
 from .utils import override_environ
+from requests.packages.urllib3.util import Timeout
```
Lukasa: Ah, so the tests are failing because you've aliased this name (`Timeout`) with the name of an exception. Might be worth changing this to `from requests.packages.urllib3.util import Timeout as Urllib3Timeout` or something similar.

Lukasa:
Great, thanks so much!

PR #3956, opened 2017-04-05
Fixes DeprecationWarnings in Python 3.6 due to invalid escape sequences

### LINKED ISSUE #3954: requests shows lots of DeprecationWarning on Python 3.6
Running tests on my project with the -Wd option now displays lots of DeprecationWarnings coming from requests.

From the Python 3.6 Changelog (https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior):
```A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases. (Contributed by Emanuel Barry in bpo-27364.)```

For reference, this is the commit that fixed those DeprecationWarnings in Django 1.11: https://github.com/django/django/commit/8119b679eb85cdc0ae3d321e54d06dd0200a1e82

rubendura:
Fixes #3954

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -59,7 +59,7 @@ def request(method, url, **kwargs):
 
 
 def get(url, params=None, **kwargs):
-    """Sends a GET request.
+    r"""Sends a GET request.
 
     :param url: URL for the new :class:`Request` object.
     :param params: (optional) Dictionary or bytes to be sent in the query string for the :class:`Request`.
@@ -73,7 +73,7 @@ def get(url, params=None, **kwargs):
 
 
 def options(url, **kwargs):
-    """Sends a OPTIONS request.
+    r"""Sends a OPTIONS request.
 
     :param url: URL for the new :class:`Request` object.
     :param \*\*kwargs: Optional arguments that ``request`` takes.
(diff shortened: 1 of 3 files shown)


sigmavirus24:
> Using raw strings for docstrings tends to surprise people.

How so?

sigmavirus24:
> My experience has been that people don't really grasp that you can prefix the string literal for docstrings, so they expect them to be plain strings.

My (admittedly anecdotal) experience has been with tools that suggest using raw-strings for this exact use-case (e.g., pydocstyle, etc.). *shrug* I don't particularly care either way.

rubendura:
I personally prefer raw strings, they remove any ambiguity and make it clear that escape sequences do not work inside the string. They also remove noise which makes them easier to read. But I don't really care either.

Should I squash this commits or is it fine this way?

Lukasa:
Squashing the commits might be nice.

nateprewitt:
I'm not partial to either approach but as a data point [PEP257](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring) seems to suggest raw docstrings are the way to go in the case of backslashes.

> Use r"""raw triple double quotes""" if you use any backslashes in your docstrings.

Lukasa:
Hrm. I'll accept that as a preponderance of evidence that I'm an idiot then. ;) I'm outvoted 2-to-1, so back to raw strings we go. Sorry @rubendura!

rubendura:
No worries. I'll revert that now.

rubendura:
Done. Should be ready now :)

Lukasa:
Yup, looks good. Thanks @rubendura!

Issue #3959, opened 2017-04-09
UnicodeDecodeError on redirect response rfc1804 to non ascii location


shmuelamar:
This issue happend to me on our production servers while requesting URL from the wild using requests.

When a server returns redirect response that follows rfc1804 (i.e. Location: //URI) and the URI contains non ASCII chars, requests fails with UnicodeDecodeError. 

This issue only happens on python2 (in python3 unicode is the default text string so all headers encoded into unicode, including the Location header).

test to reproduce:

```python
def test_redirect_rfc1804_to_non_ascii_location():
    def redirect_resp_handler(sock):
        request_content = consume_socket_content(sock, timeout=0.5)
        sock.send(text_301)
        return request_content

    close_server = threading.Event()
    server = Server(redirect_resp_handler, wait_to_close_event=close_server)
    text_301 = (
        u'HTTP/1.1 301 Moved Permanently\r\n'
        u'Location: //{0}:{1}/?Š\r\n'
        u'\r\n'.format(server.host, server.port)
    ).encode('utf8')

    with server as (host, port):
        url = u'http://{0}:{1}/'.format(host, port)
        r = requests.head(url=url, allow_redirects=True)
        assert r.status_code == 200
        assert len(r.history) == 1
        assert r.history[0].status_code == 301
        close_server.set()
```

AFAIK from [here](http://stackoverflow.com/questions/4400678/what-character-encoding-should-i-use-for-a-http-header), header fields should be US-ASCII encoded and must not contain non-ASCII chars, but the internet is a wild place so this things happen.

IMHO its the server response to return the location in urlencoded format but requests shouldn't fail on this error (maybe more informative error) or at least be consistent with python2 and python3 behavior.

**I can create a PR for fixing the issue but I'm not sure what is the best solution**, options are (what I can think of):
1. raise custom exception when Location field is not ASCII (might be inconsistent with previous versions)
2. try encode the Location field to utf8 and raise exception if encoding fails
3. on python2 decode all headers into unicode, similar to python3 (might break previous versions compatibility)

**stacktrace (on master branch):**

```python
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-4-24640a4df79e> in <module>()
----> 1 resp = requests.head('https://www.autobazar.eu/skoda-fabia-19-tdi-pd-rs-id9323271.html', allow_redirects=True)

/home/shmulik/so/requests/requests/api.py in head(url, **kwargs)
     96 
[...]

Issue #3888, opened 2017-02-24
Non-ascii (iso-8859-1) location headers are handled incorrectly on handling a redirect


atleta:
Hi,

Akkording to the RFC (2616, https://www.ietf.org/rfc/rfc2616.txt) the location header MAY contain non-ascii characters if encoded with iso-8859-1. However, this case is handled incorrectly by requests. 

httplib will correctly parse the header as iso-8859-1 (see parse_headers in http/client.py, around line 214 for python 3.5), but the resulting string, when following the redirect, is quoted without specifying an encoding by requests (see requote_uri in requests/utils.py):
```
def requote_uri(uri):
    """Re-quote the given URI.

    This function passes the given URI through an unquote/quote cycle to
    ensure that it is fully and consistently quoted.

    :rtype: str
    """
    safe_with_percent = "!#$%&'()*+,/:;=?@[]~"
    safe_without_percent = "!#$&'()*+,/:;=?@[]~"
    try:
        # Unquote only the unreserved characters
        # Then quote only illegal characters (do not quote reserved,
        # unreserved, or '%')
        print('Yo')
        return quote(unquote_unreserved(uri), safe=safe_with_percent)
    except InvalidURL:
        # We couldn't unquote the given URI, so let's try quoting it, but
        # there may be unquoted '%'s in the URI. We need to make sure they're
        # properly quoted so they do not cause issues elsewhere.
        return quote(uri, safe=safe_without_percent)
```

The calls to quote should have ```encoding='iso-8859-1'``` as a parameter otherwise quote will assume utf-8 and that will result in a wrong url. I.e.:

```
return quote(unquote_unreserved(uri), encoding='iso-8859-1', safe=safe_with_percent)
```

I have looked into fixing it, which I'm willing to do, but not 100% sure where to add the test. My best guess is doing it in test_lowlevel.py using the test_server (httpbin can't be used for this as it quotes the non-ascii characters).

sigmavirus24:
Hi there @atleta,

First, RFC 2616 has been made obsolete by RFCs [7230][], [7231][], [7232][], [7233][], [7234][], and [7235][].

As you'll see when you familiarize yourself with those RFCs, the *default* encoding for Headers is no longer ISO-8859-1 (a.k.a., Latin-1).

I suspect that if you have a specific encoding you want to do you should disallow redirects and take control over URL encoding yourself. @Lukasa may disagree, though.

[7230]: https://tools.ietf.org/html/rfc7230
[...]

PR #3963, opened 2017-04-12
fix unicode decode error on py2 when handling redirect without scheme

shmuelamar:
as discussed on #3959 , fix the UnicodeDecodeError happening because implicit decoding of str string into unicode while using python2.

This issue happens when handling redirection without scheme  (i.e. Location: //URL) on python2 and only if the location is unicode.

First time applying a PR on GitHub so let me know if there is something I can do to make it better :).

Thanks,
Shmulik

### DIFF
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 
-import os
 import pytest
 import threading
 import requests
+from requests.compat import quote, is_py3
 
 from tests.testserver.server import Server, consume_socket_content
 
@@ -204,3 +204,40 @@ def test_use_proxy_from_environment(httpbin, var, scheme):
 
         # it had actual content (not checking for SOCKS protocol for now)
         assert len(fake_proxy.handler_results[0]) > 0
+
+
+def test_redirect_rfc1808_to_non_ascii_location():
+    path = u'š'
+    expected_path = quote(path.encode('utf8')).encode('ascii')
+    expected_path_py3 = b'%C3%85%C2%A1'
+    redirect_request = []  # stores the second request to the server
+
+    def redirect_resp_handler(sock):
+        consume_socket_content(sock, timeout=0.5)
+        location = u'//{0}:{1}/{2}'.format(host, port, path)
+        sock.send(
+            b'HTTP/1.1 301 Moved Permanently\r\n'
+            b'Content-Length: 0\r\n'
+            b'Location: ' + location.encode('utf8') + b'\r\n'
+            b'\r\n'
+        )
+        redirect_request.append(consume_socket_content(sock, timeout=0.5))
+        sock.send(b'HTTP/1.1 200 OK\r\n\r\n')
+
+    close_server = threading.Event()
+    server = Server(redirect_resp_handler, wait_to_close_event=close_server)
+
+    with server as (host, port):
+        url = u'http://{0}:{1}'.format(host, port)
+        r = requests.get(url=url, allow_redirects=True)
+        assert r.status_code == 200
+        assert len(r.history) == 1
+        assert r.history[0].status_code == 301
+
+        # currently Python3 not handling non-ASCII redirects (issue #3888)
+        if is_py3:
+            assert redirect_request[0].startswith(b'GET /' + expected_path_py3 + b' HTTP/1.1')
+        else:
+            assert redirect_request[0].startswith(b'GET /' + expected_path + b' HTTP/1.1')
+
+        close_server.set()
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -179,4 +179,4 @@ Patches and Suggestions
 - Matt Kohl (`@mattkohl <https://github.com/mattkohl>`_)
 - Jonathan Vanasco (`@jvanasco <https://github.com/jvanasco>`_)
 - David Fontenot (`@davidfontenot <https://github.com/davidfontenot>`_)
-
+- Shmuel Amar (`@shmuelamar <https://github.com/shmuelamar>`_)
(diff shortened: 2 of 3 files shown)


tests/test_lowlevel.py
```diff
@@ -204,3 +203,26 @@ def test_use_proxy_from_environment(httpbin, var, scheme):
...
+
+    close_server = threading.Event()
+    server = Server(redirect_resp_handler, wait_to_close_event=close_server)
+
+    with server as (host, port):
+        url = u'http://{0}:{1}'.format(host, port)
+        r = requests.head(url=url, allow_redirects=True)
+        assert r.status_code == 200
+        assert len(r.history) == 1
+        assert r.history[0].status_code == 301
```
nateprewitt: @shmuelamar could we confirm that the redirected URL is what we expect here?

tests/test_lowlevel.py
```diff
@@ -204,3 +203,26 @@ def test_use_proxy_from_environment(httpbin, var, scheme):
...
         # it had actual content (not checking for SOCKS protocol for now)
         assert len(fake_proxy.handler_results[0]) > 0
+
+
+def test_redirect_rfc1808_to_non_ascii_location():
+    def redirect_resp_handler(sock):
+        consume_socket_content(sock, timeout=0.5)
+        sock.send(
+            b'HTTP/1.1 301 Moved Permanently\r\n'
+            b'Location: //%s:%d/?\xc5\xa0\r\n'
```
nateprewitt: Unfortunately, neither `%` or `format` work on byte strings in Python 3 prior to 3.5. This means this string will need to be constructed another way. Either using `join` on multiple byte strings, or the `+` operator will probably be the best way to approach this for all Python versions.

tests/test_lowlevel.py
```diff
@@ -204,3 +203,26 @@ def test_use_proxy_from_environment(httpbin, var, scheme):
...
+        )
+        consume_socket_content(sock, timeout=0.5)
+        sock.send(b'HTTP/1.1 200 OK\r\n\r\n')
+
+    close_server = threading.Event()
+    server = Server(redirect_resp_handler, wait_to_close_event=close_server)
+
+    with server as (host, port):
+        url = u'http://{0}:{1}'.format(host, port)
+        r = requests.head(url=url, allow_redirects=True)
```
Lukasa: Any reason this shouldn't be a get?

tests/test_lowlevel.py
shmuelamar: Hey @nateprewitt thanks for the quick review!

I would be glad to make it happen but the issue is that current decoding into native string of headers makes Location header on python2 encoded string and on python3 decoded string. so this test won't be consistent across python version.

before this PR we get for the same test the following redirected URL:
python3: http://localhost:58792/?%C3%85%C2%A0
python2: UnicodeDecodeError (due to implicit decoding into unicode on string formatting)

this PR only fixes the python2 UnicodeDecodeError (python3 behaves the same).

there is an encoding issue #3888 which leads to inconsistency between python versions as each python version handles different type of strings.

I agree that this PR is far from solving #3888 (which I'll be happy to discuss about possible solutions).

My ideas are (just note a few, regardless of order):

1. keep this PR as is
2. add xfail for this test
3. test only `the resolve_redirects()` function
4. discuss possible for solution for #3888 that I might implement in the same PR
5. leave all as is and ignore this PR

of course I am open to any suggestions :smile:

shmuelamar:
agree, its probably better - changed

tests/test_lowlevel.py
nateprewitt: Thanks for the suggestions, @shmuelamar! It's a bit unfortunate that we're further exposing the inconsistent functionality between versions by fixing this. That said, this fixes the problem it's intended to address so I guess we can pass on this check until #3888 is fixed.

Lukasa: I'd be ok adding a check on Python version to let us make the correct assertion.

shmuelamar: add the assertion for url.
please let me know if there is a better way or more common way to ignore this assertion for python3.

please let me know if any other changes needed :smile:

tests/test_lowlevel.py
```diff
@@ -204,3 +204,35 @@ def test_use_proxy_from_environment(httpbin, var, scheme):
...
+        sock.send(
+            b'HTTP/1.1 301 Moved Permanently\r\n'
+            b'Content-Length: 0\r\n'
+            b'Location: ' + location.encode('utf8') + b'\r\n'
+            b'\r\n'
+        )
+        second_request = consume_socket_content(sock, timeout=0.5)
+
+        # currently Python3 not handling non-ASCII redirects (issue #3888)
+        assert is_py3 or second_request.startswith(b'GET /' + expected_path)
```
Lukasa: Right now we should probably assert what Python 3 *does* do, even if it's wrong, and just put a note on that assertion to fix it up when we change #3888.

shmuelamar: agree, changed it

tests/test_lowlevel.py
```diff
@@ -204,3 +204,39 @@ def test_use_proxy_from_environment(httpbin, var, scheme):
...
+            b'Location: ' + location.encode('utf8') + b'\r\n'
+            b'\r\n'
+        )
+        second_request = consume_socket_content(sock, timeout=0.5)
+
+        # currently Python3 not handling non-ASCII redirects (issue #3888)
+        if is_py3:
+            assert second_request.startswith(b'GET /' + expected_path_py3 + b' HTTP/1.1')
+        else:
+            assert second_request.startswith(b'GET /' + expected_path + b' HTTP/1.1')
```
Lukasa: Ok, one last note. It's slightly nicer to save this into a mutable object in the enclosing scope (e.g. a list) and then put the assertion in the enclosing function rather than the socket handler. assertions in the socket handler *do* work, but it's usually nicer to let the socket handler run to completion rather than blow it up and possibly leak it.

shmuelamar: once again this is probably better :smile: . I hope I understand you correctly (changed it again).

PR #3969, opened 2017-04-17
fix handle of non-ascii location on redirects issue #3888

shmuelamar:
fix for #3888

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -12,7 +12,7 @@ from collections import Mapping
 from datetime import datetime
 
 from .auth import _basic_auth_str
-from .compat import cookielib, OrderedDict, urljoin, urlparse
+from .compat import cookielib, is_py3, OrderedDict, urljoin, urlparse
 from .cookies import (
     cookiejar_from_dict, extract_cookies_to_jar, RequestsCookieJar, merge_cookies)
 from .models import Request, PreparedRequest, DEFAULT_REDIRECT_LIMIT
@@ -90,7 +90,16 @@ class SessionRedirectMixin(object):
     def get_redirect_target(self, resp):
         """Receives a Response. Returns a redirect URI or ``None``"""
         if resp.is_redirect:
-            return resp.headers['location']
+            location = resp.headers['location']
+            # Currently the underlying http module on py3 decode headers
+            # in latin1, but empirical evidence suggests that latin1 is very
+            # rarely used with non-ASCII characters in HTTP headers.
+            # It is more likely to get UTF8 header rather than latin1.
+            # This causes incorrect handling of UTF8 encoded location headers.
+            # To solve this, we re-encode the location in latin1.
+            if is_py3:
+                location = location.encode('latin1')
+            return to_native_string(location, 'utf8')
         return None
 
     def resolve_redirects(self, resp, req, stream=False, timeout=None,
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -3,7 +3,6 @@
 import pytest
 import threading
 import requests
-from requests.compat import quote, is_py3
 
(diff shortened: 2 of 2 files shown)


requests/sessions.py
```diff
@@ -90,7 +90,13 @@ class SessionRedirectMixin(object):
     def get_redirect_target(self, resp):
         """Receives a Response. Returns a redirect URI or ``None``"""
         if resp.is_redirect:
-            return resp.headers['location']
+            location = resp.headers['location']
+            # currently the underling http module on py3 decode headers
```
Lukasa: s/underling/underlying/

requests/sessions.py
```diff
@@ -90,7 +90,13 @@ class SessionRedirectMixin(object):
...
+            # in latin-1. this causes incorrect handling of UTF8 encoded
+            # location headers. to solve this we re-encode the location
```
Lukasa: So I think this comment needs to have a bit more elaboration. Specifically, we need to talk about why we think we're more likely to get UTF-8 encoded location headers than Latin-1: specifically, because empirical evidence suggests that latin-1 is very rarely used with non-ASCII characters in HTTP headers.

tests/test_lowlevel.py
```diff
@@ -209,7 +209,6 @@ def test_use_proxy_from_environment(httpbin, var, scheme):
 def test_redirect_rfc1808_to_non_ascii_location():
     path = u'š'
     expected_path = quote(path.encode('utf8')).encode('ascii')
```
nateprewitt: It seems like it might be more resilient/explicit to hard code the expected path here? It would divest us from relying on `quote` in the test, saves us an import, and hopefully clear up this line a bit.

At the very least, a comment on what we're doing here seems good.

tests/test_lowlevel.py
```diff
@@ -233,11 +232,6 @@ def redirect_resp_handler(sock):
         assert r.status_code == 200
         assert len(r.history) == 1
         assert r.history[0].status_code == 301
-
-        # currently Python3 not handling non-ASCII redirects (issue #3888)
-        if is_py3:
-            assert redirect_request[0].startswith(b'GET /' + expected_path_py3 + b' HTTP/1.1')
-        else:
-            assert redirect_request[0].startswith(b'GET /' + expected_path + b' HTTP/1.1')
+        assert redirect_request[0].startswith(b'GET /' + expected_path + b' HTTP/1.1')
```
nateprewitt, line 234: We're kind of testing the test here. Perhaps also adding a check that `r.url` is what we expect would be helpful?

requests/sessions.py
nateprewitt: I agree with @Lukasa. Also this is a minor nit, but capitals and periods would be great here 😊

shmuelamar: added more info about the issue.

requests/sessions.py
shmuelamar: yes :)

shmuelamar:
thanks for the quick reply, if you think we should make the test parameterized with some more location headers on it, let me know.

requests/sessions.py
shmuelamar: added capitals and periods @nateprewitt, im not native english so feel free to help me rephrase it.

tests/test_lowlevel.py
shmuelamar: agree changed

tests/test_lowlevel.py
shmuelamar: agree, added assert for url

nateprewitt:
Yep, looks good to me too!

Lukasa:
Alright let's do it!

shmuelamar:
great thanks for merging and the fast responses! 😃

Issue #3976, opened 2017-04-20
Don't use `urllib.proxy_bypass` on Windows


schlamar:
The `urllib.proxy_bypass` implementation on Windows is using forward and reverse DNS requests. This can result in major performance issues (I guess this depends on the DNS server, right now I have seen delays up to 15 seconds!). Additionally, this behavior is contrary to all major browser implementations.

This could even be a security issue as people might rely on the proxy as a privacy feature and in this case the client shouldn't do any DNS requests as this could leak private information.

I would recommend using a custom implementation which isn't doing any DNS request.

References:

* https://github.com/kennethreitz/requests/issues/2988
* http://bugs.python.org/issue29533

If you are in favor of this proposal I might be able to do a PR.

Lukasa:
In principle we are not attached to the specific implementation of `proxy_bypass`, only the net result. Any change away would be required to reproduce all relevant functionality and to provide complete test coverage. If you're open to doing that work we'll happily consider the PR.

schlamar:
Basically, this would be the copy/pasted Windows versions of [ `urllib.proxy_bypass`](https://github.com/python/cpython/blob/98b1c82/Lib/urllib.py#L1633) and [`urllib.proxy_bypass_registry`](https://github.com/python/cpython/blob/98b1c82/Lib/urllib.py#L1581), the latter without the DNS lookups.

Not quite sure how to test that in a sensible way. Maybe mocking/monkeypatching registry access (`_winreg`)?

schlamar:
Mainly note for myself: `urllib.proxy_bypass_registry` with host `172.16.1.12` and proxyOverride `172.16.1.1` evaluates to True. Not sure if this is a bug or implementation detail (matching start of string).

schlamar:
proxyOverride `192.168.` (without star) and host `192.168.0.1` also evaluates to True. 

In the config dialog it says "do not use proxy server for addresses **beginning with**" (screenshot [here](https://blogs.msdn.microsoft.com/askie/2015/10/12/how-to-configure-proxy-settings-for-ie10-and-ie11-as-iem-is-not-available/)). So I assume the urllib matching behavior is correct and matches the Window behavior.

schlamar:
#3979

PR #3979, opened 2017-04-20
improve proxy bypass on Windows 

schlamar:
See #3976

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -37,7 +37,9 @@ except (ImportError, SyntaxError):
 # ---------
 
 if is_py2:
-    from urllib import quote, unquote, quote_plus, unquote_plus, urlencode, getproxies, proxy_bypass
+    from urllib import (
+        quote, unquote, quote_plus, unquote_plus, urlencode, getproxies,
+        proxy_bypass, proxy_bypass_environment, getproxies_environment)
     from urlparse import urlparse, urlunparse, urljoin, urlsplit, urldefrag
     from urllib2 import parse_http_list
     import cookielib
@@ -54,7 +56,7 @@ if is_py2:
 
 elif is_py3:
     from urllib.parse import urlparse, urlunparse, urljoin, urlsplit, urlencode, quote, unquote, quote_plus, unquote_plus, urldefrag
-    from urllib.request import parse_http_list, getproxies, proxy_bypass
+    from urllib.request import parse_http_list, getproxies, proxy_bypass, proxy_bypass_environment, getproxies_environment
     from http import cookiejar as cookielib
     from http.cookies import Morsel
     from io import StringIO
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -14,6 +14,7 @@ import collections
 import contextlib
 import io
 import os
+import platform
 import re
 import socket
 import struct
@@ -26,7 +27,8 @@ from ._internal_utils import to_native_string
 from .compat import parse_http_list as _parse_list_header
 from .compat import (
     quote, urlparse, bytes, str, OrderedDict, unquote, getproxies,
-    proxy_bypass, urlunparse, basestring, integer_types)
+    proxy_bypass, urlunparse, basestring, integer_types, is_py3,
+    proxy_bypass_environment, getproxies_environment)
 from .cookies import cookiejar_from_dict
 from .structures import CaseInsensitiveDict
 from .exceptions import (
@@ -35,6 +37,54 @@ from .exceptions import (
 NETRC_FILES = ('.netrc', '_netrc')
 
 DEFAULT_CA_BUNDLE_PATH = certs.where()
+
+
+if platform.system() == 'Windows':
+    # provide a proxy_bypass version on Windows without DNS lookups
+
+    def proxy_bypass_registry(host):
+        if is_py3:
+            import winreg
+        else:
+            import _winreg as winreg
+        try:
+            internetSettings = winreg.OpenKey(winreg.HKEY_CURRENT_USER,
+                r'Software\Microsoft\Windows\CurrentVersion\Internet Settings')
+            proxyEnable = winreg.QueryValueEx(internetSettings,
+                                              'ProxyEnable')[0]
+            proxyOverride = winreg.QueryValueEx(internetSettings,
+                                                'ProxyOverride')[0]
+        except OSError:
+            return False
+        if not proxyEnable or not proxyOverride:
+            return False
+
+        # make a check value list from the registry entry: replace the
+        # '<local>' string by the localhost entry and the corresponding
+        # canonical entry.
+        proxyOverride = proxyOverride.split(';')
+        # now check if we match one of the registry values.
+        for test in proxyOverride:
+            if test == '<local>':
+                if '.' not in host:
+                    return True
+            test = test.replace(".", r"\.")     # mask dots
+            test = test.replace("*", r".*")     # change glob sequence
+            test = test.replace("?", r".")      # change glob char
+            if re.match(test, host, re.I):
+                return True
+        return False
+
+    def proxy_bypass(host):  # noqa
+        """Return True, if the host should be bypassed.
+
+        Checks proxy settings gathered from the environment, if specified,
+        or the registry.
+        """
+        if getproxies_environment():
+            return proxy_bypass_environment(host)
+        else:
+            return proxy_bypass_registry(host)
 
 
 def dict_to_sequence(d):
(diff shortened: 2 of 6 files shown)


schlamar:
As first step I added tests to check the current behavior (they take around 30 seconds to run on my system) so we can detect regressions in the new implementation. 

Tested with Python 2.7 and 3.5.

schlamar:
Well coverage drops obviously as these tests are only run on Windows :-/

From my POV this is ready to get merged, please review.

requests/utils.py
```diff
@@ -39,6 +40,65 @@
...
+            # Std modules, so should be around - but you never know!
+            return 0
+        try:
+            internetSettings = winreg.OpenKey(winreg.HKEY_CURRENT_USER,
+                r'Software\Microsoft\Windows\CurrentVersion\Internet Settings')
+            proxyEnable = winreg.QueryValueEx(internetSettings,
+                                              'ProxyEnable')[0]
+            proxyOverride = str(winreg.QueryValueEx(internetSettings,
+                                                    'ProxyOverride')[0])
+            # ^^^^ Returned as Unicode but problems if not converted to ASCII
```
Lukasa: This comment is confusing: what is it trying to say?

requests/utils.py
```diff
@@ -39,6 +40,65 @@
...
+            if test == '<local>':
+                if '.' not in rawHost:
+                    return 1
+            test = test.replace(".", r"\.")     # mask dots
+            test = test.replace("*", r".*")     # change glob sequence
+            test = test.replace("?", r".")      # change glob char
+            for val in host:
+                if re.match(test, val, re.I):
+                    return 1
+        return 0
```
Lukasa: We may as well just use booleans.

requests/utils.py
```diff
@@ -39,6 +40,65 @@
...
+                                                    'ProxyOverride')[0])
+            # ^^^^ Returned as Unicode but problems if not converted to ASCII
+        except OSError:
+            return 0
+        if not proxyEnable or not proxyOverride:
+            return 0
+
+        # try to make a host list from name and IP address.
+        rawHost, port = splitport(host)
+        host = [rawHost]
```
Lukasa: What is this for?

requests/utils.py
```diff
@@ -39,6 +40,65 @@
...
+        # make a check value list from the registry entry: replace the
+        # '<local>' string by the localhost entry and the corresponding
+        # canonical entry.
+        proxyOverride = proxyOverride.split(';')
+        # now check if we match one of the registry values.
+        for test in proxyOverride:
```
Lukasa, line 67: We can probably unify these two lines.

tests/test_utils.py
```diff
@@ -598,4 +601,53 @@ def test_should_bypass_proxies_no_proxy(
...
+            ('http://google.com:5000/v1.0/', False),
+    ))
+def test_should_bypass_proxies_win_registry(url, expected, monkeypatch):
+    """Tests for function should_bypass_proxies to check if proxy
+    can be bypassed or not with Windows registry settings
+    """
+    if compat.is_py3:
+        import winreg
+    else:
+        import _winreg as winreg
```
Lukasa: May as well just import this directly from compat.

schlamar:
The code is copy pasted from Python 3.5's urlib/request.py with the lines doing DNS lookups (and appending to `host`) removed. I would recommend leaving it as is (maybe add a clarifying comment about this).

Lukasa:
I'm a bit disinclined to do that unless we think this is going to be temporary code (which it won't be until we drop 2.7 support at least). If we are going to support it I'd like the code to be sensible and comprehensible. 😃

schlamar:
@Lukasa Please give me a note when AppVeyor tests are working. Then I'll udpate this PR.

tests/test_utils.py
schlamar: Then we would need a platform check in compat and winreg could be `None` on non Windows platforms...

schlamar:
@Lukasa Should I revert c121b98c4eb1e4ee8927995d9b9f1d6bdccd9349? With this PR there is no reason anymore for caching the results...

Lukasa:
@schlamar Yeah, might be worth doing.

kennethreitz:
close/reopened to get appveyor to re-run

Lukasa:
It'll need to be rebased on top of master.

requests/utils.py
```diff
@@ -37,6 +38,60 @@
...
+
+        # make a check value list from the registry entry: replace the
+        # '<local>' string by the localhost entry and the corresponding
+        # canonical entry.
+        proxyOverride = proxyOverride.split(';')
+        # now check if we match one of the registry values.
+        for test in proxyOverride:
+            if test == '<local>':
+                if '.' not in host:
+                    return 1
```
nateprewitt: Is there a reason this is `1` and not `True` like the other return values? If so, we should probably add a comment explaining why.

requests/compat.py
```diff
@@ -37,7 +37,8 @@
 # ---------
 
 if is_py2:
-    from urllib import quote, unquote, quote_plus, unquote_plus, urlencode, getproxies, proxy_bypass
+    from urllib import quote, unquote, quote_plus, unquote_plus, urlencode, getproxies, proxy_bypass, \
```
nateprewitt: It may be better to follow something closer to the import convention of the other modules, rather than use `\`.
```python
from urllib import (
    quote, unquote, quote_plus, unquote_plus, urlencode, getproxies,
    proxy_bypass, proxy_bypass_environment, getproxies_environment)
```

requests/utils.py
```diff
@@ -37,6 +38,60 @@
...
+            if re.match(test, host, re.I):
+                return True
+        return False
+
+    def proxy_bypass(host):  # noqa
+        """Return a dictionary of scheme -> proxy server URL mappings.
+
+        Returns settings gathered from the environment, if specified,
+        or the registry.
+
```
nateprewitt: Could we remove this extra line here.

requests/utils.py
```diff
@@ -37,6 +38,60 @@
...
+        """
+        if getproxies_environment():
+            return proxy_bypass_environment(host)
+        else:
+            return proxy_bypass_registry(host)
```
nateprewitt, line 87: This seems to return a boolean but the docstring states it returns a dictionary of scheme to proxy mappings. It may be a good idea to clarify this case.

requests/utils.py
schlamar: Nope, I forgot this one 🙈

requests/utils.py
schlamar: Somehow I copied the wrong docstring, I'll fix this :)

schlamar:
@Lukasa @nateprewitt Updated, please review again.

schlamar:
BTW, coverage drops because reverting 8e6e47af43d1d89a0201bd58e97c700bebbddd5e and c121b98c4eb1e4ee8927995d9b9f1d6bdccd9349 removes a lot of covered lines. My changes alone do not negatively affect coverage.

Lukasa:
@schlamar Based on codecov's output the actual reason you're reducing coverage is that the diff isn't covered enough to avoid regressing it. To match coverage you'd need to cover 90% of the diff, but only 86% is covered. You can see the lines we're missing [here](https://codecov.io/gh/kennethreitz/requests/compare/26ab951300d192ed3ab62fdb1f0548c961731e6a...32a34722be08aecb6af10ff249480bb255831ff3/diff): what are the odds that we can write tests that appropriately cover them?

requests/utils.py
```diff
@@ -37,6 +38,59 @@
 DEFAULT_CA_BUNDLE_PATH = certs.where()
 
 
+if os.name == 'nt':
```
Lukasa: Rather than use `os.name`, might it be better to use `platform.system`?

requests/utils.py
```diff
@@ -37,6 +38,59 @@
...
+
+    def proxy_bypass_registry(host):
+        try:
+            if is_py3:
+                import winreg
+            else:
+                import _winreg as winreg
+
+        except ImportError:
+            # Std modules, so should be around - but you never know!
```
schlamar: @Lukasa Should I remove this one, I don't think this is really necessary...

Lukasa: Yeah, let's do that: there's no need to tolerate this `ImportError` I don't think.

schlamar:
@Lukasa Updated, please review again :)

schlamar:
When was the proxy bypass cache introduced? I'd like to make a reference but can't find anything in the changelog :)

nateprewitt:
@schlamar, I believe it was only a couple months ago (#3885).

schlamar:
Ah yes, this wasn't even released yet. @Lukasa in this case I would ignore this?!

Lukasa:
Yeah, you would. =)

sethmlarson:
:tada:

PR #3983, opened 2017-04-21
Add appveyor testing of Requests.

Lukasa:
In general it'd be good to have more testing, but it's also possible that via PR #3979 we'll get unexpected different behaviour on Windows than on other platforms. For this reason, we should start also testing on Windows.

This PR takes the configuration used for urllib3, updates it for Requests, and then provides it. We'll still need to actually enable Appveyor so that it'll try to build, which I cannot do as I don't have admin on the repo. @kennethreitz, mind giving that a try?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,6 +30,11 @@ environment:
       PYTHON_ARCH: "64"
       TOXENV: "py35"
 
+    - PYTHON: "C:\\Python36-x64"
+      PYTHON_VERSION: "3.6.x"
+      PYTHON_ARCH: "64"
+      TOXENV: "py36"
+
 install:
   # Install Python (from the official .msi of http://python.org) and pip when
   # not already installed.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,56 @@
+# AppVeyor.yml from https://github.com/ogrisel/python-appveyor-demo
+# License: CC0 1.0 Universal: http://creativecommons.org/publicdomain/zero/1.0/
+
+build: off
+
+environment:
+  matrix:
+    - PYTHON: "C:\\Python266-x64"
+      PYTHON_VERSION: "2.6.6"
+      PYTHON_ARCH: "64"
+      TOXENV: "py26"
+
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.x"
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -0,0 +1,56 @@
+# AppVeyor.yml from https://github.com/ogrisel/python-appveyor-demo
+# License: CC0 1.0 Universal: http://creativecommons.org/publicdomain/zero/1.0/
+
+build: off
+
+environment:
+  matrix:
+    - PYTHON: "C:\\Python266-x64"
```
sigmavirus24, line 8: Didn't we drop support for 2.6?

appveyor.yml
```diff
@@ -0,0 +1,56 @@
...
+
+    - PYTHON: "C:\\Python34-x64"
+      PYTHON_VERSION: "3.4.x"
+      PYTHON_ARCH: "64"
+      TOXENV: "py34"
+
+    - PYTHON: "C:\\Python35-x64"
+      PYTHON_VERSION: "3.5.x"
+      PYTHON_ARCH: "64"
+      TOXENV: "py35"
```
sigmavirus24, line 31: Is 3.6 not supported on AppVeyor?

appveyor.yml
nateprewitt: Not until 3.0 was the decision I believe.

Lukasa: Correct.

appveyor.yml
Lukasa: It should be, this was just an oversight.

Lukasa: Ok, added

kennethreitz:
i'm very skeptical, but let's give it a shot

kennethreitz:
doesn't travis have a windows option?

kennethreitz:
build failed, looks like it's because of 2.6 maybe. https://ci.appveyor.com/project/kennethreitz/requests

Lukasa:
Travis does not have a Windows option and does not include one in their roadmap.

Lukasa:
So the reason those builds are failing is because I have to work out where the hell GNU make is on those systems. Given that our CI builds use make, I need to find where it is. If you're happy to be a bit patient, I'll open a PR to fix it in the next few hours

Issue #3980, opened 2017-04-21
iter_lines() sometimes yields broken results when chunk_size is small


PCMan:
Here is a small test program to demonstrate the bug:
```
import requests

url = "http://lohas.pixnet.net/blog"
r = requests.get(url)
iter_lines = [line for line in r.iter_lines(chunk_size=7, decode_unicode=False)]
split_lines = r.content.splitlines()
for index, (iline, sline) in enumerate(zip(iter_lines, split_lines)):
	if iline != sline:
		print("line {} is broken".format(index))
		print(iline, len(iline))
		print(sline, len(sline))
		break
```

Expected behavior:
r.iter_lines() should give the same result as r.content.splitlines()

Actual behavior:
The test program generates the following output:
```
line 275 is broken
b'' 0
b'\t\t\t<tbody>' 10
```

Changing chunk_size can break different lines.

PCMan:
It's not caused by incorrect unicode decoding.
I tested this with both `decode_unicode=True` and `decode_unicode=False`. In both cases the bug can be reproduced if chunk_size is small enough.

PCMan:
I currently use the following code snip as a drop-in replacement.
```
def my_iter_lines(result):
            buf = b''
            for chunk in result.iter_content(chunk_size=64 * 1024, decode_unicode=False):
                buf += chunk
                pos = 0
                while True:
                    eol = buf.find(b'\n', pos)
                    if eol != -1:
                        yield buf[pos:eol]
                        pos = eol + 1
                    else:
                        buf = buf[pos:]
                        break
            if buf:
                yield buf
```

Lukasa:
This is not a breakage, it's entirely expected behaviour. `iter_lines` takes a `chunk_size` argument that limits the size of the chunk it will return, which means it will occasionally yield before a line delimiter is reached. This is the behaviour `iter_lines` has always had and is expected to have by the vast majority of requests users.

To avoid this issue, you can set the chunk_size to be very large indeed. Alternatively, you'll need to process it by wrapping the `read` method on `response.raw` yourself and handling the line splitting using the IO module.

PCMan:
@Lukasa no this is not. See the API doc.
```
iter_lines(chunk_size=512, decode_unicode=None, delimiter=None)[source]
Iterates over the response data, one line at a time. When stream=True is set on the request, this avoids reading the content at once into memory for large responses.
```
[...]

PR #3984, opened 2017-04-21
Fix the additional newline generated by iter_lines() caused by a '\r\n' pair being separated in two different chunks.

PCMan:
When a "\r\n" CRLF pair is splitted into two chunks accidentally, iter_lines() generates two line breaks instead of one. This patch fixes the incorrect behavior and closes issue #3980.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -776,7 +776,11 @@ class Response(object):
 
         .. note:: This method is not reentrant safe.
         """
+        carriage_return = u'\r' if decode_unicode else b'\r'
+        line_feed = u'\n' if decode_unicode else b'\n'
+
         pending = None
+        last_chunk_ends_with_cr = False
 
         for chunk in self.iter_content(chunk_size=chunk_size,
                                        decode_unicode=decode_unicode):
@@ -796,6 +800,23 @@ class Response(object):
             if delimiter:
                 lines = chunk.split(delimiter)
             else:
+                # Python splitlines() supports the universal newline (PEP 278).
+                # That means, '\r', '\n', and '\r\n' are all treated as end of
+                # line. If the last chunk ends with '\r', and the current chunk
+                # starts with '\n', they should be merged and treated as only
+                # *one* new line separator '\r\n' by splitlines().
+                # This rule only applies when splitlines() is used.
+
+                # The last chunk ends with '\r', so the '\n' at chunk[0]
+                # is just the second half of a '\r\n' pair rather than a
+                # new line break. Just skip it.
+                skip_first_char = last_chunk_ends_with_cr and chunk.startswith(line_feed)
+                last_chunk_ends_with_cr = chunk.endswith(carriage_return)
+                if skip_first_char:
+                    chunk = chunk[1:]
+                    # it's possible that after stripping the '\n' then chunk becomes empty
+                    if not chunk:
+                        continue
                 lines = chunk.splitlines()
 
             # Calling `.split(delimiter)` will always end with whatever text
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1303,57 +1303,76 @@ class TestRequests:
         might not be lined up ideally.
         """
         mock_chunks = [
-            'This \r\n',
-            '',
-            'is\r',
-            '\n',
-            'a',
-            ' ',
-            '',
-            '',
-            'test.',
-            '\r',
-            '\n',
-            'end.',
+            b'This \r\n',
+            b'',
+            b'is\r',
+            b'\n',
+            b'a',
+            b' ',
+            b'',
+            b'',
+            b'test.',
+            b'\r',
+            b'\n',
+            b'end.',
         ]
-        mock_data = ''.join(mock_chunks)
+        mock_data = b''.join(mock_chunks)
+        unicode_mock_data = mock_data.decode('utf-8')
 
-        mock_iter_content = lambda *args, **kwargs: (e for e in mock_chunks)
+        def mock_iter_content(*args, **kwargs):
+            if kwargs.get("decode_unicode"):
+                return (e.decode('utf-8') for e in mock_chunks)
+            return (e for e in mock_chunks)
 
         r = requests.Response()
         r._content_consumed = True
         r.iter_content = mock_iter_content
 
-        assert list(r.iter_lines(delimiter='\r\n')) == mock_data.split('\r\n')
+        # decode_unicode=None, output raw bytes
+        assert list(r.iter_lines(delimiter=b'\r\n')) == mock_data.split(b'\r\n')
 
-        # Because '\n' is a single line-end, when `iter_lines()` receives
-        # the chunks containing a single '\n', it emits '' as a line -- whereas
-        # `.splitlines()` combines with the '\r' and splits on `\r\n`.
+        # decode_unicode=True, output unicode strings
+        assert list(r.iter_lines(decode_unicode=True, delimiter=u'\r\n')) == unicode_mock_data.split(u'\r\n')
+
+        # When delimiter is None, we should yield the same result as splitlines()
+        # which supports the universal newline.
+        # '\r', '\n', and '\r\n' are all treated as one line break.
+
+        # decode_unicode=None, output raw bytes
         result = list(r.iter_lines())
-        assert result != mock_data.splitlines()
-        assert result[2] == ''
-        assert result[4] == ''
+        assert result == mock_data.splitlines()
+
+        # decode_unicode=True, output unicode strings
+        result = list(r.iter_lines(decode_unicode=True))
+        assert result == unicode_mock_data.splitlines()
+
         # If we change all the line breaks to `\r`, we should be okay.
-        mock_chunks = [chunk.replace('\n', '\r') for chunk in mock_chunks]
-        mock_data = ''.join(mock_chunks)
+        # decode_unicode=None, output raw bytes
+        mock_chunks = [chunk.replace(b'\n', b'\r') for chunk in mock_chunks]
+        mock_data = b''.join(mock_chunks)
         assert list(r.iter_lines()) == mock_data.splitlines()
+
+        # decode_unicode=True, output unicode strings
+        unicode_mock_data = mock_data.decode('utf-8')
+        assert list(r.iter_lines(decode_unicode=True)) == unicode_mock_data.splitlines()
 
 
     @pytest.mark.parametrize(
         'content, expected_no_delimiter, expected_delimiter', (
-            ([''], [], []),
-            (['line\n'], ['line'], ['line\n']),
-            (['line', '\n'], ['line'], ['line\n']),
-            (['line\r\n'], ['line'], ['line', '']),
+            ([b''], [], []),
+            ([b'line\n'], [u'line'], [u'line\n']),
+            ([b'line', b'\n'], [u'line'], [u'line\n']),
+            ([b'line\r\n'], [u'line'], [u'line', u'']),
             # Empty chunk in the end of stream, same behavior as the previous
-            (['line\r\n', ''], ['line'], ['line', '']),
-            (['line', '\r\n'], ['line'], ['line', '']),
-            (['a\r', '\nb\r'], ['a', '', 'b'], ['a', 'b\r']),
-            (['a\n', '\nb'], ['a', '', 'b'], ['a\n\nb']),
-            (['a\r\n','\rb\n'], ['a', '', 'b'], ['a', '\rb\n']),
-            (['a\nb', 'c'], ['a', 'bc'], ['a\nbc']),
-            (['a\n', '\rb', '\r\nc'], ['a', '', 'b', 'c'], ['a\n\rb', 'c']),
-            (['a\r\nb', '', 'c'], ['a', 'bc'], ['a', 'bc'])  # Empty chunk with pending data
+            ([b'line\r\n', b''], [u'line'], [u'line', u'']),
+            ([b'line', b'\r\n'], [u'line'], [u'line', u'']),
+            ([b'a\r', b'\nb\r'], [u'a', u'b'], [u'a', u'b\r']),
+            ([b'a\r', b'\n', b'\nb'], [u'a', u'', u'b'], [u'a', u'\nb']),
+            ([b'a\n', b'\nb'], [u'a', u'', u'b'], [u'a\n\nb']),
+            ([b'a\r\n', b'\rb\n'], [u'a', u'', u'b'], [u'a', u'\rb\n']),
+            ([b'a\nb', b'c'], [u'a', u'bc'], [u'a\nbc']),
+            ([b'a\n', b'\rb', b'\r\nc'], [u'a', u'', u'b', u'c'], [u'a\n\rb', u'c']),
+            ([b'a\r\nb', b'', b'c'], [u'a', u'bc'], [u'a', u'bc'])  # Empty chunk with pending data
         ))
     def test_response_lines_parametrized(self, content, expected_no_delimiter, expected_delimiter):
         """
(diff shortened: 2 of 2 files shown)


requests/models.py
```diff
@@ -820,6 +827,9 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             for line in lines:
                 yield line
 
+            # check if the current chunk ends with '\r'
+            last_chunk_ends_with_cr = chunk.endswith('\r' if decode_unicode else b'\r')
```
Lukasa: Let's hoist the result of these two `else` calls to the top of the function, rather than check it twice each time around the loop:

```python
carriage_return = u'\r' if decode_unicode else b'\r'
line_feed = u'\n' if decode_unicode else b'\r'
```

requests/models.py
```diff
@@ -796,6 +797,12 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             if delimiter:
                 lines = chunk.split(delimiter)
             else:
+                # Edge case: if the last chunk ends with '\r', and the current chunk starts with \n,
+                # they should be merged and treated as only "one" new line separator '\r\n'.
+                # So the first '\n' in this chunk should be skipped since it's just the second half of 
+                # the CRLF pair ('\r\n') rather than another new line break.
```
Lukasa: We should expand this comment to explain that this only affects the `splitlines` case because `splitlines` will treat any of `\r`, `\r\n`, and `\n` as newlines, and so splitting `\r\n` into two chunks will get misleading results.

requests/models.py
Lukasa: It would also be good to hoist this into the block that holds the `splitlines` call, as we don't need to do this processing in any other situation.

requests/models.py
PCMan: OK, will do that.

PCMan:
@Lukasa Thanks for the review and I already modified the code by following your instructions.
However there seems to be some issues with the unit tests. I'll take a look on them.

PCMan:
@Lukasa While trying to fix the unit tests, I noted that treating "\r\n" as two line breaks is actually the intended behavior based on the current unit tests. While this behavior is incompatible with the default of python (universal newline) and is not explained in the requests API doc either, what `delimiters=None` does should be well defined and documented before we try to fix this. The unit tests should be fixed to reflect the change as well.

* If treating "\r\n" as two line breaks is intended, then the fix is avoiding the use of splitlines() and split the lines on either '\r' or '\n' outselves. Then, make this documented in the API doc.
* Otherwise, we should simulate the behavior of splitlines() and follow python universal newline handling, and fix all the unit tests and API doc to reflect this behavior change.

Either way, the current code needs some fix since different chunk size can yield different results.

Another issue with the current uint tests is it only works with python2. For working with python3, delimiter should be bytes rather than str if decode_unicode is not True, which is not addressed in the unit tests.
@Lukasa Your suggestions are needed. Thank you.

requests/models.py
```diff
@@ -796,6 +800,17 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
...
+                # That means, '\r', '\n', and '\r\n' are all treated as end of 
+                # line. If the last chunk ends with '\r', and the current chunk
+                # starts with \n, they should be merged and treated as only 
+                # "one" new line separator '\r\n' by splitlines().
+                # This rule only applies when splitlines() is used.
+                if last_chunk_ends_with_cr and chunk.startswith(line_feed):
+                    # The last chunk ends with '\r', so the '\n' at chunk[0]
+                    # is just the second half of a '\r\n' pair rather than a
+                    # new line break. Just skip it.
+                    chunk = chunk[1:]
```
nateprewitt: One thing to keep in mind here is the final line of the file ending in `\r\n`. If we split at `\r`, we get a final one byte line of `\n`. This code makes chunk equivalent to `''` and `''.splitlines()` will result in an empty list. You'll get an index out of bounds with the call `lines[-1]` below.

Probably a rarer edge case but something we should cover.

Lukasa:
The unit tests should be updated: `splitlines` is clearly intended to be part of the behaviour of the code and so the unit tests as written should be considered as codifying the v2 behaviour, rather than guiding the v3. We can change them for v3.

>  For working with python3, delimiter should be bytes rather than str if decode_unicode is not True, which is not addressed in the unit tests.

Yup, that should change as well.

nateprewitt:
It seems sensible to have the test data default be in bytes since that's what we should be receiving from urllib3. We can have a `parametrize` toggle if we want to test both True and False for `decode_unicode`.

PCMan:
@Lukasa @nateprewitt Just fixed all of the unit tests and the mentioned chunk[1:] problems. All related tests should pass now. Also provide a small test program to demonstrate the bug.
```
import requests

test_content = b"line1\r\nline2"

def mock_iter_content(chunk_size=1, decode_unicode=None):
    for i in range(0, len(test_content), chunk_size):
        yield test_content[i: i + chunk_size]

r = requests.Response()
r._content_consumed = True
r.iter_content = mock_iter_content

assert list(r.iter_lines(chunk_size=6)) == test_content.splitlines()
```
Thanks!

requests/models.py
PCMan: Fixed. Thanks.

requests/models.py
```diff
@@ -813,13 +831,17 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             #
             # If we're using `splitlines()`, we only do this if the chunk
             # ended midway through a line.
-            incomplete_line = lines[-1] and lines[-1][-1] == chunk[-1]
+            incomplete_line = lines[-1] and lines[-1][-1] and lines[-1][-1] == chunk[-1]
```
Lukasa: Why has the condition been doubled up here? Under what circumstances will `line[-1][-1]` be falsey?

requests/models.py
```diff
@@ -813,13 +831,17 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
...
             if delimiter or incomplete_line:
                 pending = lines.pop()
 
             for line in lines:
                 yield line
 
+            # check if the current chunk ends with '\r'
+            if delimiter is None:
+                last_chunk_ends_with_cr = chunk.endswith(carriage_return)
```
Lukasa: Again, this should be brought into the if branch above.

tests/test_requests.py
```diff
@@ -1303,57 +1303,75 @@ def test_response_lines(self):
...
 
-        assert list(r.iter_lines(delimiter='\r\n')) == mock_data.split('\r\n')
+        # decode_unicode=None, output raw bytes
+        assert list(r.iter_lines(delimiter=b'\r\n')) == mock_data.split(b'\r\n')
 
-        # Because '\n' is a single line-end, when `iter_lines()` receives
-        # the chunks containing a single '\n', it emits '' as a line -- whereas
-        # `.splitlines()` combines with the '\r' and splits on `\r\n`.
+        # decode_unicode=True, output unicode strings
+        assert list(r.iter_lines(decode_unicode=True, delimiter='\r\n')) == unicode_mock_data.split('\r\n')
```
Lukasa: These tests run on Python 2, so Unicode strings must be prefixed with `u`

tests/test_requests.py
```diff
@@ -1303,57 +1303,75 @@ def test_response_lines(self):
...
-            (['a\n', '\rb', '\r\nc'], ['a', '', 'b', 'c'], ['a\n\rb', 'c']),
-            (['a\r\nb', '', 'c'], ['a', 'bc'], ['a', 'bc'])  # Empty chunk with pending data
+            ([b'line\r\n', b''], ['line'], ['line', '']),
+            ([b'line', b'\r\n'], ['line'], ['line', '']),
+            ([b'a\r', b'\nb\r'], ['a', 'b'], ['a', 'b\r']),
+            ([b'a\n', b'\nb'], ['a', '', 'b'], ['a\n\nb']),
+            ([b'a\r\n',b'\rb\n'], ['a', '', 'b'], ['a', '\rb\n']),
+            ([b'a\nb', b'c'], ['a', 'bc'], ['a\nbc']),
+            ([b'a\n', b'\rb', b'\r\nc'], ['a', '', 'b', 'c'], ['a\n\rb', 'c']),
+            ([b'a\r\nb', b'', b'c'], ['a', 'bc'], ['a', 'bc'])  # Empty chunk with pending data
```
Lukasa: Again, your test seems to show that all the unprefixed strings are intended to be Unicode: they need to be prefixed

requests/models.py
PCMan: Sorry I forgot to remove them. It should be prevented by the check for chunk[1:].

PCMan:
@Lukasa Thanks for the detailed review. All problems are cleared now.

requests/models.py
```diff
@@ -796,7 +800,23 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             if delimiter:
                 lines = chunk.split(delimiter)
             else:
+                # Python splitlines() supports the universal newline (PEP 278).
```
sigmavirus24, line 803: Wasn't universal newline support dropped in Python 3? I know you can no longer get a TextIOBuffer object that supports that. How does that affect splitting strings?

requests/models.py
```diff
@@ -813,7 +833,7 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             #
             # If we're using `splitlines()`, we only do this if the chunk
             # ended midway through a line.
-            incomplete_line = lines[-1] and lines[-1][-1] == chunk[-1]
+            incomplete_line = lines[-1] and  lines[-1][-1] == chunk[-1]
```
sigmavirus24: Why is there an extra space between `and` and `lines[-1][-1]`?

requests/models.py
PCMan: Well, splitlines() actually supports a superset of universal newline.
FYI: https://docs.python.org/3/library/stdtypes.html#str.splitlines
The unicode strings supports a superset of universal newline.
However, the new python 3 bytes behaves differently.
https://docs.python.org/3/library/stdtypes.html#bytes.splitlines
In both cases, however, the '\r', '\n', and '\r\n' rule still apply.

requests/models.py
PCMan: removed.

requests/models.py
```diff
@@ -796,7 +800,23 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             if delimiter:
                 lines = chunk.split(delimiter)
             else:
+                # Python splitlines() supports the universal newline (PEP 278).
+                # That means, '\r', '\n', and '\r\n' are all treated as end of 
+                # line. If the last chunk ends with '\r', and the current chunk
+                # starts with \n, they should be merged and treated as only
```
Lukasa: Minor nit: as there are inverted commas around `'\r'` in the line above, they should be added here around `\n`.

requests/models.py
```diff
@@ -796,7 +800,23 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             if delimiter:
                 lines = chunk.split(delimiter)
             else:
+                # Python splitlines() supports the universal newline (PEP 278).
+                # That means, '\r', '\n', and '\r\n' are all treated as end of 
+                # line. If the last chunk ends with '\r', and the current chunk
+                # starts with \n, they should be merged and treated as only 
+                # "one" new line separator '\r\n' by splitlines().
```
Lukasa: Minor nit: should be `*one*`, not `"one"`.

PCMan:
@Lukasa Done! :)

requests/models.py
```diff
@@ -796,7 +800,23 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
...
+                    # The last chunk ends with '\r', so the '\n' at chunk[0]
+                    # is just the second half of a '\r\n' pair rather than a
+                    # new line break. Just skip it.
+                    chunk = chunk[1:]
+                    # it's possible that after stripping the '\n' then chunk becomes empty
+                    if not chunk:
+                        continue
                 lines = chunk.splitlines()
+                # check if the current chunk ends with '\r'
+                last_chunk_ends_with_cr = chunk.endswith(carriage_return)
```
Lukasa: Oooh, just noticed a subtle misbehaviour. Imagine we get the following chunks:

```python
[b'this is a string\r', b'\n', b'\nso is this']
```

This will not do the right thing: we'll strip *two* newline characters, rather than emit an empty line. That is, we'll emit: `[b'this is a string', b'so is this']` instead of `[b'this is a string', b'', b'so is this']`, which is what we *should* emit.

This is because a chunk that is *just* a `\n` causes an early continue, above, without resetting the value of the `last_chunk_ends_with_cr` boolean. Essentially it gets completely ignored. That's not right.

It'd be good to add a test case (or modify an existing one) to include that case, and then fix it up by hoisting this flag up before the `if not chunk: continue` clause.

tests/test_requests.py
```diff
@@ -1303,57 +1303,75 @@ def test_response_lines(self):
...
+        # When delimiter is None, we should yield the same result as splitlines()
+        # which supports the universal newline.
+        # '\r', '\n', and '\r\n' are all treated as one line break.
+
+        # decode_unicode=None, output raw bytes
         result = list(r.iter_lines())
-        assert result != mock_data.splitlines()
-        assert result[2] == ''
-        assert result[4] == ''
+        assert result == mock_data.splitlines()
```
Lukasa, line 1343: We can probably collapse this line into the one above it as with the other assertions:

```python
assert list(r.iter_lines()) == mock_data.splitlines()
```

PCMan: This piece of code seems to be outdated. I didnt' see it in the current HEAD.

requests/models.py
PCMan: @Lukasa Nice catch! Sorry that I overlooked such a simple check. Just add a simple fix along with a test case for it. BTW, I did not hoist the flag per your request. It will look somewhat like this:
```
                # a temp variable is needed here and I cannot find a good variable naming that is not confusing
                old_last_chunk_ends_with_cr = last_chunk_ends_with_cr
                last_chunk_ends_with_cr = chunk.endswith(carriage_return)
                if old_last_chunk_ends_with_cr and chunk.startswith(line_feed):
                    chunk = chunk[1:]
                    if not chunk:
                        continue
```
Also this yields an unnecessary method call endwiths('\r') since when chun[1:] is empty we already know the result is False. So I prefer just setting the flag to False though it looks like code duplication at first glance.
```
if not chunk:
    last_chunk_ends_with_cr = False
```
Simple and readable.

Lukasa: Hrm. I don't know if I think it's more readable to do it this way. It means there are two places in the code that set this variable, instead of just one, which forces the reader to ask themselves *why* it was done this way. I don't know that the efficiency gains are worth it. 🤔

tests/test_requests.py
Lukasa: Yup, you're right.

Lukasa:
You'll also need to rebase on top of the current master, please, to get the appveyor tests running.

sigmavirus24:
> So this is now looking good except for my possible concern about the readability of our two alternative approaches.

I think I'm missing something. I've looked through the PR but I don't see alternative approaches or readability issues. This could just be the woefully immature GitHub Review system hiding it from me though.

Lukasa:
Yeah, it is. See [this discussion](https://github.com/kennethreitz/requests/pull/3984#discussion_r112929801).

requests/models.py
```diff
@@ -796,7 +800,24 @@ def iter_lines(self, chunk_size=ITER_CHUNK_SIZE, decode_unicode=None, delimiter=
             if delimiter:
                 lines = chunk.split(delimiter)
             else:
+                # Python splitlines() supports the universal newline (PEP 278).
+                # That means, '\r', '\n', and '\r\n' are all treated as end of
+                # line. If the last chunk ends with '\r', and the current chunk
+                # starts with '\n', they should be merged and treated as only
+                # *one* new line separator '\r\n' by splitlines().
+                # This rule only applies when splitlines() is used.
+                if last_chunk_ends_with_cr and chunk.startswith(line_feed):
```
sigmavirus24: Now that I understand what concerns @Lukasa has, I agree. I think this could be better written as:

```py
chunk_startswith_line_feed = last_chunk_ends_with_cr and chunk.startswith(line_feed)
last_chunk_ends_with_cr = chunk.endswith(carriage_return)
if chunk_startswith_line_feed:
    chunk = chunk[1:]
    if not chunk:
        continue
lines = chunk.splitlines()
```

This properly *explains* the condition on this line with a descriptive variable name. It avoids the `old_last_chunk_ends_with_cr` variable and it keeps things concise and easy to reason about.

nateprewitt: Yep, I'm on board with @sigmavirus24's suggestion. That seems the most readable option out of what's been suggested.

The only minor input I have is standardizing the use of `_` in `ends_with` vs `startswith` in the variable names.

PCMan:
@Lukasa Fixed as suggested by @sigmavirus24 .

PCMan:
@Lukasa @sigmavirus24 git squashed. Thanks for your patience.

PR #3988, opened 2017-04-24
[WIP] Fix up appveyor testing.

Lukasa:
This PR will track work as I try to work out where the hell `make` is on appveyor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python:
   - "2.6"
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -13,6 +13,8 @@ Release History
   Advanced users will be able to process malformed redirects more easily.
 - Changed the internal calculation of elapsed request time to have higher
   resolution on Windows.
+- Added ``win_inet_pton`` as conditional dependency for the ``[socks]`` extra
+  on Windows with Python 2.7.
 
 2.13.0 (2017-01-24)
 +++++++++++++++++++
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -11,6 +11,8 @@ Release History
   ``get_redirect_target(response)`` instead of directly
   querying ``Response.is_redirect`` and ``Response.headers['location']``.
   Advanced users will be able to process malformed redirects more easily.
+- Changed the internal calculation of elapsed request time to have higher
+  resolution on Windows.
 
 2.13.0 (2017-01-24)
 +++++++++++++++++++
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ init:
 	pip install pipenv
 	pipenv lock
 	pipenv install --dev
+	pipenv run pip install -e .[socks]
 
 test:
 	# This runs all of the tests. To run an individual test, run py.test with
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,11 +5,6 @@ build: off
 
 environment:
   matrix:
-    - PYTHON: "C:\\Python266-x64"
-      PYTHON_VERSION: "2.6.6"
-      PYTHON_ARCH: "64"
-      TOXENV: "py26"
-
     - PYTHON: "C:\\Python27-x64"
       PYTHON_VERSION: "2.7.x"
       PYTHON_ARCH: "64"
@@ -30,6 +25,11 @@ environment:
       PYTHON_ARCH: "64"
       TOXENV: "py35"
 
+    - PYTHON: "C:\\Python36-x64"
+      PYTHON_VERSION: "3.6.x"
+      PYTHON_ARCH: "64"
+      TOXENV: "py36"
+
 install:
   # Install Python (from the official .msi of http://python.org) and pip when
   # not already installed.
@@ -47,10 +47,10 @@ install:
   # Upgrade to the latest version of pip to avoid it displaying warnings
   # about it being out of date.
   - "pip install --disable-pip-version-check --user --upgrade pip"
-  - "mingw32-make"
+  - "C:\\MinGW\\bin\\mingw32-make"
 
 test_script:
-  - "mingw32-make coverage"
+  - "C:\\MinGW\\bin\\mingw32-make coverage"
 
 on_success:
-  - "pipenv run codecov"
+  - "pipenv run codecov -f coverage.xml"
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -8,8 +8,10 @@ This module provides a Session object to manage and persist settings across
 requests (cookies, auth, proxies).
 """
 import os
+import platform
+import time
 from collections import Mapping
-from datetime import datetime
+from datetime import timedelta
 
 from .auth import _basic_auth_str
 from .compat import cookielib, is_py3, OrderedDict, urljoin, urlparse
@@ -37,6 +39,15 @@ from .status_codes import codes
 from .models import REDIRECT_STATI
 
 REDIRECT_CACHE_SIZE = 1000
+
+# Preferred clock, based on which one is more accurate on a given system.
+if platform.system() == 'Windows':
+    try:  # Python 3.3+
+        preferred_clock = time.perf_counter
+    except AttributeError:  # Earlier than Python 3.
+        preferred_clock = time.clock
+else:
+    preferred_clock = time.time
 
 
 def merge_setting(request_setting, session_setting, dict_class=OrderedDict):
@@ -622,13 +633,14 @@ class Session(SessionRedirectMixin):
         adapter = self.get_adapter(url=request.url)
 
         # Start time (approximately) of the request
-        start = datetime.utcnow()
+        start = preferred_clock()
 
         # Send the request
         r = adapter.send(request, **kwargs)
 
         # Total elapsed time of the request (approximately)
-        r.elapsed = datetime.utcnow() - start
+        elapsed = preferred_clock() - start
+        r.elapsed = timedelta(seconds=elapsed)
 
         # Response manipulation hooks
         r = dispatch_hook('response', hooks, r, **kwargs)
--- a/setup.py
+++ b/setup.py
@@ -97,6 +97,7 @@ setup(
     extras_require={
         'security': ['pyOpenSSL>=0.14', 'cryptography>=1.3.4', 'idna>=2.0.0'],
         'socks': ['PySocks>=1.5.6, !=1.5.7'],
+        'socks:platform_system == "Windows" and python_version<"3.3"': ['win_inet_pton'],
     },
 )
 


Lukasa:
Ok, some progress has been made. We are now getting build failures due to our dependencies.

Python 2.6 is failing because of click, presumably because click doesn't support Python 2.6 + Windows:

```
pipenv lock
  File "c:\python266-x64\lib\runpy.py", line 122, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\python266-x64\lib\runpy.py", line 34, in _run_code
    exec code in run_globals
  File "C:\Python266-x64\Scripts\pipenv.exe\__main__.py", line 5, in <module>
  File "c:\python266-x64\lib\site-packages\pipenv\__init__.py", line 14, in <module>
    from .cli import cli
  File "c:\python266-x64\lib\site-packages\pipenv\cli.py", line 11, in <module>
    import click
  File "c:\python266-x64\lib\site-packages\pipenv\vendor\click\__init__.py", line 18, in <module>
    from .core import Context, BaseCommand, Command, MultiCommand, Group, \
  File "c:\python266-x64\lib\site-packages\pipenv\vendor\click\core.py", line 8, in <module>
    from .types import convert_type, IntRange, BOOL
  File "c:\python266-x64\lib\site-packages\pipenv\vendor\click\types.py", line 4, in <module>
    from ._compat import open_stream, text_type, filename_to_ui, \
  File "c:\python266-x64\lib\site-packages\pipenv\vendor\click\_compat.py", line 536, in <module>
    from ._winconsole import _get_windows_console_stream
  File "c:\python266-x64\lib\site-packages\pipenv\vendor\click\_winconsole.py", line 19, in <module>
    from ctypes import byref, POINTER, c_int, c_char, c_char_p, \
ImportError: cannot import name c_ssize_t
```

Python 2.7 and later are failing at install time:

```
pipenv lock
Creating a virtualenv for this project...
Warning! If you are running on Windows, you should use the --python option instead.
Using base prefix 'c:\\python33-x64'
New python executable in C:\Users\appveyor\.virtualenvs\requests-hjhYtdkP\Scripts\python.exe
Installing setuptools, pip, wheel...done.
Virtualenv location: C:\Users\appveyor\.virtualenvs\requests-hjhYtdkP
Locking [dev-packages] dependencies...
  File "c:\python33-x64\lib\runpy.py", line 160, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\python33-x64\lib\runpy.py", line 73, in _run_code
    exec(code, run_globals)
  File "C:\Python33-x64\Scripts\pipenv.exe\__main__.py", line 9, in <module>
  File "c:\python33-x64\lib\site-packages\pipenv\vendor\click\core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "c:\python33-x64\lib\site-packages\pipenv\vendor\click\core.py", line 697, in main
    rv = self.invoke(ctx)
  File "c:\python33-x64\lib\site-packages\pipenv\vendor\click\core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\python33-x64\lib\site-packages\pipenv\vendor\click\core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\python33-x64\lib\site-packages\pipenv\vendor\click\core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "c:\python33-x64\lib\site-packages\pipenv\cli.py", line 893, in lock
    do_lock(no_hashes=no_hashes)
  File "c:\python33-x64\lib\site-packages\pipenv\cli.py", line 440, in do_lock
    results = get_downloads_info(names_map, 'dev-packages')
  File "c:\python33-x64\lib\site-packages\pipenv\cli.py", line 405, in get_downloads_info
    name = list(convert_deps_from_pip(names_map[fname]))[0]
KeyError: 'Babel-2.4.0-py2.py3-none-any.whl'
```

My *suspicion* here is that this is pipenv related. @nateprewitt, any insights into what's happening there?

Lukasa:
As to the Python 2.6 thing, how severe that is is unclear. We technically count 2.6 as a supported platform, so it's pretty ungraceful that you can't run our tests on it. On the other hand, pipenv doesn't support Python 2.6 on Windows (because, transitively, click doesn't), so maybe that's just not a surprise?

nateprewitt:
@Lukasa yep, that's definitely something wrong in Pipenv. I'm unsure if it's a transient issue or something more. Can we kick the build? I don't seem to have permissions on appveyor, or perhaps it doesn't have an option to rebuild the test.

Lukasa:
Yup, this is a duplicate of kennethreitz/pipenv#213. @nateprewitt, on that issue you said you wanted a repro environment: I think I got you one. ;)

Lukasa:
@nateprewitt The easiest thing to do, honestly, if @kennethreitz doesn't mind too much, is to just open a new PR and push to it repeatedly to kick off builds. That'll hopefully allow you to play around and work out what's happening.

Failing that, I think you'd either have to spin up a cloud Windows machine or wait for about an hour for me to get to my local Windows box.

nateprewitt:
Aha! Good catch, I'll spin up appveyor for pipenv after lunch and try to get that pinned down.

Lukasa:
@nateprewitt You may not be able to enable it for pipenv: we needed Kenneth to do it for Requests.

nateprewitt:
@lukasa I have 2.7 running tests on Windows now (although they're failing on Socks support) with a patched branch of pipenv. It's a casing conflict for file names. I think we can get that fix rolled out pretty quickly.

The click issue seems to be noted in pallets/click#735 but hasn't received any response. Not sure if they have any intent to support 2.6 in this environment.

Lukasa:
Ok cool, I'll strip 2.6 support when the pipenv fix lands.

nateprewitt:
Alright @Lukasa, sorry for the delay. I've cut 3.6.0 for pipenv, so this issue should be fixed if you want to rebase or push another commit.

There are a few more errors when you get past the pipenv failure that I haven't had time to fully investigate. From a cursory glance though, this is what I've found so far.

 * 2.6 needs to be removed from the test list for appveyor.
 * `win_inet_pton` appears to be a requirement for PySocks to run in Python 2.X on Windows.
 * The `test_time_elapsed_blank` was failing with what appeared to be a division rounding issue, but my latest builds this morning aren't producing this. Something to keep an eye out for though.

Lukasa:
Thanks @nateprewitt, I'll start working through that stuff now. =)

Lukasa:
Hey look, this is why we should be testing on Windows! It seems that `elapsed` doesn't work as expected on Windows, because `datetime.datetime.utcnow` has a pretty low resolution, meaning that sufficiently fast requests may be labelled as taking zero time.

I believe this will work more sensibly if we use a different function to build the timedelta object.

Lukasa:
Hrm. Sadly, using `time.time()` also doesn't fix the problem on Windows: the two timestamps remain the same. I think it's possible that we cannot get an appropriately accurate clock on a Windows machine for this test to run as expected.

Lukasa:
So, on Windows, we need to use `time.clock` to get the accurate timestamps we want. In this case, we probably need to use `time.time` and `time.clock` to get the accuracy we want, and then create the `timedelta` ourselves. Should be easy enough.

sethmlarson:
Probably shouldn't rely on `time.clock` as it's been [deprecated](https://docs.python.org/3/library/time.html#time.clock) since 3.3.
Is it valid to use `time.perf_counter` when available in this context?

Lukasa:
Oh god, who knows? I guess, based on the reading of that documentation, the answer is yes?

Lukasa:
Hrm, these clock changes do not appear to be helping. I wonder what that's about.

Lukasa:
Uh...awkward. Let's not talk about what I got wrong there.

Lukasa:
@nateprewitt Is it possible to declare a conditional dependency in a Pipfile?

nateprewitt:
@Lukasa I was thinking about that last week with the pton dependency requirement. We don't currently have a way to handle something like that, but from what I can tell neither does pip without conditional code in setup.py.

Perhaps we define the list used for `socks` in `extras_require` earlier and then append `win_inet_pton`  if `platform.system()` is Windows? That should work consistently for users without any syntax changes, and pipenv already uses `extras_requires` from setup.py.

Lukasa:
I think pip can do it fine with markers. I tried to add something in 63e5b80a6b1b386f9f43ac07948306d0fa2d333b but have yet to test it live: will do so shortly. @dstufft, am I right in assuming that I can add a marker that says "if you do `pip install requests[socks]` on Windows, install this other dependency?"

@nateprewitt Adding dependencies in code in `setup.py` is almost always a bad idea, because wheels don't execute `setup.py`. That means whether or not users get the Windows dependency boils down to the system on which the wheel was built: that is, if I build the wheel on my Mac then no-one gets the Windows dep, while if I build it on my Windows box *everyone* does. That's not really a winning strategy. =)

nateprewitt:
@Lukasa 😬 I'm glad you're here to vet my advice. So the marker is evaluated at install time even for wheels? Good to know!

nateprewitt:
@Lukasa, sorry I should have had my coffee before replying to this. This is probably what you were referencing in your original question, but we aren't installing anything from setup.py in the current Makefile.

Since we can't declare a conditional dependency in Pipfile, this will either need to be handled in the Makefile or appveyor.yml, rather than setup.py. It seems pretty kludgy to shove the conditional into `init`, but `make` should give you what you need to run the test suite. appveyor.yml would fix the testing on our end, but probably result in tickets complaining about tests not working locally on Windows.

Lukasa:
Hrm.

Well, in the first instance we should consider this a feature request for conditional dependencies in Pipfile. I've used them a bit in Python, they're a fairly useful feature, so Pipfile should probably aim to support it at some point. :wink:

Doing this in the Makefile is gonna suck, so for now we'll do it in appveyor.yml.

Lukasa:
Hrm, it looks a lot like appveyor isn't reporting coverage. I wonder why that is.

sethmlarson:
What happens if we just run `codecov`? I don't know much about `pipenv`.

Lukasa:
We need to run it via `pipenv`, that will automatically use the virtualenv in which codecov is installed.

sethmlarson:
*Realizing I should probably read more on `pipenv`*
:thinking::thought_balloon:

Lukasa:
There we go, that appeared to work. Ok, so let's not merge this yet: I want to confirm with @dstufft that the marker I put in place here actually, you know, works.

schlamar:
I assume the marker will install `win_inet_pton` also on Python 3, which is not quite correct?!

Lukasa:
Ah, good spot, you're quite right.

Lukasa:
I remain extremely nervous about that marker, mostly because our testing does not use `tox` and so does not actually test the install process of the package.

schlamar:
What about a `pipenv run python setup.py develop` in Makefile init (and removing the part installing win_inet_pton in appveyor.yml)?

Lukasa:
So we could do that, and that would catch egregious packaging errors. Might be worth considering.

schlamar:
Ah well, this doesn't pull the socks extra. `install -e .[socks]` should do the trick.  

However, I'm not sure if this would make sense overall...

schlamar:
Maybe you should move this issue with the setup.py marker to a separate PR and merge the old version without changes to `setup.py` so AppVeyor testing is working for now.

Lukasa:
Amusingly, we've been installing the PyPI version of Requests in our virtualenv all this time anyway, because one of our dependencies uses it. I find that to be pretty amusing.

schlamar:
> Amusingly, we've been installing the PyPI version of Requests in our virtualenv all this time anyway, because one of our dependencies uses it. I find that to be pretty amusing.

So the develop install is really a good thing, as you make sure that you are testing against the local version and not against the PyPI version :)

Lukasa:
Sadly, no. That only works if you use a `src` directory, or in some way prevent the `requests` directory being importable. Python will always prefer the local shadow to anything installed via `pip`.

Lukasa:
Ok, this broadly looks right to me. Can I get another contributor to review?

Lukasa:
Good spot, I'll clean that up.

kennethreitz:
\o/

kennethreitz:
I'll go ahead and merge this then.

PR #3991, opened 2017-04-25
Remove some unused imports.

chrisgavin:


requests/packages/chardet/chardetect.py
```diff
@@ -17,7 +17,6 @@
 
 import argparse
 import sys
-from io import open
```
Lukasa: Can you remove the changes from this file please? This is a third-party package we vendor unchanged.

chrisgavin: Thanks, I hadn't spotted that.

PR #3996, opened 2017-05-02
Allowing users to override the default agent using env variable

amir-hadi:
We need to override the default user agent that requests uses. I think its ok to allow overriding the default user agent using environment variables.

I didn't find other options (except monkey patching default_user_agent function in utils), maybe this can help other users as well.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -685,10 +685,12 @@ def select_proxy(url, proxies):
 def default_user_agent(name="python-requests"):
     """
     Return a string representing the default user agent.
+    Allows overriding the default user agent using the REQUESTS_DEFAULT_USER_AGENT
+    environment variable.
 
     :rtype: str
     """
-    return '%s/%s' % (name, __version__)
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this!

I don't really understand why Requests should merge this into the tree. If you want to have this behaviour in your application, you can just use `Session` objects and set `Session.headers['User-Agent']` to whatever you like.

In fact, I'd say that having this functionality in Requests as provided here is not good. Mostly this is because this ignores the `trust_env` variable that Requests normally uses, preventing applications from opting-out of the environment. I am also concerned about unintended side-effects: environment variables are a very sledgehammer approach to this problem, such that any application using Requests with the default user-agent variable will be affected, even if that's not the one the user was concerned about. It's almost always better for applications to work with environment variables, rather than have the libraries go looking around for environment variables to control their behaviour.

So I'm afraid I don't think we'll be merging this in its current form, sorry!

PR #3998, opened 2017-05-04
Enable sessions to bypass creating/parsing cookies.

rowillia:
We use requests for some service-to-service communication, neither side
of which utilizes cookies.  When we profiled our services, we see ~13%
of the time spend sending requests is either creating cookie jars or
extracting cookies to jars.

This PR will enable us to disable cookies for a given `requests.Session`
by setting `session.cookies = None`.  If `sessions.cookies` is None, we
won't create any cookies for the request nor will we parse them on the
response.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -43,8 +43,6 @@ Release History
   check doesn't use forward and reverse DNS requests anymore
 - URLs with schemes that begin with ``http`` but are not ``http`` or ``https``
   no longer have their host parts forced to lowercase.
-- Enable ``Session`` objects to skip creating or parsing cookies all together.
-  This will improve performance for users who don't use cookies.
 
 **Bugfixes**
 
@@ -1369,3 +1367,4 @@ This is not a backwards compatible change.
 
 * Frustration
 * Conception
+
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,12 @@
 Release History
 ---------------
 
+**Unreleased**
++++++++++++++++++++
+
+- Enable ``Session`` objects to skip creating or parsing cookies all together.
+  This will improve performance for users who don't use cookies.
+
 2.14.2 (2017-05-10)
 +++++++++++++++++++
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -43,6 +43,8 @@ Release History
   check doesn't use forward and reverse DNS requests anymore
 - URLs with schemes that begin with ``http`` but are not ``http`` or ``https``
   no longer have their host parts forced to lowercase.
+- Enable ``Session`` objects to skip creating or parsing cookies all together.
+  This will improve performance for users who don't use cookies.
 
 **Bugfixes**
 
@@ -1367,4 +1369,3 @@ This is not a backwards compatible change.
 
 * Frustration
 * Conception
-
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -277,7 +277,8 @@ class HTTPAdapter(BaseAdapter):
             response.url = req.url
 
         # Add new cookies from the server.
-        extract_cookies_to_jar(response.cookies, req, resp)
+        if not req.discard_cookies:
+            extract_cookies_to_jar(response.cookies, req, resp)
 
         # Give the Response some context.
         response.request = req
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -248,7 +248,7 @@ class HTTPAdapter(BaseAdapter):
                 raise IOError("Could not find the TLS key file, "
                               "invalid path: {0}".format(conn.key_file))
 
-    def build_response(self, req, resp):
+    def build_response(self, req, resp, discard_cookies=False):
         """Builds a :class:`Response <requests.Response>` object from a urllib3
         response. This should not be called from user code, and is only exposed
         for use when subclassing the
@@ -277,7 +277,7 @@ class HTTPAdapter(BaseAdapter):
             response.url = req.url
 
         # Add new cookies from the server.
-        if not req.discard_cookies:
+        if not discard_cookies:
             extract_cookies_to_jar(response.cookies, req, resp)
 
         # Give the Response some context.
@@ -384,7 +384,8 @@ class HTTPAdapter(BaseAdapter):
 
         return headers
 
-    def send(self, request, stream=False, timeout=None, verify=True, cert=None, proxies=None):
+    def send(self, request, stream=False, timeout=None, verify=True, cert=None,
+             proxies=None, discard_cookies=False):
         """Sends PreparedRequest object. Returns Response object.
 
         :param request: The :class:`PreparedRequest <PreparedRequest>` being sent.
@@ -516,4 +517,4 @@ class HTTPAdapter(BaseAdapter):
             else:
                 raise
 
-        return self.build_response(request, resp)
+        return self.build_response(request, resp, discard_cookies)
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -123,8 +123,8 @@ def extract_cookies_to_jar(jar, request, response):
     :param request: our own requests.Request object
     :param response: urllib3.HTTPResponse object
     """
-    if not (hasattr(response, '_original_response') and
-            response._original_response):
+    if jar is None or not (hasattr(response, '_original_response') and
+                           response._original_response):
         return
     # the _original_response field is the wrapped httplib.HTTPResponse object,
     req = MockRequest(request)
@@ -524,12 +524,19 @@ def merge_cookies(cookiejar, cookies):
     :param cookiejar: CookieJar object to add the cookies to.
     :param cookies: Dictionary or CookieJar object to be added.
     """
+
+    if cookiejar is None:
+        if cookies is None:
+            return None
+        cookiejar = RequestsCookieJar()
+
     if not isinstance(cookiejar, cookielib.CookieJar):
         raise ValueError('You can only merge into CookieJar')
 
     if isinstance(cookies, dict):
         cookiejar = cookiejar_from_dict(
             cookies, cookiejar=cookiejar, overwrite=False)
+
     elif isinstance(cookies, cookielib.CookieJar):
         try:
             cookiejar.update(cookies)
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -123,8 +123,10 @@ def extract_cookies_to_jar(jar, request, response):
     :param request: our own requests.Request object
     :param response: urllib3.HTTPResponse object
     """
-    if jar is None or not (hasattr(response, '_original_response') and
-                           response._original_response):
+    if jar is None:
+        return
+    if not (hasattr(response, '_original_response') and
+            response._original_response):
         return
     # the _original_response field is the wrapped httplib.HTTPResponse object,
     req = MockRequest(request)
@@ -536,7 +538,6 @@ def merge_cookies(cookiejar, cookies):
     if isinstance(cookies, dict):
         cookiejar = cookiejar_from_dict(
             cookies, cookiejar=cookiejar, overwrite=False)
-
     elif isinstance(cookies, cookielib.CookieJar):
         try:
             cookiejar.update(cookies)
--- a/requests/models.py
+++ b/requests/models.py
@@ -293,15 +293,18 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         self.hooks = default_hooks()
         #: integer denoting starting position of a readable file-like body.
         self._body_position = None
+        #: boolean denoting whether or not to discard cookies from the response.
+        self.discard_cookies = False
 
     def prepare(self, method=None, url=None, headers=None, files=None,
-        data=None, params=None, auth=None, cookies=None, hooks=None, json=None):
+        data=None, params=None, auth=None, cookies=None, hooks=None, json=None,
+        discard_cookies=False):
         """Prepares the entire request with the given parameters."""
 
         self.prepare_method(method)
         self.prepare_url(url, params)
         self.prepare_headers(headers)
-        self.prepare_cookies(cookies)
+        self.prepare_cookies(cookies, discard_cookies)
         self.prepare_body(data, files, json)
         self.prepare_auth(auth, url)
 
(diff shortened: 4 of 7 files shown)


Lukasa:
Thanks for this suggestion!

So, before I go any further, I'd like to get an understanding of your results. You say that 13% of the time is spent "either creating cookie jars or extracting cookies to jars", but also that your services never use cookies. I find that somewhat tricky to believe: in the event that there are no cookies there will be relatively little processing of the absent cookies.

Firstly, the most expensive things we do in general are related to environment evaluation: checking for proxies, reading `.netrc` files, etc. So I assume you're already setting `trust_env` to `False` to avoid that overhead.

Secondly, as an attempt to reproduce your results, I profiled Requests making 1000 requests through a single Session to a Twisted instance serving a directory on my machine, with `trust_env=False`. This sets no cookies. The total runtime of the test was 4 seconds. Here's the profiling output, sorted by cumulative time, for any function that we spend more than 100ms in:

```

        1    0.018    0.018    4.038    4.038 test.py:1(<module>)
     1000    0.007    0.000    3.942    0.004 sessions.py:492(get)
     1000    0.020    0.000    3.935    0.004 sessions.py:411(request)
     1000    0.029    0.000    3.283    0.003 sessions.py:569(send)
     1000    0.019    0.000    3.007    0.003 adapters.py:375(send)
     1000    0.032    0.000    2.536    0.003 connectionpool.py:446(urlopen)
     1000    0.047    0.000    2.099    0.002 connectionpool.py:321(_make_request)
     1000    0.018    0.000    1.709    0.002 httplib.py:1084(getresponse)
     1000    0.022    0.000    1.674    0.002 httplib.py:431(begin)
     8000    0.053    0.000    1.503    0.000 socket.py:410(readline)
     1000    0.022    0.000    1.477    0.001 httplib.py:392(_read_status)
     1000    1.409    0.001    1.409    0.001 {method 'recv' of '_socket.socket' objects}
     1000    0.020    0.000    0.570    0.001 sessions.py:371(prepare_request)
     1000    0.008    0.000    0.341    0.000 models.py:297(prepare)
     1000    0.005    0.000    0.305    0.000 httplib.py:1040(request)
     1000    0.018    0.000    0.300    0.000 httplib.py:1067(_send_request)
     1000    0.015    0.000    0.291    0.000 adapters.py:240(build_response)
     5001    0.020    0.000    0.203    0.000 structures.py:42(__init__)
     1000    0.013    0.000    0.199    0.000 response.py:437(from_httplib)
     7000    0.020    0.000    0.192    0.000 sessions.py:42(merge_setting)
     1000    0.002    0.000    0.191    0.000 httplib.py:1025(endheaders)
     3009    0.010    0.000    0.190    0.000 socket.py:227(meth)
     1000    0.005    0.000    0.189    0.000 httplib.py:867(_send_output)
     1000    0.007    0.000    0.182    0.000 httplib.py:840(send)
     1000    0.169    0.000    0.169    0.000 {method 'sendall' of '_socket.socket' objects}
     1000    0.010    0.000    0.162    0.000 mimetools.py:24(__init__)
    85456    0.066    0.000    0.155    0.000 {isinstance}
     8004    0.041    0.000    0.140    0.000 _abcoll.py:548(update)
     1000    0.007    0.000    0.137    0.000 models.py:784(content)
     1000    0.004    0.000    0.135    0.000 rfc822.py:88(__init__)
     1000    0.045    0.000    0.132    0.000 httplib.py:271(readheaders)
     2000    0.015    0.000    0.129    0.000 cookies.py:121(extract_cookies_to_jar)
     1000    0.005    0.000    0.129    0.000 connectionpool.py:214(_get_conn)
    19027    0.008    0.000    0.122    0.000 {method 'join' of 'str' objects}
     1000    0.016    0.000    0.122    0.000 _collections.py:307(from_httplib)
     1000    0.006    0.000    0.121    0.000 adapters.py:277(get_connection)
     2000    0.004    0.000    0.114    0.000 models.py:715(generate)
      999    0.004    0.000    0.110    0.000 connection.py:7(is_connection_dropped)
     2000    0.005    0.000    0.109    0.000 response.py:411(stream)
     1000    0.022    0.000    0.109    0.000 models.py:350(prepare_url)
      999    0.002    0.000    0.105    0.000 wait.py:29(wait_for_read)
      999    0.010    0.000    0.102    0.000 wait.py:9(_wait_for_io_events)
     1000    0.019    0.000    0.100    0.000 models.py:591(__init__)
     1000    0.019    0.000    0.100    0.000 response.py:342(read)
```

As you can see, cookies *are* in there, but only in one place (`extract_cookies_to_jar`), with a total runtime of 129ms. Now, admittedly, I suspect the content we're sending back and forth is longer in my case than in yours, so we can assume that you'd spend less time in `socket.recv`, but even if we assumed you spent *zero* time there we don't get anywhere near to cookies being 30% of the overhead. In the case of my test, the other cookie-based code that is not part of `extract_cookies_to_jar` adds maybe another 80ms of overhead, leading to 209ms. To be pessimistic, let's double that estimate to 420ms. 420ms spread across 1000 requests is 420μs: not an entirely unreasonable overhead, given that we spend about 1/4 of that time checking that that the connection is still down.

So, certainly if we were going to add this feature I'd want the code to be *vastly* simpler. I'd also want to ask whether or not it's easier in your case simply to patch the `extract_cookies_to_jar` function to be a no-op.

rowillia:
@Lukasa Thanks a ton for your detailed analysis!  I really appreciate all of the work you put into evaluating this 😄 .

Here's a snippit of flame graph (Generated using https://github.com/uber/pyflame) from one of our services running in production, filtering out to only focus on `requests`:
[only_requests.zip](https://github.com/kennethreitz/requests/files/979777/only_requests.zip)

Here's a screenshot, highlighting all stacks that contain the word `cookies`:
screen shot 2017-05-05 at 9 53 53 am

(If you're not comfortable opening random zips, here's the raw output -
 https://gist.github.com/rowillia/e7232fc12e9010c4663811d2166a1228 , run it through https://github.com/brendangregg/FlameGraph to get an SVG)

Thanks for the `trust_env` tip, I noticed how expensive `netrc` and merging environment settings were but haven't had a chance to change things in production.

Any pointers on how we can make the code simpler?  This is my first requests PR so I'm not super familiar with the coding style here beyond what's in the docs.

Lukasa:
Making it simpler would focus on moving the complexity into helper functions that deal with this. For example, `extract_cookies_to_jar` could be enhanced to handle `None` arguments, keeping this complexity out of the main code paths.

rowillia:
Thanks for the suggestion @Lukasa, I've simplified the implementation to encapsulate some of the complexity inside of `cookies.py`.

One other thing I wanted to run by you: setting `cookies` to `None` to avoid parsing/creating cookies feels somewhat magic to me.  Do you think it would be better if there were simply a `discard cookies` flag?  The main motivator for structuring like this is the behavior I'd expect if the `session` has no cookies but the `request` does is we would pass through the requests cookies and having `discard cookies` set to `True` while specifying cookies would be illegal (I tend to be a big fan of the idea that illegal states should be unrepresentable when possible).

rowillia:
@Lukasa one additional thought I just had reading your comment - I'm most worried about optimizing CPU consumption, not wall time.  Totally agree that this will have minimal impact on wall time, but for some services that simply farm out a bunch of requests and wait for their response (using gevent) the CPU consumed here is non-trivial.

For servers using threads, I worry that the RLock on the cookiejar is creating contention so it'd be best if we could avoid it all together.

Lukasa:
> Do you think it would be better if there were simply a discard cookies flag?

No. =) We strongly, strongly resist adding API surface because it adds to the possible confusion of interacting with Requests. This goes double for situations like yours which are extremely uncommon: adding an entire additional API property for a bit of function that fewer than 0.5% of users will ever use is just complete overkill. I'd much rather add a subtle use of an existing property.

> (I tend to be a big fan of the idea that illegal states should be unrepresentable when possible).

What makes this state illegal? Setting `session.cookies = None` requests that no cookies be *persisted*, but there's no reason to prevent you putting cookie headers in the request. In fact, this is a good example as to why we don't want to add such a flag: it is unclear what exactly it applies to, and there are numerous other flags and options that it *could* apply to, so I'd say it impedes clarity more than anything else.

> For servers using threads, I worry that the RLock on the cookiejar is creating contention so it'd be best if we could avoid it all together.

It probably is, but not by much. After all, we have a lock in the connection pool too, though admittedly it's one in the default configuration that you can't contend on (more a semaphore than a lock in that case). Merely *having* the lock, even if you don't contend on it, is very expensive in Python.

I agree that minimising CPU time is important, but if you want to do that I *strongly* encourage you to look away from Requests. Requests' attempts to provide you with a helpful, intuitive API involve a *lot* of Python code that is strictly overhead. If you can ensure that your HTTP data will always be ready to go and require basically zero transformation you may find urllib3 is a better fit for your needs.

HISTORY.rst
```diff
@@ -17,6 +17,8 @@ Release History
   on Windows with Python 2.7.
 - Changed the proxy bypass implementation on Windows: the proxy bypass
   check doesn't use forward and reverse DNS requests anymore
+- Enable ``Session`` objects to skip creating or parsing cookies all together.
+  for users who don't need cookies this will improve performance.
```
Lukasa: Capital letter at the start of a sentence please.

HISTORY.rst
```diff
@@ -1320,4 +1322,3 @@ This is not a backwards compatible change.
 
 * Frustration
 * Conception
-
```
Lukasa: No need to remove this.

requests/models.py
```diff
@@ -293,15 +293,18 @@ def __init__(self):
         self.hooks = default_hooks()
         #: integer denoting starting position of a readable file-like body.
         self._body_position = None
+        #: boolean denoting whether or not to discard cookies from the response.
+        self.discard_cookies = False
```
Lukasa: As discussed, I'd rather not have this flag. It'd be better to be able to pass it directly into the adapter than to force the request to carry it around.

requests/cookies.py
```diff
@@ -123,8 +123,8 @@ def extract_cookies_to_jar(jar, request, response):
     :param request: our own requests.Request object
     :param response: urllib3.HTTPResponse object
     """
-    if not (hasattr(response, '_original_response') and
-            response._original_response):
+    if jar is None or not (hasattr(response, '_original_response') and
```
Lukasa: For clarity's sake it'd be better to have this be two separate `if` statements, rather than one gigantic one that evaluates all possible early-exit conditions.

requests/cookies.py
```diff
@@ -524,12 +524,19 @@ def merge_cookies(cookiejar, cookies):
...
+            return None
+        cookiejar = RequestsCookieJar()
+
     if not isinstance(cookiejar, cookielib.CookieJar):
         raise ValueError('You can only merge into CookieJar')
 
     if isinstance(cookies, dict):
         cookiejar = cookiejar_from_dict(
             cookies, cookiejar=cookiejar, overwrite=False)
+
```
Lukasa: No need for this extra blank line.

tests/test_requests.py
```diff
@@ -351,6 +351,22 @@ def test_request_cookie_overrides_session_cookie(self, httpbin):
...
+        # Session cookie should still be None
+        assert s.cookies is None
+
+    def test_session_with_explicitly_none_cookies(self, httpbin):
+        s = requests.session()
+        s.cookies = None
+        r = s.get(httpbin('cookies'))
+        assert r.json()['cookies'] == {}
+        # Session cookie should still be None
+        assert s.cookies is None
```
Lukasa, line 368: It'd be good to add a test where httpbin sets some cookies to prove that they also don't show up.

HISTORY.rst
rowillia: Done

requests/cookies.py
rowillia: Done

HISTORY.rst
rowillia: Done

requests/cookies.py
rowillia: Done

Lukasa:
I don't think you pushed your changes. :smile:

rowillia:
@Lukasa yeah jumped the gun on the comments 😄

Lukasa:
Can you rebase this onto the current master for me? It's having some trouble with testing because it's not mergeable.

requests/adapters.py
```diff
@@ -383,7 +384,8 @@ def proxy_headers(self, proxy):
 
         return headers
 
-    def send(self, request, stream=False, timeout=None, verify=True, cert=None, proxies=None):
+    def send(self, request, stream=False, timeout=None, verify=True, cert=None,
```
sigmavirus24, line 387: I'm really not quite fond of changing this signature. Several people rely on this signature to write their own Adapters and if this is now required, it adds extra work for adapter authors.

requests/adapters.py
```diff
@@ -277,7 +277,8 @@ def build_response(self, req, resp):
             response.url = req.url
 
         # Add new cookies from the server.
-        extract_cookies_to_jar(response.cookies, req, resp)
+        if not discard_cookies:
```
sigmavirus24, line 280: Similarly I'm not fond of having this conditional logic in here.

requests/cookies.py
```diff
@@ -524,6 +526,12 @@ def merge_cookies(cookiejar, cookies):
     :param cookiejar: CookieJar object to add the cookies to.
     :param cookies: Dictionary or CookieJar object to be added.
     """
+
+    if cookiejar is None:
+        if cookies is None:
```
sigmavirus24, line 531: I don't think this needs to be inside this other condition.

```py
if cookies is None:
    return None

if cookiejar is None:
    cookiejar = RequstsCookieJar()
```

requests/models.py
```diff
@@ -552,13 +552,16 @@ def prepare_cookies(self, cookies):
         """Prepares the given HTTP cookie data.
 
         This function eventually generates a ``Cookie`` header from the
-        given cookies using cookielib. Due to cookielib's design, the header
+        given cookies using cookielib. If `cookies` is None, no ``Cookie``
```
sigmavirus24, line 555: Your backtick surrounded `cookies` here should have two backticks.

kennethreitz:
I really like this suggestion, as well as the API proposed. Very Requests.

kennethreitz:
add to whitelist

kennethreitz:
add to whitelist

kennethreitz:
add to whitelist

nateprewitt:
Given that it's been almost a year since we've seen progress on this PR, I'm going to close this out. Please feel free to reopen if you'd like to continue. Thanks!

PR #4001, opened 2017-05-07
Specify that the timeout parameter is in seconds.

bowlofeggs:
Signed-off-by: Randy Barlow <randy@electronsweatshop.com>


kennethreitz:
thanks!

PR #4004, opened 2017-05-09
Prepare for 2.14 release.

Lukasa:
Changelog updated, vendored libraries all updated. Let's see how this goes. Looks like it'll be a good release!


Lukasa:
Ok, so this patch has changed the way we normalize URLs of non-HTTP schemes. Specifically, we no longer lowercase `http+x` URLs. This is because of shazow/urllib3#1089. My reading of [this discussion](https://github.com/kennethreitz/requests/pull/3738#issuecomment-268805728) is that we did this more-or-less on purpose, and that this is an intended effect: we now preserve the case of these schemes.

I think this deserves to be called out specifically in the changelog, and that the tests should be updated to make it clear that this is intentional behaviour. I'll do that now, but I'd like the other maintainers to weigh in.

nateprewitt:
@Lukasa, that's correct. Host names were changed to be invariably lowercased in urllib3 1.17 (Requests 2.12) which broke unix sockets for some. The patch you've linked to scoped the normalization a bit more tightly to only schemes we are certain must be case-insensitive.

sethmlarson:
Any particular reason for specifying packages manually over using `setuptools.find_packages()`? The list is getting quite long.

Lukasa:
@SethMichaelLarson Uh basically historical reasons?

sethmlarson:
It's fine to leave it, I just know I've forgotten to add a package once on a release and had to revert and re-release.

Lukasa:
Yeah, it's an ongoing problem here. PR for the next release? ;)

sethmlarson:
Sounds good to me :)

HISTORY.rst
```diff
@@ -17,6 +25,29 @@ Release History
...
+
+- Much improved handling of non-ASCII ``Location`` header values in redirects.
+  Fewer ``UnicodeDecodeError``s are encountered on Python 2, and Python 3 now
+  correctly understands that Latin-1 is unlikely to be the correct encoding.
+- If an attempt to ``seek`` file to find out its length fails, we now
+  appropriately handle that by aborting our content-length calculations.
+- Restricted ``HTTPDigestAuth`` to only respond to auth challenges made on 4XX
+  responses, rather than to all auth challenges.
+- Fixed some code that was firing ``DeprecationWarning`` on Python 3.6.
+- The dismayed person emoticon (``/o\\``) no longer has a big head. I'm sure
```
sethmlarson, line 41: Can this be promoted to **Improvements**? ;)

Lukasa: No. =P It's a bugfix.

Lukasa:
Ok, I'm calling this good. Let's see how many people's computers we can break.

Lukasa:
And done. Welcome to our exciting 2.14 world!

sethmlarson:
Huzzah! :tada:

PR #4014, opened 2017-05-10
Remove range operators from markers.

### LINKED ISSUE #4013: Amazon Linux easy_install incompatible with 2.14.1
I've been following #4006 #4012 and I'm aware the guidance is to update to a newer pip, and that easy_install is not supported. I just wanted to make the case that not all of us have a choice on python package manager.

Take my case - I'm using Amazon Linux bootstrapped by Amazon CloudFormation. We give the CloudFormation initializer code (cfn-init for short) a list of python packages to install, and it installs them in any order it wants, using any mechanism it wants. Studying the logs, it looks like under the hood it's using easy_install.

I'm not even trying to install requests. I'm getting a floating version of it transitively through trying to install docker-py. So I see:

`[ERROR] easy_install failed. Output: Searching for docker-py==1.7.2`
...
`Searching for requests>=2.5.2`
...
`Best match: requests 2.14.1`
...
```
Running requests-2.14.1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-wfkGhb/requests-2.14.1/egg-dist-tmp-H3aUMy
error: Setup script exited with error in requests setup command: Invalid environment marker: sys_platform == "win32" and python_version<"3.3"
```

Version information:
Amazon Linux AMI release 2016.09
Python 2.7.12
```
easy_install --version
setuptools 12.2
```

So at the moment I have a cumbersome workaround where I run cfn-init in two phases, first where I expressly install requests 2.13.0, then next where I install docker-py and it's happy with the requests already installed. But I just wanted to raise awareness that there is a category of people who are seeing this error through transitive use of requests (I didn't even know we were using it) and for which upgrading or changing package managers is non-trivial.

Lukasa:
Resolves #4013.

It turns out that older setuptools were pretty stupid and couldn't handle range operators, but they *can* handle parentheses and the word "or". So let's do that instead.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,7 @@ setup(
     extras_require={
         'security': ['pyOpenSSL>=0.14', 'cryptography>=1.3.4', 'idna>=2.0.0'],
         'socks': ['PySocks>=1.5.6, !=1.5.7'],
-        'socks:sys_platform == "win32" and python_version<"3.3"': ['win_inet_pton'],
+        'socks:sys_platform == "win32" and (python_version == "2.7" or python_version == "2.6")': ['win_inet_pton'],
     },
 )
 


Lukasa:
Cool, ok. Let's ship a 2.14.2 with this and see if this cleans up the last few lingering pain points.

PR #4033, opened 2017-05-17
Test that the readme renders in CI.

### LINKED ISSUE #4030: fix pypi page
markup is bad 

https://pypi.python.org/pypi/requests/

Lukasa:
Resolves #4030.


.travis.yml
```diff
@@ -14,6 +14,7 @@ python:
 install: "make"
 # command to run tests
 script:
+  - make test-readme
```
sigmavirus24: This shouldn't be run on Python 2.6. It seems `readme-renderer` doesn't support 2.6.

Lukasa: Agh, that's a pain. Ok.

sigmavirus24:
Merge this when everything else is ✔️

PR #4036, opened 2017-05-17
Pin sphinx

nateprewitt:
This will fix the build issues in Python 2.6 due to Sphinx's new requirement of `typing` in versions 1.6+. This pinning (capping) should only be in place until a solution for representing markers is introduced in Pipfile.

While we're here doing cleanup, I've also removed the no longer needed `pipenv lock` before each build. Some preliminary tests show it saving 15-20% on the total build time for CI. We can scrap that though if it's not inline with the goal of the PR, getting the build working.


Makefile
```diff
@@ -2,7 +2,6 @@
 
 init:
 	pip install pipenv
-	pipenv lock
```
Lukasa, line 5: Is this necessary? This should only be run on `init`.

nateprewitt: Sorry, are you asking if the removal is necessary, or the locking?

We can leave this in, but it's an extra 4-5 minutes of CI time that doesn't actually need to run.  Pipenv is no longer worrying about hashes on a package specific level, so platform conflicts aren't a problem now. If there is an actual change to the Pipfile not represented, Pipenv will rebuild things without this command, otherwise it uses the current Pipfile.lock.

Like I said, we can not include this commit here, but this line should be removed at some point.

Edit: `init` runs every time we start a build with `make` which means we perform the same expensive action on every build.

Lukasa: Ok, I don't mind removing it. =)

PR #4042, opened 2017-05-21
Persist Session-level CookiePolicy

nateprewitt:
This is a followup for #3463, but refactored without the persistence flag which makes this a breaking change. Currently, Requests silently throws away any cookie policies set on a CookieJar as noted in #3416. This patch will preserve the CookiePolicy set on a session-level CookieJar for both redirected and subsequent requests.

The only caveat here is this won't work for a non-RequestsCookieJar since the policy isn't exposed publicly. We may want to add some documentation about this caveat but given the rarity of this case, the information in the issue tracker may be sufficient.

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -412,7 +412,7 @@ class RequestsCookieJar(cookielib.CookieJar, collections.MutableMapping):
 
     def copy(self):
         """Return a copy of this RequestsCookieJar."""
-        new_cj = RequestsCookieJar()
+        new_cj = RequestsCookieJar(self._policy)
         new_cj.update(self)
         return new_cj
 
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -16,7 +16,8 @@ from datetime import timedelta
 from .auth import _basic_auth_str
 from .compat import cookielib, OrderedDict, urljoin, urlparse, is_py3, str
 from .cookies import (
-    cookiejar_from_dict, extract_cookies_to_jar, RequestsCookieJar, merge_cookies)
+    cookiejar_from_dict, extract_cookies_to_jar, RequestsCookieJar,
+    merge_cookies, _copy_cookie_jar)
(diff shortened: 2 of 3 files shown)


nateprewitt:
@Lukasa we're going to need to merge in master to get the builds running in 3.0.0 again. I can do it but probably won't get to it till tomorrow. I'll need to strip out get_policy too because it's no longer needed either.

nateprewitt:
Alright, I've got the 3.0.0 branch hiccups worked out and I think this is ready for a peek. It's a trivial patch at this point with the flag gone and works across anything that looks like a cookielib.CookieJar.

sigmavirus24:
I'm confused. Does this mean that we expect people to set the cookie policy on the cookie jar directly or specify a new RequestsCookieJar with a policy that persists that way? I expected that this added an attribute to the Session that was used to create cookie-jars. Also, since we're adding this to 3.0.0, I'd like us to use a stricter cookie policy. The default one is very relaxed.

nateprewitt:
@sigmavirus24, having the user either set it with `set_policy` or creating a new `RequestsCookieJar(policy)` will work.

There are two related problems this is trying to solve. The first is the CookieJar is currently copied on a redirect, but our `copy` method throws away the policy so we get a new `RequestsCookieJar` with the default policy. The second is we merge our session cookies into an empty `RequestsCookieJar` before merging in request level cookies, and that also silently discards a user set policy.

This patch is going with the assumption that if you assign a CookieJar to the Session, we continue using everything that was set on that instance of the CookieJar. We could add an attr to the Session, but I *think* that's unneeded surface area at this point. If you don't like the policy on the CookieJar you're handing us, then simply change it back to the default with `set_policy`.

I agree with upping the default policy for the RequestsCookieJar but don't know if that's directly related to #3416. I'm happy to spin off another PR for that, unless you feel it needs to be done here.

Issue #4039, opened 2017-05-19
GAE - not retrieving cookies from the response


cmin764:
I've found an issue when using requests with toolbelt through Google App Engine, because of the urllib3 patching with urlfetch. The problem is that the cookie jar is not populated due to a missing attribute (original_response) into the urllib3 response (when switching from sockets to urlfetch and using the HTTPAdapter through the toolbelt monkeypatch).

The bug is self explained within this patch that I've made and works flawlessly into the GAE production and development environments:

```python
"""Runtime monkey-patching for various problematic libraries."""


import re

import requests
from requests.cookies import MockRequest


class MockResponse(object):
    """Wraps a `httplib.HTTPMessage` to mimic a `urllib.addinfourl`.

    ...what? Basically, expose the parsed HTTP headers from the server response
    the way `cookielib` expects to see them.
    """

    SEPARATOR = re.compile(r",\s*(?=[\w-]+=)")

    def __init__(self, response):
        """Make a MockResponse for `cookielib` to read.

        :param response: a HTTPResponse or analogous carrying the headers
        """
        self._response = response

    def info(self):
        return self

    def getheaders(self, name):
        headers = self._response.getheader(name)
        return self.SEPARATOR.split(headers) if headers else []


def extract_cookies_to_jar(jar, request, response):
    """Extract the cookies from the response into a CookieJar.

    :param jar: cookielib.CookieJar (not necessarily a RequestsCookieJar)
    :param request: our own requests.Request object
    :param response: urllib3.HTTPResponse object
    """
    # the _original_response field is the wrapped httplib.HTTPResponse object,
    req = MockRequest(request)
    # pull out the HTTPMessage with the headers and put it in the mock:
    res = MockResponse(response)
    jar.extract_cookies(res, req)


def _monkeypatch_requests():
    """Monkey patch that problematic cookie extraction function."""
    modules = [
        # Patch the entire namespace within Python.
        requests.adapters,
        requests.auth,
        requests.cookies,
        requests.sessions,
    ]
    for module in modules:
        setattr(module, "extract_cookies_to_jar", extract_cookies_to_jar)


def monkeypatch():
    _monkeypatch_requests()
```

https://pastebin.com/FgDZfT8y

You should integrate this somehow because is affecting everyone (that's the reason they switch to urlfetch, standard urllib or other primitives). Thanks!

Lukasa:
[...]

PR #4044, opened 2017-05-21
Fix cookies storage on GAE

cmin764:
When using requests on GAE, the urllib3 response containing cookies
is a little bit different, therefore we have to extract the cookies
from the returned headers directly from the response object.

Fix issue #4039


Lukasa:
Just to clarify: are you interested in making further amendments to this patch, or would you like core devs or volunteers to build on it instead?

cmin764:
No, I would like to let the core devs take it and adapt it to what is right. What I want most, is to be aware of the fact that this is a serious issue affecting all GAE projects using requests with cookies.

Lukasa:
No-one is disputing that this is an issue. =)

kennethreitz:
Is this the case on the latest offering (I fail to recall the name) on GAE, or only the legacy offering?

Lukasa:
Latest. =)

kennethreitz:
If I recall correctly, GAE now has a "container" style offering that offers much more compatibility with standard Python code.

kennethreitz:
Ah, thanks! Just checking. Good to know. 

Disappointing, but good to know :-/

Lukasa:
Hey @jonparrott, I went to update this and bumped into this error:

```
_____________________ test_requests_are_updated_each_time ______________________
[gw3] darwin -- Python 3.6.0 /Users/cory/Documents/Python/requests_org/requests/.tox/py36/bin/python
httpbin = <function prepare_url.<locals>.inner at 0x111194d08>

    def test_requests_are_updated_each_time(httpbin):
        session = RedirectSession([303, 307])
        prep = requests.Request('POST', httpbin('post')).prepare()
        r0 = session.send(prep)
        assert r0.request.method == 'POST'
        assert session.calls[-1] == SendCall((r0.request,), {})
        redirect_generator = session.resolve_redirects(r0, prep)
        default_keyword_args = {
            'stream': False,
            'verify': True,
            'cert': None,
            'timeout': None,
            'allow_redirects': False,
            'proxies': {},
        }
>       for response in redirect_generator:

tests/test_requests.py:2145: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
requests/sessions.py:184: in resolve_redirects
    extract_cookies_to_jar(prepared_request._cookies, req, resp.raw)
requests/cookies.py:141: in extract_cookies_to_jar
    jar.extract_cookies(res, req)
/Users/cory/.pyenv/versions/3.6.0/lib/python3.6/http/cookiejar.py:1664: in extract_cookies
    for cookie in self.make_cookies(response, request):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <RequestsCookieJar[]>
response = <requests.cookies.MockResponse object at 0x110dc7a90>
request = <requests.cookies.MockRequest object at 0x110dc7c88>

    def make_cookies(self, response, request):
        """Return sequence of Cookie objects extracted from response object."""
        # get cookie-attributes for RFC 2965 and Netscape protocols
        headers = response.info()
>       rfc2965_hdrs = headers.get_all("Set-Cookie2", [])
E       AttributeError: 'MockResponse' object has no attribute 'get_all'
```

Now, for the mainline urllib3 case I can just proxy through to the base httplib class, but is there an equivalent method for GAE that I can use?

kennethreitz:
add to whitelist

kennethreitz:
add to whitelist

kennethreitz:
add to whitelist

theacodes:
@kennethreitz requests works fine on "flexible" (the container offering) because none of this silly urlfetch crap exists there. This only affects the "standard" environment.

@Lukasa help me understand what's going on here- why does cookie jar need to poke at the `_original_response` object (which by all accounts is *private*)? Why couldn't it just use the public methods?

Lukasa:
@jonparrott Because cookielib is fundamentally written to work with *httplib*, which means it expects an httplib response object.

theacodes:
@Lukasa well that's fun! I guess we could update `AppEngineManager` to also set the `_original_response` object for compatibility.

Lukasa:
I think that might be the better bugfix, tbh.

theacodes:
@Lukasa wanna file a bug on urllib3 and assign it to me? pypa is getting most of my OSS time right now, but I can probably do it within a few weeks.

Lukasa:
Alright, closing in favour of shazow/urllib3#1193.

trickidicki:
Unfortunately this is still an issue...

nateprewitt:
@trickidicki, as I believe you've seen, there is an active PR (shazow/urllib3#1283) to address this and once that's merged it will likely be available in Requests next release. Thanks for taking the time to check in.

euri10:
just running into this today, as best as I understand seems like https://github.com/urllib3/urllib3/pull/1283 got merged, would be amazing that requests incorporates it @nateprewitt

nateprewitt:
Hey @euri10, we’re currently waiting for a release from urllib3 and then I’m hoping to have a Requests release soon™ after.

PR #4047, opened 2017-05-22
More Verbose Execption Handling

bobstrecansky:
So I tried to add more verbose exception handling without breaking current functionality (this feels really hacky).  It seems to me that it may be pertinent to break these out and not raise a connection error for each request (as denoted in : https://github.com/kennethreitz/requests/issues/2876)

In the current implementation, a ConnectionError is raised for each call, as well as a max retries message.  It seems like the max retries message error should either be removed or made as more of a general string response statement.  This general string response is here:
https://github.com/kennethreitz/requests/blob/master/requests/adapters.py#L503

Doing so would require removing that line and adding explicit responses within that max retry block, but this might break current functionality if people are depending on the response from requests to make other decisions.

Looking forward to a discussion on this topic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .coverage
+.cache/
 MANIFEST
 coverage.xml
 nosetests.xml
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .coverage
-.cache/
 MANIFEST
 coverage.xml
 nosetests.xml
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -493,13 +493,12 @@ class HTTPAdapter(BaseAdapter):
                 # TODO: Remove this in 3.0.0: see #2811
                 if not isinstance(e.reason, NewConnectionError):
                     raise ConnectTimeout(e, request=request)
-
-            if isinstance(e.reason, ResponseError):
+            elif isinstance(e.reason, ResponseError):
                 raise RetryError(e, request=request)
-
-            if isinstance(e.reason, _ProxyError):
+            elif isinstance(e.reason, _ProxyError):
                 raise ProxyError(e, request=request)
-
+            elif isinstance(e.reason,RetryError):
+                raise RetryError(e, request=request)
             raise ConnectionError(e, request=request)
 
         except ClosedPoolError as e:
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -493,11 +493,14 @@ class HTTPAdapter(BaseAdapter):
                 # TODO: Remove this in 3.0.0: see #2811
                 if not isinstance(e.reason, NewConnectionError):
                     raise ConnectTimeout(e, request=request)
-            elif isinstance(e.reason, ResponseError):
+
+            if isinstance(e.reason, ResponseError):
                 raise RetryError(e, request=request)
-            elif isinstance(e.reason, _ProxyError):
+
+            if isinstance(e.reason, _ProxyError):
                 raise ProxyError(e, request=request)
-            elif isinstance(e.reason,RetryError):
+
+            if isinstance(e.reason,RetryError):
                 raise RetryError(e, request=request)
             raise ConnectionError(e, request=request)
 
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -518,8 +518,7 @@ class TestRequests:
         ))
     def test_errors(self, url, exception):
         with pytest.raises(exception):
-            r = requests.get(url, timeout=1)
-            assert r.exception == exception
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 .coverage
+.cache/
```
sigmavirus24: This is unrelated. Please remove it.

requests/adapters.py
```diff
@@ -493,13 +493,12 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
                 # TODO: Remove this in 3.0.0: see #2811
                 if not isinstance(e.reason, NewConnectionError):
                     raise ConnectTimeout(e, request=request)
-
-            if isinstance(e.reason, ResponseError):
+            elif isinstance(e.reason, ResponseError):
```
sigmavirus24: Why are these being changed to `elif` s? In the event that `e.reason` is a `ConnectTimeoutError` but not a `NewConnectionError` we'll now be unable to re-raise in the event that it's one of these other exeption types.

tests/test_requests.py
```diff
@@ -518,7 +518,8 @@ def test_basicauth_encodes_byte_strings(self):
         ))
     def test_errors(self, url, exception):
         with pytest.raises(exception):
-            requests.get(url, timeout=1)
+            r = requests.get(url, timeout=1)
+            assert r.exception == exception
```
sigmavirus24: We do not attach exceptions to Response objects. Why are you testing this here?

.gitignore
bobstrecansky: Was also including it to help with the unit tests storage; will open a separate PR for it.

requests/adapters.py
bobstrecansky: @sigmavirus24 - was just trying to come up with a way to not always also return the connection max retries.  Do you have another idea on how we could handle that?

tests/test_requests.py
bobstrecansky: Sorry if i'm not following normal convention.  Is there a method you'd like me to use to test this?

requests/adapters.py
Lukasa: How does changing `if` to `elif` achieve that?

bobstrecansky: I was (probably incorrectly) trying to catch only one exception for each of these, rather than multiple.

sigmavirus24: Once we raise a new exception, @bobstrecansky, we won't raise a new exception. If we do not actually raise a new exception then we need these fall-through conditions to maintains backwards compatibility.

tests/test_requests.py
sigmavirus24: Well I was more concerned that you were beginning to attach exceptions to Response objects. You don't need this assertion. `with pytest.raises(exception)` is the assertion in this test.

requests/adapters.py
bobstrecansky: There have been a couple issues opened where users of requests have seen that this clustering of multiple error messages (joining the initial error message alongside the max requests retried) can be confusing to the end user, because it isn't abundantly clear with error message is causing the problem.  Can we come up with a proposed solution for that?  That was my initial intent while writing this PR.

tests/test_requests.py
bobstrecansky: Sorry for the confusion, i'll remove this case.

requests/adapters.py
sigmavirus24: @bobstrecansky while I understand that was your intent and that this was an attempt to fix that, your attempt has done little, if anything, to actually fix that. I think you're missing a critical part of the problem here:

Requests lives at the top-level of several layers of libraries. Immediately below Requests is urllib3. *Most* of the exceptions we handle and then *re-wrap* are urllib3 exceptions. Some of those exceptions are simple exceptions raised by urllib3. Others are exceptions raised by *other* libraries that urllib3 relies on and are re-wrapped. For example, sometimes urllib3 may raise an `SSLError` that is wrapping an exception from the standard library's `ssl` module. Othertimes the `SSLError` may be wrapping an exception from `OpenSSL`. It depends on the user's installation. It's also plausible for urllib3 to be wrapping an exception from `httplib` or `http.client` (depending on Python 2 or Python 3). Beyond that each of those libraries may be wrapping some other library's exception.

The confusing part here stems from two things:

1. Errors wrapped in errors wrapped in errors wrapped in ....
2. Overly generic exceptions being raised in Requests or urllib3 that while technically accurate do not immediately portray the whole story.

bobstrecansky:
Closing this in favor of parsing the execution in the client code.

PR #4048, opened 2017-05-23
Add an enhancement for transforamtion between dict and cookiejar.

justdoit0823:
When using dict_from_cookiejar, I can only get name and value from cookie object, otherwise other cookie attributes are lost. So I want to make an enhancement about two methods dict_from_cookiejar and cookiejar_from_dict, which can be more flexible and safe to do cookie transformation and persistence like the following code.

```
import requests
from requests.utils import dict_from_cookiejar

res = requests.get('https://github.com/justdoit0823/notes')

dict_from_cookiejar(res.cookies)

dict_from_cookiejar(res.cookies, with_cookie_attr=True)
```

And the output:

```
#  with_cookie_attr is False
{'_gh_sess': 'somevalue', 'logged_in': 'no'}

#  with_cookie_attr is True
{'_gh_sess': {'discard': True,
  'domain': 'github.com',
  'name': '_gh_sess',
  'path': '/',
  'rest': {'HttpOnly': None},
  'rfc2109': False,
  'secure': True,
  'value': 'somevalue',
  'version': 0},
 'logged_in': {'discard': False,
  'domain': '.github.com',
  'expires': 2126671044,
  'name': 'logged_in',
  'path': '/',
  'rest': {'HttpOnly': None},
  'rfc2109': False,
  'secure': True,
  'value': 'no',
  'version': 0}}
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -181,3 +181,4 @@ Patches and Suggestions
 - Jonathan Vanasco (`@jvanasco <https://github.com/jvanasco>`_)
 - David Fontenot (`@davidfontenot <https://github.com/davidfontenot>`_)
 - Shmuel Amar (`@shmuelamar <https://github.com/shmuelamar>`_)
+- SenbinYu (`@justdoit0823 <https://github.com/justdoit0823>`_)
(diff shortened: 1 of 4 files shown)


Lukasa:
Thanks for this patch! I'm afraid I don't think it's necessary: these are primarily intended as internal utilities and so we're very disinclined to add features to them. Sorry about that! :sparkles:

PR #4050, opened 2017-05-23
AttributeError error raised when `files` file-pointer (fp) resolves to None

mayani:
The library raises NoneType error when file-pointer (fp) resolves to None.

```
>>> from requests import post
>>> r = post("https://example.com", files={"file-name": None})
AttributeError: 'NoneType' object has no attribute 'read'
```

However, when a param value or json field is None they are not included in the request body.

```
>>> from requests import get
>>> r = get("https://example.com", params={"file-name": None})
>>> r.request.url
```

This commit makes the beahviour consistent for files.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -155,8 +155,12 @@ class RequestEncodingMixin(object):
 
             if isinstance(fp, (str, bytes, bytearray)):
                 fdata = fp
-            else:
+            elif hasattr(fp, 'read'):
                 fdata = fp.read()
+            elif fp is None:
+                continue
+            else:
+                fdata = fp
 
             rf = RequestField(name=k, data=fdata, filename=fn, headers=fh)
             rf.make_multipart(content_type=ft)
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -187,3 +187,4 @@ Patches and Suggestions
 - Nehal J Wani (`@nehaljwani <https://github.com/nehaljwani>`_)
 - Demetrios Bairaktaris (`@DemetriosBairaktaris <https://github.com/demetriosbairaktaris>`_)
 - Darren Dormer (`@ddormer <https://github.com/ddormer>`_)
+- Rajiv Mayani (`@mayani <https://github.com/mayani>`_)
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -23,6 +23,7 @@ dev
 - DNS resolution in proxy bypass no longer includes the username and password in
   the request. This also fixes the issue of DNS queries failing on macOS.
 - Properly normalize adapter prefixes for url comparison.
+- Passing ``None`` as a file pointer to the ``files`` param no longer raises an exception.
 
 2.18.4 (2017-08-15)
 +++++++++++++++++++
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -676,6 +676,14 @@ class TestRequests:
         with pytest.raises(ValueError):
             requests.post(url, files=['bad file data'])
 
+    def test_invalid_files_input(self, httpbin):
+
+        url = httpbin('post')
+        post = requests.post(url,
+                             files={"random-file-1": None, "random-file-2": 1})
+        assert b'name="random-file-1"' not in post.request.body
+        assert b'name="random-file-2"' in post.request.body
+
     def test_POSTBIN_SEEKED_OBJECT_WITH_NO_ITER(self, httpbin):
 
         class TestStream(object):


requests/models.py
```diff
@@ -154,8 +154,10 @@ def _encode_files(files, data):
 
             if isinstance(fp, (str, bytes, bytearray)):
                 fdata = fp
-            else:
+            elif hasattr(data, 'read'):
```
Lukasa: I'd rather we explicitly look for `None`. Do you mind making this change?

kennethreitz:
Love this!

requests/models.py
```diff
@@ -154,8 +154,12 @@ def _encode_files(files, data):
 
             if isinstance(fp, (str, bytes, bytearray)):
                 fdata = fp
-            else:
+            elif hasattr(fp, 'read'):
```
ueg1990, line 158: Sorry that I randomly come out of nowhere once in a blue moon to give review comments so ignore my comment if you like :) But hasattr has a small edge case which can cause an unexpected error in Python 2. You can refer to the link here:
https://hynek.me/articles/hasattr/

The article does mention that if you know what you are doing then its all good. My recommendation is to use getattr with default value of None to avoid AttributeError

kennethreitz: @ueg1990 that's because properties in python2 aren't attributes. very known and undertstood functionality.

nateprewitt:
Hi @mayani, sorry this has been in limbo for so long. Would you mind rebasing this into a single commit on the current master branch? Otherwise, I'll look into cherry-picking this commit onto the master branch. Thanks!

sethmlarson:
You should probably rebase, there's a lot of unnecessary changes after syncing. :)

mayani:
@nateprewitt I am unsure how to rebase after a sync. Might be best to cherry pick the commits or I can submit a fresh pull request.

nateprewitt:
@mayani, you can use squash to condense them into a single commit and then either use the `--onto` option of rebase or `cherry-pick` with the squashed hash will work too.

You’ll need to reset your squashed commit before merging in current master if you go the cherrypick route.

requests/models.py
```diff
@@ -155,8 +155,12 @@ def _encode_files(files, data):
...
                 fdata = fp.read()
+            elif fp is None:
+                continue
+            else:
```
nateprewitt, line 162: In this case, are we expecting the objects `__str__` representation to be sufficient? We currently blow up on things like `datetime` or `dict`, but this may have unexpected results.

If we're OK with the above, would it make more sense to reorg the conditionals to check for `read`, `None` and then everything else would be `fdata = fp`?

e.g.
```python
if hasattr(fp, 'read'):
    fdata = fp.read()
elif fp is None:
    continue
else:
    fdata = fp
```

sigmavirus24: So the code that will eventually handle this is [here](https://github.com/urllib3/urllib3/blob/c8c98c473ab30f427fe05057f2ceded10ec93153/urllib3/filepost.py#L78..L90). 

Honestly, at this point, we *could* simplify the conditional but I'd rather leave it and let it blow up for folks. In 3.0, I'd love to have this else branch raise an exception, though.

nateprewitt: Alright, I'm on board with that. @mayani would you mind removing this `else` conditional to keep our existing functionality. We'll look at adding a more explicit exception in 3.0. Thanks!

mayani: @nateprewitt 

```python
else:
    fdata = fp

rf = RequestField(name=k, data=fdata, filename=fn, headers=fh)
```

Getting rid of else part will result in fdata being an unbound variable. Should we set it to fdata=None in else and let it blow up or leave things as is?

mayani:
Hi @nateprewitt, I did the following
```
git fetch upstream
git branch none_filepointer
```
I cherry picked the commits from mayani:master to non_filepointer. However, I won't be able to push as I don't have permissions on main repo's none_filepointer branch.

nateprewitt:
@mayani, apologies if the instructions weren’t very clear. I’ve already gathered your commits and squashed them into a single commit at the top of none_filepointer. You should be able to replace this branch with a copy of master and then cherry pick that commit on top. Then use a force push to replace this branch.

requests/models.py
nateprewitt: Urgh, yep. Ok so @sigmavirus24 we may just want to put the exception in now. Non-file objects were already exploding on the read call so they never worked.

We *could* make sure the exception inherits from AttributeError if we want to ensure we’re not breaking anyone currently relying on this defect for flow control.

mayani:
@nateprewitt Since the else part is now not being changed, where does it leave us with the none_filepointer branch?

"You should be able to replace this branch with a copy of master and then cherry pick that commit on top." Did you mean rename the branches to make local_nonepointer my new master?

Sorry, my git skills are pretty basic.

requests/models.py
sigmavirus24: @nateprewitt I'd rather never inherit from `AttributeError` if possible. That is just a world of pain. We've done this in the past, but that means if people are putting too much code in their `try/except` block, we may cause them to enter an exception handling block.

In short, I think the code as is may be safe right now. If we want to be extra magical, we could do `fdata = str(fp)` but I really don't want to add that behaviour at all. Let's just leave this as is, I guess and plan to add the exception in 3.0

nateprewitt: Sounds good! I think I misread your first response, so we should be set then.

nateprewitt:
@mayani, it looks like you created the pull request against your master branch. The easiest option is to replace it and then force-push to your repo. The following commands should do it.

```
git branch -f master none_filepointer
git push -f origin master
```

Otherwise if we’re not making changes, I can just open up a PR with the none_filepointer branch and that will merge your commit in.

mayani:
@nateprewitt Thanks. I have executed the commands as suggested

```bash
git branch -f master none_filepointer
git push -f origin master
```

nateprewitt:
Thanks @mayani!

PR #4056, opened 2017-05-24
Disable pytest messing with warnings.

Lukasa:
We may be able to remove this if pytest resolves pytest-dev/pytest#2430, but until that point we need to turn the old behaviour back on.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,2 @@
-include README.rst LICENSE NOTICE HISTORY.rst requests/cacert.pem
+include README.rst LICENSE NOTICE HISTORY.rst pytest.ini requests/cacert.pem
 recursive-include tests *.py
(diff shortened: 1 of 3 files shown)


Lukasa:
I want to point out that it's pretty stupid that we have to do this, but there we are.

nicoddemus:
> I want to point out that it's pretty stupid that we have to do this, but there we are.

Hi @Lukasa, sorry to hear that this affected you guys so. We are having some discussion related to this on pytest-dev/pytest#2430, it mean a lot if you could join the discussion and share your thoughts. 👍

Lukasa:
@niccoddemus Yeah, I saw that thread. I am not sure how much value I can provide by weighing in: your warnings filter as added prevented us from tweaking the filter ourselves for testing purposes. I can drop in though.

PR #4064, opened 2017-05-25
Avoid breaking people using requests.__version__

sigmavirus24:
Using __version__ as a module breaks users who expect
requests.__version__ to be the version string.


kennethreitz:
Thanks!

kennethreitz:
i think `requests.__version__` would have worked fine before, but it just wasn't importable.

kennethreitz:
this required some additional work, but we should be all set now

kennethreitz:
this was actually a problem with `import *`, not the name of the file

Issue #4065, opened 2017-05-27
Idea: Response.next() method, for when allow_redirects=False


kennethreitz:

sigmavirus24:
I really don't like this. Responses will only keep a connection when you're streaming and that's it. If you're planning on attaching a session, that seems really ill-considered.

kennethreitz:
I really need to figure out why all my Requests notifications are going to spam

ueg1990:
its possible you once put a previous request notification to spam and your email service may have assumed all subsequent requests notifications to spam

PR #4066, opened 2017-05-27
API improvements

kennethreitz:
- `Response.next`

Related: #4065.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -711,15 +711,17 @@ class Response(object):
 
     @property
     def is_permanent_redirect(self):
-        """True if this Response one of the permanent versions of redirect"""
+        """True if this Response one of the permanent versions of redirect."""
         return ('location' in self.headers and self.status_code in (codes.moved_permanently, codes.permanent_redirect))
 
     def next(self):
+        """Returns a PreparedRequest for the next request in a redirect chain, if there is one."""
         if self.is_redirect:
             return self._next
+
     @property
     def apparent_encoding(self):
-        """The apparent encoding, provided by the chardet library"""
+        """The apparent encoding, provided by the chardet library."""
         return chardet.detect(self.content)['encoding']
 
     def iter_content(self, chunk_size=1, decode_unicode=False):
--- a/requests/models.py
+++ b/requests/models.py
@@ -577,6 +577,10 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         for event in hooks:
             self.register_hook(event, hooks[event])
 
+    def send(self, session):
+        """Sends the given PreparedRequest to the given session."""
+        return session.send(self)
+
 
 class Response(object):
     """The :class:`Response <Response>` object, which contains a
@@ -593,6 +597,7 @@ class Response(object):
 
         self._content = False
         self._content_consumed = False
+        self._next = None
 
         #: Integer Code of responded HTTP Status, e.g. 404 or 200.
         self.status_code = None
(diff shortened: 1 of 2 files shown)


requests/models.py
```diff
@@ -577,6 +577,10 @@ def prepare_hooks(self, hooks):
         for event in hooks:
             self.register_hook(event, hooks[event])
 
+    def send(self, session):
```
sigmavirus24: Why have this when we already have `session.send`?

kennethreitz: more intuitive. i dabbled about it alot when i was writing it originally, and i opted out, decided that i could always add it back in later — i think it's time for it to go in. it's just a nice shortcut.

kennethreitz: `Response.next().send(session)` is kinda nice

kennethreitz: a lot better than `session.send(Response.next())` imo

requests/sessions.py
```diff
@@ -114,8 +114,8 @@ def get_redirect_target(self, resp):
         return None
 
     def resolve_redirects(self, resp, req, stream=False, timeout=None,
-                          verify=True, cert=None, proxies=None, **adapter_kwargs):
-        """Receives a Response. Returns a generator of Responses."""
+                          verify=True, cert=None, proxies=None, yield_responses=True, **adapter_kwargs):
```
sigmavirus24: Yield requests maybe?

requests/models.py
kennethreitz: i can peel that out if you're really against it — i think it's harmless though, and just smooths out the api a bit

requests/sessions.py
kennethreitz: sure!

requests/models.py
sigmavirus24: I see no difference, except that one is more intuitive for people already using that API. Sessions send requests, requests don't send sessions.

requests/models.py
```diff
@@ -706,12 +711,17 @@ def is_redirect(self):
 
     @property
     def is_permanent_redirect(self):
-        """True if this Response one of the permanent versions of redirect"""
+        """True if this Response one of the permanent versions of redirect."""
         return ('location' in self.headers and self.status_code in (codes.moved_permanently, codes.permanent_redirect))
 
+    def next(self):
```
sigmavirus24: If this isn't doing anything, why isn't it just a property?

kennethreitz: hmmmm

requests/models.py
sigmavirus24: "Harmless" seems in conflict with "Extra surface area that doesn't seem to provide much benefit"

requests/models.py
sigmavirus24: Also, `next_request` would be more intuitive.

requests/models.py
kennethreitz: i remember why i left this off now.

kennethreitz: if responses recorded their sent arguments, this would have made sense.

requests/sessions.py
```diff
@@ -203,22 +203,26 @@ def resolve_redirects(self, resp, req, stream=False, timeout=None,
...
-                stream=stream,
-                timeout=timeout,
-                verify=verify,
-                cert=cert,
-                proxies=proxies,
-                allow_redirects=False,
-                **adapter_kwargs
-            )
+            if yield_requests:
+                yield req
```
Lukasa, line 207: Ok, so I just stumbled across this in the code, and this is *very* confusing. Despite this being a generator, it is a generator that can literally be iterated only once in this mode before it starts doing something stupid (where "stupid" is that it repeatedly runs the same processing). That's not actually written down anywhere, so it would probably help us to at least say that the "yield requests" mode can only yield *one* request.

PR #4098, opened 2017-05-29
move functionlocal imports in utils.py and models.py to global scope

NiklasRosenstein:
This avoids runtime warnings when using the `requests` module for the first time when it is no longer in `sys.modules`. 

```
[...]\Lib\site-packages\requests\utils.py:164: RuntimeWarning: Parent module 'requests' not found while handling absolute import
  from netrc import netrc, NetrcParseError
[...]\Lib\site-packages\requests\utils.py:49: RuntimeWarning: Parent module 'requests' not found while handling absolute import
  import _winreg as winreg
```

Seeing this warning is very common for me, since I develop Python plugins for an application that embeds Python. To ensure that no dependencies between plugins are conflicting, every plugin develiers its own dependencies and makes sure they leave no trace in `sys.modules` and `sys.path`.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -16,6 +16,13 @@ import sys
 # such as in Embedded Python. See https://github.com/kennethreitz/requests/issues/3578.
 import encodings.idna
 
+try:
+    import idna as _idna
+except ImportError as _exc:
+    # Re-raised when the idna module is actually needed.
+    _idna = None
+    _idna_err = _exc
+
 from urllib3.fields import RequestField
 from urllib3.filepost import encode_multipart_formdata
 from urllib3.util import parse_url
@@ -334,11 +341,11 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
 
     @staticmethod
     def _get_idna_encoded_host(host):
-        import idna
-
+        if _idna is None:
+            raise _idna_err
         try:
-            host = idna.encode(host, uts46=True).decode('utf-8')
-        except idna.IDNAError:
+            host = _idna.encode(host, uts46=True).decode('utf-8')
+        except _idna.IDNAError:
             raise UnicodeError
         return host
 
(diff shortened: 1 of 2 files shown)


kennethreitz:
add to whitelist

kennethreitz:
add to whitelist

kennethreitz:
add to whitelist

Lukasa:
Hrm, I'm pretty nervous about this. It's hard for us to avoid regressing this behaviour and it's extremely likely that we'd add it somewhere else. Is it not more advisable to silence the warning?

NiklasRosenstein:
While that is possible, eg. by wrapping the requests API, the [`warnings` "Temporarily Suppressing Warnings"][0] documentation states that the behaviour is undefined in multithreaded applications.

  [0]: https://docs.python.org/2/library/warnings.html#temporarily-suppressing-warnings

Suppressing all warnings, always, is definitely not an option.

```python
def requests_get(*args, **kwargs):
  with warnings.catch_warnings():  # UB in multithreaded applications
    warnings.simplefilter('ignore')
    return requests.get(*args, **kwargs)
```

requests/models.py
```diff
@@ -16,6 +16,13 @@
 # such as in Embedded Python. See https://github.com/kennethreitz/requests/issues/3578.
 import encodings.idna
 
+try:
+    import idna as _idna
```
sigmavirus24, line 20: So this is **absolutely** _unacceptable_.

While the problems in the `idna` module may not be resolved, the fact that we've recently switched away from vendoring *known* good versions means that users will still run into trouble with this if they have an old enough version installed with requests.

This import is conditionally inside the staticmethod because it is:

1. Not always needed
2. Potentially incredibly slow (on the order of seconds if I remember correctly) with older versions of idna

requests/utils.py
```diff
@@ -42,11 +48,11 @@
 if platform.system() == 'Windows':
     # provide a proxy_bypass version on Windows without DNS lookups
 
+    if is_py3:
```
sigmavirus24, line 51: I confess to not knowhing why `winreg` or `netrc` were imported inside functions but I suspect we've had issues in the past. Delaying imports until we have determined them to be necessary is a *valid* use-case for not globally importing modules.

kennethreitz:
I'm going to close this.

PR #4108, opened 2017-05-30
flake8 improvements

kennethreitz:


kennethreitz:
man, that tool catches some great stuff. thanks for maintaining it, @sigmavirus24!

Issue #4112, opened 2017-05-30
please support cchardet as alternative to chardet


2mf:
aiohttp is supporting cchardet as alternative to chardet as it is faster
please do the same in requests

PR #4115, opened 2017-05-30
add cchardet support and use chardet as fallback

2mf:
I created this pull-request for #4112

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -39,15 +39,18 @@ is at <http://python-requests.org>.
 :copyright: (c) 2017 by Kenneth Reitz.
 :license: Apache 2.0, see LICENSE for more details.
 """
+def get_version(package):
+    version = package.__version__.split('.')[:3]
+    if len(version) == 2:
+          version.append('0')
+    major, minor, patch = version
+    major, minor, patch = int(major), int(minor), int(patch)
+    return major, minor, patch
 
 # Check urllib3 for compatibility.
 import urllib3
-urllib3_version = urllib3.__version__.split('.')
-# Sometimes, urllib3 only reports its version as 16.1.
-if len(urllib3_version) == 2:
-    urllib3_version.append('0')
-major, minor, patch = urllib3_version
-major, minor, patch = int(major), int(minor), int(patch)
+
+major, minor, patch = get_version(urllib3)
 # urllib3 >= 1.21.1, < 1.22
 try:
     assert major == 1
@@ -58,11 +61,6 @@ except AssertionError:
 
 import warnings
 
-def get_version(package):
-    major, minor, patch = package.__version__.split('.')[:3]
-    major, minor, patch = int(major), int(minor), int(patch)
-    return major, minor, patch
-
 try:
     # Check cchardet for compatibility.
     import cchardet as chardet
@@ -71,12 +69,13 @@ try:
     try:
         assert major == 2
         assert minor >= 1
+        assert minor < 2
         assert patch >= 0
     except AssertionError:
-        warnings.warn('Requests dependency \'cchardet\' must be version >= 2.1.0! Falling back to chardet')
+        warnings.warn('Requests dependency \'cchardet\' must be version >= 2.1.0, < 2.2.0! Falling back to chardet')
         raise
 
-except (ImportError, SyntaxError, AssertionError) as e:
+except (ImportError, AssertionError):
     # Check chardet for compatibility.
     import chardet
     major, minor, patch = get_version(chardet)
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -56,18 +56,34 @@ try:
 except AssertionError:
     raise RuntimeError('Requests dependency \'urllib3\' must be version >= 1.21.1, < 1.22!')
 
+import warnings
 
-# Check chardet for compatibility.
-import chardet
-major, minor, patch = chardet.__version__.split('.')[:3]
-major, minor, patch = int(major), int(minor), int(patch)
-# chardet >= 3.0.2, < 3.1.0
 try:
-    assert major == 3
-    assert minor < 1
-    assert patch >= 2
-except AssertionError:
-    raise RuntimeError('Requests dependency \'chardet\' must be version >= 3.0.2, < 3.1.0!')
+    # Check chardet for compatibility.
+    import cchardet as chardet
+    major, minor, patch = chardet.__version__.split('.')[:3]
+    major, minor, patch = int(major), int(minor), int(patch)
+    # cchardet >= 2.1.0
+    try:
+        assert major == 2
+        assert minor >= 1
+        assert patch >= 0
+    except AssertionError:
+        warnings.warn('Requests dependency \'cchardet\' must be version >= 2.1.0! Falling back to chardet')
+        raise
+
+except (ImportError, SyntaxError, AssertionError) as e:
+    # Check chardet for compatibility.
+    import chardet
+    major, minor, patch = chardet.__version__.split('.')[:3]
+    major, minor, patch = int(major), int(minor), int(patch)
+    # chardet >= 3.0.2, < 3.1.0
+    try:
+        assert major == 3
+        assert minor < 1
+        assert patch >= 2
+    except AssertionError:
+        raise RuntimeError('Requests dependency \'chardet\' must be version >= 3.0.2, < 3.1.0!')
 
 # Attempt to enable urllib3's SNI support, if possible
 try:
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -58,11 +58,15 @@ except AssertionError:
 
 import warnings
 
-try:
-    # Check chardet for compatibility.
-    import cchardet as chardet
-    major, minor, patch = chardet.__version__.split('.')[:3]
+def get_version(package):
+    major, minor, patch = package.__version__.split('.')[:3]
     major, minor, patch = int(major), int(minor), int(patch)
+    return major, minor, patch
+
+try:
+    # Check cchardet for compatibility.
+    import cchardet as chardet
+    major, minor, patch = get_version(chardet)
     # cchardet >= 2.1.0
     try:
         assert major == 2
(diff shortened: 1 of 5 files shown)


2mf:
sure :-)
I have to update the test settings as well to ensure chardet and cchardet run both

requests/packages.py
```diff
@@ -3,8 +3,17 @@
 # This code exists for backwards compatibility reasons.
 # I don't like it either. Just look the other way. :)
 
-for package in ('urllib3', 'idna', 'chardet'):
-    locals()[package] = __import__(package)
+for _package in ('urllib3', 'idna', ('cchardet', 'chardet')):
+    if isinstance(_package, tuple):
+        package = _package[1]
+        try:
+            locals()[package] = __import__(package[0])
```
sethmlarson: Doesn't look like this line is correct, should be `__import__(_package[0])`? Might want to unpack the tuple right away and then use proper variable names so the logic is easier to follow.

tox.ini
```diff
@@ -1,8 +1,14 @@
 [tox]
-envlist = py26,py27,py33,py34,py35,py36
+envlist = py{26,27,33,34,35,36}-{chardet,cchardet}
 
 [testenv]
 
+deps =
+    py{26,27,33,34,35,36}-chardet: chardet
```
nateprewitt: I'm not super fond of doubling the number of test instances we're running.

requests/__init__.py
```diff
@@ -56,18 +56,37 @@
 except AssertionError:
     raise RuntimeError('Requests dependency \'urllib3\' must be version >= 1.21.1, < 1.22!')
 
+import warnings
+
+def get_version(package):
```
nateprewitt: If we are going to abstract this out to a function, maybe we add it to the urllib3 import to reduce the duplicated code to a single place?

requests/__init__.py
```diff
@@ -56,18 +56,37 @@
...
+    # Check cchardet for compatibility.
+    import cchardet as chardet
+    major, minor, patch = get_version(chardet)
+    # cchardet >= 2.1.0
+    try:
+        assert major == 2
+        assert minor >= 1
+        assert patch >= 0
+    except AssertionError:
+        warnings.warn('Requests dependency \'cchardet\' must be version >= 2.1.0! Falling back to chardet')
```
Lukasa: This isn't specifying the upper pin, which is `< 3.0.0`.

nateprewitt: We probably don't want to fire this warning every time we import requests. `cchardet` should be viewed as an optional addition, not the default imo.

requests/__init__.py
```diff
@@ -56,18 +56,37 @@
...
+        raise
+
+except (ImportError, SyntaxError, AssertionError) as e:
```
nateprewitt: What's the scenario that raises a `SyntaxError`? Also if we're not using `e` for anything, we probably don't need to set it.

requests/packages.py
```diff
@@ -3,8 +3,16 @@
 # This code exists for backwards compatibility reasons.
 # I don't like it either. Just look the other way. :)
 
-for package in ('urllib3', 'idna', 'chardet'):
-    locals()[package] = __import__(package)
+for package in ('urllib3', 'idna', ('chardet', 'cchardet')):
+    if isinstance(package, tuple):
+        package, alt_package = package
```
kennethreitz, line 8: we need to be *extremely* careful about touching this code, at all.

kennethreitz:
This whole thing scares me.

Lukasa:
I for one am not wedded to doing it: I'm nervous about the amount of code it adds as well.

kennethreitz:
@Lukasa i feel like this should be a two-liner, if anything.

kennethreitz:
@2mf do you have any metrics on how much faster cchardet is than chardet? e.g. is it worth the trouble?

tox.ini
2mf: I'm not quite sure, how I can test against ccharset and charset using existing cases

requests/__init__.py
2mf: sure I'll do that

requests/__init__.py
2mf: I copied that from other places in requests, but from my view point it is not needed

2mf:
@kennethreitz a benchmark can be seen at https://pypi.python.org/pypi/cchardet

> Python 2.7.12
> 	Request (call/s)
> chardet	         0.26
> cchardet	1341.81
>
> Python 3.6.0
> 	Request (call/s)
> chardet	        0.26
> cchardet	1472.43

2mf:
@kennethreitz @Lukasa 
If you both don't like the idea of including cchardet. I don't have a problem with that but then there must be a possibility to use cchardet by monkey patching but this is prevented by other things like:

shouldn't be the version matching in the requirements of setup.py?
```
# Check chardet for compatibility.
import chardet
major, minor, patch = chardet.__version__.split('.')[:3]
major, minor, patch = int(major), int(minor), int(patch)
# chardet >= 3.0.2, < 3.1.0
try:
    assert major == 3
    assert minor < 1
    assert patch >= 2
except AssertionError:
    raise RuntimeError('Requests dependency \'chardet\' must be version >= 3.0.2, < 3.1.0!')
```
and
```
for package in ('urllib3', 'idna', 'chardet'):
    locals()[package] = __import__(package)
    # This traversal is apparently necessary such that the identities are
    # preserved (requests.packages.urllib3.* is urllib3.*)
    for mod in list(sys.modules):
        if mod == package or mod.startswith(package + '.'):
            sys.modules['requests.packages.' + mod] = sys.modules[mod]
```

kennethreitz:
That is significantly faster!

sigmavirus24:
> shouldn't be the version matching in the requirements of setup.py?

No, @2mf it shouldn't. Requests needs *very* specific versions of dependencies. Some of those dependencies are popular and we need a way to tell the user "You're using a version that will horribly break your version of Requests"

requests/__init__.py
```diff
@@ -56,18 +59,33 @@
...
+    import cchardet as chardet
+    major, minor, patch = get_version(chardet)
+    # cchardet >= 2.1.0
+    try:
+        assert major == 2
+        assert minor >= 1
+        assert minor < 2
+        assert patch >= 0
+    except AssertionError:
+        warnings.warn('Requests dependency \'cchardet\' must be version >= 2.1.0, < 2.2.0! Falling back to chardet')
```
sigmavirus24, line 75: This is not a dependency of Requests, it's an optional speed-up. Please revise

requests/__init__.py
```diff
@@ -56,18 +59,33 @@
...
-    assert patch >= 2
-except AssertionError:
-    raise RuntimeError('Requests dependency \'chardet\' must be version >= 3.0.2, < 3.1.0!')
+    # Check cchardet for compatibility.
+    import cchardet as chardet
+    major, minor, patch = get_version(chardet)
+    # cchardet >= 2.1.0
+    try:
+        assert major == 2
+        assert minor >= 1
```
sigmavirus24, line 71: Why not:

```py
assert 1 <= minor < 2
```

requests/__init__.py
```diff
@@ -56,18 +59,33 @@
...
+        assert minor < 2
+        assert patch >= 0
+    except AssertionError:
+        warnings.warn('Requests dependency \'cchardet\' must be version >= 2.1.0, < 2.2.0! Falling back to chardet')
+        raise
+
+except (ImportError, AssertionError):
+    # Check chardet for compatibility.
+    import chardet
+    major, minor, patch = get_version(chardet)
```
sigmavirus24, line 81: `chardet` shouldn't be imported in this except block. This will cause extremely confusing tracebacks for users if they can't import either.

setup.py
```diff
@@ -93,6 +93,7 @@ def run_tests(self):
     cmdclass={'test': PyTest},
     tests_require=test_requirements,
     extras_require={
+        'cchardet:(python_version == "2.7" or python_version >= "3.4")': ['cchardet>=2.1.0'],
```
sigmavirus24, line 96: If we're enforcing (in `__init__.py`) that `cchardet` be between `2.1.0` and `2.2.0` then that should be written here.

requests/packages.py
sigmavirus24: I agree. This *looks* safe, but it'd be so much better if this were testable.

sigmavirus24:
> That is significantly faster!

And harder to install and the benchmark, as I understand it, is based off a *very* old version of chardet and hasn't been updated since.

kennethreitz:
It sounds like we're against this change overall. Sorry @2mf.

sigmavirus24:
I'm on board with doing this, but the current methodology isn't 100% correct and I'm merely pointing out that those benchmarks are older and potentially misleading. @dan-blanchard has done a lot of fantastic work on chardet recently

dan-blanchard:
I feel the need to point out that chardet is in the midst of gaining support for more codecs and languages than cchardet supports. I'm retraining all the models for greater accuracy too. Speed isn't everything with encoding detection.

2mf:
Just for the sake of a recent benchmark
I did the same benchmark for chardet and cchardet that where done by @PyYoshi (on my machine) and the difference is still in the same dimension.

> Python 3.6.1
> chardet 3.0.3
> cchardet 2.1.0

> python test_chardet.py 
> chardet: 0.24908731023612882 call(s)/s
> 
> python test_cchardet.py 
> chardet: 1161.9214360906421 call(s)/s
> 

anyway, It is not worth to continue on this one while running against so many walls

dan-blanchard:
I hadn't actually looked at the benchmark code cChardet uses before, but it just [runs the same small SJIS text file through both modules 100 times](https://github.com/PyYoshi/cChardet/blob/master/src/tests/bench.py#L9). I know cChardet will pretty much always be faster than chardet—in fact, if you look at old chardet issues, you'll even see that I once toyed with the idea of trying to merge the projects—but that is not an accurate benchmark, especially for something like requests where a good amount of the time is spent downloading the page content anyway. A requests-based benchmark would make sense for deciding if it makes a difference here.

PR #4118, opened 2017-05-30
Drop python2.6 support

### LINKED ISSUE #3928: Drop 2.6 support
Our main dependent that still supports 2.6, pip, has just dropped support (pypa/pip#4343). We are now free to do the same.

hobarrera:
Drop support for python2.6. Fixes #3928

Note that this targets the 3.0.0 branch, not master.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -478,11 +478,10 @@ class HTTPAdapter(BaseAdapter):
 
                     # Receive the response from the server
                     try:
-                        # For Python 2.7+ versions, use buffering of HTTP
-                        # responses
+                        # For Python 2.7, use buffering of HTTP responses
                         r = low_conn.getresponse(buffering=True)
                     except TypeError:
-                        # For compatibility with Python 2.6 versions and back
+                        # For Python 3.3+ versions, this is the default
                         r = low_conn.getresponse()
 
                     resp = HTTPResponse.from_httplib(
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -187,8 +187,7 @@ class HTTPAdapter(BaseAdapter):
         self.init_poolmanager(pool_connections, pool_maxsize, block=pool_block)
 
     def __getstate__(self):
-        return dict((attr, getattr(self, attr, None)) for attr in
-                    self.__attrs__)
+        return {attr: getattr(self, attr, None) for attr in self.__attrs__}
 
     def __setstate__(self, state):
         # Can't handle by adding 'proxy_manager' to self.__attrs__ because
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(
         'Natural Language :: English',
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2.6',
+        'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.3',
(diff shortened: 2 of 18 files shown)


WhyNotHugo:
Not entirely sure why CI is failing -- is this branch not updated for it?
```
+ make ci
make: *** No rule to make target 'ci'.  Stop.
```

Lukasa:
@hobarrera No, it looks like it isn't. I'll try to resolve that problem today.

WhyNotHugo:
Thanks!

setup.py
```diff
@@ -84,7 +84,6 @@ def run_tests(self):
         'Natural Language :: English',
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2.6',
```
jdufresne, line 83: Good that this one was removed, but there should exist a trove classifier for `'Programming Language :: Python :: 2'` as requests still supports Python 2 in some capacity.

kennethreitz: +1

WhyNotHugo: Added `'Programming Language :: Python :: 2'` Keep in mind that `2.7` is still there as well.

requests/adapters.py
```diff
@@ -484,11 +484,10 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
...
                     # Receive the response from the server
                     try:
-                        # For Python 2.7+ versions, use buffering of HTTP
-                        # responses
+                        # For Python 2.7, use buffering of HTTP responses
                         r = low_conn.getresponse(buffering=True)
                     except TypeError:
-                        # For compatibility with Python 2.6 versions and back
+                        # For Python 3.3+ versions, this is the default
                         r = low_conn.getresponse()
```
WhyNotHugo, line 484: The comments for this bit were oddly out of date; only this branch works on python3, while the above fails (with `TypeError`).

nateprewitt:
Hey @hobarrera, it looks like 2.6 is still in the tox.ini file. If we can pull that out, I think this should be good to go.

WhyNotHugo:
Tests are "passing". Just 2.6 failed, as expected.

WhyNotHugo:
Whoops! There we go. Jenkins is still configured to run 2.6 though.

nateprewitt:
@kennethreitz, are the versions running on Jenkins hardcoded for the server? Do we have any way to delegate this to something like tox?

kennethreitz:
@nateprewitt they are currently pinned

Lukasa:
"work around" is a bit tricky here, because they're required builds. I think we need a way to turn them off for this branch.

WhyNotHugo:
The only workarounds I can think of are extremely hacky things, like raising a "SkipTest" exception before most imports run, or things like that.

nateprewitt:
@Lukasa 😕 well that's unfortunate. I talked briefly with Kenneth last night and it seemed like there was some difficulty differing build requirements between branches. We'll probably need to investigate configuration options for [Jenkins with tox](https://tox.readthedocs.io/en/latest/example/jenkins.html), or sit on this branch until we're ready to shut off 2.6 tests across the board.

@hobarrera this is definitely an infrastructure issue on our end. I'd *really* like to avoid making any code changes in Requests' tests to deal with a CI server.

kennethreitz:
so, i have a solution to this — we just need a separate job for the 3.0.0 branch, which removes python 2.6. I'll see if I can get to this today.

WhyNotHugo:
I guess I'll just leave this here until Jenkins is ready? I don't think there's anything else I can do. 🙂

kennethreitz:
The "workaround"' here is to allow administrators to merge anyway. I'll turn that on now.

kennethreitz:
there, that was easy.

WhyNotHugo:
Thanks!

PR #4120, opened 2017-05-31
Bring 3.0 up-to-date with master.

Lukasa:
This was a gnarly merge, and the odds of it going right first time are pretty low, so let's have the CI prove or disprove the success of it for me.


sigmavirus24:
@Lukasa you're merging this into master?

nateprewitt:
Other than definitely not wanting to merge this into master, this looks like it's right. It would be nice to see a smaller diff against proposed/3.0.0 though.

sigmavirus24:
Okay, I updated the branch for you

requests/__version__.py
```diff
@@ -0,0 +1,14 @@
+# .-. .-. .-. . . .-. .-. .-. .-.
+# |(  |-  |.| | | |-  `-.  |  `-.
+# ' ' `-' `-`.`-' `-' `-'  '  `-'
+
+__title__ = 'requests'
+__description__ = 'Python HTTP for Humans.'
+__url__ = 'http://python-requests.org'
+__version__ = '2.17.3'
+__build__ = 0x021703
```
nateprewitt: 0x030000

requests/__version__.py
```diff
@@ -0,0 +1,14 @@
+# .-. .-. .-. . . .-. .-. .-. .-.
+# |(  |-  |.| | | |-  `-.  |  `-.
+# ' ' `-' `-`.`-' `-' `-'  '  `-'
+
+__title__ = 'requests'
+__description__ = 'Python HTTP for Humans.'
+__url__ = 'http://python-requests.org'
+__version__ = '2.17.3'
```
nateprewitt: This should to be 3.0.0, I believe.

requests/models.py
```diff
@@ -335,17 +337,11 @@ def prepare_method(self, method):
         self.method = method
         if self.method is None:
             raise ValueError('Request method cannot be "None"')
-        self.method = to_native_string(self.method).upper()
+        self.method = to_native_string(self.method.upper())
```
nateprewitt, line 340: Looks like parens may have been unintentionally moved here? I don't know if it's actually going to affect the final output but may cause issues in future merges.

Lukasa: I *think* it doesn't matter? `.upper()` is safe on all types of strings.

nateprewitt: Yep, this is purely stylistic I think. Just wanted to make sure we knew the change was being reverted.

requests/__version__.py
sigmavirus24: 👍

Lukasa:
FYI, the windows tests won't finish because they're blocked. I pinged @kennethreitz about this in IRC.

Lukasa:
Ok, the Windows CI won't get working any time soon as @kennethreitz is on the road. I suggest we move on without it to unblock #4118.

nateprewitt:
@Lukasa, we have a [successful run](http://jenkins.kennethreitz.org:8080/job/requests/job/requests-windows/lastSuccessfulBuild/) of master on Windows from yesterday. Since nothing has changed on master, and I don't see anything directly related to Windows in the 3.0.0 merge, I think it's probably OK to merge this.

Lukasa:
Kenneth restarted the builder, so we should get a result shortly.

kennethreitz:
windows had a failure!

Lukasa:
It did, that's true, but that failure seems to be related to having had something wacky happen with the environment build. I'm restarting the builder.

Lukasa:
Yeah, I don't know that I trust the output of this Windows builder. It seems to not really be working properly.

.travis.yml
```diff
@@ -16,7 +16,17 @@ install: "make"
 script:
   - |
     if [[ "$TRAVIS_PYTHON_VERSION" != "2.6" ]] ; then make test-readme; fi
-  - make coverage
-after_success:
-  - pipenv run codecov
+  - make ci
 cache: pip
+jobs:
```
WhyNotHugo, line 21: You don't want the jobs AND the script above, it's kinda redundant.

WhyNotHugo: Oh, this is actually the same on master though, so it's probably worth fixing there.

Lukasa:
@kennethreitz What are we doing about this Windows builder?

kennethreitz:
@Lukasa not sure why the build is hanging on Windows. I'll take a peek around. It runs fine on master.

nateprewitt:
Looks like we're good to go here.

WhyNotHugo:
I've rebased #4118 onto this locally, should I push that update?

Lukasa:
@hobarrera Go for it

PR #4124, opened 2017-06-03
Add support for MessagePack

suzaku:


sigmavirus24:
Hi there @suzaku!

Thanks for sending this pull request! We're so glad you've chosen to contribute to Requests and we appreciate all the time and effort you've put in to get to the point of sending this PR.

Sadly, however, it's highly unlikely that we will accept this pull request. Here's why:

Requests focuses on satisfying the needs of 98% of its users. As it stands, roughly that percentage of our users will utilize and take advantage of the `json` keyword. Sadly, we've not received nearly as much feedback about providing the same functionality for `msgpack`.

I'll leave this open for other core reviewers to comment on, but I wanted to set your expectations appropriately.

Cheers,
Ian

kennethreitz:
Ian is correct. This would make a great extension module, however!

sigmavirus24:
Thanks again @suzaku

suzaku:
@sigmavirus24    Thanks. I will handle msgpack in my application then.
@kennethreitz Good idea.

PR #4134, opened 2017-06-06
Raise exceptions in utils.set_environ

### LINKED ISSUE #4127: UnboundLocalError: local variable 'bypass' referenced before assignment
When using requests with GAE dev_appserver sandbox, this variable gets referenced before assignment when making requests to the FB graph API

UnboundLocalError: local variable 'bypass' referenced before assignment
https://github.com/requests/requests/blob/master/requests/utils.py#L680

mlcrazy:
fixes #4127

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -612,18 +612,18 @@ def set_environ(env_name, value):
     the environment variable 'env_name'.
 
     If 'value' is None, do nothing"""
-    if value is not None:
+    value_changed = value is not None
+    if value_changed:
         old_value = os.environ.get(env_name)
         os.environ[env_name] = value
     try:
         yield
     finally:
-        if value is None:
-            return
-        if old_value is None:
-            del os.environ[env_name]
-        else:
-            os.environ[env_name] = old_value
+        if value_changed:
+            if old_value is None:
+                del os.environ[env_name]
+            else:
+                os.environ[env_name] = old_value
 
 
 def should_bypass_proxies(url, no_proxy):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import os
+import copy
 from io import BytesIO
 
 import pytest
@@ -17,7 +18,7 @@ from requests.utils import (
     requote_uri, select_proxy, should_bypass_proxies, super_len,
     to_key_val_list, to_native_string,
     unquote_header_value, unquote_unreserved,
-    urldefragauth, add_dict_to_cookiejar)
+    urldefragauth, add_dict_to_cookiejar, set_environ)
 from requests._internal_utils import unicode_is_ascii
 
 from .compat import StringIO, cStringIO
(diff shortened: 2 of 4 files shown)


requests/utils.py
```diff
@@ -615,15 +615,15 @@ def set_environ(env_name, value):
...
-    try:
+        try:
+            yield
+        finally:
+            if old_value is None:
+                del os.environ[env_name]
+            else:
+                os.environ[env_name] = old_value
+    else:
         yield
```
Lukasa: Hrm, I don't love the two separate `yield` statements here. Do you mind refactoring the old `finally` statement to invert the conditional on `if value is None: return` and to instead have `if value is not None` then do the dance with `old_value`.

mlcrazy: @Lukasa will set this refactor up and add a test

mlcrazy: @Lukasa refactored the finally statement, let me know if this is more what you would like to see there, also added tests

requests/utils.py
```diff
@@ -612,18 +612,18 @@ def set_environ(env_name, value):
     the environment variable 'env_name'.
 
     If 'value' is None, do nothing"""
+
```
Lukasa: We don't need this newline.

requests/utils.py
```diff
@@ -612,18 +612,18 @@ def set_environ(env_name, value):
...
     if value is not None:
-        old_value = os.environ.get(env_name)
+        old_value = os.environ.get(env_name, None)
```
Lukasa: `None` is implicit in `get`, you can leave it out.

requests/utils.py
```diff
@@ -618,12 +618,11 @@ def set_environ(env_name, value):
     try:
         yield
     finally:
-        if value is None:
-            return
-        if old_value is None:
-            del os.environ[env_name]
-        else:
-            os.environ[env_name] = old_value
+        if value is not None:
```
nateprewitt: @Lukasa what are your thoughts on adding `value_changed = value is not None`, and using that single value for both conditionals? The fact these two checks are bound is implicit and has the potential to get out of sync as soon as one changes.

tests/test_utils.py
```diff
@@ -651,4 +652,31 @@ def QueryValueEx(key, value_name):
...
+
+@pytest.mark.parametrize(
+    'env_name, value', (
+            ('no_proxy', '192.168.0.0/24,127.0.0.1,localhost.localdomain'),
+            ('no_proxy', None),
+            ('a_new_key', '192.168.0.0/24,127.0.0.1,localhost.localdomain'),
+            ('a_new_key', None),
+    ))
+def test_set_environ(env_name, value):
+    """Tests set_environ will set environ values and will restore the environ
```
nateprewitt: Could we end this with a period and either bring the `"""` up to the same line or wrap the text down? We don't have a hard cap at 80 characters in Requests.

tests/test_utils.py
```diff
@@ -651,4 +652,31 @@ def QueryValueEx(key, value_name):
...
+    environ_copy = copy.deepcopy(os.environ)
+    with set_environ(env_name, value):
+        assert os.environ.get(env_name) == value
+
+    assert os.environ == environ_copy
+
+
+def test_set_environ_raises_exception():
+    """Tests set_environ will raise exceptions in context when the
+    value parameter is None
```
nateprewitt: `.`

requests/utils.py
Lukasa: Works for me!

mlcrazy:
@Lukasa @nateprewitt, did some commit fixups, not sure if I did the HISTORY.rst part correctly

Lukasa:
There are conflicts in the AUTHORS.rst file. :smile:

PR #4137, opened 2017-06-06
Allow Requests.Response to be used as a context manager

### LINKED ISSUE #4136: Allow Requests.Response to be used as a context manager in `with` statements
The [docs for streaming responses](http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow) currently state:

> If you find yourself partially reading request bodies (or not reading them at all) while using stream=True, you should consider using contextlib.closing ([documented here](http://docs.python.org/2/library/contextlib.html#contextlib.closing)), like this:
> ```python
> from contextlib import closing
> 
> with closing(requests.get('http://httpbin.org/get', stream=True)) as r:
>     # Do things with the response here.
> ```

It would be great to be able to omit `contextlib.closing()` entirely, and use the call to requests directly within a `with` statement, like so:

```python
with requests.get('http://httpbin.org/get', stream=True) as r:
    # Do things with the response here.
```

However that currently results in `AttributeError: __exit__`, since [`Requests.Response`](https://github.com/requests/requests/blob/v2.17.3/requests/models.py#L576-L942) doesn't have the `__enter__` and `__exit__` methods expected of a context manager:
https://docs.python.org/3/reference/datamodel.html#context-managers

I think it would be worthwhile to add them, to allow safe streaming without the extra boilerplate. (There's also prior art - `Requests.Session()` [can be used as a context-manager](https://github.com/requests/requests/blob/v2.17.3/requests/sessions.py#L389-L393).)

edmorley:
This saves having to wrap the call to requests with `contextlib.closing()`, allowing it to be used directly in a `with` statement, like so:

```python
with requests.get('http://httpbin.org/get', stream=True) as r:
    # Do things with the response here.
```

Fixes #4136.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -183,3 +183,4 @@ Patches and Suggestions
 - Shmuel Amar (`@shmuelamar <https://github.com/shmuelamar>`_)
 - Gary Wu (`@garywu <https://github.com/garywu>`_)
 - Ryan Pineo (`@ryanpineo <https://github.com/ryanpineo>`_)
+- Ed Morley (`@edmorley <https://github.com/edmorley>`_)
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,11 @@ Release History
 dev
 +++
 
+**Improvements**
+
+- ``Response`` is now a context manager, so can be used directly in a `with` statement
+  without first having to be wrapped by ``contextlib.closing()``.
+
 **Bugfixes**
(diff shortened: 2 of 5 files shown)


Lukasa:
FYI, @kennethreitz, as you can see here the master branch *is* failing on the Windows builder. @requests/core for the moment we can ignore that failing Windows test and continue to merge until @kennethreitz works out what's going on with the builder. ;)

PR #4146, opened 2017-06-09
Revert "Convert version compatibility checks to warnings"

kennethreitz:
Reverts requests/requests#4144

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -40,44 +40,34 @@ is at <http://python-requests.org>.
 :license: Apache 2.0, see LICENSE for more details.
 """
 
+# Check urllib3 for compatibility.
 import urllib3
-import chardet
-import warnings
-from .exceptions import RequestsDependencyWarning
-
-
-def check_compatibility(urllib3_version, chardet_version):
-    urllib3_version = urllib3_version.split('.')
-    assert urllib3_version != ['dev']  # Verify urllib3 isn't installed from git.
-
-    # Sometimes, urllib3 only reports its version as 16.1.
-    if len(urllib3_version) == 2:
-        urllib3_version.append('0')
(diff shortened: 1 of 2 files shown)


kennethreitz:
continuation of https://github.com/requests/requests/pull/4144#issuecomment-307426091

kennethreitz:
restarting for windows

sigmavirus24:
So we definitely don't want a complete revert of this. I do think we we want a selective revert though. We absolutely want to be more resilient to versions we can't parse. I don't think we want to emit warnings though.

kennethreitz:
windows passed!

nateprewitt:
Alright, I'd like to wrap this up so these commits aren't stuck in limbo. @sigmavirus24, my take away from the discussion is you'd still like to see exceptions over the warnings but allow for more resilient checks. What do you see happening when we can't parse a dependency? Is that still an exception with the unparsed version included in the message? How would we allow something like `python-telegram-bot` to work if they force a vendored copy of urllib3 onto the path?

You also said you'd rather not use `assert` statements to the `python -O` concerns. I'm assuming you'd rather see this as a conditional check that raises the exception?

If that all sounds correct, and @Lukasa/@kennethreitz are on board, I'll put the follow-up together.

sigmavirus24:
> How would we allow something like python-telegram-bot to work if they force a vendored copy of urllib3 onto the path?

Briefly responding to one question here: The point is to not allow that. I have to wonder if the exception shouldn't also include the location to the module so people can see "Hey, telegram bot is doing something real stupid and it's breaking requests."

mongarneb:
Hola estamos en un curso

mongarneb:
Hola estamos en un curso

PR #4150, opened 2017-06-10
delete prepare_auth func uless param

phantooom:
url  is a  useless param.


Lukasa:
Thanks for this!

However, while we don't use this it's possible that anyone subclassing might. I'm very reluctant to remove this with no good reason, as it's definitely a breaking change.

mongarneb:
hola

PR #4173, opened 2017-06-24
Pass pool_kwargs rather than updating connection_pool_kw

### LINKED ISSUE #3633: HTTPS requests through proxies in proposed/3.0.0 aren't configured correctly

jeremycline:
This addresses an issue where making HTTPS through proxies used the
default urllib3 connection pool settings.

fixes #3633

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -10,10 +10,6 @@ and maintain connections.
 
 import os.path
 import socket
-try:
-    from threading import RLock
-except ImportError:  # threading is an optional module and may not be present.
-    from dummy_threading import RLock
 
 from urllib3.poolmanager import PoolManager, proxy_from_url
 from urllib3.response import HTTPResponse
@@ -126,7 +122,6 @@ class HTTPAdapter(BaseAdapter):
         self._pool_connections = pool_connections
         self._pool_maxsize = pool_maxsize
         self._pool_block = pool_block
-        self._pool_kw_lock = RLock()
 
         self.init_poolmanager(pool_connections, pool_maxsize, block=pool_block)
 
@@ -139,7 +134,6 @@ class HTTPAdapter(BaseAdapter):
         # self.poolmanager uses a lambda function, which isn't pickleable.
         self.proxy_manager = {}
         self.config = {}
-        self._pool_kw_lock = RLock()
 
         for attr, value in state.items():
             setattr(self, attr, value)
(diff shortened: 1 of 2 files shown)


requests/adapters.py
```diff
@@ -204,11 +198,13 @@ def proxy_manager_for(self, proxy, **proxy_kwargs):
 
         return manager
 
-    def _update_poolmanager_ssl_kw(self, verify, cert):
-        """Update the :class:`PoolManager <urllib3.poolmanager.PoolManager>`
-        connection_pool_kw with the necessary SSL configuration. This method
-        should not be called from user code, and is only exposed for use when
-        subclassing the :class:`HTTPAdapter <requests.adapters.HTTPAdapter>`.
+    @staticmethod
+    def _pool_kwargs(verify, cert):
```
Lukasa: Don't have static methods: they're just not very useful. Either we should decide users can't override this, in which case this should be a function, or we should decide they can and it becomes an instance method.

jeremycline: I can't think of a particularly good reason for someone to want to override this, so a function sounds fine me. If someone really objects to the behaviour for some unforeseen reason, they can override ``get_connection`` and build the pool kwargs their own way.

jeremycline:
This is only a problem in the proposed/3.0.0 branch so even though this breaks the API that's okay, right?

requests/adapters.py
Lukasa: Yeah, that works for me. Function is the way to go then!

Lukasa:
Yup, no problem at all.

jeremycline:
I have made it so.

Lukasa:
Failing tests. =)

jeremycline:
Oh yeah, that's Python 2.6 because I used a dictionary comprehension in the tests. I can switch it to not use a comprehension, I just thought 2.6 was getting dropped in 3.0.0 and couldn't resist the opportunity to use them :smile:

Lukasa:
heh, right now we can't remove them from the test matrix, so you might have to just tolerate not having it. :grin:

jeremycline:
It's left me with a heavy heart, but the dictionary comprehension has been excised and tests pass on Python 2.6 again.

PR #4175, opened 2017-06-26
add missing packages when parse document

wangwenpei:


Lukasa:
Thanks for this! However, there is no need for it: when you install the Requests package by pip or setup.py those packages will be installed for you.

9nix00:
Ah, I see. it looks my requests version is too old. I use `requests==2.13.0`.

maybe we should add some note in document to notice user ?

Lukasa:
What would that look like?

9nix00:
I think maybe we can add it in [release-and-version-history](http://docs.python-requests.org/en/master/community/updates/#release-and-version-history) ?

parse document should be no error. if user didn't install `requests` and they just need document. 
they should know which package depends on. I still think a better way is to create a `requirements.txt` for document. this is a normal way.e.g:[scrapy](https://github.com/scrapy/scrapy/tree/master/docs)

of course, different community,different rule. respect community leaders choice.

Lukasa:
It is considered a bad idea to create `requirements.txt` for this. `requirements.txt` is really considered to be used only for a *specific installation*, not for a general statement of what is needed. `setup.py` is pretty much the standard way to do this.

If you want to build the docs then you *have* to install Requests, as we use Sphinx autodoc to build our API documentation, and that only works if you've installed the package.

PR #4182, opened 2017-07-03
Add idna version info to requests.help

alexwlchan:
Resolves #4180.

PR #4183, opened 2017-07-03
Fix help ssl check

nateprewitt:
This will resolve the test failures discovered in #4182. We were retrieving the `OPENSSL_VERSION_NUMBER` from the system `ssl` module in all cases before now. This attribute doesn't exist in Python 2.6 causing new tests for the `info()` method to fail. This patch will return an empty value for Python 2.6 since we don't seem to have a way to reliably check the openssl version.

@Lukasa @sigmavirus24, I've pushed this branch up to the requests repo directly, so feel free to chop this up or do anything you feel is necessary to get #4182 moving. I don't know that the provided test is especially useful, but it does show that the fix here works.

### DIFF
--- a/tests/test_help.py
+++ b/tests/test_help.py
@@ -1,0 +1,21 @@
+# -*- encoding: utf-8
+
+import sys
+
+import pytest
+
+from requests.help import info
+
+
+@pytest.mark.skipif(sys.version_info[:2] != (2,6), reason="Only run on Python 2.6")
+def test_system_ssl_py26():
+    """OPENSSL_VERSION_NUMBER isn't provided in Python 2.6, verify we don't
+    blow up in this case.
+    """
+    assert info()['system_ssl'] == {'version': ''}
+
+
+@pytest.mark.skipif(sys.version_info < (2,7), reason="Only run on Python 2.7+")
+def test_system_ssl():
+    """Verify we're actually setting system_ssl when it should be available."""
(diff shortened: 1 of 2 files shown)


tests/test_help.py
```diff
@@ -0,0 +1,15 @@
...
+
+from requests.help import info
+
+
+@pytest.mark.skipif(sys.version_info[:2] != (2,6), reason="Only run on Python 2.6")
+def test_system_ssl_py26():
+    """OPENSSL_VERSION_NUMBER isn't provided in Python 2.6, verify we don't
+    blow up in this case.
+    """
+    assert info()['system_ssl'] == {'version': ''}
```
alexwlchan, line 15: This test file would also pass if there was never anything interesting in `system_ssl`.

How about a second test that asserts that on Pythons >2.6, the `version` key is non-empty?

alexwlchan:
Also, does this need a changelog entry?

tests/test_help.py
nateprewitt: Yep, that seems reasonable if we're going to be testing for this specifically. I've added that into the test commit, if we decide to merge that part of the PR.

nateprewitt:
@alexwlchan, we historically haven't done changelog entries with PRs in Requests, but I know @Lukasa has chosen to ask for that in other projects. Perhaps it's worth asking if we want to consider moving to that here, or continue to compile them once before a release.

Issue #4158, opened 2017-06-14
Making requests to URLs with username and password does not respect no_proxy evironment settings


hodossy:
## Summary

When determining whether proxies should be bypassed or not, netloc is considered, however it contains the username and password separated by a colon, thus proxy settings cannot be applied, as the either username or host[:port] is checked among the no_proxy hosts.

## Expected Result

Proxy settings are respected

## Actual Result

Following errors are thrown:
```
OSError: Tunnel connection failed: 504 Gateway Timeout
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='localhost', port=8080): Max retries exceeded with url: / (Caused by ProxyError('Cannot connect to proxy.', OSError('Tunnel connection failed: 504 Gateway Timeout',)))

```
## Reproduction Steps

```python
import requests
import os

os.environ['https_proxy'] = #insert your proxy here
os.environ['no_proxy'] = 'localhost,127.0.0.1'

requests.get('https://username:password@localhost:8080/')
```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.3"
  },
  "cryptography": {
    "version": ""
  },
  "implementation": {
    "name": "CPython",
    "version": "3.6.0"
  },
  "platform": {
    "release": "3.16.0-4-amd64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.17.3"
  },
  "system_ssl": {
    "version": "1000114f"
  },
  "urllib3": {
    "version": "1.21.1"
  },
  "using_pyopenssl": false
}
```

## Workaround
proxies kwarg can be used to bypass this issue, e.g. by setting {'https': None}. Should be used with caution.

### Problem in my opinion:
should_bypass_proxies(url, no_proxy) function in utils.py:

network location is used, which matches the following pattern: [user[:password]@]host[:port]

- ip is determined based on ':' position, which is not valid in this case, as ip will be the username: `ip = netloc.split(':')[0]`
- when checking if proxy should be applied, two things are checked: `if netloc.endswith(host) or netloc.split(':')[0].endswith(host):`, neither of which will be true, as host may not contain port, and username surely does not end with the host.

I propose to use `domain = urlparse(url).hostname` instead of `netloc = urlparse(url).netloc` when checking against the no_proxy environment variable.

Lukasa:
[...]

PR #4188, opened 2017-07-08
Make username:pass@host URLs play nice with NO_PROXY environment variable

Rjevski:
Fixes issue #4158.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -639,28 +639,30 @@ def should_bypass_proxies(url, no_proxy):
     no_proxy_arg = no_proxy
     if no_proxy is None:
         no_proxy = get_proxy('no_proxy')
-    netloc = urlparse(url).netloc
+    parsed = urlparse(url)
 
     if no_proxy:
         # We need to check whether we match here. We need to see if we match
-        # the end of the netloc, both with and without the port.
+        # the end of the hostname, both with and without the port.
         no_proxy = (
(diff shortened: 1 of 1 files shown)


requests/utils.py
```diff
@@ -639,28 +639,30 @@ def should_bypass_proxies(url, no_proxy):
...
                     # If no_proxy ip was defined in plain IP notation instead of cidr notation &
                     # matches the IP of the index
                     return True
         else:
+            port = parsed.port or (80 if parsed.scheme == "http" else 443)
+            hostport = parsed.hostname + ":" + str(port)
+
             for host in no_proxy:
-                if netloc.endswith(host) or netloc.split(':')[0].endswith(host):
+                if hostport.endswith(host) or parsed.hostname.endswith(host):
```
Lukasa: I think we need some better testing here: `hostport` will now always contain a port where they wouldn't have before.

ghost: For some reason I overlooked that - I assumed netloc would always contain a port as well. Would making hostport contain a port only if explicitly specified in the URL be alright?

Lukasa: I'm honestly not sure: we may need some tests to validate this.

tonka3000:
Is there any progress so far? we also having problem with that bug!

nateprewitt:
It looks like this ended up getting resolved in #4427, so I'm closing this as resolved. Thanks for the effort here @ghost.

PR #4194, opened 2017-07-19
compat.py uses try/except to prevent linter errors

cclauss:
flake8 testing of https://github.com/requests/requests on Python 3.6.2

$ __time flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./requests/compat.py:53:11: F821 undefined name 'unicode'
    str = unicode
          ^

./requests/compat.py:54:18: F821 undefined name 'basestring'
    basestring = basestring
                 ^

./requests/compat.py:55:27: F821 undefined name 'long'
    numeric_types = (int, long, float)
                          ^

./requests/compat.py:56:27: F821 undefined name 'long'
    integer_types = (int, long)
                          ^

./requests/models.py:357:19: F821 undefined name 'unicode'
            url = unicode(url) if is_py2 else str(url)
                  ^
```

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -32,6 +32,21 @@ except (ImportError, SyntaxError):
     # because of u'...' Unicode literals.
     import json
 
+try:
+    integer_types = (int, long)  # Python 2
+    numeric_types = (int, long, float)
+except NameError:
+    integer_types = (int, )      # Python 3
+    numeric_types = (int, float)
+    
+builtin_str = str
+try:
+    basestring                   # Python 2
+    bytes = str
+    str = unicode
+except NameError:
+    basestring = (str, bytes)    # Python 3
+
 # ---------
 # Specifics
 # ---------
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this! However, I'm not concerned about the linter errors here because any moderately intelligent static analysis tool could tell you that they're not a concern. So I don't think this patch provides any particular value to the project. Sorry!

cclauss:
Understood.  Is there a better linter that you would recommend for Python projects?

Lukasa:
Not particularly: this linter is fine. It's just ill-suited to this kind of file.

As a more general note, you'll find you may encounter hostility when you open PRs that do nothing more than make linter changes to files. If the project isn't using the linter in CI, that's usually a signal that they aren't interested in having their project linted. :smile:

cclauss:
Agreed.  It is always difficult to know who are those who know linting is possible but don't want to do it and those who don't even know that it is possible and also don't know of the lintable problems that lurk in their code.

Thanks for the awesome work -- this project is essential to tons of developers!

PR #4204, opened 2017-07-26
Correct urllib3 compatibility comment

browniebroke:
So that it's in line with the code. 

I saw the warning raised recently and had to dive into the code for exact compatibility, but I stopped after reading the comment rather than the code. 

Maybe a better solution would be to surface the supported versions in the warning? However, I see it was changed recently in https://github.com/requests/requests/pull/4144/, any specific reason for removing that?


Lukasa:
The primary reason is just trying to reduce the maintenance area of this code: it's painful to need to update in lots of places. Ideally we'd have a nice refactor that will work here to minimise the number of places we need to write our version constraints: if someone wants to do that, it might be a worthwhile project.

PR #4206, opened 2017-07-26
some docs on request redirect detection

jvanasco:
This is just a small docstring that I wanted to suggest for inclusion. I should have offered it in my earlier commit.

The purpose of the docstring: we had some unit tests to ensure that `get_redirect_target` was being called the appropriate number of times. They broke because I calculated the number incorrectly.  It took a while to re-familiarize myself with the code and figure out what was going on.


kennethreitz:
this looks solid to me!

PR #4231, opened 2017-08-09
Drop Python 3.3 support.

Lukasa:
At this point, urllib3 is dropping 3.3 support. I think we should do so as well: 3.3 is long past support and never had the usage that 2.6 did, so we should probably abandon it. As a bonus reason, our CI is breaking on 3.3, so we should probably take the easy route out of that problem.


kennethreitz:
Nice :)

Don't forget to update the docs!

Lukasa:
Good call, updated.

westover:
This is also currently blocking other builds from succeeding #4234 because of dependency issues in 3.3 specifically!

Issue #4233, opened 2017-08-10
Suggestion: Improve the raise for status to include response body if there is one


westover:
Looking for more details on the HTTPError exception. The API I am working with sends back a 400 but also has a text/json body in it. It would be nice if the raise_for_status() function included the response text if there was any.

## Expected Result

That the body would be included

## Actual Result

No response body

## Reproduction Steps
N/A Feature Request! Current workaround:
```python
import requests
from requests.exceptions import HTTPError
def expanded_raise_for_status(res):
    """
    Take a "requests" response object and expand the raise_for_status method to return more helpful errors
    @param res:
    @return: None
    """
    try:
        res.raise_for_status()
    except HTTPError as e:
        if res.text:
            raise HTTPError('{} Error Message: {}'.format(str(e.message), res.text))
        else:
           raise e
    return
```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": "2.0.3"
  },
  "idna": {
    "version": ""
  },
  "implementation": {
    "name": "CPython",
    "version": "2.7.13"
  },
  "platform": {
    "release": "16.7.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "1010006f",
    "version": "16.2.0"
  },
  "requests": {
    "version": "2.18.3"
  },
  "system_ssl": {
    "version": "100020cf"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": true
}
```

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).

westover:
Added a pull with an implementation

Lukasa:
Let's close this to focus on the pull itself: it's always nicer to discuss an idea with actual code, rather than in the abstract. :smile: Closing to focus on #4234.

PR #4234, opened 2017-08-10
Feature Suggestion: Include response body in raise_for_status method

westover:
#4233 

Potential implementation of my feature suggestion

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -924,6 +924,18 @@ class Response(object):
                 reason = self.reason.decode('iso-8859-1')
         else:
             reason = self.reason
+        
+        if isinstance(self.text, bytes):
+            # We attempt to decode utf-8 first because some servers
+            # choose to localize their reason strings. If the string
+            # isn't utf-8, we fall back to iso-8859-1 for all other
+            # encodings. (See PR #3538)
+            try:
+                body_text = self.text.decode('utf-8')
+            except UnicodeDecodeError:
+                body_text = self.text.decode('iso-8859-1')
+        else:
+            body_text = self.text
 
         if 400 <= self.status_code < 500:
             http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
@@ -932,6 +944,8 @@ class Response(object):
             http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
 
         if http_error_msg:
+            if isinstance(body_text, basestring):
+                http_error_msq += u' Response Body: %s' % body_text
             raise HTTPError(http_error_msg, response=self)
 
     def close(self):
--- a/requests/models.py
+++ b/requests/models.py
@@ -945,7 +945,7 @@ class Response(object):
 
         if http_error_msg:
             if isinstance(body_text, basestring):
-                http_error_msq += u' Response Body: %s' % body_text
+                http_error_msg += u' Response Body: %s' % body_text
             raise HTTPError(http_error_msg, response=self)
 
     def close(self):
--- a/requests/models.py
+++ b/requests/models.py
@@ -909,7 +909,7 @@ class Response(object):
 
         return l
 
-    def raise_for_status(self):
+    def raise_for_status(self, include_text=False, max_text_length=500):
         """Raises stored :class:`HTTPError`, if one occurred."""
 
         http_error_msg = ''
@@ -925,18 +925,6 @@ class Response(object):
         else:
             reason = self.reason
         
-        if isinstance(self.text, bytes):
-            # We attempt to decode utf-8 first because some servers
-            # choose to localize their reason strings. If the string
-            # isn't utf-8, we fall back to iso-8859-1 for all other
-            # encodings. (See PR #3538)
-            try:
-                body_text = self.text.decode('utf-8')
-            except UnicodeDecodeError:
-                body_text = self.text.decode('iso-8859-1')
-        else:
-            body_text = self.text
-
         if 400 <= self.status_code < 500:
             http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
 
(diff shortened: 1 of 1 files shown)


requests/models.py
```diff
@@ -924,6 +924,18 @@ def raise_for_status(self):
                 reason = self.reason.decode('iso-8859-1')
         else:
             reason = self.reason
+        
+        if isinstance(self.text, bytes):
```
Lukasa: This doesn't do anything sensible. `self.text` is definitionally a unicode string, so this block will never be entered. We can safely skip it.

*However*, this is also not necessarily a safe thing to do: see my more general feedback.

requests/models.py
```diff
@@ -932,6 +944,8 @@ def raise_for_status(self):
             http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
 
         if http_error_msg:
+            if isinstance(body_text, basestring):
+                http_error_msg += u' Response Body: %s' % body_text
```
Lukasa: The body text may be *enormous*: megabytes in size. We absolutely do not want to build a string that long. If we're adding the body text to the exception message, we should add only a segment of it.

sigmavirus24:
> The response object is already attached to the exception, which means it's fairly trivial for user code to access this data in a way that is going to be safe for their application. What benefit do we obtain by placing the body into the exception message, and does it outweigh the risks?

This is exactly my question. And in all fairness, I'm fairly certain the benefit is "nothing". This will also only hurt folks who already depend on our current _decently_ formatted messages when logging response errors (i.e., so many command-line clients that would break).

I'm pretty sure what's actually happening here is that our documentation is not sufficiently clear enough to indicate that the response is attached to the exception.

I'm pretty strongly convinced this is a bad idea.

requests/models.py
westover: Just following the above pattern from self.reason

requests/models.py
westover: I agree we can definitely only take a snippet. But recall this would only be triggered IF we were dealing with a 4XX or 5XX status which shouldn't have an enormous body. Limiting the size to a few hundred characters seems like a reasonable thing for those bad cases.

westover:
As for the merits let me give you a pretty clear example:
Imagine you are wrapping an existing API because its interface is terrible or has secret keys you don't want to expose client side. The existing API gives back reasonable error information in the body when an 'illegal' value is passed but has no way to know what values it will accept for certain query parameters.
So currently requests raises an error message like this 
```
HTTPError: 400 Client Error: Bad Request for url: https://PROTECTED_TO_HIDE_SECRETS
```
Which frankly isn't enough information to pass along for troubleshooting while trying to stay true to wrapping the existing API and not adding a lot of boiler plate error handling that is ALREADY in requests "raise_for_status" method.
However like said above the wrapped API has a reasonable error message 
```
{"faultCode":"DIAGNOSTIC CODE","responseCode":"400","message":"This release is not available or does not cover the requested index."}
```
As in the original feature suggestion issue: wouldn't it be nice to see that without needing more boilerplate error handling?

And attaching the response to the exception still means I need boilerplate just to get that message.

To the statement of it breaking existing folks, sure I can buy that, so we can use an keyword in invocation of the function to select whether or not to include that text.

Lukasa:
> As in the original feature suggestion issue: wouldn't it be nice to see that without needing more boilerplate error handling?

Sure, but maybe your API returns the message in headers. Or maybe it's a very large JSON document where only one part is relevant. Or maybe it's like [GitHub's 404 page](https://github.com/dddd/dddd), a 126kB HTML document that would all of a sudden turn this exception text:

```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://github.com/dddd/dddd
```

Into

```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://github.com/dddd/dddd Response Body: <!DOCTYPE html>\n<html>\n  <head>\n    <meta http-equiv="Content-type" content="text/html; charset=utf-8">\n    <meta http-equiv="Content-Security-Policy" content="default-src \'none\'; base-uri \'self\'; connect-src \'self\'; form-action \'self\'; img-src data:; script-src \'self\'; style-src \'unsafe-inline\'">\n    <meta content="origin" name="referrer">\n    <title>Page not found &middot; GitHub</title>\n    <style type="text/css" media="screen">\n      body {\n        background-color: #f1f1f1;\n        margin: 
```

I contend that this is very much not an improvement, especially as it'd have to be rendered in 80-character wide terminals and so would end up taking multiple lines.

The general problem of this approach is that while in *some* situations it works nicely, in many others it works very badly. The negative experience from the bad cases vastly outweighs the convenience in the good cases, IMO, so that suggests that you should probably tolerate the two lines of boilerplate required to get the exception message you'd actually like to see.

westover:
@Lukasa ... Well considering that the use case I am looking at has nothing to do with terminals then sure but rather browsers themselves. As I let the error bubble back up to the client browser. 

And in response to your earlier piece. I made the behavior optional and limited so who it would "work really badly" for is somewhat elusive.

And not to be a jerk but I think the whole 80 character thing really shows a disconnect with reality. I honestly don't know a single developer who uses the 80 character terminal width on a regular basis. So IMO this whole line of reasoning is totally bunk.

Lastly, if the tagline is really about HTTP for Humans. Then you tell me what is more actionable: "400 Client Error" or "400 Client Error: ERROR TEXT". The 404 case is silly here because 404 has a specific meaning. 400 is a general 'caller screwed up' message.

Alright climbing off that soapbox. If you don't want it totes cool with me. Good day!

squeedee:
Sorry to rip this one open again, but I actually want this behavior. Or at least, I don't want to duplicate what's in raise_for_status just to add the body.. It would be nice to have `raise_for_status(include_body=True)` or `raise_for_status('additional message')`

Would either of these satisfy both set of requirements?

sigmavirus24:
No. `raise_for_status` is meant to help with control flow, not give you a complete response body when you already have the response. Further more, if you aren't handling it and some third party is (e.g., Rollbar, Sentry, etc.) you could leak sensitive information to that service which is more than likely not what 99% of our users want. Secure by default means secure in every aspect, including trying as hard as possible to prevent people from shooting themselves in the foot with our defaults.

revmischa:
It wouldn't be a default. Let users ask to include the body if they want it, but don't make it a default.

PR #4237, opened 2017-08-10
Adds import statement to front page of docs for clarity and quicker start up

robinske:


kennethreitz:
I'm down to try this — but I may remove it later.

kennethreitz:
Current example is very intentional.

PR #4240, opened 2017-08-11
Adding header name to exception

### LINKED ISSUE #4239: Add header name to InvalidHeader exception message
requests.get('http://example.com', headers={'foo': 1})
requests.exceptions.InvalidHeader: Header value 1 must be of type str or bytes, not <class 'int'>

It would be good to add the name of the bad header to make it easier
to track this down in large bodies of code. Something like:

requests.exceptions.InvalidHeader: Header foo value 1 must be of type str or bytes, not <class 'int'>

Thanks.

Summary.

## Expected Result

What you expected.

## Actual Result

What happened instead.

## Reproduction Steps

```python
import requests

```

## System Information

    $ python -m requests.help

```
<paste here>
```

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).e

mgasvoda:
Adds the name of the header to the invalid header exception raised on TypeError.

Fixes #4239

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -868,8 +868,8 @@ def check_header_validity(header):
         if not pat.match(value):
             raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
     except TypeError:
-        raise InvalidHeader("Header value %s must be of type str or bytes, "
-                            "not %s" % (value, type(value)))
+        raise InvalidHeader("Header %s value %s must be of type str or bytes, "
+                            "not %s" % (name, value, type(value)))
 
 
 def urldefragauth(url):
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -868,7 +868,7 @@ def check_header_validity(header):
         if not pat.match(value):
             raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
     except TypeError:
-        raise InvalidHeader("Value for header {%s:%s} must be of type str or " 
+        raise InvalidHeader("Value for header {%s:%s} must be of type str or "
                             "bytes, not %s" % (name, value, type(value)))
 
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -868,8 +868,8 @@ def check_header_validity(header):
         if not pat.match(value):
             raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
     except TypeError:
-        raise InvalidHeader("Header %s value %s must be of type str or bytes, "
-                            "not %s" % (name, value, type(value)))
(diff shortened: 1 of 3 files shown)


kennethreitz:
I like this.

requests/utils.py
```diff
@@ -868,8 +868,8 @@ def check_header_validity(header):
         if not pat.match(value):
             raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
     except TypeError:
-        raise InvalidHeader("Header value %s must be of type str or bytes, "
-                            "not %s" % (value, type(value)))
+        raise InvalidHeader("Header %s value %s must be of type str or bytes, "
```
nateprewitt: This may read a little clear if the name and/or value are wrapped in quotes.

Perhaps even adjusting the wording to something like "Header {name}: {value} must have a value of string or bytes, not {type}".

mgasvoda: Makes sense, I'll add that in really quick. Thanks for the feedback!

requests/utils.py
```diff
@@ -868,8 +868,8 @@ def check_header_validity(header):
         if not pat.match(value):
             raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
     except TypeError:
-        raise InvalidHeader("Header %s value %s must be of type str or bytes, "
-                            "not %s" % (name, value, type(value)))
+        raise InvalidHeader("Value for header {%s:%s} must be of type str or "
```
nateprewitt: Mind adding a space after the colon?

requests/utils.py
```diff
@@ -868,8 +868,8 @@ def check_header_validity(header):
         if not pat.match(value):
             raise InvalidHeader("Invalid return character or leading space in header: %s" % name)
     except TypeError:
-        raise InvalidHeader("Header value %s must be of type str or bytes, "
-                            "not %s" % (value, type(value)))
+        raise InvalidHeader("Value for header {%s:%s} must be of type str or "
```
nateprewitt: Mind adding a space after the colon?

mgasvoda: On it!

Lukasa:
Can we add a test for this change?

mgasvoda:
@Lukasa I modified the existing tests to include a basic assertion that the header name is included in the exception message - If you would prefer that assertion be more specific to match the expected exception statement exactly, I can definitely add that in there as well.

mgasvoda:
History updated, thanks again for your feedback!

Lukasa:
Thanks for this! :sparkles:

PR #4245, opened 2017-08-17
Fix parse_header_links on empty header

remram44:
When receiving an empty `Link:` header, one bogus entry was returned anyway. This breaks some applications.

I agree that web servers shouldn't be sending those but they do so here we are. Also fixes docstring with proper return type.

Fixes https://github.com/halcy/Mastodon.py/issues/74

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -10,6 +10,8 @@ dev
 
 **Bugfixes**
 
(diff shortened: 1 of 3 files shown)


remram44:
Sure thing!

Lukasa:
Congratulations on your first Requests contribution! ✨🍰✨

remram44:
You guys are very friendly. Thanks for the kind words. I'll try to patch the bogus server as well.

PR #4258, opened 2017-08-26
info for #4259 DO NOT MERGE

ofek:


ofek:
@kennethreitz Hey there! Could you please re-open this? There is a hanging test I want to verify in CI.

kennethreitz:
sorry!

ofek:
@kennethreitz No problem! I submitted #4259

Lukasa:
Ok, so the first note is that I don't reproduce this failure locally. That suggests that the failure is environmental: either OS-based (I'm on macOS) or something more nuanced. My first guess is that it's related to proxies, so I'm going to chase that rabbit hole.

Lukasa:
Ok if we fail out of the `mixed_case_scheme_acceptable` test the next test hangs instead. Let me see if I can escape that one too.

Lukasa:
Ok so what seems to be happening is that the HTTP server is blocking forever. Let's try to find out why.

Lukasa:
Hrm, here's a weirdness: this doesn't appear to happen when I run the tests under tox.

Lukasa:
Ohhhhh I know what this will be. Ok, one sec.

Lukasa:
Yup, got it. Ok, will open a PR that fixes this problem.

ofek:
Between this and that macOS memory allocation thing, you're the King of Bug Hunting!

Lukasa:
Heh, thanks!

PR #4261, opened 2017-08-27
Pin to httpbin 0.5.0.

### LINKED ISSUE #4259: tests are broken
Specifically, this hangs https://github.com/requests/requests/blob/5dcbdb50fc9652cbeffeec22a21d1196e938bfca/tests/test_requests.py#L151

It can be seen in #4258 https://travis-ci.org/requests/requests/jobs/268764155

This occurs locally for me as well.

Lukasa:
The release of httpbin 0.6.0 seems to have broken pytest-httpbin in a surprising way. We want to pin that version out, and probably also alert @kevin1024 about the problem.

Resolves #4259.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,3 +14,4 @@ docutils
 flake8
 tox
 detox
+httpbin==0.5.0


kevin1024:
OK, this is a really crazy rabbit hole - I think @Lukasa is actually more up what's going on than I.

I think the problem is caused by this commit to httpbin by @kennethreitz:

https://github.com/kennethreitz/httpbin/commit/2e94212886e4c9531b8ba7f55f4fb0b8134e10ac

which is an attempt to fix kennethreitz/httpbin#340

BUT I think it only fixes it on Gunicorn?  And pytest-httpbin uses werkzeug which has chunked encoding issues so it ends up causing problems: https://github.com/pallets/werkzeug/issues/1149

kennethreitz:
@kevin1024 we can remove that — I don't think it fixed anything.

kevin1024:
@kennethreitz sweet, I think this problem should magically disappear in that case 🌟

PR #4262, opened 2017-08-27
[do not merge] Try to fix hanging tests

kevin1024:
The tests are hanging sometimes, see #4259 and #4261.  I wonder if it would work to unpin both packages since I think I fixed a bug with pytest-httpbin hanging awhile back.


kevin1024:
welp, that doesn't appear to be working.

PR #4277, opened 2017-09-05
Update allow_redirects docstrings defaults

geokala:


requests/api.py
```diff
@@ -33,7 +33,7 @@ def request(method, url, **kwargs):
         before giving up, as a float, or a :ref:`(connect timeout, read
         timeout) <timeouts>` tuple.
     :type timeout: float or tuple
-    :param allow_redirects: (optional) Boolean. Enable/disable GET/OPTIONS/POST/PUT/PATCH/DELETE/HEAD redirection. Defaults to ``True``.
+    :param allow_redirects: (optional) Boolean. Enable/disable GET/OPTIONS/POST/PUT/PATCH/DELETE/HEAD redirection. Defaults to ``False`` for HEAD, ``True`` for all others.
     :type allow_redirects: bool
```
Lukasa, line 37: This is not correct. `requests.head` defaults this to `False`, but all other uses (including `requests.request('HEAD', ...)` default it to `True`.

requests/sessions.py
```diff
@@ -465,7 +465,8 @@ def request(self, method, url,
             data before giving up, as a float, or a :ref:`(connect timeout,
             read timeout) <timeouts>` tuple.
         :type timeout: float or tuple
-        :param allow_redirects: (optional) Set to True by default.
+        :param allow_redirects: (optional) Set to False by default for HEAD,
+            True by default for all others.
```
Lukasa, line 469: Same note here, this docstring does not require updating.

requests/api.py
geokala: OK, understood- and sorry for the delay, busy week. I'll look to get this updated before the end of the weekend.

Lukasa: No worries: we're all busy people!

Unit03:
This is still relevant imho, do you think we can reopen this PR, @kenneth-reitz @Lukasa ?

geokala:
Well, that's a blast from the past. I'm happy to fix it- does it just require the removal of that second, incorrectly updated, docstring?

Unit03:
@geokala I suppose it is. :) If it's of no relevance for you at this point, I will be equally happy to take this over.

henribru:
Just ran into this myself, would be great if this could be merged so as to not cause confusion.

geokala:
@Unit03 Happy for you to take it over if you're happy to.

Unit03:
@geokala I will do just that, then, thank you for your input. :)

@Lukasa I've addressed your remarks and opened a new PR (to not take more of @geokala's time): #5127 - but I'd be ok to contribute this change in some other way if you think it's appropriate.

Issue #4214, opened 2017-08-01
Returning cookies from Adapters


Singletoned:
I've been trying to create a subclass of `BaseAdapter` for Requests that supports WSGI applications, (preferably based on the requests-wsgi-adapter package, or from scratch if need be).

The sticking point has been cookies and session handling.  From reading the Requests code, my understanding is that cookies are automatically extracted from the reponse object returned from the adapter's `send` method.  In order for the cookies to be automatically extracted the response needs to have a `r.raw._original_response.msg` chain of attributes and this should return an object with an `info` method that should return an `email.message.Message` instance (largely because of  `cookies.py:extract_cookies_to_jar`, but also because of how `http.cookiejar` behaves).

This seems like a needlessly complicated series of steps to go through, particularly as I'm not currently clear on how to correctly convert my response body to an `email.message.Message` (and a quick scan of the Python documentation makes this look non-trivial).

Have I misunderstood?  Are there any helper functions for performing these steps?

Would you consider a patch to `extract_cookies_to_jar` that also looks, for example, for a `_cookies_dict` property on the response, at the point that it currently silently returns if there is no `_original_response` attribute?

Alternatively, would you consider supporting something like https://pypi.python.org/pypi/cookies/ with a friendlier interface (kind of "Requests for cookies")?  I haven't dug into it a lot, but it looks relatively easy to add optional support for it (which I'd be happy to contribute).

Lukasa:
> Have I misunderstood? Are there any helper functions for performing these steps?

You have not: there are no helper functions for performing these steps.

> Alternatively, would you consider supporting something like https://pypi.python.org/pypi/cookies/ with a friendlier interface (kind of "Requests for cookies")?

Nope. =) Note the README on that project:

> This doesn’t compete with the cookielib (http.cookiejar) module in the Python standard library, which is specifically for implementing cookie storage and similar behavior in an HTTP client such as a browser. Things cookielib does that this doesn’t:
>
> - Write to or read from browsers’ cookie stores or other proprietary formats for storing cookie data in files
> - Handle the browser/client logic like deciding which cookies to send or discard, etc.

`cookies` is not an appropriate replacement for `cookielib`. We need to continue to use `cookielib` for the moment.
> Would you consider a patch to `extract_cookies_to_jar` that also looks, for example, for a `_cookies_dict` property on the response, at the point that it currently silently returns if there is no `_original_response` attribute?

[...]

Issue #3957, opened 2017-04-07
comma in case: join multiple Set-Cookie Headers


jake491:
Many web-apps send cookie in multiple Set-Cookie headers, one header for one cookie. Requests joins this headers in one big header separate by comma. 

From same webapp to requests.get
```
Set-Cookie: ASP.NET_SessionId=token1; path=/; HttpOnly
Set-Cookie: Cookie1=token2; path=/ecp
Set-Cookie: X-BEResource=WIN-RBFR0BDA7V7.testlab.net~1; path=/ecp/15.0.516.30; HttpOnly
Set-Cookie: X-BackEndCookie=token3; expires=Mon, 03-Apr-2017 12:25:07 GMT; path=/ecp; HttpOnly
```
In requests:
resp.headers  this look like
```
Set-Cookie: ASP.NET_SessionId=token1; path=/; HttpOnly,    
  Cookie1=token2; path=/ecp, 
  X-BEResource=WIN-RBFR0BDA7V7.testlab.net~1; path=/ecp/15.0.516.30; HttpOnly, 
  X-BackEndCookie=token3; expires=Thu, 06-Apr-2017 08:27:22 GMT; path=/ecp; HttpOnly
```
Firefox, IE, Chrome invalid interpreted this cookies because separater is comma but in  field EXPIRES  comma separate day and date. 

Principal problem in modification traffic. In my opinion then server send multiple headers lib must return multiple headers

Lukasa:
Requests makes no promises about leaving headers in the form they are received from the server. Right now, I highly recommend you use the cookiejar Requests uses to store cookies rather than extracting them from the headers: we correctly insert them in there.

Longer term, we should probably add a fix to exclude the `Set-Cookie` header from joining in this way.

daniel-kun:
This has bitten us and cost us quite a few hours. I'd like to vote to leave `Set-Cookie` intact. Why does requests fiddle with the headers anyways? Is there some good reason for it? IMHO it only complicates things for everybody. What am I missing?
Thanks for your good work!

Ousret:
Hi, there is a way to parse headers correctly with https://github.com/Ousret/kiss-headers `requests` won't be able to change their representation for headers any time soon.

apeschar:
In order to access the unmerged header lines, use the underlying `urllib3` response object ([HTTPResponse](https://urllib3.readthedocs.io/en/latest/reference/urllib3.response.html#urllib3.response.HTTPResponse)):

~~~py
[...]

PR #4282, opened 2017-09-10
New HTTPHeaderDict class for response headers with multiple header value support

the-allanc:
I personally wanted something like this while writing [my own Requests-related library](https://github.com/the-allanc/pyriform), but I can see other people also have wanted something like this - #3957 and #4214 both mention it in terms of handling `Set-Cookie` headers.

I've written an subclass implementation of `CaseInsensitiveDict` which is compatible with all the current usage (according to the tests), but adds new methods for clients that want to deal directly with unjoined headers. I've also added quite a number of tests and made slight changes to the documentation in relation to it.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,10 @@ dev
 
 **Improvements**
 
+- Use `HTTPHeaderDict` for response headers, allowing easier access to
+  individual values when multiple response headers are sent using the same
+  header name.
+
 **Bugfixes**
 
 - Parsing empty ``Link`` headers with ``parse_header_links()`` no longer return one bogus entry
@@ -52,7 +56,7 @@ dev
 
 **Bugfixes**
 
-- Fix an error in the packaging whereby the *.whl contained incorrect data that
+- Fix an error in the packaging whereby the \*.whl contained incorrect data that
   regressed the fix in v2.17.3.
 
 2.18.0 (2017-06-14)
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -416,6 +416,10 @@ represented in the dictionary within a single mapping, as per
     of the message, by appending each subsequent field value to the combined
     field value in order, separated by a comma.
 
+If you do need to access each individual value sent with the same header, then
+you can use the ``multiget`` method to get a sequence of all the values returned
+for a particular header.
+
 Cookies
 -------
 
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -417,7 +417,7 @@ represented in the dictionary within a single mapping, as per
     field value in order, separated by a comma.
 
 If you do need to access each individual value sent with the same header, then
-you can use the ``multiget`` method to get a sequence of all the values returned
+you can use the ``getlist`` method to get a sequence of all the values returned
 for a particular header.
 
 Cookies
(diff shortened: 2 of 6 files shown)


kennethreitz:
I generally like this. 

This should be a PR against the `proposed/3.0.0` branch, though, not master.

kennethreitz:
I also wouldn't be opposed to the name `CaseInsensitiveMultiDict`

kennethreitz:
I haven't checked, but I'd like this API to be compatible with Flask/Werkzeug's MultiDict, if possible.

kennethreitz:
In general though, this looks real solid! ✨🍰✨

kennethreitz:
Good job with the tests too :cookie:

kennethreitz:
You get a cookie.

kennethreitz:
I also really like that this is a subclass of `CaseInsensitiveDict`.

the-allanc:
With regards to MultiDict in werkzeug, it uses `getlist` and `setlist`. HTTPHeaderDict in urllib3 also uses `getlist` (it has no `setlist` equivalent). I was going to use the same naming convention, except that I'm storing and returning tuples.

I did have concerns about the lists being modified outside of HTTPHeaderDict (which tries to enforce that only string types are present). So it's either a case of::
- Leaving the names as they are.
- Using "list" names even though we're returning tuples.
- Rename the methods and return lists (either copies or the underlying sequence).

the-allanc:
Renaming `HTTPHeaderDict` to `MultiCaseInsensitiveDict` is a bit more contentious - because it's not a generalised class (it's intended for string values only).

The main intent was to keep the exact same behaviour that we had in requests for response header processing, but to handle the concatenation of values ourselves (rather than letting urllib3 to do it for us). The concatenation of values is specific to HTTP headers (comma separation).

I could refactor part of the code and extract the `Multi...` part of the class into a separate class, and then inherit from that. We would need to decide what happens for general item access (I note that `MultiDict` returns just the first item).

kennethreitz:
I think what you have is good, then, unless we cast to lists.

kennethreitz:
Which seems silly to me.

kennethreitz:
I just wanted to provide a very familiar API to those who have used similar APIs before.

kennethreitz:
I've just always liked the name `CaseInsensitiveMultiDict`, and wish we had it. This is good though.

the-allanc:
OK, I've renamed the methods to `getlist` and `setlist`, and they now return lists, so that's more consistent with other implementations. Both urllib3 and werkzeug provide copies of the sequences, so we also do the same (we still use tuples under the hood, which makes me happy).

I was going to try and extract some of the code into a `CaseInsensitiveMultiDict` class, but there's too much nuance in trying to determine the best behaviour for such a generalised class (especially for the `extend` method, which tries to distinguish between items and sequences - how should it behave if you're using the dictionary to store sequences of sequences?)

I've made sure that this class is entirely backwards compatible with what we were using before - I wrote some of the tests using `CaseInsensitiveDict` and they still worked when I switched to `HTTPHeaderDict` for responses. Are you sure that you still want it against the `proposed/3.0.0` branch?

kennethreitz:
@the-allanc definitely :)

P.S. good work!

the-allanc:
In that case, I'm closing this pull request in favour of #4290.

PR #4283, opened 2017-09-10
Pep8ify

Gnouc:
PEP8 forbids inventing __double_leading_and_trailing_underscore__
name, so don't do it.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -104,7 +104,7 @@ class HTTPAdapter(BaseAdapter):
       >>> a = requests.adapters.HTTPAdapter(max_retries=3)
       >>> s.mount('http://', a)
     """
-    __attrs__ = ['max_retries', 'config', '_pool_connections', '_pool_maxsize',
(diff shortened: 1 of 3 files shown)


Lukasa:
The first rule of PEP8 is that a foolish consistency is the hobgoblin of little minds. In this case, the double-underscore is safe and reasonable. :smile:

cuonglm:
@Lukasa Why is it safe since when we don't know when Python core add such thing?

It also makes someone new to Python think that `__attrs__` is a magic method when it doesn't. `requests` code base is great and recommended to read in many sources, IMHO, it's a poor choice for chosing such name.

Lukasa:
Because there's no need for us to be predictive. Requests is one of the most widely-used Python packages: we'll know when such a thing is added, if for no other reason than because most of the Python core team uses Requests and will spot the problem if and when it's introduced. So for now we're in a safe place.

kennethreitz:
e.g. people probably rely on this — we don't want to change anything, especially for pedantic reasons, unless we have a good reason to.

PR #4290, opened 2017-09-12
New HTTPHeaderDict class for response headers with multiple header value support

the-allanc:
Same changes as my other pull request #4282, but this against the proposed/3.0.0 branch.

### DIFF
--- a/3.0-HISTORY.rst
+++ b/3.0-HISTORY.rst
@@ -66,5 +66,9 @@
 
 - ``Response.raise_for_status()`` now returns the response object for good responses
 
+- Use ``HTTPHeaderDict`` for response headers, allowing easier access to
+  individual values when multiple response headers are sent using the same
+  header name.
+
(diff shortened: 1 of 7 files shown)


the-allanc:
The CI builds are failing on tests which were failing before my changes were pushed (honest!)

kennethreitz:
We know :)

@Lukasa what do you think of this? I'm very :+1: personally.

kennethreitz:
I think this will be a major improvement to Requests.

ml31415:
Why reinventing the wheel? `urllib3` has a multiheader-ready headers implementation for ages. Stable, fast, reviewed and well tested. If I remember correctly, it was already suggested a while ago to reuse it also in requests, so that it could be avoided to copy and re-read all the headers, which would also be a considerable speed improvement on creating a Response object.

kennethreitz:
Merged into '3.0' branch. ✨🍰✨

kennethreitz:
Pending further modifications.

PR #4302, opened 2017-09-22
Update no_proxy logic to handle user:password@ in the url.

bobh66:
Also updated tests to include user@ and user@password: scenarios

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -649,7 +649,7 @@ def should_bypass_proxies(url, no_proxy):
         # We need to see if we match the end of the netloc, accounting for a
         # 'username[:password]@' at the beginning and a ':port' at the end.
 
-        host_or_ip = netloc.split('@')[-1].split(':')[0]
+        host_or_ip = netloc.rsplit('@')[-1].split(':')[0]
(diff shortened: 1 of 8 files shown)


requests/utils.py
```diff
@@ -642,25 +642,26 @@ def should_bypass_proxies(url, no_proxy):
...
         no_proxy = (
             host for host in no_proxy.replace(' ', '').split(',') if host
         )
 
-        ip = netloc.split(':')[0]
-        if is_ipv4_address(ip):
+        # We need to see if we match the end of the netloc, accounting for a
+        # 'username[:password]@' at the beginning and a ':port' at the end.
+
+        host_or_ip = netloc.split('@')[-1].split(':')[0]
```
sigmavirus24: People are rather inconsistent as to whether or not they properly encode their password. If they have an `@` in their password, and they've not encoded it appropriately, this will cause issues. It would be better to `rsplit` here as a result. It'll cause less teeth-gnashing.

nateprewitt:
It looks like this ended up getting resolved in #4427, so I'm going to close this out. Thanks for the PR @bobh66!

PR #4305, opened 2017-09-23
Fix the invalid restructured text in HISTORY file

sigmavirus24:
This was breaking the display of our long_description on PyPI.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -52,8 +52,8 @@ dev
 
 **Bugfixes**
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
It may be worth, in the future, moving the changelog out of our long description. Flake8 has done this to great effect and made the Changelog far more navigable.

sigmavirus24:
This is presently failing because of https://github.com/kennethreitz/httpbin/issues/380 which I'm working on a fix for.

PR #4307, opened 2017-09-25
add library authors instructions for depending on requests

graingert:


Lukasa:
Why are we adding this? It’s a bit of a surprise to find we need to document how versions work.

sigmavirus24:
I agree with @Lukasa. It's not our responsibility to teach users to responsibly use their dependencies. It's also not in our best interest to tell users to cap themselves at version 3.

PR #4316, opened 2017-09-30
Avoid importing platform

methane:
platform module is relatively large: it takes about 5ms to import.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -8,7 +8,7 @@ This module provides a Session object to manage and persist settings across
 requests (cookies, auth, proxies).
 """
 import os
-import platform
+import sys
 import time
 from collections import Mapping
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Is `sys.platform` always equal to `win32` even on 64bit windows installations?

sethmlarson:
Might wanna do `sys.platform.startswith('win')`? I've seen that in another library, can't recall which.

methane:
It is always 'win32'.
https://docs.python.org/3/library/sys.html#sys.platform

https://github.com/python/cpython/blob/d6bb65f378e34fe0c11fdb39588357ecf22964eb/PC/pyconfig.h#L297-L301

sigmavirus24:
Once https://github.com/kennethreitz/httpbin/pull/392 merges and Kevin releases a pytest-httpbin package, the tests here should start passing.

nateprewitt:
@methane would you mind rebasing this onto the current master branch?

methane:
done

sethmlarson:
:tada: LGTM

PR #4326, opened 2017-10-05
Remove redundant Python 2.6 stuff

hugovk:
Python 2.6 support was dropped in https://github.com/requests/requests/pull/4118.

This removes some other redundant 2.6 code. 

Also mention Python 3.7 as officially supported, to reflect the README.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -465,8 +465,6 @@ class HTTPAdapter(BaseAdapter):
                     low_conn.send(b'0\r\n\r\n')
 
                     # Receive the response from the server
-                    # For Python 2.7+ versions, use buffering of HTTP
-                    # responses
                     r = low_conn.getresponse(buffering=True)
 
                     resp = HTTPResponse.from_httplib(
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -465,7 +465,11 @@ class HTTPAdapter(BaseAdapter):
                     low_conn.send(b'0\r\n\r\n')
 
                     # Receive the response from the server
-                    r = low_conn.getresponse(buffering=True)
+                    try:
+                        r = low_conn.getresponse(buffering=True)
+                    except TypeError:
+                        # For compatibility with Python 3.6 versions and up
+                        r = low_conn.getresponse()
 
                     resp = HTTPResponse.from_httplib(
                         r,
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -465,13 +465,9 @@ class HTTPAdapter(BaseAdapter):
                     low_conn.send(b'0\r\n\r\n')
 
                     # Receive the response from the server
-                    try:
-                        # For Python 2.7+ versions, use buffering of HTTP
-                        # responses
-                        r = low_conn.getresponse(buffering=True)
-                    except TypeError:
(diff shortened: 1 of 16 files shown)


requests/adapters.py
```diff
@@ -466,11 +466,9 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
 
                     # Receive the response from the server
                     try:
-                        # For Python 2.7+ versions, use buffering of HTTP
-                        # responses
                         r = low_conn.getresponse(buffering=True)
                     except TypeError:
-                        # For compatibility with Python 2.6 versions and back
+                        # For compatibility with Python 3.6 versions and up
```
sethmlarson, line 471: Can't this section be changed to remove the `try-except`?

requests/help.py
```diff
@@ -89,12 +89,6 @@ def info():
         'version': getattr(idna, '__version__', ''),
     }
 
-    # OPENSSL_VERSION_NUMBER doesn't exist in the Python 2.6 ssl module.
-    system_ssl = getattr(ssl, 'OPENSSL_VERSION_NUMBER', None)
-    system_ssl_info = {
-        'version': '%x' % system_ssl if system_ssl is not None else ''
-    }
```
sethmlarson: These values are being used, can't just remove them. I'd keep them as-is just in case.

hugovk: How about this?
```python
    system_ssl_info = {
        'version': getattr(ssl, 'OPENSSL_VERSION_NUMBER', '')
    }
```

requests/adapters.py
hugovk: What do you suggest?

`r = low_conn.getresponse(buffering=True)` gives `TypeError: getresponse() got an unexpected keyword argument 'buffering'` on Python 3.6.

sethmlarson: Hm, TIL. Well I guess just leave it as is then. I'll rescind my review.

sethmlarson:
Looks like some failures regarding cookies, could you check those out?

hugovk:
Similar cookie failures are happening on latest master:

https://travis-ci.org/requests/requests/builds/279556831

sethmlarson:
You're right, those will have to be fixed in a different PR. I'm :+1: from my brief look at these changes.

nateprewitt:
Hey @hugovk, thanks for taking the time to put this PR together! Unfortunately it's pointed against our master branch rather than our proposed/3.0.0 branch where the original #4118 was made. It appears some of this is redundant but there are also new changes. If you'd like to move those over to a PR against the correct branch, we'd be happy to review them. Thanks again for you work!

hugovk:
@nateprewitt Thanks, I've created https://github.com/requests/requests/pull/4328 against proposed/3.0.0.

But don't you want this stuff in master as well? Or do you have another process for merging 3.0.0 back into master at some point? 

I guess you don't want to support 2.6 in releases after 3.0.0 either :)

nateprewitt:
@hugovk, yes, we'll be merging the 3.0.0 branch back into master when we're ready to release. Until that time though, we're not planning on removing 2.6 support so these need to stay out of the master branch to allow future releases in the 2.x series.

PR #4327, opened 2017-10-05
removed python2 unicode literals from documentation examples

miguelgrinberg:
The documentation has a bunch of examples showing unicode literals in Python 2's `u'foo'` syntax, including one in the front page right below the note encouraging users to switch to Python 3.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -44,9 +44,9 @@ consumption.
     >>> r.encoding
     'utf-8'
     >>> r.text
-    u'{"type":"User"...'
+    '{"type":"User"...'
     >>> r.json()
-    {u'private_gists': 419, u'total_private_repos': 77, ...}
+    {'private_gists': 419, 'total_private_repos': 77, ...}
 
 See `similar code, sans Requests <https://gist.github.com/973705>`_.
 
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -679,12 +679,12 @@ So, GitHub returns JSON. That's great, we can use the :meth:`r.json
     >>> commit_data = r.json()
 
     >>> print(commit_data.keys())
-    [u'committer', u'author', u'url', u'tree', u'sha', u'parents', u'message']
+    ['committer', 'author', 'url', 'tree', 'sha', 'parents', 'message']
 
-    >>> print(commit_data[u'committer'])
-    {u'date': u'2012-05-10T11:10:50-07:00', u'email': u'me@kennethreitz.com', u'name': u'Kenneth Reitz'}
+    >>> print(commit_data['committer'])
+    {'date': '2012-05-10T11:10:50-07:00', 'email': 'me@kennethreitz.com', 'name': 'Kenneth Reitz'}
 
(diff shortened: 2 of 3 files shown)


sigmavirus24:
Given the currently supported versions of Python 3 (e.g., Python 3.4+), using `u''` is a cross-version compatible way of ensuring one always has text when dealing with native strings. I don't think we want to remove this.

Note: All versions of Python 3 in which unicode-literals were invalid syntax have been EOL'd. That said, the documentation examples that do

```
print(u'foo')
u'foo'
```

Are wrong on *some* of the versions of Python that Requests supports. We can't easily document both though.

---

Complete aside: Hi @miguelgrinberg! Hope all is well!

miguelgrinberg:
Hi @sigmavirus24!

Okay. Just looks weird to see the `u`s when you are used to working with Py3, specially in a highly prominent example in the front page of the docs. The PEP that introduced back the `u` was made to simplify the porting of code from Py2, not to encourage people to use it on brand new code.

Would you want to keep this PR in the back burner until 2020 and then merge it, when the `u`s become completely unnecessary?

sigmavirus24:
> The PEP that introduced back the u was made to simplify the porting of code from Py2, not to encourage people to use it on brand new code.

We're well aware of that. However, for those of us stuck writing code compatible on Python 2 and Python 3, `u` is one of the better solutions to ensuring we have text everywhere.

> Would you want to keep this PR in the back burner until 2020 and then merge it, when the `u` s become completely unnecessary?

Not particularly. We periodically go through the list and close old and stale pull requests so since there's no plan to remove this right now, it doesn't make sense to try to keep this PR alive until then. I'd be happy to merge it when we drop 2.7 but we'll need you to resubmit it then.

Thanks @miguelgrinberg

miguelgrinberg:
@sigmavirus24 I agree with the use of `u` for writing compatible code inside requests, but disagree with favoring a Python 2 idiom that is on its way out in public facing documentation and examples.

I respect your decision, but to paraphrase @kennethreitz, as a "person of excellent taste", I don't want to see those `u` prefixes any more than I have to. :)

sigmavirus24:
>  I agree with the use of u for writing compatible code inside requests

I'm talking about writing code that *uses* requests on Python 2 and 3 that is compatible with both and behaves exactly the same. Which is in better taste, code using requests on 2 and 3 that does

```py
my_maybe_text = 'foo bar bogus'
if isinstance(my_maybe_text, bytes):
    my_text = my_maybe_text.decode('utf-8')
else:
    my_text = my_maybe_text
```

or 

```py
my_text = u'foo bar bogus'
```

That's really the reason to use both. While we handle both cases well, they may be handled slightly differently from the user's perspective and advising them to be explicit and clear to themselves and others doesn't seem like the worst thing ever. It's like providing examples using `b'some bytes here'` that work the same across 2 and 3. At the moment it's a necessary evil.

kennethreitz:
It is best practice to use the unicodecode literals at this time. 

This will change, in about a year, or maybe sooner.

PR #4329, opened 2017-10-06
Python 3.7 is supported

hugovk:
The README says "Requests officially supports Python 2.6–2.7 & 3.3–3.7, and runs great on PyPy." so update the version in some other places too.


hugovk:
Should we instead remove 3.7 from the README and the commented out 3.7-dev from .travis.yml?

Alternatively: the commit message for the commenting out was to speed up the build: https://github.com/requests/requests/commit/2758b8b0f029abb9166eb1f124dcabe1bd35e949

Travis gives up to five parallel build jobs, but only four are taken per build. Should 3.7-dev be re-enabled?

Having said that, there's still 8 months until 3.7 release, so I think the best thing to do is not claim support yet, but if possible test it. See https://snarky.ca/how-to-use-your-project-travis-to-help-test-python-itself/

hugovk:
See https://github.com/requests/requests/pull/4341 to remove claimed support from the README.

PR #4331, opened 2017-10-10
Fix docs for data param in api

rymdhund:
The data param also accepts a list of tuples. It says so in the docs for request() but not for post(), put() and patch()


kennethreitz:
this is unnecessary information.

Issue #4309, opened 2017-09-28
Warn about possible slowness when using an older version of pyOpenSSL


PrimordialHelios:
[This](https://github.com/requests/requests/issues/4278#issuecomment-329516221) comment on #4278 suggests warning about possible slowness for older versions of pyOpenSSL.

I'd love to work on this, I'm just not sure which versions to warn about and would love input from some developers more experienced with the library.

nateprewitt:
So I think @Lukasa or @sigmavirus24 may have a more definitive answer to this, but it looks like `PyOpenSSL 16.0.0` was when the `Cryptography>=1.0` requirement was put in place. From my understanding that's when the speed ups were implemented.

We already require `Cryptography>=1.3.4` in both Requests and urllib3 though, so users entering this state are installing the PyOpenSSL requirement from another package. If we implement a warning, I think we should first check that PyOpenSSL is installed, and then if cryptography's version is less than 1.3.4.

ghost:
>so users entering this state are installing the PyOpenSSL requirement from another package. If we implement a warning, I think we should first check that PyOpenSSL is installed

Just to be clear, do you mean check that the OpenSSL module is available, or check specifically that the module is pyOpenSSL?

nateprewitt:
@PrimordialHelios, I *believe* that PyOpenSSL is sufficient (and much easier to check). We don't import or use cryptography directly in Requests, so the urllib3 pyopenssl contrib module should be our only point of contact.

PrimordialHelios:
@nateprewitt How would you suggest checking the Cryptography version without importing it directly? The only way I've been able to find is to `import .help` and check the version from `info()['cryptography']`, but that's a bit sub-optimal. Did I overlook anything?

nateprewitt:
This should be resolved with #4332 merged. Thanks again @PrimordialHelios!

PR #4332, opened 2017-10-10
Warn about possible slowness on older versions of pyOpenSSL

PrimordialHelios:
Give a warning when `cryptography < 1.3.4` and `pyOpenSSL < 16.0.0` as discussed in #4309

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -71,6 +71,17 @@ def check_compatibility(urllib3_version, chardet_version):
     assert patch >= 2
 
 
+def _check_cryptography(cryptography_version):
+    # cryptography < 1.3.4
+    try:
+        cryptography_version = list(map(int, cryptography_version.split('.')))
+    except ValueError:
+        return
+
+    if cryptography_version < [1, 3, 4]:
+        warning = 'Old version of cryptography ({0}) may cause slowdown.'.format(cryptography_version)
+        warnings.warn(warning, RequestsDependencyWarning)
+
 # Check imported dependencies for compatibility.
 try:
     check_compatibility(urllib3.__version__, chardet.__version__)
@@ -83,6 +94,10 @@ except (AssertionError, ValueError):
 try:
     from urllib3.contrib import pyopenssl
     pyopenssl.inject_into_urllib3()
+
+    # Check cryptography version
+    from cryptography import __version__ as cryptography_version
+    _check_cryptography(cryptography_version)
 except ImportError:
     pass
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -178,3 +178,4 @@ Patches and Suggestions
 - Ryan Pineo (`@ryanpineo <https://github.com/ryanpineo>`_)
 - Ed Morley (`@edmorley <https://github.com/edmorley>`_)
 - Matt Liu <liumatt@gmail.com> (`@mlcrazy <https://github.com/mlcrazy>`_)
+- Taylor Hoff <primdevs@protonmail.com> (`@PrimordialHelios <https://github.com/PrimordialHelios>`_)
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,8 @@ dev
 
 **Improvements**
 
+- Warn user about possible slowdown when using cryptography version < 1.3.4
+
 **Bugfixes**
 
 - Parsing empty ``Link`` headers with ``parse_header_links()`` no longer return one bogus entry


requests/__init__.py
```diff
@@ -86,6 +87,25 @@ def check_compatibility(urllib3_version, chardet_version):
 except ImportError:
     pass
 
+# Check for older version of pyOpenSSL
+def check_pyopenssl(cryptography_version, pyopenssl_version):
```
sigmavirus24: This should be a private function. We really shouldn't be exposing this as part of our API.

requests/__init__.py
```diff
@@ -86,6 +87,25 @@ def check_compatibility(urllib3_version, chardet_version):
...
+    # Only applies to pyOpenSSL
+    if sys.modules['OpenSSL'].__title__ == 'pyOpenSSL':
```
sigmavirus24: I feel like there's a bifurcation. Below we check that `'OpenSSL' in sys.modules` why not do this check there too? Further, is there another `OpenSSL` module that we might import that isn't `pyOpenSSL`? What is this check protecting against?

requests/__init__.py
```diff
@@ -86,6 +87,25 @@ def check_compatibility(urllib3_version, chardet_version):
...
+        # Warning only applies for Cryptography < 1.3.4
+        if list(map(int, cryptography_version.split('.'))) < [1, 3, 4]:
+            # Warning only applies for pyOpenSSL < 16.0.0
+            assert list(map(int, cryptography_version.split('.'))) < [16, 0, 0]
```
sigmavirus24: This will always be false since no relesaed version of cryptography is `16.0.0` or higher. I think this was meant to be `pyopenssl_version`

requests/__init__.py
```diff
@@ -86,6 +87,25 @@ def check_compatibility(urllib3_version, chardet_version):
...
+
+# Check cryptography version
+if 'OpenSSL' in sys.modules:
+    try:
+        cryptography_version = sys.modules['cryptography'].__version__
+        pyopenssl_version = sys.modules['OpenSSL'].__version__
+        check_pyopenssl(cryptography_version, pyopenssl_version)
+    except (AssertionError, ValueError):
+        warning = 'Old version of pyOpenSSL ({0}) may cause slowdown.'.format(pyopenssl_version)
```
sigmavirus24: Is the issue for old versions of `pyOpenSSL` or `cryptography`.

requests/__init__.py
ghost: `check_compatibility()` defined just above `check_pyopenssl` wasn't private, so I thought it would be best to expose this too for the sake of consistency.

requests/__init__.py
ghost: I misread a reply to the #4309 and assumed multiple packages could provide an `OpenSSL` module, but it looks like I was wrong. I'll remove that `__title__` check.

requests/__init__.py
ghost: Yeah, it's supposed to be `pyopenssl_version`, I can't believe I missed that typo!

requests/__init__.py
ghost: Old versions of `pyOpenSSL`. @nateprewitt recommended checking `Cyptography`'s version before `pyOpenSSL` in [this comment](https://github.com/requests/requests/issues/4309#issuecomment-332942003).

nateprewitt: To clarify, I believe the comment says we want to verify `pyOpenSSL` is installed, and if it is check that `cryptography` is at least 1.3.4 which is our minimum requirement. We haven't done any testing to see how far back you need to go with `cryptography` for the slowdown to occur, but 1.3.4 should be sufficient.

requests/__init__.py
```diff
@@ -86,6 +87,25 @@ def check_compatibility(urllib3_version, chardet_version):
...
+def check_pyopenssl(cryptography_version, pyopenssl_version):
+    # Only applies to pyOpenSSL
+    if sys.modules['OpenSSL'].__title__ == 'pyOpenSSL':
+        # Warning only applies for Cryptography < 1.3.4
+        if list(map(int, cryptography_version.split('.'))) < [1, 3, 4]:
+            # Warning only applies for pyOpenSSL < 16.0.0
+            assert list(map(int, cryptography_version.split('.'))) < [16, 0, 0]
+
+# Check cryptography version
+if 'OpenSSL' in sys.modules:
```
nateprewitt: I also don't think we actually need the check to be this complicated. We're already checking for `pyOpenSSL` with our `inject_pyopenssl_into_urllib3` above. A simple helper function that checks cryptography's version meets or exceeds our threshold (supposedly 1.3.4), and otherwise displays a warning, should be sufficient.

requests/__init__.py
```diff
@@ -70,6 +71,12 @@ def check_compatibility(urllib3_version, chardet_version):
     assert minor < 1
     assert patch >= 2
 
+# Check for older version of pyOpenSSL
+def __check_pyopenssl(pyopenssl_version, cryptography_version):
```
nateprewitt: This should really be _check_cryptography because that's what we're concerned about.

requests/__init__.py
```diff
@@ -70,6 +71,12 @@ def check_compatibility(urllib3_version, chardet_version):
...
+    # Warning only applies for Cryptography < 1.3.4
+    if list(map(int, cryptography_version.split('.'))) < [1, 3, 4]:
+        # Warning only applies for pyOpenSSL < 16.0.0
```
nateprewitt: So we don't really care about the version of pyOpenSSL, the only reason it's significant is 16.0 is where a version constraint for cryptography > 1.0 is put in place. Let's remove this check and just launch the warning if cryptography doesn't meet our requirement. That's the source of the slowdown.

requests/__init__.py
```diff
@@ -83,8 +90,16 @@ def check_compatibility(urllib3_version, chardet_version):
 try:
     from urllib3.contrib import pyopenssl
     pyopenssl.inject_into_urllib3()
+
+    # Check pyOpenSSL version
+    cryptography_version = sys.modules['cryptography'].__version__
```
nateprewitt: No need to pass both of these. We don't actually care about the pyOpenSSL version as I stated above, just that it's being imported for urllib3.

We can do the check with something like:
```
from cryptography import __version__ as cryptography_version
```

This allows us to have to only worry about the import error.

requests/__init__.py
```diff
@@ -83,8 +90,16 @@ def check_compatibility(urllib3_version, chardet_version):
...
+    pyopenssl_version = sys.modules['OpenSSL'].__version__
+    __check_pyopenssl(pyopenssl_version, cryptography_version)
 except ImportError:
     pass
+except (AssertionError, ValueError):
+    warning = 'Old version of pyOpenSSL ({0}) may cause slowdown.'.format(pyopenssl_version)
```
nateprewitt: Let's move this into the `_check_cryptography` function.

requests/__init__.py
ghost: Ah, I just misunderstood the cause of the slowdown. Thanks for pointing that out, I'll get on this as soon as I can!

requests/__init__.py
```diff
@@ -71,12 +71,11 @@ def check_compatibility(urllib3_version, chardet_version):
     assert minor < 1
     assert patch >= 2
 
-# Check for older version of pyOpenSSL
-def __check_pyopenssl(pyopenssl_version, cryptography_version):
-    # Warning only applies for Cryptography < 1.3.4
+def _check_cryptography(cryptography_version):
+    # Cryptography < 1.3.4
```
nateprewitt: One petty nit and we're set here I think. Mind lowecasing the cryptography package name here, below in the warning message, and in the try/except comment?

requests/__init__.py
```diff
@@ -43,6 +43,7 @@
 import urllib3
 import chardet
 import warnings
+import sys
```
nateprewitt: Sorry before we squash this import is also not being used anymore.

requests/__init__.py
```diff
@@ -70,6 +71,11 @@ def check_compatibility(urllib3_version, chardet_version):
     assert minor < 1
     assert patch >= 2
 
+def _check_cryptography(cryptography_version):
+    # cryptography < 1.3.4
+    if list(map(int, cryptography_version.split('.'))) < [1, 3, 4]:
+        warning = 'Old version of Cryptography ({0}) may cause slowdown.'.format(cryptography_version)
```
nateprewitt: And cryptography is still capitalized here.

requests/__init__.py
ghost: I *think* I figured out how to squash, but I might be wrong. If there are some weird commits, that's probably why.

ghost:
I think I squashed everything right this time. Please let me know if I missed something!

ghost:
Great! Thanks for working with me and helping me get some experience, I really appreciate it!

requests/__init__.py
```diff
@@ -70,6 +70,11 @@ def check_compatibility(urllib3_version, chardet_version):
     assert minor < 1
     assert patch >= 2
 
+def _check_cryptography(cryptography_version):
+    # cryptography < 1.3.4
+    if list(map(int, cryptography_version.split('.'))) < [1, 3, 4]:
```
Lukasa: This check will break if for any reason cryptography has a non-numerical version string. As in other places, we should aim to tolerate that by assuming that weird cryptography version numbers will be fine.

ghost:
I've updated the changelog, added myself to AUTHORS.rst, and fixed the commit message.

Issue #4318, opened 2017-10-01
netrc reading does not respect NETRC environment variable


mitar:
As described [here](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html), there is an environment variable `NETRC` which one can use to control where `.netrc` file is. Requests always searches in user's home directory.

## System Information

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.5.2"
  },
  "platform": {
    "release": "14.5.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.18.4"
  },
  "system_ssl": {
    "version": "100020cf"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": false
}
```

chairmank:
Do we still want this feature? I want this feature, but I don't know overall sentiment among users.

I see that https://github.com/psf/requests/pull/4339 was closed because it wasn't finished. I volunteer to finish the work with appropriate test coverage and documentation.

nik003:
This has been open very long without response. How will this case be handled? This would be a very good feature to implement.

sigmavirus24:
This project has generally had a terrible experience with it's netrc support. It causes confusion and isn't convenient to disable. If anything, I'd expect the current maintainers to want to drop support for finding/parsing netrc files and close this as something they won't implement. (And they'd be right to do so)

mitar:
Awesome! Thanks!

PR #4339, opened 2017-10-13
Respect the NETRC environment variable, if set.

chrislawlor:
If the NETRC environment variable is set, use it to
load the user's netrc file. If not set, default to
searching the user's home directory for  a `.netrc`
or `_netrc` file.

Issue #4318

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -34,7 +34,7 @@ from .structures import CaseInsensitiveDict
 from .exceptions import (
     InvalidURL, InvalidHeader, FileModeWarning, UnrewindableBodyError)
 
-NETRC_FILES = ('.netrc', '_netrc')
+NETRC_FILES = os.environ.get('NETRC', ('~/.netrc', '~/_netrc'))
 
 DEFAULT_CA_BUNDLE_PATH = certs.where()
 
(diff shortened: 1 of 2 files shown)


requests/utils.py
```diff
@@ -166,13 +166,14 @@ def get_netrc_auth(url, raise_errors=False):
...
 
         for f in NETRC_FILES:
-            try:
-                loc = os.path.expanduser('~/{0}'.format(f))
-            except KeyError:
-                # os.path.expanduser can fail when $HOME is undefined and
-                # getpwuid fails. See http://bugs.python.org/issue20164 &
-                # https://github.com/requests/requests/issues/1846
-                return
+            if f.startswith('~'):
```
Lukasa: This is not right: we’ll now prepend a tilde to a path we know has one. Given that `expanduser` is safe to call on paths without user flags, we can just unconditionally call it. No need for this change.

chrislawlor:
Not too sure how to go about testing this, at least with the `os.environ` check done at module load time. I'm thinking about moving it inside `get_netrc_auth` to facilitate testing. With the check done there, a test could look something like:

1. Create an .netrc file in the temp directory
1. Use `utils.set_environ` context manager to temporarily set `NETRC`
1. Call `get_netrc_auth` and test that it is using the values from the temp file, in a similar manner as the `test_basicauth_with_netrc` test.

Does that sound like a good approach? I'm not sure that I like having the `os.environ` check inside the function, but I can't think of a good alternative.

requests/utils.py
```diff
@@ -167,7 +167,7 @@ def get_netrc_auth(url, raise_errors=False):
 
         for f in NETRC_FILES:
             try:
-                loc = os.path.expanduser('~/{0}'.format(f))
+                loc = os.path.expanduser(f)
```
sigmavirus24, line 170: So to make this more test-able and to probably make this more future-proof, what if we do the following:

Let's revert the change to L37, so that it looks like:

```py
NETRC_FILES = ('.netrc', '_netrc')
```

Then let's do this on L162 or so:

```py
netrc_file = os.environ.get('NETRC')
if netrc_file is not None:
   netrc_locations = (netrc_file,)
else:
   netrc_locations = ('~/{0}'.format(f) for f in NETRC_FILES)
```

And then we can change the loop on L168 to be:

```py
for f in netrc_locations:
```

Sure it's slightly more code, but it also is far more explicit and gives us room to expand if necessary.

kennethreitz:
Closing due to inactivity.

PR #4349, opened 2017-10-18
Fix case-insensitive comparison in Session.get_adapter

alvarogzp:
While trying to get the adapter for an url, the url is lowered before comparing with the prefix, but the prefix is not, so if it contains non-lowercase characters (eg. `https://example.com/sOmE_WeiRD_pReFIX/`), it won't return the correct adapter.

The other solution is to lower the prefix while adding the adapter. I have opted for the first solution to isolate in one place the case-insensitivity feature.

Edit:
Another alternative could be to document that the prefix should be passed lowered to Session.mount() (this is what I am doing for now to work-around the issue).

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -696,7 +696,7 @@ class Session(SessionRedirectMixin):
         """
         for (prefix, adapter) in self.adapters.items():
 
-            if url.lower().startswith(prefix):
+            if url.lower().startswith(prefix.lower()):
                 return adapter
 
         # Nothing matches :-/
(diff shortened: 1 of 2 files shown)


alvarogzp:
Yes, I will update the pull request with them.

alvarogzp:
@Lukasa I have added one test to check that it works with mixed case, another to check that the comparison is case-insensitive (this two are the ones that would have been failing before), and also another test to check prefix matching works as expected, ie. returning the adapter with the longest prefix that match (as I haven't found any test about it).

Lukasa:
Mind bringing this branch up to date with master?

alvarogzp:
Done!

Lukasa:
Thanks! 🎉

PR #4352, opened 2017-10-23
Add something to the docs about hooks on Session()

alexwlchan:
I had to work this out by reading the source code for `Session` – better to put it front-and-centre in the docs.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -452,11 +452,23 @@ If the callback function returns a value, it is assumed that it is to
 replace the data that was passed in. If the function doesn't return
 anything, nothing else is effected.
 
+::
+
+    def record_hook(r, *args, **kwargs):
+        r.hook_called = True
+        return r
+
 Let's print some request method arguments at runtime::
 
     >>> requests.get('http://httpbin.org', hooks={'response': print_url})
     http://httpbin.org
     <Response [200]>
+
+You can add multiple hooks to a single request.  Let's call two hooks at once::
+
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Is http://docs.python-requests.org/en/master/user/advanced/?highlight=hooks#event-hooks not helpful? It's in the same document.

sigmavirus24:
Ah, I see, it doesn't explain you can have hooks on a session. (Sorry, still ingesting coffee)

docs/user/advanced.rst
```diff
@@ -458,6 +458,15 @@ Let's print some request method arguments at runtime::
     http://httpbin.org
     <Response [200]>
 
+You can also assign hooks to a ``Session`` instance.  The hook will then be
+called on every request made to the session.  For example::
+
+   >>> s = requests.Session()
+   >>> s.hooks = dict(response=print_url)
```
sigmavirus24: This should be:

```
s.hooks['response'].append(print_url)
```

As we keep a list of hooks

alexwlchan:
(Review markups in progress.)

I’ve pushed the change to switch to dict literals, and clarified the use of multiple hooks on a session – I’ll do the same for the functional API later.

Lukasa:
Does “later” mean “in this PR when I next have time” or “in a follow-up PR”? Just want to make sure we’re all on the same page.

alexwlchan:
Later = “in this PR, when I am slightly less crushed under PyCon UK prep”

Lukasa:
👍

alexwlchan:
Turns out the train is a convenient time for doing small changes. How’s this for an example of multiple hooks?

PR #4360, opened 2017-10-30
bugfix issue #4353 by AzePUG team member Elshad Agayev

elshadaghazade:
- Added new function validate_url into the file utils.py at the line 858
- Before to call prepend_scheme_if_needed function we are checking validity of proxy url in the file adapters.py at the line 307, if wrong then raising ProxyError exception with clear error message

by Azerbaijan Python User Group


nateprewitt:
Thanks for contributing to requests, @elshadaghazade! Unfortunately, this issue is already being addressed in #4356. I'll be closing this one in favor of the earlier PR.

Issue #4325, opened 2017-10-04
Session adapter doesn't update https connections cert data


aleperno:
Under a session, the first https request sets the HTTPSConnectionPool certs data.
If in a subsequent request, we change the certs data (eg: Set verify to True), the adaptor will update the ConnectionPool attributes, however each connection in the pool will still contain the old certs data, thus when requesting a connection from the pool, its info will differ from what we would expect.

## Expected Result

The used connection should have updated certs data

## Actual Result

The used connection have "old" certs data.

## Reproduction Steps
These tests where performed against a server with a self-signed CA.
```python
In [1]: from requests import Session

In [2]: url = "https://somepage"

In [3]: auth = ("admin", "admin123")

In [4]: s = Session()

In [5]: s.get(url, auth=auth, verify=False)

Out[5]: <Response [200]>

# This should fail, but doesnt
In [6]: s.get(url, auth=auth, verify=True)

Out[6]: <Response [200]>

# Create a new session
In [7]: Session().get(url, auth=auth, verify=True)
# Raises certificate verify failed
```

## System Information

{
  "chardet": {
    "version": "3.0.4"
  }, 
  "cryptography": {
    "version": "1.7.2"
  }, 
  "idna": {
    "version": "2.6"
  }, 
  "implementation": {
    "name": "CPython", 
    "version": "2.7.12"
  }, 
  "platform": {
    "release": "4.4.0-59-generic", 
    "system": "Linux"
  }, 
  "pyOpenSSL": {
    "openssl_version": "1000207f", 
    "version": "16.0.0"
  }, 
  "requests": {
    "version": "2.18.4"
  }, 
  "system_ssl": {
    "version": "1000207f"
  }, 
  "urllib3": {
    "version": "1.22"
  }, 
  "using_pyopenssl": true
}

Lukasa:
Yeah. Requests is using some old APIs in urllib3 that don’t interact with the connection pooling. It should update to pass the TLS data into the pool manager to ensure that the TLS info factors in.

aleperno:
@Lukasa thanks for your prompt reply

foslock:
This is also true of the certificate/key paths provided in the `cert` keyword argument. It seems that the `certfile` and `keyfile` attributes are maintained on the connection and not reset if a new `cert` kwarg is given on a subsequent request on the session.

gluemaker:
[...]

PR #4361, opened 2017-10-30
Bugfix issue #4325 by AzePUG team member Elshad Agayev

elshadaghazade:
PoolManager object was initialized once and assigned to self.poolmanager property of HTTPAdapter class. Thats why updating settings (for example - verify=False to verify=True) does not affect. I reinitialized PoolManager on every connection and problem solved.

adapters.py:166
```
def init_poolmanager(self ...):
   ....
   self.poolmanager = PoolManager(num_pools=connections, maxsize=maxsize,
                                       block=block, strict=True, **pool_kwargs)
   return self.poolmanager
```

adapters.py:320
```
def get_connection(self...):
   ...
   conn = self.init_poolmanager(self._pool_connections, self._pool_maxsize, block=self._pool_block).connection_from_url(url)
```

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -161,6 +161,10 @@ class HTTPAdapter(BaseAdapter):
         self.poolmanager = PoolManager(num_pools=connections, maxsize=maxsize,
                                        block=block, strict=True, **pool_kwargs)
 
+        #bugfix_issue4325 by AzePUG team member Elshad Agayev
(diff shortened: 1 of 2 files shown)


sigmavirus24:
I believe the issue you are attempting to fix was scheduled for Requests 3.0 and is already fixed on the branch of code dedicated to that release. As such, I'm closing this. Thanks for contributing!

PR #4368, opened 2017-11-03
Clarify behaviour of `json` parameter.

danielroseman:
`json` is ignored if `data` is not empty.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -281,6 +281,7 @@ the ``json`` parameter (added in version 2.4.2) and it will be encoded automatic
 
     >>> r = requests.post(url, json=payload)
 
+Note, the ``json`` parameter is ignored if either ``data`` or ``files`` is passed.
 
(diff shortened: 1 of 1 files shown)


ghost:
How about this?

> The `data` parameter takes precedence over the `json` parameter.

docs/user/quickstart.rst
```diff
@@ -273,6 +273,7 @@ the ``json`` parameter (added in version 2.4.2) and it will be encoded automatic
 
     >>> r = requests.post(url, json=payload)
 
+Note, the ``json`` parameter is ignored if the ``data`` parameter is not empty.
```
nateprewitt: We should probably note that both `data` *and* `files` take precedence over the `json` parameter.

danielroseman:
@nateprewitt Thanks for the comments and apologies for the delay. I've updated as suggested.

nateprewitt:
Yep, looks great to me too, thanks @danielroseman!

Issue #4373, opened 2017-11-06
On Windows, getting wrong proxy settings from Windows registry


mingyuan-xia:
For some windows machines (not all), a confusing Windows registry key type (`HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Internet Settings \ ProxyEnable`) fails the following code in `requests/requests/utils.py`.

```
            proxyEnable = winreg.QueryValueEx(internetSettings,
                                              'ProxyEnable')[0]
            proxyOverride = winreg.QueryValueEx(internetSettings,
                                                'ProxyOverride')[0]
```
The case is if proxyEnable is a string , everything is fine. But `ProxyEnable` could be `REG_SZ` on some Windows machines (no idea why, but it happens). After user deletes this key or set if to `REG_DWORD 0x00000000(0)`, things get right. But I believe better type checks in requests will help.

First found: https://stackoverflow.com/questions/46260169/python-locationvalueerror-no-host-specified
Aggregated discussion: https://github.com/conda/conda/issues/3962

## Expected Result

Type checks. no failure for  `REG_SZ` values.

Lukasa:
I'd be happy to merge a patch that resolves this issue.

mingyuan-xia:
PR: https://github.com/requests/requests/pull/4377

nehaljwani:
@mingyuan-xia Could you please open a PR with that patch? I'd really love to see that get fixed in the next release of requests.

mingyuan-xia:
https://github.com/requests/requests/pull/4377

PR #4377, opened 2017-11-07
#4373, fix possible winreg value type difference

mingyuan-xia:


mingyuan-xia:
It seems that `test_should_bypass_proxies_win_registry` only monkeypatches WinReg but never tests anything (no assert~). Will adding another monkeypatch there be sufficient?

tests/test_utils.py
```diff
@@ -677,6 +677,13 @@ def QueryValueEx(key, value_name):
             elif value_name == 'ProxyOverride':
                 return [override]
 
+    def QueryValueEx2(key, value_name):
+        if key is ie_settings:
+            if value_name == 'ProxyEnable':
+                return ["1"] # this could be a string (REG_SZ) or a 32-bit number (REG_DWORD)
+            elif value_name == 'ProxyOverride':
+                return [override]
+
```
Lukasa: This is not sufficient, as it has never been used. However, you're right that this never actually tests anything, which isn't great. This was supposed to call `should_bypass_proxies`: want to add that call, and then parametrise this test based on the return value of `QueryValueEx` with `ProxyEnable`?

mingyuan-xia: Let me try adding some real assertions here. Need to find a Windows first, though.

mingyuan-xia:
Since we already have multiple parameterized tests, I simply rotate the `ProxyEnable` Windows registry key type per test to simulate the case. Also I have added an assertion to test the proxy selection as it was supposed to be there.

Lukasa:
Looks like the tests are failing: mind addressing the test failures?

sigmavirus24:
Hi @mingyuan-xia the tests are passing but the branch isn't up-to-date with master. If you can fix that up, I think we can finally accept this work. 👍

mingyuan-xia:
Ok I will pull from the master and please squash my commits into one to make things more concise.

PR #4378, opened 2017-11-09
for RFC-7616 add SHA-256 and SHA-512

linuxlizard:
While working with some FIPS libraries that disabled MD5, we ran into a problem with our Requests based tools not being able to talk to our server. 

screen shot 2017-11-08 at 4 19 47 pm

I haven't fully read the RFC-7616 SHA-256 and SHA-512 standards but this patch works.

Thanks for Requests!

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -153,6 +153,18 @@ class HTTPDigestAuth(AuthBase):
                     x = x.encode('utf-8')
                 return hashlib.sha1(x).hexdigest()
             hash_utf8 = sha_utf8
+        elif _algorithm == 'SHA-256':
+            def sha256_utf8(x):
+                if isinstance(x, str):
+                    x = x.encode('utf-8')
+                return hashlib.sha256(x).hexdigest()
+            hash_utf8 = sha256_utf8
+        elif _algorithm == 'SHA-512':
+            def sha512_utf8(x):
+                if isinstance(x, str):
+                    x = x.encode('utf-8')
+                return hashlib.sha512(x).hexdigest()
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Could you add test cases for this, please?

linuxlizard:
Can do.

linuxlizard:
Status: I'm reaching out to Requests developers listed in "contributing" for some assistance adding tests.

linuxlizard:
Status update: Twitter reply from @lukasaoz 
"  | Lukasaoz @linuxlizard I doubt httpbin supports these auth modes yet so you may need to begin by adding support there. Alternatively you can use code like test_lowlevel.py to force the correct responses.Nov 14, 2017, 12:01 PM
-- | --

I'm going to investigate that approach.

linuxlizard:
Status update: Twitter reply from @lukasaoz 
"  | Lukasaoz @linuxlizard I doubt httpbin supports these auth modes yet so you may need to begin by adding support there. Alternatively you can use code like test_lowlevel.py to force the correct responses.Nov 14, 2017, 12:01 PM
-- | --

I'm going to investigate that approach.

linuxlizard:
Created pull request for httpbin to add SHA-512 support.
https://github.com/kennethreitz/httpbin/pull/401

linuxlizard:
Thanksgiving over. httpbin patch accepted. Back to work on learning how to add tests to Requests.

linuxlizard:
Shutting down this pull request. Creating a new pull request with test code.

PR #4380, opened 2017-11-10
utils: winreg module may not exist like on windows universal platform.

afedchin:
UWP has no API to get access to windows registry, so python has no winreg module on UWP.
This fixes unexpected error on UWP.


afedchin:
@Lukasa I've updated the PR

afedchin:
@Lukasa done

afedchin:
rebased on current master

PR #4390, opened 2017-11-20
[docs; tiny] Clarify that Response.ok will *only* return True/False

alexwlchan:
This is a tiny documentation fix, just to be super-duper explicit about the return value of `response.ok`.

(I wanted to check I could use its value in a function that’s meant to return a boolean, but the docs weren’t clear about what a non-Truthy return value would be.)

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -686,11 +686,11 @@ class Response(object):
 
     @property
     def ok(self):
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Thanks @alexwlchan

PR #4409, opened 2017-11-29
Use single default ssl_context in HTTPAdapter connection pool manager

jmcackles:
Creates default SSLContext object for each HTTPAdapter that is expected to be used for https requests.

There are just a few small changes for:
- Adding a default SSLContext object only to the https HTTPAdapter in each session (not the http one)
- Creating the default object using the same urllib3 function that would be used to create a default SSLContext for each connection
- Adding ssl_context to poolmanager's connection_pool_kw to get passed down to all connections
- Pickling HTTPAdapter with the new attribute
- Breaking up the HTTPAdapter's send method a little (this is not related, I was just there anyway.  I'm happy to undo that if it's not wanted.  There are separate functions for getting the timeout and sending chunked now.)

Any feedback is welcome!  This is for https://github.com/requests/requests/issues/4322

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -16,6 +16,7 @@ from urllib3.response import HTTPResponse
 from urllib3.util import parse_url
 from urllib3.util import Timeout as TimeoutSauce
 from urllib3.util.retry import Retry
+from urllib3.util.ssl_ import create_urllib3_context as default_ssl_context
 from urllib3.exceptions import ClosedPoolError
 from urllib3.exceptions import ConnectTimeoutError
 from urllib3.exceptions import HTTPError as _HTTPError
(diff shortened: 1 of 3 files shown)


Lukasa:
It’d be great to separate out the unrelated refactoring and supply it in a separate patch, please. 

A single SSLContext per adapter is not right, I’m afraid. An adapter can and will run across many different requests with different TLS settings. To get this to work properly you will need to transfer all TLS settings from Requests onto the adapter. You will also need to make this patch against the 3.0 branch, as it will break backwards compat.

jmcackles:
Thank you for the fast response!  I apologize for not looking at this close enough-- I can close this PR and make a new one against update-3.0 (?).

I think if I'm understanding right, it should still be OK to have one SSLContext per adapter, but make sure if 'verify' or 'cert' arguments are truthy for the 'cert_verify' method, I set conn.ssl_context = None there to avoid modifying the adapter's SSLContext.  Does that seem pretty reasonable, or do you mind pointing me where I should be looking for any other TLS settings?  (Sorry to bug, I appreciate the time to answer).

Lukasa:
I think if the goal is to have the minimal number of contexts, we need to key off the various bits of TLS configuration and create the contexts lazily.

PR #4424, opened 2017-12-11
Fix: Removed incorrect plural of 'status'

SHSauler:
The plural of `status` is `statuses` or `statūs`. Please disregard if it's here for a particular reason. 

https://en.wiktionary.org/wiki/status#English 
https://en.wiktionary.org/wiki/status#Etymology_2_2

### DIFF
--- a/requests/status_codes.py
+++ b/requests/status_codes.py
@@ -34,7 +34,7 @@ _codes = {
     204: ('no_content',),
     205: ('reset_content', 'reset'),
     206: ('partial_content', 'partial'),
-    207: ('multi_status', 'multiple_status', 'multi_stati', 'multiple_stati'),
+    207: ('multi_status', 'multiple_status'),
     208: ('already_reported',),
     226: ('im_used',),
 


sethmlarson:
Doesn't this break backwards compatibility? I'm doubtful someone uses those values though.

sigmavirus24:
Hi there @SHSauler! Thanks for sending your first pull request to Requests. We really appreciate your contribution. In this case, I don't think we can actually accept this pull request. In this case, there are a few reasons why:

1. As @SethMichaelLarson mentioned, it breaks backwards compatibility. There may not be many people using it, but our API is stable and frozen which makes this a hard change to accept.

2. These aliases pre-date me, but I believe that they're actually in use in the wider internet (which is probably why they were added in the first place)

3. I think there are other places where we use "stati" and "statuses" interchangeably so it's not unusual for this.

4. The internet is a weird place. There's a header spelt "Referer" (instead of "Referrer") which is standard and so misspellings can be both common and de facto (or de jure) standards.

All that aside, the you're entirely correct about the proper pluralization of status. Sadly, fixing it here is not probably. We *could* probably fix this for v3.0 though. There's already a branch where we're hacking on changes for that version so if you want to resubmit this there, we can definitely dig more into reasons 2, 3, and 4 and decide if they block that change there.

Thanks again 🎉

SHSauler:
Thank you for your time to review this. I completely understand your reasoning.

PR #4427, opened 2017-12-12
Use hostname not netloc

### LINKED ISSUE #4158: Making requests to URLs with username and password does not respect no_proxy evironment settings

### LINKED ISSUE #4275: Requests tries to resolve `user:pass@host` on OS X
If you have proxy bypass settings, requests tries to resolve the hostname before stripping credentials from the URL.

It looks like this happens because the [proxy bypass](https://github.com/requests/requests/blob/4bda7b66e7ece5be51b459edd046a70915b4792c/requests/utils.py#L78) logic is called before [urldefragauth](https://github.com/requests/requests/blob/b9638ec7be9134786e28684ca17f3d161297ac5e/requests/utils.py#L879) strips the credentials from the URL.

## Expected Result

Should not try to resolve an invalid hostname.

## Actual Result

DNS resolution for `user:pass@host` is made:

screenshot 2017-09-05 11 04 11

## Reproduction Steps

On OS X with proxy bypass hosts configured:

screenshot 2017-09-04 18 08 46

```python
import requests
requests.get('http://user:pass@localhost')
```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "2.7.10"
  },
  "platform": {
    "release": "15.6.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.18.4"
  },
  "system_ssl": {
    "version": "100020cf"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": false
}
```

ddormer:
Fixes #4275 and fixes #4158 by passing only the hostname/IP to urllib's `proxy_bypass` function which was timing out further down in `socket.gethostbyname` when passed a hostname or IP that contained authentication credentials.

While fixing the parameter passed to `proxy_bypass`, I also took the chance to replace other uses of `netloc` with `hostname` in the `should_bypass_function`.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -703,28 +703,31 @@ def should_bypass_proxies(url, no_proxy):
     no_proxy_arg = no_proxy
     if no_proxy is None:
         no_proxy = get_proxy('no_proxy')
-    netloc = urlparse(url).netloc
+    parsed = urlparse(url)
 
     if no_proxy:
         # We need to check whether we match here. We need to see if we match
-        # the end of the netloc, both with and without the port.
+        # the end of the hostname, both with and without the port.
         no_proxy = (
             host for host in no_proxy.replace(' ', '').split(',') if host
         )
 
-        ip = netloc.split(':')[0]
-        if is_ipv4_address(ip):
+        if is_ipv4_address(parsed.hostname):
             for proxy_ip in no_proxy:
                 if is_valid_cidr(proxy_ip):
-                    if address_in_network(ip, proxy_ip):
+                    if address_in_network(parsed.hostname, proxy_ip):
                         return True
-                elif ip == proxy_ip:
+                elif parsed.hostname == proxy_ip:
                     # If no_proxy ip was defined in plain IP notation instead of cidr notation &
                     # matches the IP of the index
                     return True
         else:
+            host_with_port = parsed.hostname
+            if parsed.port:
+                host_with_port += ':{0}'.format(parsed.port)
+
             for host in no_proxy:
-                if netloc.endswith(host) or netloc.split(':')[0].endswith(host):
+                if parsed.hostname.endswith(host) or host_with_port.endswith(host):
                     # The URL does match something in no_proxy, so we don't want
                     # to apply the proxies on this URL.
                     return True
@@ -737,7 +740,7 @@ def should_bypass_proxies(url, no_proxy):
     # legitimate problems.
     with set_environ('no_proxy', no_proxy_arg):
         try:
-            bypass = proxy_bypass(netloc)
+            bypass = proxy_bypass(parsed.hostname)
         except (TypeError, socket.gaierror):
             bypass = False
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -182,3 +182,4 @@ Patches and Suggestions
 - Arthur Vigil (`@ahvigil <https://github.com/ahvigil>`_)
 - Nehal J Wani (`@nehaljwani <https://github.com/nehaljwani>`_)
 - Demetrios Bairaktaris (`@DemetriosBairaktaris <https://github.com/demetriosbairaktaris>`_)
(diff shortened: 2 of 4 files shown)


requests/utils.py
```diff
@@ -679,28 +679,28 @@ def should_bypass_proxies(url, no_proxy):
...
-                    if address_in_network(ip, proxy_ip):
+                    if address_in_network(parsed.hostname, proxy_ip):
                         return True
-                elif ip == proxy_ip:
+                elif parsed.hostname == proxy_ip:
                     # If no_proxy ip was defined in plain IP notation instead of cidr notation &
                     # matches the IP of the index
                     return True
         else:
+            host_with_port = '{}:{}'.format(parsed.hostname, parsed.port)
```
nateprewitt: A couple notes here. The `port` attribute is either an `int` or `None`, which means a case like http://localhost being passed to `urlparse` would yield `localhost:None` here. While that's unlikely to affect matching, I'm not sure that's what we want to be doing.

The second thing is we're still planning to support python 2.6 until 3.0.0 is released which means this format statement won't work as is. We'll need to provide something along the lines of `{0}:{1}` instead to avoid the ValueError.

ddormer:
@nateprewitt Thanks for the review! Below I've listed the changes I made before I squashed my commits.

- Check if the port exists before adding it to prevent `host:None`.
- Use `{0}` to support Python 2.6.
- Updated the HISTORY.rst.

sigmavirus24:
@ddormer Can you rebase this so it's up-to-date with master?

ddormer:
@sigmavirus24 Thanks for reviewing this, I've rebased so it's up to date with master.

PR #4428, opened 2017-12-13
Reuse session SSLContexts for same TLS settings

jmcackles:
Creates SSLContext objects in HTTPAdapter and uses a single SSLContext for connections with the same TLS settings (ca_cert_dir, ca_certs, cert_file, and key_file).

This is another try for https://github.com/requests/requests/issues/4322

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -15,6 +15,7 @@ from urllib3.poolmanager import PoolManager, proxy_from_url
 from urllib3.response import HTTPResponse
 from urllib3.util import Timeout as TimeoutSauce
 from urllib3.util.retry import Retry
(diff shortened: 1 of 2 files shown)


jmcackles:
My best guess is that the failure is related to https://bugs.python.org/issue9949, but I'm really not sure.  The test is failing because using a relative path and absolute path to the same cert file creates 2 different SSLContexts in the windows tests.  I'm not sure if this is an important case, but I had added a test for it.

jmcackles:
I was able to look on Windows and it was clearer, the test was failing because of letter case issues, which is fixed now, I'm sorry about that.  I think symlinks are still an issue for Windows, but I haven't added another test case or anything to fix it.

sethmlarson:

PR #4439, opened 2017-12-28
Fix broken link to Certifi

nkonin:
certifi.io doesn't available at least since 07.11.2017
(https://github.com/certifi/certifi.io/issues/16)

### DIFF
--- a/docs/community/recommended.rst
+++ b/docs/community/recommended.rst
@@ -15,7 +15,7 @@ Certifi CA Bundle
 validating the trustworthiness of SSL certificates while verifying the
 identity of TLS hosts. It has been extracted from the Requests project.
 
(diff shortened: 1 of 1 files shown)


docs/community/recommended.rst
```diff
@@ -15,7 +15,7 @@ Certifi CA Bundle
 validating the trustworthiness of SSL certificates while verifying the
 identity of TLS hosts. It has been extracted from the Requests project.
 
-.. _Certifi: http://certifi.io/en/latest/
+.. _Certifi: https://github.com/certifi/
```
sigmavirus24: Should be https://github.com/certifi/python-certifi

nateprewitt:
@nkonin, would you mind rebasing this onto the current master branch? I think we should be set after that.

Issue #4313, opened 2017-09-30
Importing requests takes additional 13MB, more than doubling Python's mem usage


erezsh:
Simple running:

    import requests

In a new, blank process, raises memory consumption from ~11MB to ~24MB

## Expected Result

Even big libraries like Django raise it by less than 1MB

## Reproduction Steps

```python
import requests

```
:)

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.5.3"
  },
  "platform": {
    "release": "4.10.13-200.fc25.x86_64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.18.4"
  },
  "system_ssl": {
    "version": "100020bf"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": false
}
```
(Note that pyopenssl doesn't seem to be the cause this time)

erezsh:
P.S. the same effect happens in Python 2

kengruven:
I've been looking into this, and it's about what you'd expect: 'requests' uses a lot of libraries, and some of them use a lot of memory.

(In comparison, Django uses very few memory-hungry modules.)

The biggest offenders that I've found in 'requests':


 1. urllib3 (> 6 MB)

I know this one makes dealing with URLs and HTTP a million times easier, and there's no chance 'requests' will stop using it, but there could be sub-dependencies here that have memory optimization opportunities.

For example, 'urllib3' (eagerly) imports 'uuid', for creating unique boundaries when encoding multipart/form-data.  The 'uuid' module adds almost 2 MB.  Is there an alternative that could work?  For example, plain 'random' is about half that size.  Or import it lazily.


 2. cgi (> 4 MB)

This is only used for one method (parse_header), which is about 20 lines of code.

Idea: write our own.  Requests doesn't even use the full parse output -- a regex that looked at two parts of the string could do the trick, I think.


 3. chardet (> 3 MB)

This one seems to include a lot of character frequency tables.  In the common case for 'requests' (where the server supplies a Content-Type), it's not even used.

Idea: lazy load this, when needed.  It seems to do some fancy runtime version checking in `__init__`, so I'm not sure what's involved here.


### Regarding lazy loading:

[...]

PR #4442, opened 2017-12-31
Reduce overall memory usage by removing cgi dependency and writing a protected function

DemetriosBairaktaris:
In response to issue #4313 , I removed the cgi module import and added a parse_header function nested inside get_encoding_from_headers in utils.py.  I have run the tests both before and after making these changes.  The test_util cases all passed before and they still do now.  Please let me know about any suggestions to this code.


requests/utils.py
```diff
@@ -453,13 +452,27 @@ def get_encoding_from_headers(headers):
...
+        content_type_and_params_delimiter = ';'
+        content_type += content_type_and_params_delimiter
+
+        tokens = content_type.split(content_type_and_params_delimiter)
+        content_type_index = 0
+        params_index = 1
+
+        content_type = tokens[content_type_index]
+        params = tokens[params_index]
+        params_dict = dict(param.split('=') for param in params.split())
```
sigmavirus24: I'm not convinced this is going to be correct. I think having an `=` inside the parameter value is acceptable if it's quoted. We should probably then use `param.split('=', 1)` to ensure we only split on the first found =.

requests/utils.py
```diff
@@ -453,13 +452,27 @@ def get_encoding_from_headers(headers):
...
+        return content_type,params_dict
```
sigmavirus24: Please add a space, e.g.,

```py
return content_type, params_dict
```

requests/utils.py
```diff
@@ -453,13 +452,27 @@ def get_encoding_from_headers(headers):
     :param headers: dictionary to extract encoding from.
     :rtype: str
     """
+    def parse_header(content_type):
+        #Inner function to parse header
```
sigmavirus24: Why does this need to be an inner function? Why can't we define it at the module level as `_parse_header`?

requests/utils.py
DemetriosBairaktaris: Thanks for your suggestions.

requests/utils.py
DemetriosBairaktaris: Ill check it out. Thank you

DemetriosBairaktaris:
@sigmavirus24 Thoughts on this?  I figured since the since we are only worrying about parsing content-type headers, we should focus the function on that format.

requests/utils.py
```diff
@@ -459,7 +488,7 @@ def get_encoding_from_headers(headers):
     if not content_type:
         return None
 
-    content_type, params = cgi.parse_header(content_type)
+    content_type, params = _parse_content_type_header(content_type)
```
sigmavirus24, line 486: If `content_type` is checked above and this is a private function, do we need the guard at the top of the function above? If the check above is ever removed, we could end up with a `TypeError` when we  effectively do `content_type, params = None`.

requests/utils.py
```diff
@@ -447,6 +446,36 @@ def get_encodings_from_content(content):
...
+    """Returns content type and parameters from given header
+
+    :param header: string
+    :return: tuple containing content type and dictionary of
+         parameters
+    """
+    if not header:
+        return None
+    # append delimiter on end to ensure at least two elements when split by ';'
+    header += ';'
```
sigmavirus24: I don't think this is necessary. Can we do

```py
tokens = header.split(';', 1)
content_type, params_list = tokens[0].strip(), tokens[1:]
if  len(params_list) < 1:
    return content_type, {}
params = params_list[0]
```

If I understand correctly that's the effect of this otherwise. It's simple, clearer, and doesn't require us to string concatenation which results in a new allocation and isn't terribly efficient.

```py
>>> a = [1]
>>> b, c = a[0], a[1:]
>>> b, c
(1, [])
```

requests/utils.py
```diff
@@ -447,6 +446,36 @@ def get_encodings_from_content(content):
...
+    # split content type's main value from params
+    tokens = header.split(';', 1)
+    content_type_index = 0
+    params_index = 1
+
+    content_type = tokens[content_type_index].strip()
+    params = tokens[params_index]
+    params_dict = dict()
+
+    for param in params.split(';'):
```
sigmavirus24: The above said, if we weren't restricting it to split once we could simplify the code further and do:

```
tokens = header.split(';')
content_type, params = tokens[0].strip(), tokens[1:]
params_dict = {}. # Using dict is actually slower than a dictionary literal. Weird but tru
for param in params:
    # ... What you have below

return content_type, params_dict
```

requests/utils.py
DemetriosBairaktaris: That is pretty nice.  Didn't know a[1:] would not throw an index out of bounds... nice.

DemetriosBairaktaris:
@sigmavirus24 Ok, I pushed some changes with your suggestions, it's looking better for sure.

requests/utils.py
```diff
@@ -447,6 +446,29 @@ def get_encodings_from_content(content):
...
+    :return: tuple containing content type and dictionary of
+         parameters
+    """
+
+    tokens = header.split(';')
+    content_type, params = tokens[0].strip(), tokens[1:]
+    params_dict = {} 
+
+    for param in params:
+        if param and not param.isspace():
```
sethmlarson: Can we not call `param = param.strip()` before the `if param` and save a call to `param.isspace()`?

ie Is this:
```python
for param in params:
    param = param.strip()
    if param:
        # Continue here...
```
the same as this
```python
for param in params:
    if param and not param.isspace():
        param = param.strip()
        # Continue here...
```
in all situations?

requests/utils.py
```diff
@@ -447,6 +446,29 @@ def get_encodings_from_content(content):
...
+            param = param.strip()
+            key, value = param, True
+            if '=' in param:
```
sethmlarson: Might be able to get rid of the list comprehension here by using `param.find('=')` and checking for a non-negative? Definitely would be faster. Something to the tune of:

```python
key, value = param, True
index = param.find('=')
if index != -1:
    key = param[:index].strip('\'" ')
    value = param[index+1:].strip('\'" ')
```

tests/test_utils.py
```diff
@@ -470,6 +470,33 @@ def test_parse_dict_header(value, expected):
...
+        (
+            'application/json ; charset=utf-8',
+            ('application/json', {'charset': 'utf-8'})
+        ),
+        (
+            'text/plain',
+            ('text/plain', dict())
+        ),
+        (
+            'multipart/form-data; boundary = something ; \'boundary2=something_else\' ; no_equals ',
```
sethmlarson: Can you also add a test case with double-quotes around the value? ie: `boundary = "something";` should return `{'boundary': 'something'}`.

DemetriosBairaktaris:
@SethMichaelLarson Good suggestions, I pushed some changes.

requests/utils.py
```diff
@@ -456,15 +456,19 @@ def _parse_content_type_header(header):
...
+        param = param.strip()
+        if param:
             key, value = param, True
-            if '=' in param:
-                param_tokens = [x.strip('\'" ') for x in param.split('=', 1)]
-                key, value = param_tokens[0], param_tokens[1]
+            index_of_equals = param.find("=")
+            if index_of_equals != -1:
+                before_equals = slice(0, index_of_equals)
+                after_equals = slice(index_of_equals + 1, len(param))
```
SethMichaelLarson: Can you remove the `slice`s here? Just use index slice notation, (`param[:index_of_equals]` and `param[index_of_equals:]`) much more readable IMO.

requests/utils.py
```diff
@@ -447,6 +446,33 @@ def get_encodings_from_content(content):
...
+    params_dict = {}
+    items_to_strip = "\"' "
+
+    for param in params:
+        param = param.strip()
+        if param:
+            key, value = param, True
+            index_of_equals = param.find("=")
+            if index_of_equals != -1:
+                before_equals = slice(0, index_of_equals)
```
sethmlarson: Can you remove the `slice()` usage here? Using bracket-slice notation is more readable and faster.

`param[:index_of_equals]` and `param[index_of_equals+1:]`

sethmlarson:
Besides my two last comments this looks good to me. I'll leave this review for @sigmavirus24 now. :)

tests/test_utils.py
sethmlarson: Still would like to see a test-case containing this scenario.

sigmavirus24:
Thanks for your patience @DemetriosBairaktaris ! This looks excellent!

Issue #4443, opened 2018-01-01
fragment not append on redirect


stcioc:
In the following case (see detailed explanation and test case at http://www.webdbg.com/test/redir/fragment/):
- try to open a link that contains a fragment (say http://foo/#SomeInfo)
- the server sends a 302 Redir to http://bar

## Expected Result

The expected result is that requests library follows the redirection and appends the original fragment from the link (e.g. it redirects to http://bar/#SomeInfo)

## Actual Result

The redirect link is followed without the fragment (e.g. requests tries to open http://bar)

## Reproduction Steps

```python
import requests
response = requests.get("https://www.bayden.com/test/redir/fragment/redir-1.asp#/HereIsSomeInfo")
html = response.text
print(response.url)
```
output is "http://www.enhanceie.com/test/redir/fragment/final.asp"
output should be "http://www.enhanceie.com/test/redir/fragment/final.asp#HereIsSomeInfo"

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.6.2"
  },
  "platform": {
    "release": "10",
    "system": "Windows"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.18.4"
  },
  "system_ssl": {
    "version": "100020bf"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": false
}
```

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).

sigmavirus24:
This isn't how redirects work. The RFCs (3986, 7230, 7231, 7232, 7233, 7234, 7235) do not require or prescribe retaining the fragment. If you want us to retain the fragment you'll need to handle redirects yourself. The server is telling us what to redirect to and we're supposed to follow that.

Thanks for opening this issue to clarify.

sigmavirus24:
Actually this was formalized in the rewrite of HTTP/1.1 (see also https://stackoverflow.com/questions/2286402/url-fragment-and-302-redirects)

sigmavirus24:
This is a fantastic way to leak some data to an untrusted server, however, and is probably far from ideal. Oh well, we'll have to fix this.

sigmavirus24:
[...]

PR #4452, opened 2018-01-07
append previous url fragment on redirect

nateprewitt:
This PR should address #4443 by appending the original request's fragment to subsequent redirects, unless a new fragment is provided in the Location header. This behaviour should bring us into compliance with [RFC 7230 § 7.1.2](https://tools.ietf.org/html/rfc7231#section-7.1.2). I also wrote a test to confirm we don't send fragment information to new servers, our underlying implementation strips that.

One outstanding question I do have though is about chained redirects. 

-----

Let's say we make a request (`http://url#alice`) and get a 302 response (`http://new_url#bob`). We would leave `#bob` as the fragment for the second request. When we request `http://new_url#bob`, we get back another 301 to `http://final_url/`. Do we append `alice` or `bob` at this point? 

I went with the assumption that the "original" request is the first one in the chain. I could see an argument that all requests are stateless though, and that the `new_url#bob` is the only originating request in scope for the second redirect.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -123,6 +123,7 @@ class SessionRedirectMixin(object):
         hist = []  # keep track of history
 
         url = self.get_redirect_target(resp)
+        previous_fragment = urlparse(req.url).fragment
         while url:
             prepared_request = req.copy()
 
@@ -147,8 +148,12 @@ class SessionRedirectMixin(object):
                 parsed_rurl = urlparse(resp.url)
                 url = '%s:%s' % (to_native_string(parsed_rurl.scheme), url)
 
-            # The scheme should be lower case...
+            # Normalize url case and attach previous fragment if needed (RFC 7231 7.1.2)
             parsed = urlparse(url)
+            if parsed.fragment == '' and previous_fragment:
+                parsed = parsed._replace(fragment=previous_fragment)
+            elif parsed.fragment:
+                previous_fragment = parsed.fragment
             url = parsed.geturl()
 
             # Facilitate relative 'location' headers, as allowed by RFC 7231.
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -235,3 +235,75 @@ def test_redirect_rfc1808_to_non_ascii_location():
         assert r.url == u'{0}/{1}'.format(url, expected_path.decode('ascii'))
 
         close_server.set()
+
+def test_fragment_not_sent_with_request():
+    """Verify that the fragment portion of a URI isn't sent to the server."""
+    def response_handler(sock):
+        req = consume_socket_content(sock, timeout=0.5)
+        sock.send(
+            b'HTTP/1.1 200 OK\r\n'
+            b'Content-Length: '+bytes(len(req))+b'\r\n'
+            b'\r\n'+req
+        )
+
+    close_server = threading.Event()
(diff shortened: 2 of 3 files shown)


nateprewitt:
I went ahead and did [some testing](https://gist.github.com/nateprewitt/233301041bbd70cc53ef31a807d85bc7) to see how other user agents handle this. All of the major browsers except Safari will append `#bob`, so I've updated the logic and extended the tests to reflect that.

I could see having the expected fragment being modified in transit being confusing for end-users, but it's probably best we behave like the rest of the ecosystem unless we have a strong reason not to.

tests/testserver.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+import threading
+import requests
+
+from tests.testserver.server import Server, consume_socket_content
```
sigmavirus24: I'm confused how this import is working right now. Maybe it's too late for me to be reviewing code?

nateprewitt: 😬 this is me running `git add tests` and not paying attention to garbage in the directory. Disregard this file.

sigmavirus24: I await your force-push to remove it 😄

nateprewitt: Ok, we should be updated now.

requests/sessions.py
```diff
@@ -147,8 +148,12 @@ def resolve_redirects(self, resp, req, stream=False, timeout=None,
                 parsed_rurl = urlparse(resp.url)
                 url = '%s:%s' % (to_native_string(parsed_rurl.scheme), url)
 
-            # The scheme should be lower case...
+            # Normalize url case and attach previous fragment if needed (RFC 7231 7.1.2)
             parsed = urlparse(url)
+            if parsed.fragment == '' and previous_fragment:
+                parsed = urlparse('{0}#{1}'.format(url, previous_fragment))
```
sigmavirus24: Can we do `parsed._replace(fragment='previous_fragment')`?

sigmavirus24: My concern is that `url` might end with `#` which should be equivalent to `parsed.fragment == ''`. What do we do there?

nateprewitt: Yep, this is definitely a gap. It looks like `geturl()` will also strip empty fragments from redirect urls which probably isn’t the behavior we want.

I’m a little leery of using private methods, but given that `_replace` has been stable for a decade it’s probably safe. I’m wondering given the trailing hash issue if it may be easier to append the fragment option (`’’`, `#`, `#fragment`) after we call `geturl`. I’ll try to come up with something tonight.

sigmavirus24: `_replace` is actually part of `namedtuple` which `ParseResult` inherits from. If we wanted to, we could switch wholesale to github.com/hyper/rfc3986 which has a proper public API for this but that is definitely more work than this.

nateprewitt:
@sigmavirus24 sorry for the delay. I've been playing around with rfc3986 and empty fragments. The rfc3986 conversion is a little more than I've had time to do right now and seems like a good followup PR.

I have empty fragments working on another branch, but the solution is cluttered and the behaviour around them is mostly undefined.Given that we don't support them currently, I'm going to propose this is the solution we proceed with. If you'd prefer we maintain empty fragments, I'll push up the other branch.

nateprewitt:
Alright, rebased and merging.

Issue #4455, opened 2018-01-09
no response object captured in error of endless redirect chain and confusing message


jvanasco:
I found an incredibly bad URL that causes a handful of issues.  the url below jumps between a bunch of SocialFlow (trib.al) and Bitly url shorteners for tracking, then ends up on a connection timeout that also looks like a max retries exceeded (which makes no sense to me)

The exception that is raised under Py2 and Py3 does not have a response object, although one should exist – there are a long history of redirects that gets lost and makes this difficult to debug the url's problem.

   HTTPSConnectionPool(host='this.cm', port=443): Max retries exceeded with url: /slate/follow (Caused by ConnectTimeoutError(<urllib3.connection.VerifiedHTTPSConnection object at 0x1021788d0>, 'Connection to this.cm timed out. (connect timeout=2)'))

```python
import requests

try:
	url = 'http://slate.me/1KDocND'
	r = requests.get(url, timeout=2)
	print("success")
	print(r.response)
except Exception as error:
	print("error")
	print(error)
	print(type(error.response))
```

sigmavirus24:
The `max retries exceeded with url` is due entirely to the fact that Requests uses a subset of urllib3 at this point. Requests' usage of retries were wholly limited to the API that urllib3 provided prior to creating the Retry object. It's been updated little since. As a result, we don't allow urllib3 to perform its own retries so when something like this fails that is potentially retryable from the perspective of urllib3 but retries is at 0 already, we get that exception. It's plagued us for years. That said, I think we can probably do better here, but @nateprewitt or I will have to do some digging to see why we're missing the work of appending a response to the exception. Thanks for the report and the detail provided. 🎉

jvanasco:
I did some digging myself and figured a few things out...

1. The "max retries" text is from `requests/adapters.py` lines 497-514.  `urllib3` reports the connection timeout as a `MaxRetryError`, then wraps that error in a `ConnectionError` from `requests/exceptions.py`.  

2. Based on your info above, I traced a bunch of stuff back to resolve_redirects - which i'm very familiar with.  I will generate a PR of a starting point for a possible approach (it's just going to be a rough sketch).  I'm not familiar with enough of requests to take this on myself, but I may be able to jumpstart someone else.

PR #4456, opened 2018-01-10
an idea on persisting redirect history when exceptions are encountered

jvanasco:
this is a quick idea on an approach to address #4455

within `resolve_redirects` exceptions raised by`Session.send` are caught. If they match a given set of tests (currently just the base `ConnectionError` class), the `resp` is replaced with an instance of an exception class used to hold bad response data.  Then a new exception is raised. 

This makes handling these types of errors similar to handling items without redirects for most users (i did not consider the legacy ways of using requests)

The immediate ideas/concerns I have on this:

1. It might make sense to have a global "NotARealResponse" object that stands in for a `Response` object and can contain connection/history info when there is no immediate response.  I'm not sure this is the only use-case for this.

2. I thought about extending the current exception with request/response data, but this seemed more appropriate.

Just to be clear, this is only submitted for discussion/illustration of the other ticket.  This is not a PR for merging.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -137,12 +137,6 @@ class SessionRedirectMixin(object):
         while url:
             prepared_request = req.copy()
 
-            # is the response a redirected exception? If so, exit early.
-            # this should only match if set within the loop
-            if isinstance(resp, _RedirectedExceptionData):
-                # it would be best to raise the wrapped exception with these new attributes. but for now this works.
-                raise ConnectTimeout('The was an error on a redirected url.', response=resp, request=req)
-
             # Update history and keep track of redirects.
             # resp.history must ignore the original request in this loop
             hist.append(resp)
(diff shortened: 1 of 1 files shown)


requests/sessions.py
```diff
@@ -47,6 +48,16 @@
     preferred_clock = time.time
 
 
+class _RedirectedExceptionData(Exception):
```
sigmavirus24, line 51: I understand this is all just an idea, but I'm curious why this inherits from Exception here.

jvanasco: Just an artifact.  An earlier version was `_RedirectedException()` and was raised. After I got things working, I realized i should try to standardize behavior between an error in redirect chain (which requests detects) and an error in an immediately requested url (which urllib3 detects).  That meant a placeholder object (not raised) and the led to the idea of having a faux response-like object (which could be appropriate for exceptions raised in `adapters.py`.

sethmlarson:

PR #4459, opened 2018-01-11
Pass python_requires argument to setuptools

jdufresne:
Helps pip decide what version of the library to install.

https://packaging.python.org/tutorials/distributing-packages/#python-requires

> If your project only runs on certain Python versions, setting the
> python_requires argument to the appropriate PEP 440 version specifier
> string will prevent pip from installing the project on other Python
> versions.

https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords

> python_requires
>
> A string corresponding to a version specifier (as defined in PEP 440)
> for the Python version, used to specify the Requires-Python defined in
> PEP 345.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,6 +72,7 @@ setup(
     package_data={'': ['LICENSE', 'NOTICE'], 'requests': ['*.pem']},
     package_dir={'requests': 'requests'},
     include_package_data=True,
+    python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     install_requires=requires,
     license=about['__license__'],
     zip_safe=False,


sigmavirus24:
Is there a minimum version of setuptools necessary to support this?

sethmlarson:
@sigmavirus24 It was [added in 24.2.0](https://github.com/pypa/setuptools/blob/56bf1a9c12fd881fe6f69c3693cc426acf8012c3/CHANGES.rst#v2420) which was released on [2016-07-20](https://pypi.python.org/pypi/setuptools/json).

jdufresne:
Additionally, `setup()` ignores unknown keyword arguments. So, while only users with setuptools 24.2+ can take advantage of this feature, it is no harm to users of previous versions.

sigmavirus24:
@jdufresne can you update the branch? I'm totally on-board with this given your commentary. Cheers!

jdufresne:
I have rebased on to the latest master. Thanks.

PR #3733, opened 2016-11-29
remove weird characters from README.rst

jorgesumle:

PR #4490, opened 2018-01-31
removed invalid characters

royopa:


nateprewitt:
Hey @royopa, these are valid utf-8 characters and intended to be in the document. You can find more information in #3733 which attempted to do that same. Thanks for checking in on this though!

PR #4496, opened 2018-02-02
Pass kwargs from adapter send call to add_headers, per documentation

jonblum:
The docstring for `HTTPAdapter.add_headers` implies that keyword args from the `send` call will be passed into `add_headers` – this change passes them in.  This would be helpful, particularly for kwargs like `timeout`, which may influence the headers that need to be added by an `HTTPAdapter` subclass.


sigmavirus24:
This looks reasonable. Thanks @jonblum

PR #4497, opened 2018-02-03
Clarify docs that Requests always uses certs from certifi

jdufresne:
Since commit 0d7de6430eef0cf09f9662746daf0c28d83f144e, certifi is always used for certificates. Certify became a hard dependency of the package in 628633143d5b8590b1dbdf5371fe81fb8250dffd.

Now update the docs to clarify that Request will always use certificates from certifi.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -274,15 +274,14 @@ If you specify a wrong path or an invalid cert, you'll get a SSLError::
 CA Certificates
 ---------------
 
-By default, Requests bundles a set of root CAs that it trusts, sourced from the
-`Mozilla trust store`_. However, these are only updated once for each Requests
-version. This means that if you pin a Requests version your certificates can
-become extremely out of date.
+Requests uses certificates from the package `certifi`_. This allows for users
+to update their trusted certificates without changing the version of Requests.
 
-From Requests version 2.4.0 onwards, Requests will attempt to use certificates
-from `certifi`_ if it is present on the system. This allows for users to update
-their trusted certificates without having to change the code that runs on their
-system.
+Before version 2.16, Requests bundled a set of root CAs that it trusted,
+sourced from the `Mozilla trust store`_. The certificates were only updated
+once for each Requests version. When ``certifi`` was not installed, this led to
+extremely out-of-date certificate bundles when using significantly older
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hey @jdufresne, thanks for updating the docs! I think this is good for the current version of Requests. I do wonder if we may want some historical information though. Noting this changed with the release of 2.16 and leaving a small blurb on how it worked prior may be helpful.

jdufresne:
I added the suggested blurb. Let me know what you think. I'm happy to tweak or reword if you think it could be improved. Thanks.

jdufresne:
Updated to remove suggested lines. Thanks.

jdufresne:
Rebased to resolve merge conflicts.

docs/user/advanced.rst
```diff
@@ -274,15 +274,13 @@ If you specify a wrong path or an invalid cert, you'll get a SSLError::
...
-from `certifi`_ if it is present on the system. This allows for users to update
-their trusted certificates without having to change the code that runs on their
-system.
+Requests uses certificates from the package `certifi`_. This allows for users
+to update their trusted certificates without changing the version of Requests.
+
+Before version 2.16, Requests bundled a set of root CAs that it trusted,
+sourced from the `Mozilla trust store`_. The certificates were only updated
+once for each Requests version. This meant that the certificates of a pinned
+version of Requests could become extremely out of date.
```
sigmavirus24: They would be out of date if and only if users didn't separately install certifi. This is a blanket statement that is plainly untrue.

jdufresne: This is why I included "could". It is true under some conditions, but not others.

How would you like to see it altered?

What about?

> This meant that the certificates of a pinned version of Requests could become extremely out of date if certi was not installed.

Or should the sentence be removed entirely? Any other suggestions?

sigmavirus24: I think the `could` is too subtle. What about:

```
When `certifi` was not installed, this led to extremely out-of-date certificate bundles when using significantly older versions of Requests.
```

jdufresne: Sounds good and works for me. I've changed the wording as suggested in the latest revision of the PR.

PR #4501, opened 2018-02-05
Separate collections from collections.abc

rhettinger:
In Python 3.7, there is a warning that the abstract bases in collections.abc will no longer be accessible through the regular collections module in Python 3.8.  This patch future-proofs requests.

I wasn't able to run the test suite locally, so this needs a thorough test going forward.

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -43,6 +43,7 @@ if is_py2:
     import cookielib
     from Cookie import Morsel
     from StringIO import StringIO
+    from collections import Callable, Mapping, MutableMapping
 
     from urllib3.packages.ordered_dict import OrderedDict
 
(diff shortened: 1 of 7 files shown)


requests/compat.py
```diff
@@ -60,6 +61,7 @@
     from http.cookies import Morsel
     from io import StringIO
     from collections import OrderedDict
+    from collections.abc import Callable, Mapping, MutableMapping
```
sigmavirus24, line 64: @rhettinger on which versions of Python 3 will this actually succeed on? It was my impression that `collections.abc` was introduced in 3.6 or 3.7.

sigmavirus24:
This branch is also out-of-date with master and will need to be updated

requests/compat.py
nateprewitt: @sigmavirus24 I'd tested it back to 3.4 without issue, so I think this should work.

rhettinger: This works back to Python 3.3.

nateprewitt:
Hey @rhettinger, would you mind rebasing this onto current master? Assuming @sigmavirus24 doesn't have further issues, I think we're set to merge this.

PR #4505, opened 2018-02-07
use dumb reason decoding on Python 2

### LINKED ISSUE #4489: UnicodeEncodeError on localised response.reason message (bad use of IOError)
## Expected result

Python 2.7, requests 

A website returns 'Non-Autoris\xe9' (non authorized, in French) as response.reason on a 401 error. Before #3554 (https://github.com/requests/requests/pull/3554/files) a response.raise_for_error() returns this : 
```python
import requests
resp = requests.get('http://www.example.net/admin/')  # a website returning « 401 Non Autorisé »
try:
    resp.raise_for_error()
except Exception as e:
    pass
print e
```
401 Client Error: Non-Autoris for url: http://www.example.net/admin/

(\xe9 is just removed, not really an issue)

## Actual Result

After #3554 : 

UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 29: ordinal not in range(128)

## Why ?

It's because HTTPError is in fact an IOError : the first init argument can't be a unicode string. See https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError

```python
>>> unicode(IOError(12, u'é'))
u'[Errno 12] \xe9'
>>> unicode(IOError(u'é'))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 0: ordinal not in range(128)
>>> unicode(Exception(u'é'))
u'\xe9'
```

thomasnoel:
Try to fix #4489 : Python 2.x IOError cannot handle unicode argument, and HTTPError is a IOError.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -914,14 +914,17 @@ class Response(object):
 
         http_error_msg = ''
         if isinstance(self.reason, bytes):
-            # We attempt to decode utf-8 first because some servers
-            # choose to localize their reason strings. If the string
-            # isn't utf-8, we fall back to iso-8859-1 for all other
-            # encodings. (See PR #3538)
-            try:
-                reason = self.reason.decode('utf-8')
-            except UnicodeDecodeError:
-                reason = self.reason.decode('iso-8859-1')
+            if is_py2:
+                reason = self.reason.decode('utf-8', 'ignore')
+            else:
+                # We attempt to decode utf-8 first because some servers
+                # choose to localize their reason strings. If the string
+                # isn't utf-8, we fall back to iso-8859-1 for all other
+                # encodings. (See PR #3538)
+                try:
+                    reason = self.reason.decode('utf-8')
+                except UnicodeDecodeError:
+                    reason = self.reason.decode('iso-8859-1')
         else:
             reason = self.reason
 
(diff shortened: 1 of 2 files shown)


requests/models.py
```diff
@@ -914,14 +914,17 @@ def raise_for_status(self):
...
-            # We attempt to decode utf-8 first because some servers
-            # choose to localize their reason strings. If the string
-            # isn't utf-8, we fall back to iso-8859-1 for all other
-            # encodings. (See PR #3538)
-            try:
-                reason = self.reason.decode('utf-8')
-            except UnicodeDecodeError:
-                reason = self.reason.decode('iso-8859-1')
+            if is_py2:
+                reason = self.reason.decode('utf-8', 'ignore')
```
sigmavirus24, line 918: If `IOError` will not accept unicode on Python 2, why are we bothering to decode it at all here?

thomasnoel: Don't really know, only try to get back to pre- #3554 situation. I'm not good about encoding & such...

You can reject this PR as requests is OK with Python 3, and it's 2018.

sigmavirus24: Sorry, let me try to be clearer:

```py
>>> # Python 2
>>>'foo'.decode('utf-8')
u'foo'
```

```py
>>> # Python 2
>>> u'foo'.encode('utf-8')
'foo'
```

I think you want `encode` in this case

MrKiven: it's like a formula in py2... just, `str.decode -> unicode` & `unicode.encode -> str`

nateprewitt:
I think we’re unlikely to fix this at this point given the usage of Python 2.7 and lack of activity on the PR.

Issue #4434, opened 2017-12-20
iOS issue 'No child processes'


mucke5:
I am using Kodi and got the following error. Could you advise me how to solve the issue?

System: iOS 9.3.3 | Kodi 17.6 (arm64)

File "/var/mobile/Library/Preferences/Kodi/addons/script.module.requests/lib/requests/__init__.py", line 97, in <module>
from . import utils
File "/var/mobile/Library/Preferences/Kodi/addons/script.module.requests/lib/requests/utils.py", line 42, in <module>
if platform.system() == 'Windows':
File "/Users/Shared/jenkins/workspace/IOS-ARM64/tools/depends/xbmc-depends/iphoneos8.1_arm64-target/lib/python2.7/platform.py", line 1263, in system
File "/Users/Shared/jenkins/workspace/IOS-ARM64/tools/depends/xbmc-depends/iphoneos8.1_arm64-target/lib/python2.7/platform.py", line 1230, in uname
File "/Users/Shared/jenkins/workspace/IOS-ARM64/tools/depends/xbmc-depends/iphoneos8.1_arm64-target/lib/python2.7/platform.py", line 965, in _syscmd_uname
IOError: (10, 'No child processes')

sigmavirus24:
This is an odd error. At the moment we're not compatible with iOS and that was never really a target that we defined as wanting to be compatible with. It seems that Kodi doesn't allow python to spawn subprocesses in order to determine the system name. As such, when we try to make decisions based on the Operating System, we fail out. I don't think there's presently an ideal way of fixing this.

mucke5:
I tried some approaches from here: https://www.bountysource.com/issues/51501895-kodi-17-5-ios-2-18-4-failing-for-youtube-addon

I can open the broken Addon 'Backup' immediately after installing:
1- Release v2.13.0
2- Release v2.18.4 with replaced code [if platform.system() == 'Windows'] to [if 'win32' in sys.platform]

Restart Kodi leads to a crash. Therefore I need to delete the folder [script.module.requests] manually to use Kodi again. So these are only temporary solutions.

Is that a possible fix to make decisions based on the Operating System?
https://github.com/analogue/mythbox/blob/master/resources/src/mythbox/platform.py#L31

sigmavirus24:
[...]

PR #4510, opened 2018-02-09
Fix for iOS

mucke5:
In addition to  #4434 I found a possible solution not only for Kodi. Could you try such idea jsergio123/script.module.resolveurl@2e3f773 to solve the iOS issue 'No child processes' ?


sigmavirus24:
@mucke5 did you mean to open a pull request here?

PR #4513, opened 2018-02-14
Copy cookie policy when copying a CookieJar

laurentb:


nateprewitt:
Hey @laurentb, thanks for contributing to Requests! This is something we've looked to support in the past but it's had some nuances that have prevented possible solutions being merged.

When we perform a [request](https://github.com/requests/requests/blob/3c1d36b827417fdeaf5a1c106129de30dac371d7/requests/sessions.py#L423) (or [redirect](https://github.com/requests/requests/blob/3c1d36b827417fdeaf5a1c106129de30dac371d7/requests/sessions.py#L189)), a function called `merge_cookies` will create a new cookie jar and then merge both the session and local cookies together. This loses the policy information because we aren't guaranteed both objects being RequestsCookieJars. #3463 was an attempt to solve some of this issue but there was some pushback because session-level cookies are problematic in regards to domain specifications and security.

All that said, if you're really just looking to carry over the policy in the `copy` method, I think something we could consider. We may want to create a wrapper on RequestsCookieJar though so we're not directly calling a CookieLib private variable. We should probably remove the second test too because it suggests this patch fixes something it doesn't. If you run the test without the policy change, it will still pass.

laurentb:
Thanks, I have indeed a weird issue with redirections and cookies that I'm trying to fix. It's good to see there are issues similar to mine, I found the `merge_cookies` too but wasn't sure what to do with it yet.

In the meantime I'm at least removing my useless test.

Since I'm already using a subclass of RequestsCookieJar (https://github.com/laurentb/weboob/blob/ae2a105b61d86072248aeb269bfb6049f0b4632e/weboob/browser/cookies.py#L44) I don't really need the fix in copy(). We do a lot of stuff around requests (e.g. https://github.com/laurentb/weboob/blob/ae2a105b61d86072248aeb269bfb6049f0b4632e/weboob/browser/browsers.py#L323) that may also explain why the fix is enough for me.

> We may want to create a wrapper on RequestsCookieJar though so we're not directly calling a CookieLib private variable.

Yes, that bugged me too :) I'm adding a last commit to make it public. I suppose that one would be useful to everyone, so I can split it off to a separate pull request if needed.

nateprewitt:
Kicking tests. If you get a chance can you rebase this onto master, @laurentb?

nateprewitt:
Hi again @laurentb, would you mind rebasing this one more time for us. Apologies on the delays.

/cc @sigmavirus24 if you feel like giving this a quick peak. I think I'm still happy with where this is at though.

laurentb:
Done

PR #4522, opened 2018-02-26
Add PURGE verb

val314159:
Many packages have established PURGE as a defacto HTTP verb for flushing caches.
This implements the new verb.


nateprewitt:
Hey @val314159, thanks for putting this together and contributing to Requests! Unfortunately, Requests has a frozen API because we consider it mostly feature complete. You can see in [this section](http://docs.python-requests.org/en/master/user/advanced/#custom-verbs) of the documentation how to use custom verbs with Requests, which we've deemed sufficient for the time being. Hopefully that helps with your use case.

PR #4546, opened 2018-03-15
3.0

kennethreitz:
I'll be logging what I'm working on in this pull request. 

So far: 

- Removed all Python 2-specific code.
- Ran codebase through Black (White, actually) (new `setup.py format` command).
- Converted many `str.format`s to f-strings.
- Removed many "if bytes" code paths.
- Integrated rfc3986 for URL Parsing (needs to be done in `utils.py` still.
- Integrated new `HTTPHeaderDict`. 
- Typed out `api.py` and `utils.py` (I think, correctly), added a new types module, `types.py`. 
- Updated test suite to use a session fixture. 
- Added `__slots__` to most heavily-used classes.

**This pull request is not a request for feedback.**

All tests passing.

Think that's everything. Will add comments as I go.


kennethreitz:
Need to clean up AppVeyor, apparently.

kennethreitz:
AppVeyor's working! ✨🍰✨

kennethreitz:
Name–squatted "requests3" on PyPI. I'm liking that idea more and more for v3.

dsully:
Are the imports and checks of chardet, pyopenssl & cryptography still needed once Python 2 is dropped?

I ask, as my CLI applications are really sensitive to import time / human perception of how fast the command is, and these modules add a fair amount of import overhead, as found via "python3.7 -Ximporttime"

Thanks

sigmavirus24:
> Are the imports and checks of chardet, pyopenssl & cryptography still needed once Python 2 is dropped?

- chardet is required across Py2 and 3.
- `pyopenssl`/`cryptography` will likely be needed as we continue to support older Python 3 versions that aren't as secure as the latest.

As for the `checks`, I'm guessing you mean version checks. In which case, we will not drop those because it's our only mechanism for telling users they're using unsupported software configurations that mean we will not help them with bugs.

> I ask, as my CLI applications are really sensitive to import time / human perception of how fast the command is, and these modules add a fair amount of import overhead, as found via "python3.7 -Ximporttime"

Yes, cryptography adds import time and so does chardet. I'm surprised you haven't complained about idna as well since that also adds a lot of overhead just by importing it. There's little we can do to reduce that without outright removing features (automatic character encoding detection (chardet), security (pyopenssl/cryptography), internationalized domain name handling (idna)) that are standard for HTTP clients in any language.

dsully:
@sigmavirus24 I agree overall (and I'm not sure why idna didn't show up in my imports), however - for environments where none of that functionality is needed, it would be nice to have a way to disable those imports.

I'm happy to block them via a sys.modules hack, but the version checking & other usage code (such as "def apparent_encoding" assume these modules are there. If you would be open to selective loading or deferred loading, I can create a PR.

requests/structures.py
```diff
@@ -86,6 +86,128 @@ def __repr__(self):
         return str(dict(self.items()))
 
 
+class HTTPHeaderDict(CaseInsensitiveDict):
```
dbaxa, line 89: I would like to suggest making use of urllib3's `HTTPHeaderDict` instead o requests having a similar but different version.

nateprewitt:
Closing this as it hasn’t been actioned on in a few years.

PR #4560, opened 2018-03-27
Session timeout / Added the ability to set a default session timeout

yuvalbar:
This is a very minimal PR that aims to allow developers to assign a default timeout to requests made using `Requests`, while not breaking any existing usage of the library. I open it following bumping into the problem of a request hanging indefinitely several times and reading through the open issues - specifically @kennethreitz 's comment here regarding assigning a default timeout to the lib: https://github.com/requests/requests/issues/3070#issuecomment-318926146

In essence, it allows a user to set a default timeout for a `Session` object so that any request made using that session will default to the session timeout unless specifically assigned a timeout of its own. I attempted to circumvent the entire issue of breaking existing code or deciding on the right default timeout for the library itself, instead just giving the user a tool to set it per session where necessary, instead of once per request.

The only open issue I am aware of is that if a request is specifically assigned to never timeout (`timeout=None`) and the session it is sent from has a default timeout set, then the session timeout will take. However, I believe that in such a case it makes sense to send the request from a new session.

Happy to receive any feedback. Thanks in advance!

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -397,6 +397,9 @@ class Session(SessionRedirectMixin):
         self.mount('https://', HTTPAdapter())
         self.mount('http://', HTTPAdapter())
 
+        # Set default timeout
+        self.timeout = None
+
     def __enter__(self):
         return self
 
@@ -506,7 +509,7 @@ class Session(SessionRedirectMixin):
 
         # Send the request.
         send_kwargs = {
-            'timeout': timeout,
+            'timeout': timeout or self.timeout,
             'allow_redirects': allow_redirects,
         }
         send_kwargs.update(settings)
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Thanks for putting this together. I don’t think it will be accepted. In the past this library supported session-level timeouts but this proved to be unintuitive to most users as they expected various different things. We intentionally removed the feature, with prejudice. Adding it back at this time is unlikely.

yuvalbar:
@sigmavirus24 Thanks for addressing my PR so promptly. I find it hard to advocate adding this when I don't have any knowledge of past issues with such a mechanism. I failed to find resources for this when looking through the issues - are there any specific issue threads you can point me to? Is proper implementation of this feature in a non-confusing way something that you're even willing to discuss?

PR #4573, opened 2018-04-04
add HTTP 419 to requests.codes

### LINKED ISSUE #4572: Add HTTP 419 support to requests.codes
Very simply, I would like to have HTTP 419 status code represented as a constant in the `requests.codes` module, perhaps as `requests.codes.AUTHENTICATION_TIMEOUT`.

While this code is not officially (to my knowledge) part of an RFC / standard, it seems to have colloquial use / adoption on many Internet-facing web services.

Here's another framework who is considering this support (granted, the context is a bit different WRT requests):
GetBlimp/django-rest-framework-jwt#158

I'll post a basic PR to illustrate the request; no pressure to merge it as I am unsure what requests' policy on RFCs/standards is (whether or not dogmatic adherence thereto is a tenet/goal of this project).

mxmader:
via status_codes.py.
fixes #4572

### DIFF
--- a/requests/status_codes.py
+++ b/requests/status_codes.py
@@ -70,6 +70,7 @@ _codes = {
     416: ('requested_range_not_satisfiable', 'requested_range', 'range_not_satisfiable'),
     417: ('expectation_failed',),
     418: ('im_a_teapot', 'teapot', 'i_am_a_teapot'),
+    419: ('authentication_timeout', 'auth_timeout'),
     421: ('misdirected_request',),
     422: ('unprocessable_entity', 'unprocessable'),
     423: ('locked',),


nateprewitt:
Hey @mxmader, thanks for taking the initiative to put this together and start the conversation While there are some uses of 419 on the web, it’s not standardized as you noted in the related issue. I’d personally prefer we avoid introducing things like this (there are a handful of non-standard status codes) into Requests. Primarily because we’ll get follow on PRs asking where the line is for status code x, y, z and if the IETF ever decides to use any of these codes we can’t “unrelease” versions of Requests with bad constants. Since this constant list has no effect on how Requests handles a received 419, I think we’re likely to pass on this. I’ll let @sigmavirus24  weigh in before we close this though.

mxmader:
no problem. this PR was part thought experiment, part search for convenience. i'll close it.

Issue #4557, opened 2018-03-26
wrong condition returning true only if there's the same domain repeated


kamiswim:
https://github.com/requests/requests/blob/b66908e7b647689793e299edc111bf9910e93ad3/requests/cookies.py#L295

nateprewitt:
Hi @kamiswim, you’re right that this doesn’t look correct. The behavior doesn’t match the doctstring and doesn’t make much sense.

We don’t utilize domains in our cookie implementation but it’s been a goal for a while. It’s likely this method has seen little usage which is why it’s escaped notice this long. I think we should definitely fix this in 3.0 but I’m not sure if we have users that are depending on this current behavior now.

If you would like to open a PR fixing this to the proposed/3.0.0 branch, we’re happy to review that. That is unless @sigmavirus24 thinks we should fix it on master.

inishchith:
closing due to inactivity

PR #4581, opened 2018-04-10
Update to return correct value for multiple_domains().

ghost:
requests/cookies.py

Adjust flow to add domains to list as the first operation in the loop, then convert to a set and check the length of that set is greater than 1.

Add tests to pass for a single domain, two cookies with the same domain, and multiple domains.

Perhaps not the most efficient method, but it works.

### DIFF
--- a/requests/cookies.py
+++ b/requests/cookies.py
@@ -290,11 +290,13 @@ class RequestsCookieJar(cookielib.CookieJar, collections.MutableMapping):
 
         :rtype: bool
         """
-        domains = []
+        domains = set()
         for cookie in iter(self):
-            if cookie.domain is not None and cookie.domain in domains:
+            if cookie.domain:
+                domains.add(cookie.domain)
+
+            if len(domains) > 1:
                 return True
-            domains.append(cookie.domain)
         return False  # there is only one domain in jar
 
     def get_dict(self, domain=None, path=None):
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -452,6 +452,24 @@ class TestRequests:
         # Make sure the cookie was sent
         assert r.json()['cookies']['foo'] == 'bar'
 
+    def test_cookies_multiple_domains_returns_correct_value(self, httpbin):
+        """Tests that multiple_domains() returns the correct value. 
+
+        See GH #4557
+        """
+        s = requests.session()
+        r = s.get(httpbin('cookies'))
+
+        r.cookies.set('foo', 'bar', domain='.google.com')
+        assert r.cookies.multiple_domains() == False
+
+        r.cookies.set('baz', 'foo')
+        assert r.cookies.multiple_domains() == False
+
+        r.cookies.set('baz', 'foo', domain='.google.com')
+        r.cookies.set('zin', 'foo', domain='.github.com')
+        assert r.cookies.multiple_domains() == True
+
     def test_cookielib_cookiejar_on_redirect(self, httpbin):
         """Tests resolve_redirect doesn't fail when merging cookies
         with non-RequestsCookieJar cookiejar.
--- a/3.0-HISTORY.rst
+++ b/3.0-HISTORY.rst
@@ -73,5 +73,8 @@
 
 - ``Response.raise_for_status()`` now returns the response object for good responses
 
+- ``RequestCookieJar.multiple_domains()`` will return the correct boolean value
+  when exectued. See `#4557`_.
+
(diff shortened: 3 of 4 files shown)


ghost:
#4557

requests/cookies.py
```diff
@@ -292,9 +292,11 @@ def multiple_domains(self):
         """
         domains = []
         for cookie in iter(self):
-            if cookie.domain is not None and cookie.domain in domains:
-                return True
             domains.append(cookie.domain)
+            domains_set = set(domains)
```
nateprewitt: A few quick notes here.

We’re rebuilding the set every loop which probably isn’t necessary. Since it’s serving the same purpose of `domains`, what do you think about making `domains` a set?

The other issue I’m seeing is the conditional. Let’s take an example set of domains `['google.com', None, 'google.com']`. The first two loops will add both 'google.com' and None to `domains` and the third loop will satisfy the criteria despite there only being one domain. It may be better if we only add if the cookie.domain is not None and then have the single check for `len(domains)` being > 1

ghost: I think you're correct there, I'll update the pr tomorrow.

ghost:
@nateprewitt Thank you kindly for the helpful feedback!

I've adjusted the method to make use of set in place of the domains list.  I've also adjusted the structure of the conditional a bit and added an additional test check that null values are handled correctly by the method.

tests/test_requests.py
```diff
@@ -385,6 +385,27 @@ def test_param_cookiejar_works(self, httpbin):
...
+        r = s.get(httpbin('cookies'))
+        r.cookies.set('foo', 'bar', domain='.google.com')
+
+        assert r.cookies.multiple_domains() == False
+
+        r.cookies.set('baz', 'foo')
+
+        assert r.cookies.multiple_domains() == False
+
+        r.cookies.set('baz', 'foo', domain='.google.com')
```
nateprewitt: If we’re going to add this to the test let’s verify it’s working as expected. 

We can probably collapse a few of the new lines too if you don’t think it affects readability. Perhaps no extra newline between the cookies.set and assert lines?

ghost: Is it worth a new test method to make testing the ['.google.com',None,'.google.com'] case a bit less awkward?

Also, when you say let's verify it's working as expected are you referring to the test itself or the method being tested?

nateprewitt: If you think that makes it easier to grock, I don’t have an issue with it. I’d encourage a glance at pytest’s `parameterize` decorator which will allow you to run the same test with different input.

ghost:
@nateprewitt You're very welcome, sir.

The other updates you requested have been made.

nateprewitt:
@gahancorpcfo, It looks like you may have accidentally merged master into your commit. Would you mind only including the changes related to this PR. The easiest way to do this is likely to reset HEAD~1 and then `git add -p your-relevant-files` or cherry pick from another branch if you have it.

ghost:
I believe I've managed to produce a single commit with the relevant changes that was not merged with master. 

. . . I hope. 

@nateprewitt Thank you very much for your help with this.

nateprewitt:
@gahancorpcfo it looks like you still have some extra code in cookies.py. Would you mind removing that? `git add -p` will allow you to easily select the chunks you need.

ghost:
@nateprewitt 

The extra code has been removed from the commit now, it should be good to go.

nateprewitt:
@gahancorpcfo thanks so much for seeing this through! We have an issue right now with our Windows build environment that's failing on Python 3.4, so I'm going to wait to merge this. We'll get this in once that's resolved.

As for the issue, it looks like https://github.com/jonparrott/cmarkgfm has had issues with Windows wheel builds recently between versions 0.2.0 and 0.4.0. I'll need to reach out to Jon to see if this is supposed to be fixed. That is unless someone else would like to spend some time digging and opening a relevant issue with the repository.

nateprewitt:
Just as a follow up note, jonparrott/cmarkgfm#2 is the issue we're hitting right now. Until a wheel is built for 3.4 (hopefully in the next week), 3.4 won't be passing on Windows under appveyor.

ghost:
@nateprewitt Thanks for the heads up!  I'm happy to help.

theacodes:
Sorry it took so long, but 3.4 wheels should now be available for Windows. :)

nateprewitt:
No worries, thanks @jonparrott!

sethmlarson:

Issue #4579, opened 2018-04-09
os.environ lowercase takes precedence over uppercase


ZuZuD:
I wonder if there is a specific reason for the lowercase http_proxy to take precedence over its uppercase version (same behavior for the other os.environ). The convention is to use uppercase for environment variables and setting both could potentially lead users to unexpected behavior.

https://github.com/requests/requests/blob/master/requests/utils.py#L699

Can someone give me the original reason why things have been done like this?

Thanks!

## System Information

    $ python -m requests.help
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "2.7.13"
  },
  "platform": {
    "release": "4.9.85-38.58.amzn1.x86_64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.18.4"
  },
  "system_ssl": {
    "version": "100020bf"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": false
}

nateprewitt:
Hi @ZuZuD, to clarify, your concern is about `no_proxy` being given precedence over `NO_PROXY`? I believe the lambda function was designed to give preference to whatever casing it was handed first, and then attempt to find an uppercased version if that did not work.

The `no_proxy` check was added later (#945) and I believe chose lowercase for scoping issues. A lowercased environment variable is conventionally used to represent a "local" scope environment variable, where as an uppercased name is considered global and likely spans sessions. We're giving preference to the local scope first before looking for something in global scope, because it's considered to be more specific.

ZuZuD:
Thank you for your complete answer @nateprewitt ! By curiosity I took a look at other http libraries/software to assess their behavior regarding those env variables.

1) curl: lowercase takes precedence over uppercase https://github.com/curl/curl/blob/master/lib/url.c#L2659

```
$ ~/github/curl export HTTPS_PROXY=0.0.0.0
$ ~/github/curl curl -i https://github.com
$ curl: (7) Failed to connect to 0.0.0.0 port 1080: Connection refused

$ ~/github/curl export https_proxy=localhost
$ ~/github/curl curl -i https://github.com
$ curl: (7) Failed to connect to localhost port 1080: Connection refused
```
[...]

PR #4583, opened 2018-04-11
Add a comment about lowercase env. variables precedence over uppercase

ZuZuD:
As suggested by @sigmavirus24 for the follow-up of issue #4579

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -696,6 +696,8 @@ def should_bypass_proxies(url, no_proxy):
 
     :rtype: bool
     """
+    # Prioritize lowercase environment variables over uppercase
+    # to keep a consistent behaviour with other http projects (curl, wget).
     get_proxy = lambda k: os.environ.get(k) or os.environ.get(k.upper())
 
     # First check whether no_proxy is defined. If it is, check that the URL


requests/utils.py
```diff
@@ -696,6 +696,8 @@ def should_bypass_proxies(url, no_proxy):
 
     :rtype: bool
     """
+    # Prioritize lowercase environment variables over uppercase
+    # to keep a consistent behaviour with other http projects
```
nateprewitt: Would you mind specifying the two you checked (curl, wget) so we have some concrete examples?

ZuZuD:
Sure @nateprewitt that's updated!

nateprewitt:
It also looks like we have a windows failure in the tests but it seems like it may be related to a dependency or pipenv issue. Not really relevant for a comment change but something we’ll want to fix.

ZuZuD:
Let's keep it concise then @nateprewitt 👍

nateprewitt:
Thanks @ZuZud! Kicking the tests really quickly to verify this is an actual issue with 3.4 and not transient.

nateprewitt:
Thanks @ZuZuD!

PR #4586, opened 2018-04-12
allowing params to get requests on session

ggalves:
Simple fix to get method of Session class allowing it to receive params, just as requests.get would allow

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -514,7 +514,7 @@ class Session(SessionRedirectMixin):
 
         return resp
 
-    def get(self, url, **kwargs):
+    def get(self, url, params=None **kwargs):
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Hi @ggalves Can you share a bit more about what makes this PR necessary? `Session.get` already accepts `params` via `**kwargs` and passes them on appropriately to the best of my knowledge. Is something not behaving as you expect it to?

ggalves:
Hi @sigmavirus24, I'm sorry, you are right, it's unnecessary.

I couldn't reproduce my problem anymore, i did the PR on a hurry and must have mislooked something lol. I'll try later to reproduce it and open an issue if necessary. 

You can close this PR, my apologies.

PR #4591, opened 2018-04-16
Use pyOpenSSL only when ssl doesn't support SNI

methane:
Since recent Python supports SNI, importing pyOpenSSL
is avoidable now.


sigmavirus24:
I don't believe we only use pyOpenSSL to fix up SNI issues anymore. This needs careful review

sethmlarson:
This has already been completed in https://github.com/psf/requests/pull/5443

PR #4596, opened 2018-04-19
Update all pypi.python.org URLs to pypi.org

jdufresne:
For details on the new PyPI, see the blog post:

https://pythoninsider.blogspot.ca/2018/04/new-pypi-launched-legacy-pypi-shutting.html


sigmavirus24:
I think we can merge this without AppVeyor passing.

PR #4603, opened 2018-04-21
Nicer URL formatting with lists

sayanarijit:
# Nicer URL formatting with lists

### Purpose

Using `['https://example.com', route1, route2]` as URL parameter as compared to `'https://example.com/{}/{}'.format(route1, route2)` or `'https://example.com/'+route1+'/'+route2`

### Examples

```python
import requests

slug = 1
r = requests.get(['https://api.jikan.me/anime', slug])
r.url
# 'https://api.jikan.me/anime/1'

category, item = 'jokes', 'random'
r = requests.get(['https://api.chucknorris.io', category, item])
r.url
# 'https://api.chucknorris.io/jokes/random'
```

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -9,6 +9,7 @@ requests (cookies, auth, proxies).
 """
 import os
 import sys
+import six
 import time
 from collections import Mapping
 from datetime import timedelta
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @sayanarijit, thanks for the interest in contributing to Requests! Unfortunately, the Requests API is more or less frozen at this point, so I don't believe we'll be able to merge this. It should be easy to accomplish what you're looking for by wrapping the list in a join though, so the additional overhead for users wishing to do this is minimal. Thanks again for putting this out there!

Issue #3977, opened 2017-04-20
No Content-Length header when POSTing or PUTing an empty Django InMemoryUploadedFile


CianciuStyles:
I'm trying to send a PUT request passing as the data kwarg an instance of a InMemoryUploadedFile (which is the model Django uses for uploaded files, see for reference https://docs.djangoproject.com/en/1.10/ref/files/uploads/#django.core.files.uploadedfile.InMemoryUploadedFile), like this:

```
files = request.FILES.getlist('file')
for uploaded_file in files:
    [...]
    requests.put(upload_url, data=uploaded_file)
```
The problem is that the resulting request will not have the expected Content-Length header, but the Transfer-Encoding one. I think the problem is at https://github.com/kennethreitz/requests/blob/master/requests/models.py#L492, where requests sets one header or the other based on the `if length` check. In the case of a stream with 0 length, the check fails and the Transfer-Encoding header gets set.

The snippet above works as intended when the file is non-empty (because the `if length` check passes), and also `requests.put(upload_url, data=uploaded_file.read())` works (as this is not a stream anymore, but it could cause problems when reading a huge file in memory).

Lukasa:
Why is the transfer encoding header a problem? HTTP/1.1 servers are required to understand it.

CianciuStyles:
The problem is not in the Transfer-Encoding header, but the lack of the Content-Length header: if I try to upload the file to Amazon S3 (for example), it will reply with a 411 Content Length Missing response.

Lukasa:
Ok, so this boils down to an S3 behaviour, as S3 requires the content length. 

Requests does not in general consider this behaviour a bug: I consider S3 at fault here. However, it's understandably a problem. You should be able to get the behaviour you want by writing a small proxy object that sends all `read` calls to the Django object and that implements `__len__`, which will be enough for us to work out its length.

Lukasa:
Oh I apologise. You're posting an empty file. Requests really does not provide an easy way to override our logic here, which is very deliberate (plenty of users pass us file objects that return a length of zero but which nonetheless return data from `read()`). For this reason the easiest thing to do is likely to use the [PreparedRequest flow](http://docs.python-requests.org/en/master/user/advanced/#prepared-requests) to tweak the headers of the prepared request to suit this case.

sigmavirus24:
[...]

PR #4609, opened 2018-04-24
Fix handling of zero-length files

MarcSchmitzer:
This was previously fixed in e7c9bbb96 and broken again in
4c82dbab6fc. Add a comment this time to prevent people from breaking
it again.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -486,7 +486,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             if files:
                 raise NotImplementedError('Streamed bodies and files are mutually exclusive.')
 
-            if length:
+            if length is not None:  # length may be zero.
                 self.headers['Content-Length'] = builtin_str(length)
             else:
                 self.headers['Transfer-Encoding'] = 'chunked'
@@ -514,7 +514,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         """Prepare Content-Length header based on request method and body"""
         if body is not None:
             length = super_len(body)
-            if length:
+            if length is not None:
                 # If length exists, set it. Otherwise, we fallback
                 # to Transfer-Encoding: chunked.
                 self.headers['Content-Length'] = builtin_str(length)
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -158,10 +158,10 @@ def super_len(o):
                 except (OSError, IOError):
                     total_length = 0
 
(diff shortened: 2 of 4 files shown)


mschmitzer:
Ok, this actually runs afoul of some tests. Interesting.

nateprewitt:
Hi @MarcSchmitzer, we use content-length of 0 internally to represent data of indeterminate size. This is sent using `Transfer-Encoding: chunked`. It's advised you don't provide a Content-Length when using Requests, since it should always be handled internally. As we say in the [docs](http://docs.python-requests.org/en/master/user/quickstart/#custom-headers), we will override the Content-Length header at will.

If you MUST pass a custom content length for some reason, I'd suggest using the PreparedRequests workflow. You can modify the headers on the PreparedRequest to be what you'd like and then send it.

Thanks for taking the initiative to try to fix this though. I'm going to close this out since it's something we won't be changing in the near future.

nateprewitt:
I guess to clarify, if you've found what you think is an actual bug around zero length files, please feel free to open an issue. A few initial tests suggest everything is handled fine and we send `Content-Length: 0` when it's possible to determine.

mschmitzer:
Yes, I do believe this is a bug in requests. I am not setting the content-length header from the outside, but rather providing an empty file object as the request body. Will open an issue.

Thanks for looking at this anyway.

nateprewitt:
@MarcSchmitzer, "empty" file objects aren't necessarily always empty in Python, which is why we use `Transfer-Encoding: chunked`. This is a valid method for transferring the file to the server.

If your server or use case requires that a Content-Length must be passed, modifying the PreparedRequest is likely what you'll need.

Some other issues for reference before opening a new one:
#4404
#3977

mschmitzer:
Ok, thanks again for your responses.

If you're interested, the problem is in combination with boto3/s3transfer. You could say it is really their bug, only they're currently vendoring a requests version (2.7.0) that behaves differently and doesn't trigger the bug. I'm only running into it because Debian undoes the vendoring and the requests version on Stretch (2.12) triggers it. Guess I'm going to go talk to the boto3 people.

sigmavirus24:
That sounds like the boto3 are being explicit in what they're supporting and your linux distribution packager is ignoring that and saying "Hey this configuration seems to work!" which means that you're not really getting boto3. I'd be unsurprised if there were odd custom patches attached to your packages that you're getting as well such that requests isn't really requests either.

PR #4616, opened 2018-04-27
Add module docstring to __version__.py

Stevoisiak:
Adds a docstring to `__version__.py`.


kennethreitz:
this module is fine as–is, but thank you for you contribution!

PR #4618, opened 2018-04-27
Use multi-line import for __version__

Stevoisiak:


sigmavirus24:
Hi there @Stevoisiak !  Thanks for sending this PR along. We don’t accept stylistic changes, unfortunately as this project has its own style guide that we haven’t had a chance to document.

Stevoisiak:
@sigmavirus24 Should single-line imports over 100 characters be left alone as well?

https://github.com/requests/requests/blob/943a5c8e89db1758ae24adbbedacb3b05c32df4a/requests/compat.py#L58-L59

nateprewitt:
@Stevoisiak you can find the documentation on Kenneth Reitz’s code style [here]http://docs.python-requests.org/en/master/dev/contributing/#kenneth-reitz-s-code-style). We do allow lines over 100 and typically will not adjust the code base with linter changes unless there’s something significantly abnormal.

PR #4623, opened 2018-05-03
add long_description_content_type to setup.py

ngoldbaum:
This will make the readme render correctly on pypi.

https://packaging.python.org/tutorials/distributing-packages/#description

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,6 +65,7 @@ setup(
     version=about['__version__'],
     description=about['__description__'],
     long_description=readme + '\n\n' + history,
+    long_description_content_type='text/markdown',
     author=about['__author__'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -65,6 +65,7 @@ def run_tests(self):
     version=about['__version__'],
     description=about['__description__'],
     long_description=readme + '\n\n' + history,
+    long_description_content_type='text/markdown',
```
mgedmin, line 68: But readme is ReStructuredText!

mgedmin: The ReST rendering problem on PyPI was caused by an unescaped `*.whl` in the changelog entry for 2.18.1.  It's already fixed in master.

Issue #4613, opened 2018-04-26
Proxies from environment parsing problem


pbelskiy:
Hello, I've faced with a stupid problem, in my Ubuntu 16.04 in `/etc/environment` I wrote http proxy with end with one invisible space char at the end: `http_proxy=http://1.2.3.4:8080 `,and of course I didn't notice it when I just copy past proxies for my machines. 

And then when I used `requests.get()` I get a strange error:

```
Traceback (most recent call last):
  File "./t.py", line 5, in <module>
    x = requests.get("http://test/api")
  File "/usr/local/lib/python3.6/site-packages/requests/api.py", line 72, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/requests/api.py", line 58, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/requests/sessions.py", line 508, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python3.6/site-packages/requests/sessions.py", line 618, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/requests/adapters.py", line 405, in send
    conn = self.get_connection(request.url, proxies)
  File "/usr/local/lib/python3.6/site-packages/requests/adapters.py", line 303, in get_connection
    proxy_manager = self.proxy_manager_for(proxy)
  File "/usr/local/lib/python3.6/site-packages/requests/adapters.py", line 197, in proxy_manager_for
    **proxy_kwargs)
  File "/usr/local/lib/python3.6/site-packages/urllib3/poolmanager.py", line 440, in proxy_from_url
    return ProxyManager(proxy_url=url, **kw)
  File "/usr/local/lib/python3.6/site-packages/urllib3/poolmanager.py", line 385, in __init__
    proxy = parse_url(proxy_url)
  File "/usr/local/lib/python3.6/site-packages/urllib3/util/url.py", line 199, in parse_url
    raise LocationParseError(url)
urllib3.exceptions.LocationParseError: Failed to parse: 1.2.3.4:8080 
```
And then I found out that problem with `getproxies()` from `urllib` which used by `requests` for getting proxies. 

**Maybe we can just `strip()` each proxy string? It will be conveniently.**

## Expected Result

{'http': 'http://1.2.3.4:8080'}

## Actual Result

{'http': 'http://1.2.3.4:8080 '}

I got the string which ends with one space

[...]

PR #4627, opened 2018-05-04
Strip whitespace for proxy urls, test for same. #4613

ghost:
I'm not sure this is the right place to do this operation, but it does solve the problem. Or at least it passes the test that I wrote for it.

Quite likely I'm not reading them correctly, but as far as I can tell the other test methods related to proxies don't test something like a `requests.get()` call with the expectation of a successful result.

I found a ton of tests for error cases, though, and modeled the test for this situation on those, but it seems like there would be a better way.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -300,7 +300,7 @@ class HTTPAdapter(BaseAdapter):
         proxy = select_proxy(url, proxies)
 
         if proxy:
-            proxy = prepend_scheme_if_needed(proxy, 'http')
+            proxy = prepend_scheme_if_needed(proxy, 'http').strip()
             proxy_url = parse_url(proxy)
             if not proxy_url.host:
                 raise InvalidProxyURL("Please check proxy URL. It is malformed"
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -549,10 +549,16 @@ class TestRequests:
         with pytest.raises(InvalidProxyURL):
             requests.get(httpbin(), proxies={'http': 'http:///example.com:8080'})
 
-    def test_proxy_strip_spaces(self):
+    @pytest.mark.parametrize('proxies', (
+        {'http': 'localhost:1234'},
+        {'http': ' http://localhost:1234'},
+        None,
+        {'http': 'http://localhost:1234 '}
+    ))
+    def test_proxy_strip_spaces(self, proxies):
         # Don't yet have testing proxies sorted out.
         with pytest.raises(ProxyError):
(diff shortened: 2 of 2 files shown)


requests/adapters.py
```diff
@@ -300,7 +300,7 @@ def get_connection(self, url, proxies=None):
         proxy = select_proxy(url, proxies)
 
         if proxy:
-            proxy = prepend_scheme_if_needed(proxy, 'http')
+            proxy = prepend_scheme_if_needed(proxy, 'http').strip()
```
nateprewitt, line 303: We may want to call `strip` on the proxy itself before handing it to `prepend_scheme_if_needed`. Leading whitespace causes issues with `urlparse` and we get back a weird url which will fail later.

tests/test_requests.py
```diff
@@ -549,6 +549,11 @@ def test_proxy_error_on_bad_url(self, httpbin, httpbin_secure):
         with pytest.raises(InvalidProxyURL):
             requests.get(httpbin(), proxies={'http': 'http:///example.com:8080'})
 
+    def test_proxy_strip_spaces(self):
+        # Don't yet have testing proxies sorted out.
+        with pytest.raises(ProxyError):
+            requests.get('http://localhost:1', proxies={'http': 'http://localhost:1234 '})
```
nateprewitt: Would you mind adding a couple more tests for proxies without schemes, leading whitespace, and proxy values of `None`? You should be able to find examples of using pytest's `parametrize` decorator above to do all the tests in one function.

requests/adapters.py
sigmavirus24: I'm not even certain either of those are the right change, honestly, However, I can't think of a better change.

nateprewitt: I agree this isn’t great, but I’m not sure if we have a better place to catch both environmental and parameter proxies for sanitizing. We’d need to make a larger change to how proxies are handled I think if we wanted this to be done in a less hacky fashion.

nateprewitt: Sorry, let me redact that last statement. This doesn’t cover `request_url`, so we really need to do a single mutation on the proxies dict passed to send, or ensure the malformed uri exception is raised for the trailing white space.

ghost: Do you mean to do something like this?
```python
if proxy:
    proxy.strip()
```

tests/test_requests.py
ghost: - [ ] missing scheme
- [ ] leading whitespace
- [ ] proxy value None

sethmlarson:

PR #4629, opened 2018-05-04
Fix iter_lines boundary bug when CRLF straddles two chunks

nanotone:
In `Response.iter_lines`, if a `\r\n` sequence (the only multi-character line boundary) straddles two chunks, the `\r` and `\n` are treated as two separate boundaries. Consequently, `iter_lines` will yield one extra blank line, as an empty string.

This addresses the issue by checking if a chunk ends with `\r`, and when appropriate, leaving the trailing line in `pending` until the next chunk or EOF arrives.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -785,6 +785,7 @@ class Response(object):
         """
 
         pending = None
+        carriage_return = u'\r' if decode_unicode else b'\r'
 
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hi @nanotone, thanks for putting this together! This has actually already been addressed and is patched for Requests-3.0 (#3984). We won't be able to merge this into the current master branch because it's a breaking change for some users which isn't backwards compatible.

PR #4636, opened 2018-05-08
Add check proxies not none in line 683

### LINKED ISSUE #4635: Loop if 'proxies' is not None in requests/sessions.py line 679
function merge_environment_settings

kartikeyaj0:
fixes #4635
In function merge_environment_settings before entering the loop


nateprewitt:
Hi @kartikeyaj0, thanks for contributing to Requests. From the issue and this pull request it's not immediately clear how you're entering this state. Proxies will be assigned to an empty dictionary if `None` is passed to the session, and the docstring explicitly states it must be a dictionary.

In the event a user somehow circumvents this, I think it's better to make sure we blow up here rather than silently failing. For that reason, I don't think we're going to accept this patch at this time. If you have more information on your issue, or believe there's a compelling reason we're missing, please feel free to let us know. Thanks!

PR #4647, opened 2018-05-16
fix appveyor

nateprewitt:
I don't know what's going on with Appveyor/Pipenv but it's no longer liking `-e .` installs out of the Pipfile. I went back to a version of Pipenv that we have successful builds for and it's still broken. I'd like to get builds fixed while we determine root cause.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,24 +1,24 @@
 [[source]]
 url = "https://pypi.org/simple/"
 verify_ssl = true
+name = "pypi"
 
 [dev-packages]
-
 pytest = ">=2.8.0"
 codecov = "*"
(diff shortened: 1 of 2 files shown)


techalchemy:
@nateprewitt @sigmavirus24 I’m wondering if the old school format of extras mentioned in https://github.com/requests/requests/issues/4646#issuecomment-389411076 is breaking when parsing with pip10...

nateprewitt:
I thought it might be Pip 10 when the builds started failing but we have successful builds with the same Pipfile and Pip 10 from a few weeks ago.

https://ci.appveyor.com/project/kennethreitz/requests/build/1.0.537/job/stejqhugrbl4010i

PR #4648, opened 2018-05-16
bring Pipfile.lock up to date with current constraints

nateprewitt:
The Pipfile has strayed a bit from our current build constraints and contains information that's no longer consistent with Warehouse. Updating should help with installation and prevent problems like pypa/packaging-problems#147 from reoccurring when users pick Requests as a model for their packages.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,26 +1,14 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "72b5a08e9c266b930d308024036e928e6b99ed4b7a50f22af377a463b7867a14"
-        },
-        "host-environment-markers": {
-            "implementation_name": "cpython",
-            "implementation_version": "3.6.2",
-            "os_name": "posix",
-            "platform_machine": "x86_64",
-            "platform_python_implementation": "CPython",
-            "platform_release": "16.7.0",
-            "platform_system": "Darwin",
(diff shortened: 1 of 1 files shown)


sigmavirus24:
@nateprewitt if I remember correctly, you're a Pipenv maintainer and the semi-official position is that the project is geared towards applications, not libraries. If that's correct, do you recall the reasoning for adding it to Requests? I'm wondering if we need the hassle of constantly updating the lockfile when we're just a library.

kennethreitz:
It's just as valid for libraries as `requirements.txt` is.

sigmavirus24:
So Pipenv is pbr + requirements.txt for libraries then?

nateprewitt:
@sigmavirus24 I think the reason the term “application dependencies” was chosen is because Pipfile is not (currently) a replacement for setup.py. The Pipfile/Pipfile.lock can be used in a library similar to how requirements.txt is used but won’t be used when installing from Warehouse.

I agree having to update this is tedious and creates noisy PRs. We should avoid updating it if we don’t have to but I also know Kenneth wants Requests to be an example of good practice. With our current lockfile, we’re not adhereing to that tenant. If we can release a stable lockfile and avoid pushing broken lockfiles in the future this should be a much less common occurrence.

PR #4669, opened 2018-06-05
Allow urllib3 1.23

michael-k:
Changelog: https://github.com/urllib3/urllib3/blob/1.23/CHANGES.rst#123-2018-06-04

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ packages = ['requests']
 requires = [
     'chardet>=3.0.2,<3.1.0',
     'idna>=2.5,<2.7',
-    'urllib3>=1.21.1,<1.23',
+    'urllib3>=1.21.1,<1.24',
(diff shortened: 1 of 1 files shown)


bar:
Can you please bump `requests` with this fix so we can fix the current version conflicts?

There are a lot of other libs out there that don't cap `urllib3` meaning they are failing due to this cap

nateprewitt:
Hi @bar, I’ve just commented on a PR to address this. We usually won’t modify this until we are ready for a release. It shouldn’t break anything, but simply emits a warning.

We’ll have a release out in the next week or two once urllib3’s new version has had a chance to settle. I’ll try to get master a bit cleaner this evening.

lilydjwg:
Lots of warning emails send me here. I'm going to edit the local code now because I don't like the coming thousands of emails.

Issue #4673, opened 2018-06-06
Inconsistent urllib3 version requirements between setup.py and __init__.py


GPMueller:
With https://github.com/requests/requests/commit/ac944b7439009ffbf7a10dfee35202f1ac090e76 the maximum allowed version of urllib3 in [setup.py](https://github.com/requests/requests/blob/master/setup.py) was increased to `1.23`.
However, in https://github.com/requests/requests/blob/master/requests/__init__.py#L57-L63 the version is still required to be `<= 1.22`.

This is a bug since the release of requests 1.23 yesterday: https://pypi.org/project/urllib3/#history

## Expected Result

Installation of `requests` via `pip` should work with urllib3 version `1.23`

## Actual Result

e.g.
```
/usr/local/lib/python2.7/dist-packages/requests/__init__.py:80: RequestsDependencyWarning: urllib3 (1.23) or chardet (3.0.4) doesn't match a supported version!
  RequestsDependencyWarning)
ContextualVersionConflict: urllib3 1.23
```

## Reproduction Steps

```
pip install urllib3==1.23
pip install requests
```

```python
import requests
```

yan12125:
Maybe #4669 should be reverted until urllib3 is fully supported.

GPMueller:
It looks to me like it should suffice to simply change the [`__init__.py`](https://github.com/requests/requests/blob/master/requests/__init__.py) to allow v1.23 of urllib3, but I'm not sure since I have not tested it.

rblaine95:
Just tested changing `__init__.py:63: assert minor <= 22` to `<= 23` and then tested with `aurman -Syu` to see if the error came back.

There's no error anymore, however I'm not sure if just simply incrementing that value could cause errors further down the road.

**Edit:** I'll update this if I encounter any errors. So far, I haven't come across anything.

nateprewitt:
Hey everyone, to be clear this isn't a bug but rather a specific design decision. We cannot guarantee support for new releases of urllib3 before they've been tested with the Requests code base. In the event of a new release we haven't tested you will see a warning message informing you about the issue.

We are currently slated to release Requests 2.19 in the next week or two which will move the warning pin. Our standard procedure is to wait a week or so and see if any major bugs are raised in urllib3's new release. Things will continue to work fine with 1.22 in the meantime.

valignatev:
My program relied on stdout output, and requests warning broke it. Good to know that issue is known and will be fixed soon :)

[...]

PR #4671, opened 2018-06-05
Allow urllib 1.23 (part 2)

yan12125:
Follow-up of https://github.com/requests/requests/pull/4669

This eliminates the following warning:
```
$ python -c 'import requests'
/usr/lib/python3.6/site-packages/requests/__init__.py:80: RequestsDependencyWarning: urllib3 (1.23) or chardet (3.0.4) doesn't match a supported version!
  RequestsDependencyWarning)
```

PR #4674, opened 2018-06-06
fixed urllib3 version dependency issue

nmante:
The `requests` setup.py file allows this version range for `urllib3` `>=1.21.1,<1.24`. However, the `__init__.py` has assertion statements that only allow up to (and including) version `1.22`. In this PR, I'm updating the `assert` statements in `__init__.py` to be inline with the `setup.py` `requires`

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -57,10 +57,10 @@ def check_compatibility(urllib3_version, chardet_version):
     # Check urllib3 for compatibility.
     major, minor, patch = urllib3_version  # noqa: F811
     major, minor, patch = int(major), int(minor), int(patch)
-    # urllib3 >= 1.21.1, <= 1.22
+    # urllib3 >= 1.21.1, <= 1.23
     assert major == 1
     assert minor >= 21
-    assert minor <= 22
+    assert minor <= 23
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @nmante, thanks for the contribution to Requests! We’re aware of the recent release of urllib3 and the current pin discrepancy at the tip of master. The current mismatch between the __init__.py doesn’t affect a current release but only the master branch which will be updated.

This issue has been addressed a few times in #4669, #4671, and #4673. It will be addressed with the release of Requests 2.19.

nmante:
@nateprewitt got it! I was working with release 2.18.4 which was causing issues because of the mismatch.

nateprewitt:
Just to reiterate what’s in the other threads in case someone stumbles on this. The release of 2.18.4 has an intentional requirement for urllib3 < 1.23. If your environment modifies the dependency requirements after Requests is installed, we emit a warning stating the dependencies no longer match the installation requirements and things are no longer guaranteed to work as expected.

Everything should work fine with Requests 2.18.4 as long as the correct dependencies are installed. Thanks again @nmante for checking in!

PR #4675, opened 2018-06-07
Don't warn about compatibility with urllib3 v1.23

jeremycline:
Requests allows v1.23 in #4669, but the compatibility check at import
time still looks for 22. Bump the version to 23.

edit: I do wonder if these checks should just be removed since it's pinned in setup.py and pip doesn't let you install 23 anyway without #4669.

Signed-off-by: Jeremy Cline <jcline@redhat.com>

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -60,7 +60,7 @@ def check_compatibility(urllib3_version, chardet_version):
     # urllib3 >= 1.21.1, <= 1.22
     assert major == 1
     assert minor >= 21
-    assert minor <= 22
+    assert minor <= 23
 
     # Check chardet for compatibility.
     major, minor, patch = chardet_version.split('.')[:3]


yan12125:
FYI: similar patches at https://github.com/requests/requests/pull/4671 and https://github.com/requests/requests/pull/4674 were rejected.

nateprewitt:
Hey @jeremycline, thanks for putting this together. We’re aware the tip of master is inconsistent and merging #4669 was a bit premature. In order to prevent multiple follow up PRs for other things we missed, we (maybe just I) decided we’d just wait to do all of the upgrade steps at once with the 2.19 release as per our usual release process.

We’ve addressed this a few times in a few other PRs and issues (#4671, #4673, #4674).

jeremycline:
Ah cool. I should have checked closed PRs, sorry for the noise.

PR #4677, opened 2018-06-10
Fix unicode auth url bug in py3

mecforlove:
Just Fixed bug in [issue4649](https://github.com/requests/requests/issues/4649)

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -16,7 +16,7 @@ import warnings
 
 from base64 import b64encode
 
-from .compat import urlparse, str, basestring
+from .compat import urlparse, str, basestring, is_py3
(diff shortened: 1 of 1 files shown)


mecforlove:
I just updated one file, but why the ci can't pass?

nateprewitt:
Hi @mecforlove, thanks for contributing to Requests! The issue here is that you can't call `encode` on all valid inputs for username and password, specifically bytes. We're going to need a solution that works for all the inputs we test, as well as some new/updated tests to ensure this doesn't end up regressed later.

sigmavirus24:
Further, as explained in my last comment, we can not fix this in Requests 2. Requests 3 will be Python 3 only and will make this fixable.

PR #4679, opened 2018-06-11
idna had a release

russellballestrini:
we should upgrade our bounds to avoid people running into trouble.

https://github.com/kjd/idna/commit/f8389a87092e5a5eb2f92aabf4fbcef4fc9e9499

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,10 +44,9 @@ packages = ['requests']
 
 requires = [
     'chardet>=3.0.2,<3.1.0',
-    'idna>=2.5,<2.7',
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @russellballestrini, thanks for letting us know! This is something we’ll handle in our release process for the next formal Requests release.

adrianchifor:
This breakes most installations of 2.18.4. We fixed it in our case by adding `idna==2.6` to our `requirements.txt` before `requests`, in case someone else has this problem.

PR #4683, opened 2018-06-12
Do not alter status_codes docstring if it is None

martindurant:
Following https://github.com/requests/requests/pull/4394/files, with PYTHONOPTIMIZE=2, the current code breaks with TypeError due to None as the docstring. This broke dask tests https://github.com/dask/dask/pull/3594 .

### DIFF
--- a/requests/status_codes.py
+++ b/requests/status_codes.py
@@ -114,6 +114,7 @@ def _init():
 
     global __doc__
     __doc__ = (__doc__ + '\n' +
-               '\n'.join(doc(code) for code in sorted(_codes)))
+               '\n'.join(doc(code) for code in sorted(_codes))
+               if __doc__ is not None else None)
 
 _init()


requests/status_codes.py
```diff
@@ -114,6 +114,7 @@ def doc(code):
 
     global __doc__
     __doc__ = (__doc__ + '\n' +
```
sigmavirus24, line 116: So let's look at how this evaluates:

```
__doc__ = (None + '\n' + None)
```

In your case. Do we instead want

```py
__doc__ = (__doc__ + '\n' +
           '\n'.join(doc(code) for code in sorted(_codes))) if __doc__ is not None else None
```

Which checks if  `__doc__` is `None` prior to trying to evaluate the parenthetical?

martindurant:
```
(x + '\n' + 'code_stuff'  if x is not None else None)
```
evaluates to `None` when `x` is `None` - the addition is not evaluated.

nateprewitt: Yep, I don't think we need to worry about evaluation unless the conditional is met. This looks good from my end unless @sigmavirus24 has further comments. Thanks for the quick follow up on this @martindurant!

sigmavirus24:
@nateprewitt please release at your earliest convenience

PR #4685, opened 2018-06-13
Note that session() should not be used

Stevoisiak:
According to requests/api.rst, the `Session` class was originally named session in 0.x. [When the class name was capitalized in 1.x](https://github.com/requests/requests/blob/265ef609d5903151374fba480aa81aafe68126ff/docs/api.rst#migrating-to-1x), the `session()` method was added for backwards compatibility.

This adds a docstring note that `requests.Session` should be used instead

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -735,7 +735,12 @@ def session():
     """
     Returns a :class:`Session` for context-management.
 
+    .. deprecated:: 1.0.0
+
+        This method has been deprecated since version 1.0.0 and is only kept for
+        backwards compatibility. New code should use :class:`~requests.sessions.Session`
+        to create a session. This may be removed at a future date.
+
     :rtype: Session
     """
-
     return Session()


Stevoisiak:
@nateprewitt In that case, I'll add a [`DepreciationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) telling users to use `Session` instead.

I tried to track down when `Session` was added, but it seems to have been [sometime before `v0.8.4`](https://github.com/requests/requests/blob/v0.8.4/requests/sessions.py#L50), the oldest tagged version on GitHub.

Stevoisiak:
On closer inspection, it seems `session` was created one day *after* `Session`. ([Aug 16, 2011](https://github.com/requests/requests/commit/519b9cee969e7364654a05ed466db253d35370d3) vs [Aug 15, 2011](https://github.com/requests/requests/commit/ca428504d485b0495b5e794cdd1b47bbb0bd890f))

Stevoisiak:
@nateprewitt I've reverted the docstring changes in favor of a `DepreciationWarning`.

sigmavirus24:
I'm not sure this is worth the grief we'll receive for adding this DeprecationWarning.

nateprewitt:
I’m in agreement with @sigmavirus24. The DeprecationWarning will likely be more of a harassment for existing users than deterrent for new users.

I think the initial scope of a minor docstring change is likely the extent of what we’d want in this PR.

Stevoisiak:
Noted. I've removed the depreciation warning and updated the docstring.

requests/sessions.py
```diff
@@ -736,8 +736,10 @@ def session():
...
     Returns a :class:`Session` for context-management.
 
+    This method is kept for backwards compatibility. For new code,
+    initialize :class:`Session` directly.
+
     :rtype: Session
     """
-    warnings.warn("'requests.session' is depreciated, use 'requests.Session' instead",
-                  DeprecationWarning, stacklevel=2)
+
```
sigmavirus24: Please remove the extraneous new-line here.

requests/sessions.py
```diff
@@ -735,6 +735,9 @@ def session():
     """
     Returns a :class:`Session` for context-management.
 
+    This method is kept for backwards compatibility. For new code,
```
sigmavirus24: Let's do something different since we're using sphinx + reStructuredText:

```rst
.. deprecated:: 1.0.0

    This method has been deprecated since version 1.0.0 and is only kept for
    backwards compatibility. New could must use :class:`~requests.session.Session`
    to create a session. This will be removed at a future date.
```

requests/sessions.py
Stevoisiak: The newline was part of the original file.

There are a few other methods in `sessions.py` with an extra newline (`post`, `put`, `patch`, `delete`, `send`). Should these be changed as well?

requests/sessions.py
```diff
@@ -735,8 +735,11 @@ def session():
     """
     Returns a :class:`Session` for context-management.
 
-    This method is kept for backwards compatibility. For new code,
-    initialize :class:`Session` directly.
+    .. deprecated:: 1.0.0
+
+        This method has been deprecated since version 1.0.0 and is only kept for
+        backwards compatibility. New could must use :class:`~requests.session.Session`
```
sigmavirus24: New *code*

Stevoisiak: This was fixed in commit https://github.com/requests/requests/pull/4685/commits/5789e91ec7265d662cd0eb077db2b5bb9363b520

Stevoisiak:
Should I squash these commits together?

nateprewitt:
@Stevoisiak, please.

Stevoisiak:
@nateprewitt Squashed and rebased against master.

requests/sessions.py
```diff
@@ -735,7 +735,12 @@ def session():
     """
     Returns a :class:`Session` for context-management.
 
+    .. deprecated:: 1.0.0
+
+        This method has been deprecated since version 1.0.0 and is only kept for
+        backwards compatibility. New code should use :class:`~requests.session.Session`
```
nateprewitt: I think this might need to be`requests.sessions.Session`.

sigmavirus24: Whatever the module path is, that's what it should be. I was writing from memory and didn't intend for it to be copied verbatim.

Stevoisiak: Updated.

nateprewitt:
This looks good to me, thanks @Stevoisiak!

Stevoisiak:
There are a few files in Requests which call `session()` instead of `Session()`. Should these be updated?

PR #4697, opened 2018-06-19
Remove redundant __init__ for BaseAdapter

Stevoisiak:


nateprewitt:
@Steviosiak, I don’t know if I understand why you feel this is redundant. This is the base class that can be used for creating custom adapters.

We usually avoid making changes in the code unless they solve a specific problem and make the user experience better.

I don’t think we’re likely to accept this at this time.

Stevoisiak:
@nateprewitt I was under the impression that `BaseAdapter` was initializing an instance of itself, which is something objects do by default.

I'm not sure what purpose `__init__` is serving in this case.

Stevoisiak:
@nateprewitt 

<s>To illustrate, each of the following examples would be functionally identical.</s>

```Python
class BaseAdapter(object):
    def __init__(self):
```
```Python
class BaseAdapter(object):
    def __init__(self):
        object.__init__(self)
```
```Python
class BaseAdapter(object):
    def __init__(self):
        pass
```

Pylint warns about the first example with [`useless-super-delegation`](https://github.com/PyCQA/pylint/issues/1567)

Stevoisiak:
It seems my [previous subclass assumptions](https://github.com/requests/requests/pull/4697#issuecomment-398533573) were incorrect.

The examples I provided will behave differently if a subclass inherits from more than one parent class.

```Python
class BaseAdapter(object):
    def __init__(self):

class BaseAdapter2(object):
    def __init__(self):
        object.__init__(self)

class BaseAdapter3(object):
    def __init__(self):
        pass

class Second:
    def __init__(self):
        print("second")
        super().__init__()

class Other1(BaseAdapter, Second):
    def __init__(self):
        print("other1 init")
        super().__init__()

class Other2(BaseAdapter2, Second):
    def __init__(self):
        print("other2 init")
        super().__init__()

class Other3(BaseAdapter3, Second):
    def __init__(self):
        print("other3 init")
        super().__init__()

Other1()
Other2()
Other3()
```

Output:

```
other1 init
second
other2 init
other3 init
```

However, removing `__init__` does seem to have the same output.

```Python
class BaseAdapter(object):
    def __init__(self):

class BaseAdapter4(object):
    pass

class Second:
    def __init__(self):
        print("second")
        super().__init__()

class Other1(BaseAdapter, Second):
    def __init__(self):
        print("other1 init")
        super().__init__()

class Other4(BaseAdapter4, Second):
    def __init__(self):
        print("other4 init")
        super().__init__()

Other1()
Other4()
```

Output:

```
other1 init
second
other4 init
second
```

Related: 

- [Stack Overflow: How does Python's super() work with multiple inheritance?](https://stackoverflow.com/q/3277367/3357935)
- [Stack Overflow: Calling parent class \_\_init__ with multiple inheritance, what's the right way?](https://stackoverflow.com/a/9575426/3357935)
- [Scott Lobdell: Multiple Inheritance in Python](http://scottlobdell.me/2015/06/multiple-inheritance-python/)

Stevoisiak:
Personal testing indicates this change would improve performance when initializing BaseAdapter. (0.57 seconds vs 0.11 seconds)

```Python
import timeit

setup = """
class BaseAdapter(object):
    def __init__(self):
"""
print(timeit.timeit('BaseAdapter()', setup=setup))

setup = """
class BaseAdapter(object):
    pass
"""
print(timeit.timeit('BaseAdapter()', setup=setup))
```

Output

```
0.5758354471040419
0.10679823940642941
```

nateprewitt:
@Stevoisiak I believe the benchmark you’re providing is showing the total run time of 1,000,000 runs. That means we’re measuring an initialization variation of .4 microseconds which is an incredibly small amount of time for what is likely a one time operation in someone’s code.

As was stated earlier in the PR, we usually don’t remove code unless there’s a compelling reason to do so. I’m not convinced we need this init but it’s nearly impossible to know how people are using this BaseAdapter and I’d rather not risk breaking things.

Stevoisiak:
@nateprewitt Ah, the timings did seem a bit large. Thank you for taking the time to explain.

PR #4699, opened 2018-06-20
Test multi-class inheritance for BaseAdapter

Stevoisiak:
Ensure that `BaseAdapter` supports multi-class inheritance. This means a subclass of `BaseAdapter` will invoke all parent classes when inheriting from multiple classes.

See: https://github.com/requests/requests/pull/4697#issuecomment-398814166

    class BaseAdapter(object):  # pass
        def __init__(self):

    class BaseAdapter(object):  # pass
        pass

    class BaseAdapter(object):  # fail
        def __init__(self):
            BaseAdapter.__init__(self)

    class BaseAdapter(object):  # fail
        def __init__(self):
            object.__init__(self)

    class BaseAdapter(object):  # fail
        def __init__(self):
            pass

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -13,7 +13,7 @@ import warnings
 import io
 import requests
 import pytest
-from requests.adapters import HTTPAdapter
+from requests.adapters import BaseAdapter, HTTPAdapter
 from requests.auth import HTTPDigestAuth, _basic_auth_str
 from requests.compat import (
     Morsel, cookielib, getproxies, str, urlparse,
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @Stevoisiak, it looks like this test passes regardless of the existence of `__init__`. If that's the case, we're effectively testing a behaviour of the Python interpreter rather than any Requests code. I don't think we're likely to accept a patch around that unless there's a compelling reason to do so.

As I tried to state in the last PR, we really appreciate contributions to Requests but need to make sure they're providing benefit to our users and the code base. Test coverage is definitely an area we can do better in, but the tests should be focused on behaviours of Requests.

I'm going to close this out since I think it's unlikely to be merged.

PR #4701, opened 2018-06-20
Documentation and tests for form-encoded multivalued elements as lists

akaihola:


docs/user/quickstart.rst
```diff
@@ -261,6 +261,23 @@ useful when the form has multiple elements that use the same key::
       ...
     }
 
+An alternative method to express multiple elements with the same key is to use
+a dictionary and list the element values as a list::
+
+    >>> payload = {'key1': ['value1', 'value2']}
```
nateprewitt: It may be better to roll this into the example above. They are both performing the same action, just with different syntactic sugar.

akaihola: I rolled the examples into one in c61f36e. Does that look like what you were thinking about?

nateprewitt:
Alright, I think I'm happy with this. Would you mind rebasing onto the master branch and then we can look at merging. Thanks @akaihola!

akaihola:
@nateprewitt, I now rebased on master.

PR #4704, opened 2018-06-21
Don't use depreciated `session()` method

stevoisiak:
- References to depreciated method `session()` now initialize `Session()` directly.
- Added test for depreciated `session()` method.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -59,9 +59,9 @@ class TestRequests:
 
     def test_entry_points(self):
 
-        requests.session
-        requests.session().get
-        requests.session().head
+        requests.Session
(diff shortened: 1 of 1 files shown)


nateprewitt:
I don’t think this will necessarily hurt anything to migrate, but if we’re going to have the `session` method left in the API we should have at least some tests exercising it. If the day comes that we ever do remove `session` we will know very quickly that these tests need to be updated.

I’ll leave final judgement here to @sigmavirus24 but I’m not sure we need to do this now.

Stevoisiak:
@nateprewitt In that case, I can add a test specifically for `session()` to ensure backwards compatibility.

PR #4705, opened 2018-06-21
Remove test for Python 2.6

Stevoisiak:
Requests does not officially support 2.6, meaning this test is no longer needed.


nateprewitt:
Even if we don’t actively support 2.6 anymore, we don’t need to remove visibility on issues or intentionally break 2.6 at this time. We’ve intentionally deferred removing 2.6 specific code until the Requests 3.0 release.

PR #4711, opened 2018-06-26
Simplify chained comparison

Stevoisiak:
Improve code readability by simplifying logic for a chained comparison.


nateprewitt:
Hi @Stevoisiak, just to reiterate again, if there isn’t a compelling reason to change something that provides noticeable benefit to the user or fixes something wrong in the code, we’re unlikely to accept it. This is a stylistic change that doesn’t really fit that criteria so I believe we’re going to pass on this. Thanks for the suggestion!

PR #4717, opened 2018-06-28
Add official Python 3.7 support

ericwb:
Python 3.7 was just released [1]. This is a small change to
enable support in requests.

[1] https://docs.python.org/3.7/whatsnew/3.7.html

Signed-off-by: Eric Brown <browne@vmware.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 language: python
 matrix:
   include:
-    # - python: '2.6'
     - python: '2.7'
     - python: '3.4'
     - python: '3.5'
@@ -10,8 +9,6 @@ matrix:
     - python: '3.7'
       dist: xenial
       sudo: true
-    # - python: 'pypy'  -- appears to hang
-    # - python: 'pypy3'
 # command to install dependencies
 install: "make"
 # command to run tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,5 @@
 sudo: false
 language: python
-matrix:
-  include:
-    - python: '2.7'
-    - python: '3.4'
-    - python: '3.5'
-    - python: '3.6'
-    - python: '3.7'
-      dist: xenial
-      sudo: true
 # command to install dependencies
 install: "make"
 # command to run tests
@@ -22,8 +13,35 @@ jobs:
     - stage: test
       script:
         - |
-          if [[ "$TRAVIS_PYTHON_VERSION" != "2.6" ]] ; then make test-readme; fi
+          make test-readme
         - make ci
+      python: '2.7'
+    - stage: test
+      script:
+        - |
+          make test-readme
+        - make ci
+      python: '3.4'
+    - stage: test
+      script:
+        - |
+          make test-readme
+        - make ci
+      python: '3.5'
+    - stage: test
+      script:
+        - |
+          make test-readme
+        - make ci
+      python: '3.6'
+    - stage: test
+      script:
+        - |
+          make test-readme
+        - make ci
+      python: '3.7'
+      dist: xenial
+      sudo: true     
     - stage: coverage
       python: 3.6
       script: codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,17 @@
 sudo: false
 language: python
-python:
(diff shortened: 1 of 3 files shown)


sigmavirus24:
Travis doesnt' appear to support 3.7 yet: https://travis-ci.org/requests/requests/jobs/397660384

ericwb:
Looks like it's being worked: https://github.com/travis-ci/travis-ci/issues/9815

sigmavirus24:
I see nothing indicating that it's in progress on Travis' part other than an issue that non-Travis-CI employees are commenting on.

nateprewitt:
Hey @ericwb, we’ll likely want this enabled for appveyor as well before we claim support for 3.7.

I haven’t seen an update from them but that’d be something to look into before we merge this.

jdufresne:
According to https://github.com/appveyor/ci/issues/2475, Appveyor now supports Python 3.7.

nateprewitt:
@jdufresne great to hear! Once we get a working build with that, I’m happy to merge.

nateprewitt:
Hey @ericwb, wanted to check in on this again if you get a minute. Otherwise, I'm going to look at cherry picking these onto master with the appveyor integration later this week. Thanks!

sigmavirus24:
@ericwb would you be willing to rebase this?

nateprewitt:
And squash the commits too if you wouldn’t mind. Thanks!

.travis.yml
```diff
@@ -22,8 +13,35 @@ jobs:
     - stage: test
       script:
         - |
-          if [[ "$TRAVIS_PYTHON_VERSION" != "2.6" ]] ; then make test-readme; fi
+          make test-readme
```
jdufresne: This could drop the `|` and fit on one line

.travis.yml
```diff
@@ -22,8 +13,35 @@ jobs:
...
         - make ci
+      python: '2.7'
+    - stage: test
+      script:
+        - |
+          make test-readme
```
jdufresne: Same.

.travis.yml
```diff
@@ -22,8 +13,35 @@ jobs:
...
+        - make ci
+      python: '3.4'
+    - stage: test
+      script:
+        - |
+          make test-readme
```
jdufresne: Same.

.travis.yml
```diff
@@ -22,8 +13,35 @@ jobs:
...
+        - make ci
+      python: '3.5'
+    - stage: test
+      script:
+        - |
+          make test-readme
```
jdufresne: Same.

.travis.yml
```diff
@@ -22,8 +13,35 @@ jobs:
...
+        - make ci
+      python: '3.6'
+    - stage: test
+      script:
+        - |
+          make test-readme
```
jdufresne: Same.

.travis.yml
```diff
@@ -22,8 +13,35 @@ jobs:
...
+        - make ci
+      python: '3.7'
+      dist: xenial
+      sudo: true
```
nateprewitt: Why do we need sudo here and does this conflict with the sudo: False declaration above?

.travis.yml
```diff
@@ -1,14 +1,5 @@
 sudo: false
 language: python
-python:
```
nateprewitt, line 3: Could we get a quick explanation of the benefits/trade offs of moving from the matrix approach to this new setup?

.travis.yml
sigmavirus24: No, it does not. This is the only way to use xenial on Travis is to use sudo on this job alone. And xenial is the only target that has true Python 3.7. So it's not ideal (because enabling sudo is *slow*) but it's the best we have right now

.travis.yml
sigmavirus24: You mean moving from the old configuration to the more explicit matrix definition? Other than greater flexibility?

mrkz:
@nateprewitt any release coming soon with python3.7 support?
Thanks in advance!

nateprewitt:
@mrkz, 2.19.1 should already support 3.7. This was just adding documentation stating that.

Issue #4724, opened 2018-07-06
Getting response content starting with  b


shiveshabhishek:
Summary.
In output of content of a response , it is getting printed with starting as **b**
## Expected Result
No **b** at the starting of output
What you expected.
No **b** at the starting of output
## Actual Result
output without **b** at the beginning
What happened instead.
![image](https://user-images.githubusercontent.com/29581152/42371314-bbf171b2-812c-11e8-8eb6-cf48850c16e0.png)

## Reproduction Steps
Just print any output like : response.content
```python
import requests

```

## System Information

Linux Ubuntu 18.06
python version:3.6.5

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).

sigmavirus24:
You likely want `response.text` then if you don't want the content as raw bytes but instead as a string.

This is not a bug.

shiveshabhishek:
ohh, i see.. I thought that it's showing improper response.
Thanks for correcting my issue @sigmavirus24

PR #4725, opened 2018-07-06
Fix print output of content

shiveshabhishek:
-- Removed b at the beginning of printing of content of a response

Fixes to Issue: #4724

Signed-off-by: Shivesh <shankeyshivesh@gmail.com>


sigmavirus24:
This is not a fix for #4724

PR #4733, opened 2018-07-11
fix cookie management bug

fx-kirin:
Even I used custom cookie jar class to manage cookie values, redirecting link only used default `RequestCookieJar` class and it seemed there was not work around.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -420,7 +420,7 @@ class Session(SessionRedirectMixin):
 
         # Merge with session cookies
         merged_cookies = merge_cookies(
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @fx-kirin, thanks for contributing to Requests! Could we get a test for this so we don’t accidentally regress it in the future?

nateprewitt:
We're going to close this out since it's been pending feedback for quite some time. I think the idea is correct long term but may be breaking as the current interface always returns a RequestsCookieJar which may be relied on. Feel free to open an issue to track this if you'd like.

PR #4739, opened 2018-07-13
Document that params and data can be list of tuples everywhere

moy:
Fix a few inconsistencies in the doc found while working on [MechanicalSoup](https://github.com/MechanicalSoup/MechanicalSoup/pull/219).

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -19,7 +19,8 @@ def request(method, url, **kwargs):
     :param method: method for the new :class:`Request` object.
     :param url: URL for the new :class:`Request` object.
     :param params: (optional) Dictionary or bytes to be sent in the query string for the :class:`Request`.
-    :param data: (optional) Dictionary or list of tuples ``[(key, value)]`` (will be form-encoded), bytes, or file-like object to send in the body of the :class:`Request`.
+    :param data: (optional) Dictionary, list of tuples, bytes, or file-like
+        object to send in the body of the :class:`Request`.
     :param json: (optional) A JSON serializable Python object to send in the body of the :class:`Request`.
     :param headers: (optional) Dictionary of HTTP Headers to send with the :class:`Request`.
     :param cookies: (optional) Dict or CookieJar object to send with the :class:`Request`.
@@ -102,7 +103,8 @@ def post(url, data=None, json=None, **kwargs):
     r"""Sends a POST request.
 
     :param url: URL for the new :class:`Request` object.
-    :param data: (optional) Dictionary (will be form-encoded), bytes, or file-like object to send in the body of the :class:`Request`.
+    :param data: (optional) Dictionary, list of tuples, bytes, or file-like
+        object to send in the body of the :class:`Request`.
     :param json: (optional) json data to send in the body of the :class:`Request`.
     :param \*\*kwargs: Optional arguments that ``request`` takes.
     :return: :class:`Response <Response>` object
@@ -116,7 +118,8 @@ def put(url, data=None, **kwargs):
     r"""Sends a PUT request.
 
     :param url: URL for the new :class:`Request` object.
-    :param data: (optional) Dictionary (will be form-encoded), bytes, or file-like object to send in the body of the :class:`Request`.
+    :param data: (optional) Dictionary, list of tuples, bytes, or file-like
+        object to send in the body of the :class:`Request`.
     :param json: (optional) json data to send in the body of the :class:`Request`.
     :param \*\*kwargs: Optional arguments that ``request`` takes.
     :return: :class:`Response <Response>` object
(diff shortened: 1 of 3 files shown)


requests/api.py
```diff
@@ -18,7 +18,9 @@ def request(method, url, **kwargs):
 
     :param method: method for the new :class:`Request` object.
     :param url: URL for the new :class:`Request` object.
-    :param params: (optional) Dictionary or bytes to be sent in the query string for the :class:`Request`.
+    :param params: (optional) Dictionary or list of tuples
```
nateprewitt: This can probably be two lines instead of three, following the line length of the other entries.

moy: To use 2 lines, one needs to cut the ` ``literal`` ` block, and I've seen some RST processor yield incorrect results in this case so I preferred avoiding this cut. I can change it if you prefer.

nateprewitt: I don’t think we need to cut the literal block. This line is almost identical to the one below it which only occupies a single line.

moy:
> As for the others, it feels unnecessarily redundant for a smaller edge case.

I find it strange to have a parameter with exactly the same behavior in different functions and different documentations. If you want to avoid redundancy, then factoring the docs with links (stg like "Parameters to be sent in the query string for the :class:`Request`. See <link to proper doc> for details.") would be IMHO better.

In the existing doc, the string is already precise enough to discourage the user from trying to find another doc (it already documents the "dictionary" and "bytes" case). When I found the doc for [Session.request](http://docs.python-requests.org/en/master/api/#requests.Session.request), it didn't occur to me that the actual doc for the `params` parameter should be searched for elsewhere.

requests/api.py
moy: Indeed: my text editor wraps at 72 characters so it was cutting in the middle of the literal block. I make lines a bit longer, but in most cases it doesn't change the number of lines. Other docstrings fit on one line because they're e.g. 171 characters long.

nateprewitt: @moy yes, so if we split it across two lines it will be < 100 characters per line which is consistent with this projects typical line length.

nateprewitt:
@moy, how about we go with the below single-line copy then for the non-`request` entries.

```(optional) Dictionary, list of tuples, or bytes to be sent in the query string for the :class:`Request`.```

File: requests/api.py (hunk 1 above)
Comment (moy) [reply requests/api.py]: I'm a new contributor so I can't talk about the established practices, but the doc says "Line-length can exceed 79 characters, to 100, when convenient.", so I understand that I can make an exception to the 79 characters when I have a good reason ("when convenient"). I don't see the benefit of making an exception here.

**Comment by moy** [comment]
I'm fine with the single line, but I'm not sure everyone would understand. I would add a link to the doc of requests.request for details.

**Comment by nateprewitt** [comment]
@moy, these are thin wrappers around a single call which is for a well documented function. If users have a use case where they need tuples, we have documentation around this. If they're only looking at the code, it's a single click in most IDEs to determine its usage.

The main thing we're trying to avoid here in unnecessary bloat and having several places that need to be updated for the same information. As I said above, I'm happy to make this information clearer, but would prefer we do it without establishing a pattern of copy-pasting to every function any time we make a tweak.

**Comment by moy** [comment]
> If they're only looking at the code, it's a single click in most IDEs to determine its usage.

I did have to dig into the code to find this, yes. But to me, the point of autogenerated documentation (http://docs.python-requests.org/en/master/api/) is precisely to avoid asking the user to read the code. Why do you publish this autogenerated documentation online if the recommended way to read it is within the code with an IDE?

> would prefer we do it without establishing a pattern of copy-pasting to every function any time we make a tweak.

That's precisely why I suggest a link: to avoid duplicating, and to let the user find the centralized place more easily.

**Comment by nateprewitt** [comment]
> That's precisely why I suggest a link: to avoid duplicating, and to let the user find the centralized place more easily.

I think we may be going in circles now. I believe we agreed on your initial suggestion to linking to the `:class:Request` (with minor wording tweaks). In your current commits, this has the same effective copy as `requests.request`. Adding two links that provide the same information, and have the potential to drift, isn't ideal.


To summarize the steps I'm proposing:
* The `params` string in `requests.request` should be consistent with the `data` string.
* The `Request` class docstring references the list of tuples. Ideally something like:
```
     :param data: the body to attach to the request. If a dictionary or list of
         tuples ``[(key, value)]`` is provided, form-encoding will take place.
     [...]
     :param params: Dictionary, list of tuples, or bytes to be sent in the query string for
         the :class:`Request`.
```
* Everything else:
```
:param params: (optional) Dictionary, list of tuples, or bytes to be sent in the query string for the :class:`Request`.
:param data: (optional) Dictionary, list of tuples, bytes, or file-like object to send in the body of the :class:`Request`.
```

moy:
OK, I read you too quickly.

I pushed essentially what you're proposing. I just kept a bit more details in Request's `:param params:`.

moy:
Weird, my local version was correct, but apparently I pushed a version with incorrect cut-and-paste. #4910 should fix this.

PR #4751, opened 2018-07-25
compat.py already defines unicode() as str() on Python 3

cclauss:
It is not necessary or desirable to repeat the logic that properly belongs in compat.py.


cclauss:
Fails pretty miserably.

PR #4769, opened 2018-08-22
Use a `with` statement when sending multipart data in quickstart guide 

Ares513:
The original documentation implies that `requests` will close the file for you; this is not the case.  Since it is up to the user to release the resources, I think it should be reflected as such in the docs.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -297,7 +297,6 @@ POST a Multipart-Encoded File
 Requests makes it simple to upload Multipart-encoded files::
 
     >>> url = 'http://httpbin.org/post'
-    >>> files = {'file': open('report.xls', 'rb')}
     >>> with open('report.xls', 'rb') as f:
             r = requests.post(url, files={'file': f})
             r.text
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -311,10 +311,9 @@ Requests makes it simple to upload Multipart-encoded files::
 You can set the filename, content_type and headers explicitly::
 
     >>> url = 'http://httpbin.org/post'
-    >>> files = {'file': ('report.xls', open('report.xls', 'rb'), 'application/vnd.ms-excel', {'Expires': '0'})}
-
-    >>> r = requests.post(url, files=files)
-    >>> r.text
+    >>> with open('report.xls', 'rb') as f:
+            r = requests.post(url, files={'file': ('report.xls', f, 'application/vnd.ms-excel', {'Expires': '0'})})
+            r.text
     {
       ...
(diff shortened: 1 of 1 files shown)


sigmavirus24:
> The original documentation implies that requests will close the file for you; this is not the case. 

Can you explain where it implies it? The code is just an example, not documentation of best practices. it is demonstrating how one might send a `multipart/form-data` request with the syntactic sugar of Requests. It also demonstrates the functional API, which is great for testing and getting up-and-running, but also not a best practice. At what point does the documentation have to become solely best practices?

Ares513:
The current documentation does an excellent job of demonstrating behavior specific to Requests. That said, documentation should really be a source of truth. It should strive to demonstrate best practices, because on mature libraries, my expectation is generally that the instructions and examples it gives are the "right" way to do it. If the documentation is not showing me the right way to do it in production, without warning the user that it's only an example, then it's ineffective; I need to know what I need to do to make this work in the real world. 

I think an analogy is apt here. Suppose you get an instruction manual for a forklift. On page 93, it suggests using a particular lever in a particular way. Now, people who really know their forklifts should know better, that that's not really the right way, it's just an example. But who reads the manual? Not the experts. They already know it; the manual is most useful to the people who have the most to learn. If it is incorrect, only the experts will know. 

I'm sure you've seen this happen elsewhere, too; someone posts an example solution, and it gets copied and used in hundreds of code bases. It doesn't even really matter if it's good or bad. It's the "dictionary definition" and thus is generally accepted as truth. 

Authoritative documents that are trusted are how antipatterns spread; people trust what they say, even when they're wrong, because _what else is there to trust?_

sigmavirus24:
@Ares513 a more apt analogy, in my opinion, is someone receiving a kit to construct a bookcase and being told how screws work, how to turn a screwdriver, how to swing a hammer, etc. The Requests documentation is not the place for generic Python best practices (i.e., closing a file once it's been opened). In no way do we say that we close the file on behalf of the user. I don't even think that's entirely implied. The `>>>` preface is what you would see if you did `python` and were toying around with Requests in a terminal or interactive shell. I don't think we're required to show you how to use a file such that it auto-closes either. Yes, folks may be learning Python from our docs. No, I don't believe it's entirely our responsibility to teach them how to write perfect Python. It's only our responsibility to show them how to use Requests to do what they want.

endivan:
icon = Octicons::Octicon.new("x")
icon.options
# {:class=>"octicon octicon-x", :viewBox=>"0 0 12 16", :version=>"1.1", :width=>12, :height=>16, :"aria-hidden"=>"true"}

sethmlarson:

PR #4787, opened 2018-09-14
Fix broken essay link

vinayak-mehta:


vinayak-mehta:
Will do @nateprewitt!

vinayak-mehta:
@nateprewitt Did a `git grep` for "kennethreitz.org". Couldn't find any other link that is broken.

PR #4794, opened 2018-09-17
Proxy headersAllow headers to be passed to proxy

zulrang:
Adding a header field to proxies will pass any headers to the proxy connection.


nateprewitt:
Hi @zulrang, thanks for trying to help move this issue forward. Unfortunately, I don't think we'll be likely to take this approach. Adding a special cased "headers" entry to proxies is going to cause weird behaviour. Both because it's not a proxy, and because this is a hard thing to document and get people to use properly.

As Lukasa said in the original issue, I'm not sure if there's a clean interim solution right now. We can work towards implementing a fix in 3.0 but I don't think we want to implement a hacky solution that needs to be maintained until then.

PR #4798, opened 2018-09-19
Improve parsing of header Digest

alkino:
If this header is multiline we can found 'Digest\r\n  realm="..."\r\n ' instead of "Digest ".


alkino:
Let's discuss about it :p

sigmavirus24:
No. See what @Lukasa told you in your issue

PR #4812, opened 2018-10-01
Fix no_proxy being too greedy.

### LINKED ISSUE #4795: `no_proxy` implementation is too greedy
The implementation of `no_proxy` environment just checks that the hostname ends with some host present in the `no_proxy` variable. Although there is no clear specification for `no_proxy`, this behavior is very unexpected and even dangerous.

- It prevents you from bypassing only one domain and not its subdomains.
- If `no_proxy` is set for `requests.com` and the attacker knows it, they can register `attacker-requests.com` and successfully bypass the proxy.

## Expected Result

- `no_proxy=gle.com` should bypass ONLY the domain `gle.com`
- `no_proxy=www.gle.com` should bypass ONLY the domain `www.gle.com`
- `no_proxy=.google.com` should bypass ONLY subdomains of `google.com` (this works OK)

(just like curl does)

## Actual Result

`no_proxy=gle.com` bypasses any domain that ends with `gle.com`, including:
- `google.com`
- `www.google.com`
- `foo.gle.com`
- `foo.bar.gle.com`

## Reproduction Steps

```
$ http_proxy=http://no-proxy-here/ no_proxy=gle.com python -c "import requests; print(requests.get('http://www.google.com'))"     
<Response [200]>
```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.7"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.0"
  },
  "platform": {
    "release": "4.18.7-arch1-1-ARCH",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.19.1"
  },
  "system_ssl": {
    "version": "1010100f"
  },
  "urllib3": {
    "version": "1.23"
  },
  "using_pyopenssl": false
}
```

I can prepare a PR if needed.

gilbsgilbs:
**Summary:** Setting `no_proxy` on `requests.com` used to enable it on `somerequests.com` also, which was unexpected and potentially dangerous. (see #4795)

Note that this delegates to urllib more often, as stated [here](https://github.com/requests/requests/issues/4795#issuecomment-425886286). I don't have the history behind `should_bypass_proxies` implementation, therefore my change is the tiniest possible (minimum thing to fix the greedy behavior). I don't know exactly why requests tries some things first and then delegates to urllib though.

Fixes #4795

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [dev-packages]
-pytest = ">=2.8.0"
+pytest = ">=3.1.0"
 codecov = "*"
 pytest-httpbin = ">=0.0.7"
 pytest-mock = "*"
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -727,12 +727,16 @@ def should_bypass_proxies(url, no_proxy):
                     # matches the IP of the index
                     return True
         else:
-            host_with_port = parsed.hostname
+            host_without_port = '.' + parsed.hostname
+            host_with_port = host_without_port
             if parsed.port:
                 host_with_port += ':{0}'.format(parsed.port)
 
             for host in no_proxy:
-                if parsed.hostname.endswith(host) or host_with_port.endswith(host):
+                if not host.startswith('.'):
+                    host = '.' + host
+
+                if host_without_port.endswith(host) or host_with_port.endswith(host):
                     # The URL does match something in no_proxy, so we don't want
                     # to apply the proxies on this URL.
                     return True
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -6,6 +6,7 @@ import filecmp
 from io import BytesIO
 import zipfile
 from collections import deque
+import sys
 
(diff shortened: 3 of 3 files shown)


gilbsgilbs:
Apparently [CPython 3.4 `no_proxy` implementation was as naive as requests'](https://github.com/python/cpython/blob/cd1d5c554c40b5b348de2e77d65986aa154e4477/Lib/urllib/request.py#L2349), hence the failing build. Not sure giving the final word to `proxy_bypass` is a great idea. Can't we just use the same implementation as in [CPython 3.7](https://github.com/python/cpython/blob/655608a1112e592cd6a9155ebe774dd285f561f3/Lib/urllib/request.py#L2487)?

mirnakhori (approved):

nateprewitt:
Resolving since this was finally addressed in 2.34.0.

PR #4821, opened 2018-10-11
Add read() to Response object.

smoser:
The traditional 'read' interface is useful in some cases.
At very least it may be easier for code that expectes file-handle
like objects to just add requests support.

Due to gzip/deflate content-encoding it is not really possible for
the caller to use just use Response.raw.read().  If gzip was used
that would return compressed bytes.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -640,6 +640,9 @@ class Response(object):
         #: is a response.
         self.request = None
 
+        self._read_iter = None
+        self._read_leftover = None
+
     def __enter__(self):
         return self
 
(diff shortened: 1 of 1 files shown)


smoser:
Hi,
I realize the implementation probably needs some work, but I wanted to see what you think of the general idea before I polished it up.

I've been using this in a downstream project at https://git.launchpad.net/simplestreams/tree/simplestreams/contentsource.py#n356 for a while.

sigmavirus24:
Hi there, Requests is currently under a feature-freeze and the `read()` method isn't really the only thing that's necessary for code or libraries that require a file-like object.

Further, it is possible to get the raw, compressed bytes from `Response.raw.read` so this functionality is an unnecessary duplication of the logic in urllib3 + extra work to provide functionality that already exists.

smoser:
The problem with Response.raw.read() is that if content-encoding of gzip, it will return gzipped' content.
The user is not expecting gzipped content.

Here is attempt to use requests and 'raw.read' as you suggest.
It returns gzip compressed data even though the url opened has normal json content.  This fact is not something the user of requests should have to expect or handle.

    >>> import requests
    >>> url = "http://cloud-images.ubuntu.com/releases/streams/v1/com.brightbox:official.json"
    >>> r = requests.get(url, stream=True)
    >>> r.headers.get('content-encoding')
    'gzip'
    >>> r.raw.read(32)
    b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03\xcdWMs\x9b0\x10\xbd\xfbW0\\\x1b\x08\xb2)&\xfc\x89^z\xcf'

Here is the function that my change provides.  Now, read() returns
the uncompressed data.

    >>> r = requests.get(url, stream=True)
    >>> r.read(32)
    b'{\n   "updated": "Thu, 11 Oct 201'


With regard to feature-freeze, I don't need this to be landed right now.  I was really looking for feedback as to whether or not you were open to a 'read' interface on Response.

sigmavirus24:
> With regard to feature-freeze, I don't need this to be landed right now. I was really looking for feedback as to whether or not you were open to a 'read' interface on Response.

The feature-freeze is relevant because it defaults the answer to No

PR #4823, opened 2018-10-13
Add auth.HTTPHeaderAuth

tmshn:
Since authentication using 'Authorization' header is quite popular situation, I added `HTTPHeaderAuth` class to make it easier.

With this, instead of doing this

```python
requests.get(url, header={'Authorization': 'token xxx'})
```

(this will be overwritten by netrc!)

or

```python
def auth_header(r):
    r.haeders['Authorization'] = 'token xxx'
    return r
requests.get(url, auth=auth_header)
```

you can do

```python
requests.get(url, auth=HTTPHeaderAuth('token xxx'))
```

This is easier, clearer and no overwrite by netrc file.

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -76,7 +76,38 @@ class AuthBase(object):
         raise NotImplementedError('Auth hooks must be callable.')
 
 
-class HTTPBasicAuth(AuthBase):
+class HTTPHeaderAuth(AuthBase):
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Thanks for sending this over. Unfortunately we're under a feature freeze and cannot accept this at this time.

Issue #4315, opened 2017-09-30
Investigate users with slow import time, and attempt to mitigate in future releases


kennethreitz:
Letter from a fan:

> Hi Kenneth! 
> Lol I'm named Kenneth too :) - but just call me Ken. 
> Thanks for Requests! I use it for my bots, and it really simplifies everything. 
> It helps my life - I work with a MediaWiki wiki, and Requests condenses all the (guess what) > requests I have to make. 
> My only qualm is how long it takes to import - but I love it all the same. 
> Thank you!

This pretty much sums up many people's perception of Requests, and I'd like for us to fix this — be it pyopenssl's fault or not.

methane:
See https://gist.github.com/methane/96d58a29e57e5be97769897462ee1c7e

sigmavirus24:
@methane can you share how you generated that?

methane:
This is produced by patched Python.
See python/cpython#3490

NielsDa:
Hi 
Are there any plans to improve this issue?.
**request** is used in a CLI tool named **iocage** that is used to control virtual instances of FreeBSD,
Importing requests is significantly slower than creating, and starting a new VM by cloning 100000 files. :-)
Since iocage is started every time a command  is issued, it is like a 3-4 sec. delay is introduced into bash, ..pretty annoying..
and in most cases it actually does not need the functionality provided by requests.. :-| 
https://forums.freenas.org/index.php?threads/iocage-is-slow-at-loading-request.63109/

methane:
Is pyOpenSSL installed?
If so, could you try uninstall it?

NielsDa:
_>Is pyOpenSSL installed?_
Seems like it.
```root@tango:~/requests # find / | grep pyOpenSSL
/usr/local/lib/python3.6/site-packages/pyOpenSSL-16.2.0-py3.6.egg-info
/usr/local/lib/python3.6/site-packages/pyOpenSSL-16.2.0-py3.6.egg-info/PKG-INFO
/usr/local/lib/python3.6/site-packages/pyOpenSSL-16.2.0-py3.6.egg-info/SOURCES.txt
/usr/local/lib/python3.6/site-packages/pyOpenSSL-16.2.0-py3.6.egg-info/top_level.txt
/usr/local/lib/python3.6/site-packages/pyOpenSSL-16.2.0-py3.6.egg-info/requires.txt
[...]

PR #4825, opened 2018-10-15
Document the security extra for SNI support with legacy Python

browniebroke:
This was added back in #2195.

This started as I was trying to understand why we had PyOpenSSL installed on the project I joined recently. After a bit of digging, it seems that it came from the time when we were using Python 2 and was for Requests/urllib3.

However, looking at Requests' documentation, I couldn't find a clear explanation about this feature, so this is my attempt at fixing this.

Disclaimer: I'm by no means an expert at SNI, so please correct me on anything wrong that follows.

Here are what I could find:

- The intend was to [document this feature](https://github.com/requests/requests/pull/2195#issuecomment-54524390) after merging it, so I assume this contribution is welcome.
- Looking at [the documentation from urllib3](https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl-py2), it seems that the extra dependencies aren't needed in Python 3* and by the look of #4315 it can slow down importing Requests when we have PyOpenSSL.
- Another open pull request (#4591) is trying to change this, but @sigmavirus24 comment implies that there is more to SNI in Requests?

I'm not sure if this is the best place to document this, adding it to the "install" section feels overwhelming for beginners, and I felt like it belongs close to the SSL & certificates verification of the advanced section. Happy to move it somewhere else.

*Edit:*

(*) Actually, it seems it was implemented in [PEP 476](https://www.python.org/dev/peps/pep-0476/) which covers Python >=2.7.9 or >=[3.4.3](https://docs.python.org/3.4/whatsnew/changelog.html#python-3-4-3-final), so not any Python 3 versions.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -269,6 +269,28 @@ If you specify a wrong path or an invalid cert, you'll get a SSLError::
 .. warning:: The private key to your local certificate *must* be unencrypted.
    Currently, Requests does not support using encrypted keys.
 
+.. _sni-support:
+
+SNI support with legacy Python
+------------------------------
+
+.. versionadded:: 2.4.1
+
+It's possible to enable `Server Name Indication (SNI)`_ via `urllib3's SNI support`_
+for legacy Python versions (older than 2.7.9) by installing a few optional dependencies
+using the ``security`` extra::
+
+    $ pipenv install requests[security]
+
+This will install PyOpenSSL and its dependencies and setup ``urllib3`` automatically
+when Requests is imported.
+
+However, it should not be required if you're using Python 3 and you could save a bit
+of time when importing Requests by not including it.
+
+.. _`Server Name Indication (SNI)`: https://en.wikipedia.org/wiki/Server_Name_Indication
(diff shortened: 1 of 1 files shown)


docs/user/advanced.rst
```diff
@@ -269,6 +269,28 @@ If you specify a wrong path or an invalid cert, you'll get a SSLError::
 .. warning:: The private key to your local certificate *must* be unencrypted.
    Currently, Requests does not support using encrypted keys.
 
+.. _sni-support:
+
+SNI support with legacy Python
+------------------------------
+
+.. versionadded:: 2.4.1
```
browniebroke: I've put this `versionadded` directive, but since it's quoting a very old version, it might not make sense?

sigmavirus24: This was originally added for SNI support but further is really just a way of providing modern OpenSSL for old versions of Python too. It allows them to communicate with modern websites regardless of SNI.

browniebroke: Thanks for your feedback. I've updated the change to be less focused on SNI, and direct to the upstream documentation for `urllib3`.

sethmlarson:
Closing this as we're not recommending the security extra any longer.

PR #4827, opened 2018-10-16
CVE URLs update: www sub-subdomain no longer valid

webmaven:


nateprewitt:
Thanks @webmaven!

webmaven:
You're welcome @nateprewitt. Was wondering if adding the new CVE (http://cve.mitre.org/cgi-bin/cvename.cgi?name=2018-18074 eg. #4718) would also be an appropriate documentation change at this point, but would need to know next release's version number (eg. 2.19.2, or 2.20?) to include it.

nateprewitt:
@webmaven, thanks for the offer! We’re likely going to be doing a release sometime next week, and I think we’ll get all of that info bundled during the release process.

cachedout:
@nateprewitt and @webmaven Do you have any insight as to whether or not this change could be applied to versions 2.6.0 and 2.7.0 and, if so, would doing so address the security concern outlined in the CVE? We're distributing those versions in a public repo and are considering just applying this change instead of forcing users through an upgrade quite far forward. Any thoughts would be much appreciated. :)

nateprewitt:
Hi @cachedout, I think you *could* apply the patch in #4718 (or a derivative) to the head of 2.6 or 2.7. We don’t have any intention to maintain that in Requests though since both of those release are approaching 4 years since release and are 13 versions behind.

If you choose to go down that path for Saltstack, we probably want to make it clear that it’s a forked version of Requests at that point. If you’re already vendoring copies though, that may not be a problem.

cachedout:
@nateprewitt Totally understood and thanks for the quick reply. That gives me what I need. Thanks!

PR #4828, opened 2018-10-17
compat: py 2.7 has ordered dicts

cHYzZQo:
Use it if available rather than relying on urllib3 to provide it.

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -45,7 +45,10 @@ if is_py2:
     from StringIO import StringIO
     from collections import Callable, Mapping, MutableMapping
 
-    from urllib3.packages.ordered_dict import OrderedDict
+    try:
+        from collections import OrderedDict
+    except ImportError:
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @cHYzZQo, thanks for contributing to Requests! Unfortunately, we're already addressing the urllib3 1.24 related changes in our upcoming release patch. If you upgrade to 1.24 before we move the versioning pin, we won't be able to guarantee things work. Our current plan is to have a release ready within the next week after any major bugs in urllib3 have had a chance to surface.

cHYzZQo:
I see you got about 5 of these, so I'm fine with letting this drop. Unfortunately pip is a mess and depending on what order things happen in it doesn't always respect requests' pinning. Are your coming changes anywhere in public yet?

cHYzZQo:
nvmd just saw the v2.20.0 PR., tx

PR #4829, opened 2018-10-17
Follow suit of urllib3 and drop python2.6

asottile:
- Already untested (?)
- Resolves https://github.com/urllib3/urllib3/issues/1456

I used [pyupgrade](https://github.com/asottile/pyupgrade) for the syntax changes.  Most of the motivation was to resolve the urllib3 issue.

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -79,14 +79,14 @@ def _check_cryptography(cryptography_version):
         return
 
     if cryptography_version < [1, 3, 4]:
-        warning = 'Old version of cryptography ({0}) may cause slowdown.'.format(cryptography_version)
+        warning = 'Old version of cryptography ({}) may cause slowdown.'.format(cryptography_version)
         warnings.warn(warning, RequestsDependencyWarning)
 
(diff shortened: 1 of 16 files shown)


nateprewitt:
Hi @asottile, thanks for the pull request. As we noted in the urllib3 issue, this is already addressed and will be in our release within the next week once the changes in urllib3 have had some time to air out.

nateprewitt:
I’ve created #4830 for visibility. Thanks again @asottile!

asottile:
@nateprewitt disregarding the urllib3 issue -- is this PR not ok? does requests intend to support python2.6 (it doesn't test against it)

nateprewitt:
There’s nothing wrong with the PR, but we already did this work a year ago. We haven’t supported python2.6 in a while but hadn’t intended to forcibly break compat until 3.0.0. 

The reason we declined to merge this is it’s more work to rebase the release commits and would remove credit from the original author. It also introduces some breakages that will affect a number of our downstream dependencies which are unrelated to 2.6.

PR #4834, opened 2018-10-17
Allow urllib3 1.24

AymericDu:
Changelog: https://github.com/urllib3/urllib3/blob/1.24/CHANGES.rst#124-2018-10-16


nateprewitt:
Hi @AymericDu, thanks for contributing to Requests. We're aware of the 1.24 release and have an announcement up in the issues detailing next steps. Once we've been able to vet the release, we'll be pushing a new version of Requests with the pin move. Thanks for your patience.

PR #4835, opened 2018-10-17
v2.20.0

nateprewitt:
2.20.0 (2018-10-??)
-------------------
**Bugfixes**

-   Content-Type header parsing is now case-insensitive (e.g.
    charset=utf8 v Charset=utf8).
-   Fixed exception leak where certain redirect urls would raise
    uncaught urllib3 exceptions.
-   Requests removes Authorization headers from requests redirected
    from https to http on the same hostname. (CVE-2018-18074)
-   `should_bypass_proxies` now handles URIs without hostnames, e.g. files.

**Dependencies**
- We now support urllib3 v1.24.

**Deprecations**
- Requests has officially stopped support for Python 2.6.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,12 +6,29 @@ dev
 
 **Bugfixes**
 
+-   \[Short description of non-trivial change.\]
+
+2.20.0 (2018-10-18)
+-------------------
+
+**Bugfixes**
+
 -   Content-Type header parsing is now case-insensitive (e.g.
     charset=utf8 v Charset=utf8).
 -   Fixed exception leak where certain redirect urls would raise
     uncaught urllib3 exceptions.
+-   Requests removes Authorization header from requests redirected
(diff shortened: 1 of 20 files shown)


HISTORY.md
```diff
@@ -6,12 +6,25 @@ dev
 
 **Bugfixes**
 
+-   \[Short description of non-trivial change.\]
+
+2.20.0 (2018-10-??)
```
nateprewitt: We'll need to update the date here before merging.

kapilt:
afaics this doesn't address one major incompatibility with python 2.7 and latest urllib3

namely requests.compat still does

```
    from urllib3.packages.ordered_dict import OrderedDict
```

ie. see https://github.com/urllib3/urllib3/issues/1456

nateprewitt:
🛬Landing.

Issue #4830, opened 2018-10-17
Urllib3 1.24


nateprewitt:
Urllib3 released a new version today which has some breaking changes for Python 2.6. While we work out any other possible incompatibilities Requests will remain compatible with urllib3 1.21.1 to 1.23 for the next week.

If you’re experiencing issues due to pip’s dependency resolution, you will need to pin ˋurllib3==1.23` in your package requirements.

laurentvv:
same problem with python 3.6 and urllib3 1.24 : 
/usr/lib/python3.6/site-packages/requests/__init__.py:91: RequestsDependencyWarn                                                                                                                                                             ing: urllib3 (1.24) or chardet (3.0.4) doesn't match a supported version!
  RequestsDependencyWarning)

Only warning don't block my program

Lucas-C:
Same issue here, with this message during `pip install` : 

    requests 2.19.1 has requirement urllib3<1.24,>=1.21.1, but you'll have urllib3 1.24 which is incompatible.

and this error at execution time :
```
/usr/lib/python2.7/site-packages/gitlab/__init__.py:29: in <module>
    import requests
/usr/lib/python2.7/site-packages/requests/__init__.py:112: in <module>
    from . import utils
/usr/lib/python2.7/site-packages/requests/utils.py:24: in <module>
    from . import certs
E   ImportError: cannot import name certs
```

Pinning `urllib3==1.23` worked as a workaround

siphr:
Pinned for workaround but hopefully this is not a permanent solution.

hrxn:
\#MeToo
```powershell
PS E:\Test> pip check
requests 2.19.1 has requirement urllib3<1.24,>=1.21.1, but you have urllib3 1.24.
PS E:\Test>
```
https://pypi.org/project/urllib3/#history
https://pypi.org/project/requests/#history

New👏release👏now👏(soon?)

And if you allow me to enunciate a modest proposal:
Strive a bit for better coordination in the future in order to avoid releasing incompatible changes.
Isn't this Python after all? All grownup now, big and professional, supposedly.

nateprewitt:
@Hrxn, just to be clear, we’ve just crossed the 24 hour threshold since the urllib3 release. We’ve provided both a work around and posted a patch with the fix for review within hours. This is the same release process that’s been in place for a couple years now.

[...]

PR #4836, opened 2018-10-17
Allow urllib3 1.24

ofek:
A new version of has been released. Most notably it ships https://github.com/urllib3/urllib3/pull/1346

That, plus https://github.com/requests/requests/pull/4718 on our side will resolve https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-18074. We just need to release after merging this.

cc @sigmavirus24 @nateprewitt


nateprewitt:
Hi @ofek, we’ve got comms available in #4830 and the patch for 2.20 is staged in #4835.

PR #4838, opened 2018-10-20
Fixed sending objects with no .read and with length

tzickel:
This is a bug fix, because sending such an object currently will cause an exception:
```python
  File "test.py", line 81, in client
    r = session.post(url, data=DataWithoutLength(a)).text
  File "/Users/icebreak/source/requestssux/requests/requests/sessions.py", line 572, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/Users/icebreak/source/requestssux/requests/requests/sessions.py", line 524, in request
  File "/Users/icebreak/source/requestssux/requests/requests/sessions.py", line 637, in send
  File "/Users/icebreak/source/requestssux/requests/requests/adapters.py", line 449, in send
    timeout=timeout
  File "/Users/icebreak/source/requestssux/.venv2/lib/python2.7/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/Users/icebreak/source/requestssux/.venv2/lib/python2.7/site-packages/urllib3/connectionpool.py", line 354, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/Users/icebreak/miniconda2/lib/python2.7/httplib.py", line 1042, in request
    self._send_request(method, url, body, headers)
  File "/Users/icebreak/miniconda2/lib/python2.7/httplib.py", line 1082, in _send_request
    self.endheaders(body)
  File "/Users/icebreak/miniconda2/lib/python2.7/httplib.py", line 1038, in endheaders
    self._send_output(message_body)
  File "/Users/icebreak/miniconda2/lib/python2.7/httplib.py", line 886, in _send_output
    self.send(message_body)
  File "/Users/icebreak/miniconda2/lib/python2.7/httplib.py", line 858, in send
    self.sock.sendall(data)
  File "/Users/icebreak/miniconda2/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
TypeError: sendall() argument 1 must be convertible to a buffer, not DataWithoutLength
```
The patch makes sure that it will only fix this specific instance.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -29,7 +29,7 @@ from urllib3.exceptions import ResponseError
 from urllib3.exceptions import LocationValueError
 
 from .models import Response
-from .compat import urlparse, basestring
+from .compat import urlparse, basestring, is_py2
 from .utils import (DEFAULT_CA_BUNDLE_PATH, extract_zipped_paths,
                     get_encoding_from_headers, prepend_scheme_if_needed,
                     get_auth_from_url, urldefragauth, select_proxy)
@@ -418,6 +418,12 @@ class HTTPAdapter(BaseAdapter):
         self.add_headers(request, stream=stream, timeout=timeout, verify=verify, cert=cert, proxies=proxies)
 
         chunked = not (request.body is None or 'Content-Length' in request.headers)
+        if is_py2 and request.body and not hasattr(request.body, 'read') and hasattr(request.body, '__iter__'):
+            try:
+                # httplib's send will work if body supports the buffer protocol
+                memoryview(request.body)
+            except TypeError:
+                chunked = True
 
         if isinstance(timeout, tuple):
             try:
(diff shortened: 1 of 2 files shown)


sigmavirus24:
I believe the correct, backwards-compatible, and feature-intended behaviour is to raise a different exception that we can not and will not handle the body

nateprewitt:
Hi @tzickel,

If you read the [documentation](http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests) and the docstrings for the post method, you can see we support the following:
> :param data: (optional) Dictionary, list of tuples, bytes, or file-like object to send in the body of the :class:`Request`.

You can find the definition of file-like [here](https://docs.python.org/3/glossary.html#term-file-object).

We will have the same problem on most of our interfaces when supplied objects not meeting the expected input criteria. We could also implement a custom `Header` class, but if it doesn't conform to the mapping interface, it's unlikely to work as expected.

I believe you'll find if you replace your test case with the minimum of implementing `IOBase` from the `io` module that it fixes the error you're encountering. While @sigmavirus24 is right that we could implement exception handling, I think it's a bit of a slippery slope to start adding exact type checking to every API entry. 


This is the test code I used:
<details>

```python
class TestIteratorWithLength(io.IOBase):
    def __init__(self, data):
        self.data = data.encode()
        self.length = len(self.data)
        self.itersize = 2**20

    def __len__(self):
        return self.length

    def __iter__(self):
        return self

    def __next__(self):
        if self.data:
            ret = self.data[:self.itersize]
            self.data = self.data[self.itersize:]
            return ret
        else:
            raise StopIteration()

    next = __next__

test = TestIteratorWithLength('test')

r = requests.post('https://httpbin.org/post', data=test)
print(r.json())
```

</details>

tzickel:
@nateprewitt The issue with file object is described here:
https://github.com/requests/requests/issues/4837
mainly, it's a performance hog both I/o and memory allocation, since it will call .read with 8192 blocks instead of iterating and getting how much data it can (currently i'm streaming the data in 64MB chunks....).

Also seems pretty arbitrary to support \_\_iter__ without length, but not support it with...

tzickel:
Hmm.... this is an issue only in python2 (and maybe old python 3 versions)... newer python http handling can bypass this requests limitation, I still think requests should act the same...

Projjol-zz:
Closing due to inactivity and label

PR #4849, opened 2018-10-27
Extending Gitignore for Easier Contribution

erayerdin:
`.gitignore` has been changed.

# Rationale
Developers/contributors might use different tools. These tools might provide undesired files about tests, test reports, binary files et cetera. That's why `.gitignore` had to be updated.

# What has changed?
Old entries are still present at the end of file so that maintainers can still have the same result. The file was generated by [gitignore.io](https://www.gitignore.io). A full path to what aspects have been included can be seen in the first line of `.gitignore` file. Anyway, to be verbose:

 - **Jetbrains + Pycharm** caches, configs etc.
 - Generic **Python** files which are not supposed to be in the repository such as archived sources, compiled `pyc` files etc. This section also contains some files for `pytest`, `coverage` and `tox`.
 - **Virtualenv** has been extended. Other developers/tools might use different path for virtualenv. Pycharm default is `.venv` while current gitignore does not have that, instead, it has `env/` entry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -282,3 +282,30 @@ pip-selfcheck.json
 
 
 # End of https://www.gitignore.io/api/git,python,virtualenv,pycharm+all,jetbrains+all
+
+## old ignore ##
+.coverage
+MANIFEST
+coverage.xml
+nosetests.xml
+junit-report.xml
+pylint.txt
+toy.py
+.cache/
+cover/
+build/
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Users of these various tools should use [a global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

Issue #4850, opened 2018-10-29
Authorization header dropped accidentally because wrong port change detection logic


quickwind:
Summary.

## Expected Result

We have a library code using requests, when we get a URL with auth, it will be redirected to cas authentication url which is exactly same schema, same hostname, but for the port, our original request explicitly gave 443, but the location returned in 302 response is without ":443" (means default 443 https), in such case we expect the redirected request should not strip the authorization header. 

## Actual Result

The authorization header is stripped and causing eventually 401 error. 

## Reproduction Steps

```python
import requests
session = requests.Session()
session.auth = ('admin', 'password1')
response = session.get('https://example.com:443/foo')
```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.7"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.6.5"
  },
  "platform": {
    "release": "4.15.0-36-generic",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.20.0"
  },
  "system_ssl": {
    "version": "1010008f"
  },
  "urllib3": {
    "version": "1.24"
  },
  "using_pyopenssl": false
}
```

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).

nateprewitt:
Hi @quickwind, thanks for bringing this to our attention. I've quickly tossed together #4851 to address the problem. I'm not ecstatic about the solution, but it should resolve the issues. I'll try to give the issue some more thought in the morning and we'll work towards getting this resolved in the next bugfix release.

quickwind:
@nateprewitt Thanks for your quick response, looking forward to next patch version...

nateprewitt:
No problem @quickwind, we'll likely wait another week to see if anything else crops up and then I'll look at getting a hotfix release out. Going to close now that #4851 is merged.

PR #4851, opened 2018-10-29
proper handling for default ports in auth stripping

nateprewitt:
This is an attempt to address the default port issues presented in #4850. Our recent changes around auth stripping to handle downgrade attacks in #4718 broke compatibility for cases like http://example.com:80 -> http://example.com. This will allow the use of "default" ports and no port interchangeably.

If you get a moment @sigmavirus24, I'd like a second pair of eyes on this. I'm not ecstatic about continuing to expand this function, but I don't think there's an easier simplification of what we have.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -19,7 +19,7 @@ from .cookies import (
 from .models import Request, PreparedRequest, DEFAULT_REDIRECT_LIMIT
 from .hooks import default_hooks, dispatch_hook
 from ._internal_utils import to_native_string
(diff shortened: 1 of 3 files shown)


rishabh5j:
@nateprewitt  : hi nate. version bump not required?

nateprewitt:
@rishabh5j, this will be in the next release, likely 2.20.1.

rishabh5j:
@nateprewitt  : thanks i can see its version upped now.

PR #4855, opened 2018-10-30
Fix the syntax highlighting of README

frostming:


nateprewitt:
Hi @frostming, I believe these classes are defined for specific formatting by @kennethreitz. I'm going to err on the side of caution and pass on these for now since the current readme is valid markdown. If Kenneth has different thoughts, I'll defer to him to reopen and merge this.

Thanks for the suggestion!

PR #4856, opened 2018-10-31
Remove unnecessary travis config

sloria:
Each job already defines its `script`, so need 
to define it at the top level.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@ sudo: false
 language: python
 # command to install dependencies
 install: "make"
-# command to run tests
-script:
-  - |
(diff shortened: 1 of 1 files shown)


sloria:
Not sure why this was closed. Is there anything I should fix up?

Also, I just noticed that `sudo:` should be removed. TravisCI has migrated to the new infrastructure already: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration . It's also not necessary on [xenial for Python 3.7](https://github.com/deadsnakes/travis-ci-python3.7-example#update-2018-07-03)

PR #4858, opened 2018-11-04
standard library suppress instead of try...except...pass

Hyaxia:
Code readability :)
Just thought it would be nice to use the suppress instead of the except...pass usage.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -34,7 +34,7 @@ from ._internal_utils import to_native_string, unicode_is_ascii
 from .utils import (
     guess_filename, get_auth_from_url, requote_uri,
     stream_decode_response_unicode, to_key_val_list, parse_header_links,
-    iter_slices, guess_json_utf, super_len, check_header_validity)
+    iter_slices, guess_json_utf, super_len, check_header_validity, suppress)
 from .compat import (
     Callable, Mapping,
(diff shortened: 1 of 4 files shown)


sigmavirus24:
I'm -1ish on this change because it seems to purely be a refactoring that isn't necessary which can make archaeology on bugs significantly more difficult. I'm not a fan of change for change's sake when it comes to requests.

nateprewitt:
I think I’m in agreement with @sigmavirus24 here. While this does make a small simplification to the code it’s adding additional overhead to ensure we continue using this going forward.

I think we’ll pass on this change for now, @hyaxia, but thank you for putting the idea forward!

Hyaxia:
Thanks for replying.
@nateprewitt, Just out of curiousity, if there was no need for the python 2.7 compatibility, and you could use the "suppress" method from the standard library,
(There was no need for the function i created but just to import the class from contextlib)
 would your opinion on the matter be different?

nateprewitt:
We’ve discussed removing Python 2 support as part of Requests 3.0 but I don’t know if that decision has been finalized. We may revisit this at that point but until then I think we’re unlikely to merge a change like this.

Hyaxia:
Cool thanks!

PR #4861, opened 2018-11-06
Set iter_lines() decode_unicode default to False

altendky:
The docs just read a bit strange with `decode_unicode=None` as opposed to `False`.


sigmavirus24:
I had a suspicion that it was meaningful but it appears it isn't from a cursory look. Go ahead

nateprewitt:
Thanks again @altendky! ✨

PR #4863, opened 2018-11-08
Redirect double encoding

behanna:
This change makes requests more tolerant of a misconfigured Apache server that escapes percent signs in its redirects.  The change includes tests that fail without the change and pass with the change.


nateprewitt:
Hi @behanna, thanks for putting the time and effort into creating this patch! While we want Requests to be tolerant of common server misbehaviors, this seems be a pretty niche case. We provide the option in Requests to disable redirects with the `allow_redirects` flag for cases when users need to handle special redirect circumstances. I'd say that's probably the better approach in this case since we'd otherwise be adding somewhat magic behavior here that may not apply to all users.

PR #4869, opened 2018-11-12
Set up CI with Azure Pipelines

kaylangan:
Currently, CI and PR validation are run on AppVeyor and Travis using pretty different configurations.  This PR shows how [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/) could be used to unify the pipeline configuration and run CI and PR checks more quickly. Azure Pipelines offers unlimited minutes with 10 parallel jobs.

The first commit attempts to match the AppVeyor and Travis builds apples to apples. You can see in the [build results](https://dev.azure.com/kaylangan-github/kangantest-requests/_build/results?buildId=229&view=logs) that Azure Pipelines typically runs the Linux jobs in <2 minutes and the Windows jobs in ~5 minutes. The Linux jobs are slightly faster than on Travis; the Windows jobs, though individually slower than on AppVeyor, are faster because of the additional parallelism.

The second commit unifies the build using a single configuration. CI is run only on Linux and coverage only on Windows. Here's an example of the [build results](https://dev.azure.com/kaylangan-github/kangantest-requests/_build/results?buildId=245&view=logs). I also hooked up test reporting for the Linux builds which you can see [here](https://dev.azure.com/kaylangan-github/kangantest-requests/_build/results?buildId=245&view=ms.vss-test-web.test-result-details):
image

I was a little concerned that the Windows builds appear to fail intermittently, but I noticed that AppVeyor seems to have the same flakiness?

Would you consider trying out Azure Pipelines and adding it as additional validation for PR/CI?

Disclaimer: I'm a Program Manager on Azure Pipelines.


kennethreitz:
:sparkles: :cake: :sparkles:

We’ve been very happy with Azure Pipelines for Pipenv!

kennethreitz:
Could you sponsor the ‘kennethreitz’ org for free pipelines?

Also, I’m getting this error when setting it up:

> Unexpected vmImage

kennethreitz:
got it working!

kaylangan:
Since the project is public, you're already set up for the free 10 pipelines. Let me know if you need anything else!

kennethreitz:
I meant for parallel builds :)

kaylangan:
Yup, you should be all set! Let me know if you're seeing otherwise.

kennethreitz:
Thanks so much!

sigmavirus24:
@kennethreitz Can you add @nateprewitt and myself to the "kennethreitz" org for Pipelines? Otherwise, I'm going to ask that we remove this if we don't have a way of modifying/updating the CI bulds.

PR #4889, opened 2018-12-04
Bumped maximum version of idna

alex:


alex:
AppVeyor failures don't seem related to this patch

alex:
Failing vsts build also looks unrelated to this patch.

mazulo:
@alex The build will fail for Python 3.5 on `idna==2.8` so I think the bump isn't safe. I just saw the same at work

alex:
I'm not sure I understand? Why do you think this will fail with Python 3.5? AFAICT, idna did not drop support for it: https://github.com/kjd/idna/commit/2618ad17c86498d627ebb136143c2799ad48cbe3

mazulo:
We're using 3.5 (where also failed) and the [request-ci](https://kennethreitz.visualstudio.com/requests/_build/results?buildId=17) in this PR failed for the same version (just checked it was for Windows, tho). I'm also not sure about why. Maybe it's a coincidence and I misunderstood the log

nateprewitt:
Hey @alex, thanks for getting this put together. I agree the failures look unrelated. Kenneth recently changed our testing infrastructure and I don’t think we’ve validated it’s working since. I’ll take a look at things this evening and try to work out the issue.

nateprewitt:
So it looks like this is just a spurious pipenv failure. I've disabled appveyor but I don't seem to have any permissions for Azure. @kennethreitz do we have a way to get Ian and I access to the build infrastructure?

jakul:
Why not `<3` ? Pinning a maximum minor version causes install fails of `requests` whenever a new minor version of `idna` is released, even if that version is compatible. Instead, you could allow _any_ version of idna v2, and then there will only be issues with `requests` if there are incompatibilities in `idna`.

sigmavirus24:
@jakul we can't guarantee future compatibility with a library whose backwards compatibility we don't manage. It often occurs that people break things accidentally. That's all fine until someone upgrades their version they use with requests to find requests is broken and then we get 50 bug reports within 30 minutes of the release telling us we're broken. Most people don't immediately need the latest version of idna, they tend to use the overly eager upgrade strategy in pip meaning that they get it, think they need it, and pester us to upgrade. That's fine, because we can then fix the incompatibility when we find it, meanwhile users can restrict themselves to using the version compatible with everything.

jakul:
@sigmavirus24 Thanks for taking the time to write a detailed response. I understand your points, but I think it's a tradeoff between "the requests team have to do work everytime idna makes any release" vs "the requests team have to do work if idna does a release which actually breaks something in requests". 

That's my 2 cents, cheers.

sigmavirus24:
Yes, but that's a tradeoff we're willing to make. "Stability in the requests user experience because we can proactively fix an accidental backwards incompatibility" versus "This is an arcane instability happening in 2% of our users cases that we could have caught and now people don't trust requests to be stable because we promised forwards compatibility that we couldn't actually guarantee."

nateprewitt:
Alright we've got successful builds, so I'm going to rebase and merge. Thanks Alex!

https://ci.appveyor.com/project/kennethreitz/requests/builds/20888463

PR #4898, opened 2018-12-10
revert azure pipeline changes

nateprewitt:
Since we don't have access to the new testing infrastructure, we're going to revert the Azure changes for now. We can revisit this once the permissions issues are fixed.

Reverting commits:
fc9dfac
bc3eddc
d911f27
3da2f10
ce52825
398d110
c15d2b6
086655e


nateprewitt:
For auditing purposes, I missed e4974c7 in this commit and it's been reverted in abe950c.

PR #4909, opened 2018-12-13
#4908 json exception error

SebastianoF:


sigmavirus24:
`Response.json` is a method so that we do not need to handle the underlying exceptions. You're expected to handle them or only call `.json` when you know the response body has JSON to be decoded. Nowhere else in this library do we print anything either.

PR #4915, opened 2018-12-23
Normalize percent-encoded bytes before comparison

sethmlarson:
urllib3 replaced it's URL parser with `rfc3986` for more compliant parsing (urllib3/urllib3#1487). We've also got an effort underway to run downstream tests before releases (urllib3/urllib3#1508) and during that process we've discovered that requests requires percent-encoded bytes to keep their casing.

Adding this to the unit tests would allow either version of pre- and post-rfc3986 urllib3 to pass requests unit tests. I don't know if the `Request.url` attribute needs to maintain it's casing in this situation or not, if it does this is not the solution. Just putting this together as a potential solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,5 +38,5 @@ jobs:
       dist: xenial
       sudo: true     
     - stage: coverage
-      python: 3.6
+      python: '3.6'
       script: codecov
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: docs
 init:
 	pip install pipenv --upgrade
-	pipenv install --dev --skip-lock
+	pipenv install --dev
 test:
 	# This runs all of the tests, on both Python 2 and Python 3.
 	detox
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [dev-packages]
-pytest = ">=2.8.0"
+pytest = ">=2.8.0,<4.1"
 codecov = "*"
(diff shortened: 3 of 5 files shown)


sigmavirus24:
Closing and re-opening to trigger a new build

sethmlarson:
Is a newer pytest version not compatible with one of our plugins / our test setup somehow?

sigmavirus24:
Beats me

sethmlarson:
I guess I'll have to poke around with this more and see why we're getting these funky errors on test load.

sethmlarson:
Just as I suspected, pytest 4.1 is the problem. Had to deal with this in multiple places at work as well. :( Now to figure out why [Python 3.6 is still installing pytest 4.1.1 despite being locked to `<4.1`](https://travis-ci.org/requests/requests/jobs/481793727#L499).

...and it's installing on Python 3.7 on AppVeyor? Is this a bug with Pipenv and `--skip-lock`?

sethmlarson:
I don't know what you think about this @sigmavirus24 / @kennethreitz, I don't want to remove the `--skip-lock` because we're getting really outdated dependencies but it looks like `pipenv` is having trouble with the version specifier?

.travis.yml
```diff
@@ -38,5 +38,5 @@ jobs:
       dist: xenial
       sudo: true     
     - stage: coverage
-      python: 3.6
+      python: '3.6'
```
sigmavirus24, line 41: This is odd

sethmlarson:
Thanks for merging this!

PR #4917, opened 2018-12-28
Remove unnecessary compat shim for bytes

jdufresne:
Both Python 2.7 & Python 3 have the type `bytes`. On Python 2.7, it is an alias of `str`, same as what was defined in `compat.py`.

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -47,7 +47,6 @@ if is_py2:
 
 
     builtin_str = str
-    bytes = str
     str = unicode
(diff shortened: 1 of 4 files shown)


kennethreitz:
people rely on this code. it's cruft from our 2.5/2.6 days.

jdufresne:
Even if the import in `compat.py` sticks around for backwards compatibility, there is no reason to use the non-standard `bytes` internally. Same reasoning as applied in #4922.

PR #4919, opened 2018-12-28
Updated wording for switch devhelp

blak3irwin:
Adding file location for devhelp switch in make.bat


sigmavirus24:
Thank you, but this file is auto-generated and not manually maintained.

Issue #4876, opened 2018-11-16
chunk_size=None does not work as expected when streaming a response


hiranya911:
Summary.

I have the following code to stream an SSE response.

```
resp = session.get(self.url, stream=True)
resp_iterator = resp.iter_content(decode_unicode=True, chunk_size=None)
while _some_condition():
    nextchar = next(resp_iterator)
    _add_to_buffer(nextchar)
```

## Expected Result

According to the [documentation](http://docs.python-requests.org/en/master/api/#requests.Response.iter_content), when chunk_size is None, it should stream the content as it is received. 

## Actual Result

No content is received by the consumer. It simply hangs. If I simply remove the chunk_size argument, I can see the consumer receiving 1 byte at a time.

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  }, 
  "cryptography": {
    "version": ""
  }, 
  "idna": {
    "version": "2.7"
  }, 
  "implementation": {
    "name": "CPython", 
    "version": "2.7.13"
  }, 
  "platform": {
    "release": "4.17.0-3rodete2-amd64", 
    "system": "Linux"
  }, 
  "pyOpenSSL": {
    "openssl_version": "", 
    "version": null
  }, 
  "requests": {
    "version": "2.20.1"
  }, 
  "system_ssl": {
    "version": "1010008f"
  }, 
  "urllib3": {
    "version": "1.24.1"
  }, 
  "using_pyopenssl": false
}
```

sigmavirus24:
I suspect that this is an issue with urllib3's stream method that wasn't caught when it was changed.

https://github.com/requests/requests/blob/57d7284c1a245cf9fbcecb594f50471d86e879f7/requests/models.py#L750

sethmlarson:
The two modes within `HTTPResponse.stream()` (being `HTTPResponse.read()` and `HTTPResponse.read_chunked()`) both *appear* to support setting `amt=None` and it's the default parameter for all three. Do you know where that regression may have been introduced @sigmavirus24?

It'd be great to know what the headers are for the HTTP response you're getting here because a lot of the behavior within `HTTPResponse.stream()` is determined by a few headers.

nateprewitt:
As a quick FYI, I did some minimal debugging and this appears to only be reproducible with Requests and not with generic usage of urllib3. It's likely our adapters code may be doing something to the urllib3 Reponse object that isn't being tested for in Requests or urllib3.

**Requests Repro**
```python
[...]

PR #4920, opened 2018-12-28
iter_content() checks for < 1 chunk_size

alex-wenzel:
With requests v2.20.1, supplying a value of 0 for `chunk_size` causes `iter_content()` to hang. This made debugging code that dynamically computes a value for `chunk_size` difficult. These two lines explicitly enforce `chunk_size` being a positive integer.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -769,7 +769,7 @@ class Response(object):
             raise StreamConsumedError()
         elif chunk_size is not None and not isinstance(chunk_size, int):
             raise TypeError("chunk_size must be an int, it is instead a %s." % type(chunk_size))
-        elif chunk_size < 1:
+        elif chunk_size is not None and chunk_size < 1:
             raise ValueError("chunk_size must be a positive integer, got "+str(chunk_size))
         # simulate reading small chunks of the content
         reused_chunks = iter_slices(self._content, chunk_size)
--- a/requests/models.py
+++ b/requests/models.py
@@ -769,6 +769,8 @@ class Response(object):
             raise StreamConsumedError()
         elif chunk_size is not None and not isinstance(chunk_size, int):
             raise TypeError("chunk_size must be an int, it is instead a %s." % type(chunk_size))
+        elif chunk_size < 1:
+            raise ValueError("chunk_size must be a positive integer, got "+str(chunk_size))
         # simulate reading small chunks of the content
         reused_chunks = iter_slices(self._content, chunk_size)
 


nateprewitt:
Hi @alex-wenzel, thanks for taking a look at this. I believe this is actually a duplicate of #4876. This is a new behaviour introduced with the last release of urllib3 where 0 no longer denotes "any size". I believe this is probably a bug since it doesn't make sense to iterate in chunks of 0.

In regards to the PR, I don't think we want to handle the issue with an exception. We handle 0 correctly in `iter_slices` and preventing that would further break the API expectations. We'll track next steps in #4876 for now.

/cc @sethmlarson

alex-wenzel:
@nateprewitt That makes sense, thanks for the quick review/reply!

sethmlarson:
I'll investigate, thanks for the report.

sethmlarson:
What I'm seeing so far is that urllib3 hasn't (in recent history, tested urllib3 1.18 to 1.24.1) not-blocked indefinitely when using `amt=0` into `HTTPResponse.stream()`. Is this issue different from the one described into #4876 which appears to be about `amt=None`?

nateprewitt:
@sethmlarson after a closer look, it is not. `None` does appear to work correctly with all of the combinations I tried today. It looks like Requests has never supported this actually, at least not since 2.0.

@alex-wenzel I'm going to reopen this since I agree we can do a better job here. Part of me would be more inclined to treat 0 the same way we treat `None`. That would require special-casing logic for `0`->`None` but the alternative of requiring everyone implement this themselves (`len('')`-> `None`) seems silly. The case of `amt<0` is already handled by the standard library. @sigmavirus24, @sethmlarson, any thoughts on that?

sethmlarson:
What do other streaming interfaces typically do with a chunk size of zero? Any examples in the stdlib? I'd be inclined to copy whatever is done there.

alex-wenzel:
For what it's worth, the code I was debugging includes:

```python
file_size = int(r.headers['Content-Length'])
chunk_size = int(file_size / 50)
```

...so files smaller than 50 bytes would hang forever. I have to imagine this would be one of the most common reasons a user would inadvertently set `chunk_size=0` and in this case I think having it behave like `None` would be reasonable.

Also, slightly tangential to this thread, but it looks like my most recent changes failed in AppVeyor with an `ImportError` in Python 3.7 specifically that doesn't look related to my changes.

nateprewitt:
I took a look at getting this added into 2.27.0 but it has some unfortunate side effects. Non-streamed requests currently succeed at returning data from `iter_content` for non-positive integers. While I agree that's nonsensical, adding this behavior breaks that workflow, which we won't be able to do in a minor release.

For the time being, this is an unfortunate behavior which we could definitely document for `iter_content`. We won't be able to merge this PR until it's reworked to maintain current behavior though. Unfortunately, I don't see an immediate way to do that.

PR #4922, opened 2018-12-29
Remove unnecessary compat shim for OrderedDict

jdufresne:
The shim is the same on both Python 2 & 3. It is always `collections.OrderedDict`. Avoid the indirection and import from Python stdlib instead.

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -43,6 +43,7 @@ if is_py2:
     import cookielib
     from Cookie import Morsel
     from StringIO import StringIO
+    # Keep OrderedDict for backwards compatibility.
     from collections import Callable, Mapping, MutableMapping, OrderedDict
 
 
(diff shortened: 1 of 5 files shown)


nateprewitt:
Hey @jdufresne, thanks for putting this together. We originally had this change in the commits to remove Python 2.6, but there are at least a handful of well known projects that use Requests compat imports directly.

In order to maintain backwards compatibility, we went ahead and left these intact in Requests 2.x (at least for now) and removed them in the Requests 3.x branch. I think I’d still bias towards keeping them unless there’s a compelling reason they need to go.

jdufresne:
Thanks. That makes a lot of sense.

What do you think about keeping the compat shim, but not using it internally? That would avoid the unnecessary indirection internally, but keep the compat import for backwards compatibility. I could also add a comment to explain why it exists so others understand it isn't an oversight.

nateprewitt:
Sure, I think that sounds reasonable.

PR #4924, opened 2018-12-31
Include Python Information with default User Agent

cooperlees:
- A lot of people use default user agent with requests
- Lets include Python runtime + version information
- Allows servers to know some more useful information about their users e.g. PyPI / Warehouse

This will help pypistats.org for example - https://github.com/crflynn/pypistats.org/issues/13

Open to other ideas here. Just throwing this up to start the discussion.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -796,7 +796,13 @@ def default_user_agent(name="python-requests"):
 
     :rtype: str
     """
-    return '%s/%s' % (name, __version__)
+    try:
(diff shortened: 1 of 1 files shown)


sigmavirus24:
- https://github.com/requests/requests/pull/709
- https://github.com/requests/requests/pull/2803
- https://github.com/requests/requests/issues/2785
- https://github.com/requests/requests/issues/3182

In short, we used to include that and now don't. And we won't be adding it back in any time soon

PR #4932, opened 2019-01-08
Make requests compliant with PEP 476

ssbarnea:
Enables loading of certificate bundles from SSL_CERT_FILE or
SSL_CERT_DIR as mentioned in https://legacy.python.org/dev/peps/pep-0476/

Current patch keeps backwards compatibility by preferring the more
specific options and using the new ones only as fallbacks.

Fixes: #2899

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -701,11 +701,13 @@ class Session(SessionRedirectMixin):
             for (k, v) in env_proxies.items():
                 proxies.setdefault(k, v)
 
-            # Look for requests environment configuration and be compatible
-            # with cURL.
+            # Look for requests CA_BUNDLE configuration in the environment. Be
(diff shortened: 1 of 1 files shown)


ofek:
@Lukasa Shouldn't this already have been on master via https://github.com/requests/requests/pull/2903?

Could it be due to this? https://github.com/requests/requests/commit/85a29f1da0d0ccd79e865a7172ea39a2a7667312

nateprewitt:
@ofek this was merged into the Proposed/3.0.0 branch which isn’t merged with master yet. I’m inclined to uphold the decision Lukasa made the last time this PR was proposed. We’ll treat this as a breaking change because it does have a high probability to break existing setups.

Thanks for the proposal @ssbarnea!

Thaodan:
Will work REQUESTS_CA_BUNDLE with capath until than too?

PR #4944, opened 2019-01-21
Fixes: TestSuperLen class

deepio:
Pytest itself would fail, until I fixed the semantics of the pytest params in the decorator for `test_io_streams`. I am not certain unittests are being used currently, as many tests failed prior to changing anything. Am I not using them correctly?

I also wanted to start a discussion and share thoughts on default_ports, I believe it should be written in the same manner as default_headers.


deepio:
@sigmavirus24 Can you provide more information?

deepio:
Closed due to no feedback.

PR #4947, opened 2019-01-21
Docstring explaining when HTTPError is/is not raised

jerojasro:
This is mostly to tell the reader about when HTTP error must be handled, and to better link to the other places in the doc. where this is explained.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -538,6 +538,8 @@ to hang indefinitely::
     not time out.
 
 
+.. _errors-and-exceptions:
+
 Errors and Exceptions
 ---------------------
 
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -26,7 +26,14 @@ class RequestException(IOError):
 
 
 class HTTPError(RequestException):
-    """An HTTP error occurred."""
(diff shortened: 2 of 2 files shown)


recepsirin (approved):

sigmavirus24:
Thanks for the contribution @jerojasro and trying to improve the documentation!

Personally, I'm not sure I understand _why_ this change is helpful. From what I can understand from the analytics of the requests documentation, most people don't read up on exceptions first and instead can understand the content of this from the examples and documentation already available. I feel like this is a solution in search of a problem. I'm going to close this due to its age but it can be re-opened if you want to help me understand your reasoning here and maybe workshop a different update

jerojasro:
hello @sigmavirus24 ; my goal here is to better explain what is (and isn't) an HTTPError, since I think the current docstring is way too terse.

Reading back what I wrote, perhaps the last paragraph is unnecessary, but I think the change to the main sentence (An HTTP error (status code >= 400) occurred.) is what's important. I added that because to me (and I'm assuming here also to any non HTTP-protocol experts) is not immediately obvious what is meant by HTTP error, whether that includes timeouts, network issues, TLS issues, etc., which to a beginner might all land in the same category, even though for request's authors they do not.

PR #4954, opened 2019-01-30
[fix] Remove extra blank line

zengjuchen:
1. The removed blank line was added at commit 8a2014e26 by Kenneth Reitz at
2017-05-26 22:51:40.
The blank line under it was created at 2012, there is a 5 year gap
between them.

2. It has conflict with the style used on line 63.

So I assume it was misadded and removed it.

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -45,7 +45,6 @@ if is_py2:
     from StringIO import StringIO
     from collections import Callable, Mapping, MutableMapping, OrderedDict
 
-
     builtin_str = str
     bytes = str
     str = unicode


sigmavirus24:
Hi there,

Thanks for sending a PR to requests. Unfortunately, we have no style guide that shows one thing to be right or the other to be wrong. As such, PRs that adjust style in one way or another are not accepted.

zengjuchen:
> Hi there,
> 
> Thanks for sending a PR to requests. Unfortunately, we have no style guide that shows one thing to be right or the other to be wrong. As such, PRs that adjust style in one way or another are not accepted.

I understand your principles, but style is very important element in Python. And `Requests` package is famous of its beautiful coding. If you review the `compat.py` file now, you can see the style conflict (blank line in import area) between its `is_py2` and `is_py3` import part(in line 47, there are 2 blank lines, and in line 64, same logic, only 1 blank line). I found it was caused by the action of 
 commit `7fd9267b3bab1d` didn't fully revert the change induced in `8a2014e265d6c`.

Even though this PR can't be accepted as an official PR. I think this issue should be solved in other functional PRs.

With all my respects.

Issue #4445, opened 2018-01-04
404 responses for chunk-encoded requests may not be received


rasjidw:
404 responses (or any response sent by the server prior to the entire request being sent) for chunk-encoded requests may not be received by the caller.

This appears to happen if the server closes the connection (after sending the 404 response) prior to the completion of the chunked post / upload.

## Expected Result

Posting a chunked encoded request to an invalid URL should return the 404 response.

## Actual Result

A ```requests.exceptions.ConnectionError: [Errno 32] Broken pipe``` error is raised instead.

## Reproduction Steps

A minimal server using flask is given below:

```
#!/usr/bin/env python3

import sys

from flask import Flask, request
app = Flask(__name__)

@app.route('/')
def hello_world():
    return 'Hello, World!'

@app.route('/post_endpoint', methods=['POST'])
def post_normal_response():
    for line in request.stream:
        str_line = line.decode('utf-8')
        sys.stdout.write(str_line)
        sys.stdout.flush()
    return 'OK'

if __name__ == '__main__':
    app.run()
```

Client code illustrating the problem is:

```
#!/usr/bin/env python3

from time import sleep

import requests


VALID_URL = 'http://127.0.0.1:5000/post_endpoint'
INVALID_URL = 'http://127.0.0.1:5000/does_not_exist'


LINES = ['Line {}\n'.format(x) for x in range(5)]
DATA = ''.join(LINES).encode('utf-8') 

def chunked():
    for line in LINES:
       sleep(1)
       yield line.encode('utf-8') 

def test_url(url):
    r = requests.post(url, data=DATA)
    print(r)


    r = requests.post(url, data=chunked())
    print(r)

if __name__ == '__main__':
   test_url(VALID_URL)
   print('-----')
   test_url(INVALID_URL)
```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.5.2"
  },
  "platform": {
    "release": "4.4.0-104-generic",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.18.4"
  },
  "system_ssl": {
    "version": "1000207f"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": false
}
```

## Patch to address the issue

[...]

PR #4958, opened 2019-02-03
Delegate chunked encoding handling to urllib

leamingrad:
This MR removes the custom handling of sending chunked HTTP requests from requests to delegate to the equivalent logic in urllib3 (https://github.com/urllib3/urllib3/blob/4325867d1ae0d139a11c8689c2d2a5ba2c666c83/src/urllib3/connectionpool.py#L351).

As far as I can tell this has a couple of advantages:
* Remove code that is duplicated between the two libraries
* Allow fixes for outstanding issues in this area to be made in urllib3 so that users of both libraries benefit (I noticed #4445 when having a quick trawl of the issues but I guess there are more)
* All requests are now logged by the urllib3 debug logger (which was the issue I was originally trying to diagnose)

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -435,64 +435,19 @@ class HTTPAdapter(BaseAdapter):
             timeout = TimeoutSauce(connect=timeout, read=timeout)
 
         try:
-            if not chunked:
-                resp = conn.urlopen(
-                    method=request.method,
-                    url=url,
-                    body=request.body,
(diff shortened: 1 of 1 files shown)


leamingrad:
Quick note on testing that I forgot to put in the description: I haven't added any specific tests to cover this feature, but I think that the code is covered by existing tests from what I can see - if there are extras that would be helpful I am happy to add them.

PR #4978, opened 2019-02-12
Fix double image request

sayzlim:
Use official framework for the "custom" ad design integration.


kennethreitz:
screenshot?

sayzlim:
Same as before, only the minor changes.

Issue #2949, opened 2015-12-28
Session's Authorization header isn't sent on redirect


jwineinger:
I'm using requests to hit developer-api.nest.com and setting an Authorization header with a bearer token. On some requests, that API responds with an 307 redirect. When that happens, I still need the Authorization header to be sent on the subsequent request. I've tried using `requests.get()` as well as a session.

I suppose I could work around this by not allowing redirects, detecting the 307 and then issuing the new request myself but I'm wondering if this is a bug. Should I expect that the Authorization header would be sent on all requests made within the context of a session? 

``` python
In [41]: s = requests.Session()

In [42]: s.headers
Out[42]: {'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive', 'User-Agent': 'python-requests/2.7.0 CPython/3.4.3 Darwin/15.2.0'}

In [43]: s.headers['Authorization'] = "Bearer <snip>"

In [45]: s.get("https://developer-api.nest.com/devices/thermostats/")
Out[45]: <Response [401]>

In [46]: s.get("https://developer-api.nest.com/devices/thermostats/")
Out[46]: <Response [200]>

In [49]: Out[45].history
Out[49]: [<Response [307]>]

In [50]: Out[46].history
Out[50]: []

In [51]: Out[45].request.headers
Out[51]: {'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive', 'User-Agent': 'python-requests/2.7.0 CPython/3.4.3 Darwin/15.2.0'}

In [52]: Out[46].request.headers
Out[52]: {'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive', 'User-Agent': 'python-requests/2.7.0 CPython/3.4.3 Darwin/15.2.0', 'Authorization': 'Bearer <snip>'}
```

Lukasa:
Where is the redirect to?

jwineinger:
Ah, a different domain. firebase-apiserver03-tah01-iad01.dapi.production.nest.com

Lukasa:
Yup, so that's somewhat deliberate: we're very aggressive with stripping authorization headers when redirected to a new host. This is a safety feature to deal with [CVE 2014-1829](http://www.cve.mitre.org/cgi-bin/cvename.cgi?name=2014-1829), which was caused by us persisting headers on off-host redirects.

[...]

PR #4983, opened 2019-02-15
Add trusted domains

jaketreacher:
Solution for #2949 

Allows setting a list of domains that are "trusted", so the authorization header isn't stripped upon being redirected there.

Default functionality remains unchanged.


kennethreitz:
elegant solution! I approve.

epenet:
This seems to have broken Python 2.7 compatibility (see issue #5175 and my PR #5182)

PR #4987, opened 2019-02-17
[fix] fix failing doctests in requests/*

aless10:
## Why this PR
Running the command:
`
$ py.test --doctest-module requests
`
gives failing doctests. This PR aims to fix these failing doctests.

Issue Link: https://github.com/kennethreitz/requests/issues/4986

## How to test
`
$ py.test --doctest-module requests
`
All the 11 doctests should pass.


requests/__init__.py
```diff
@@ -9,29 +9,22 @@
...
-   >>> print(r.text)
-   {
-     ...
-     "form": {
-       "key2": "value2",
-       "key1": "value1"
-     },
-     ...
-   }
+   >>> r.json()
```
luckydenis: Hi, I would suggest to do so: 
```
   >>> print(r.text)
   ... # doctest: +ELLIPSIS
   ... # doctest: +NORMALIZE_WHITESPACE
   {
     ...
     "form": {
       "key2": "value2",
       "key1": "value1"
     },
     ...
   }
```
So we change the test minimally.

alessio-izzo: Thanks. I did not know that I could use this syntax.

aless10: Thanks. I did not know that I could use this syntax. I'm going to change it according to your suggestion.

aless10: I added NORMALIZE_WHITESPACE and ELLIPSIS to the pytest.ini file and removed from the doctest in order to use them in pytest configuration.

kennethreitz:
fantastic

PR #4999, opened 2019-02-23
Add project URLs with links to source code and docs

greysteil:
Warehouse now uses the `project_urls` provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help automation tool find the source code for Requests. For example, see Django's [setup.py](https://github.com/django/django/blob/master/setup.py) and [PyPI listing](https://pypi.org/project/Django/).


kennethreitz:
fantastic!

PR #5004, opened 2019-02-27
Lock pytest-httpbin to <1.0

sethmlarson:
[pytest-httpbin](https://github.com/kevin1024/pytest-httpbin) release 1.0.0 updated the `subjectAltName` field to the certificate which our unit tests assume doesn't have an IP Address entry in order to test warning generation.  [Changelog for 1.0.0 even mentions this breakage in our unit tests.](https://github.com/kevin1024/pytest-httpbin/commit/be3c3205972195ca632949ed7af1dc7086cb7899)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ test:
 	# This runs all of the tests, on both Python 2 and Python 3.
 	detox
 ci:
+	python -m pip freeze
 	pipenv run py.test -n 8 --boxed --junitxml=report.xml
 
 test-readme:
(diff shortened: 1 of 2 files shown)


sethmlarson:
So ran `pip freeze` after installing from pipenv and [we're still receiving pytest-httpbin==1.0.0](https://travis-ci.org/kennethreitz/requests/jobs/499055970#L551).

Don't know what to do from here. Pipenv bug?

PR #5017, opened 2019-03-09
improve error message for invalid timeouts

bennn:
Rearrange a try/except so that these two programs give similar error
messages.

Program 1:

```
import requests
requests.get("https://random.org, timeout=(True, True))
```

Program 2:

```
import requests
requests.get("https://random.org, timeout=True)
```

Before this commit, _Program 2_ raised an error in `urllib3` asking for
an "int, float or None" timeout.


bennn:
Why is this closed & not merged?

PR #5043, opened 2019-04-06
Incorporating a `__main__.py` file into request ?

Carelvd:
Python supports the inclusion of a `__main__.py` file withing ones packages. I though it might be nice to include such a  file within requests to handle the more common tasks from the command line e.g. `GET`, `HEAD`, `PUT` and `POST`. I have included an example for both `HEAD` and `GET` though I'm not certain of their utility. 

More often then not it is necessary to download a file from a given url and I have added a `download` option to assist with that as well.

### DIFF
--- a/requests/__main__.py
+++ b/requests/__main__.py
@@ -1,0 +1,60 @@
+import argparse
+import re
+from . import api
+
+def filename(request) :  
+#   print(request.headers)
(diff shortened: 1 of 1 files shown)


kennethreitz:
send this to the kennethreitz/requests3 repo!

Issue #5037, opened 2019-03-28
Add request event hook


navyad:
There was a scenario where I need to intercept the request and was searching for if there is some way can that be done, then I found the event hook but only `response` event hook is supported.
I hope adding `request` hook can be useful, if yes I would like to implement that.

Any pointers or suggestion appreciated, Thanks

abranjith:
Can you tell what exactly you mean by request hook ? What does intercepting a request mean ? It sounds like you want to do something after request has been made - for which you can use response hook

Holzhaus:
I need the same thing. A hook that retrieves the `Request` or `PreparedRequest` object in the moment when it is sent (without waiting for the response).

Holzhaus:
I implemented the hook in PR #5044.

joaonc:
What a bummer that PR #5044 was closed..   I'd love to have the request hook as well.
My scenario is that for some tests I need to manipulate the headers right before the request (after business logic is done in the REST client).

nateprewitt:
@joaonc is there a reason you can’t use the PreparedRequests workflow for this use case?

joaonc:
I guess there are several ways to skin this cat.  I have a business layer where the REST clients to our web services are defined (ex `AccountService` has a `get_account(account_id)` method) and those clients handle urls for different environments, logging, headers (we have some headers that require business logic), etc.

In some of my tests, I wanted to alter these headers.
One easy way to do it would be to add `request` to the list of hooks.
Another is to have these REST clients accept `headers` as a parameter.

I have a base REST client from which all others derive and ended up adding a check for `hooks.get('request')` and go from there, basically adding the functionality myself.  If it gets added to the `requests` library later, should be easy to switch.

sigmavirus24:
So there seem to be a couple use-cases here that I wanted to summarize and directly address:

1. A REST client library that needs to conditionally perform some logic
2. Someone wanting to handle redirects in a bespoke and unexplained way

As someone who has written multiple REST client libraries, there are many ways to skin this cat and it all depends on how well designed the client library is. For anything that's going to be remotely complex or long-lived, I personally like a layered approach. By which I mean:

- The bottom layer is a class that wraps/extends a `requests.Session` (ideally, it doesn't subclass so that you could swap it out for something similar in the future if you wanted and the rest of your library wouldn't have to know)
- The next layer is what understands URL building and other business logic that might need to happen. This could be used by power users who want to just see response data directly
[...]

PR #5044, opened 2019-04-06
Add request hook

Holzhaus:
Add a request hook that fires when a request is sent. This resolves kennethreitz/requests#5037.

### DIFF
--- a/requests/hooks.py
+++ b/requests/hooks.py
@@ -8,16 +8,19 @@ This module provides the capabilities for the Requests hooks system.
 
 Available hooks:
 
+``request``:
+    The prepared request generated when a Request is sent.
 ``response``:
     The response generated from a Request.
 """
-HOOKS = ['response']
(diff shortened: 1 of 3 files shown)


frostming:
@Holzhaus I remember the hooks were removed in some old commit https://github.com/kennethreitz/requests/commit/ee6c638e9cfdbdf39b9e499d95c353460365f42c#diff-ab3246c102b4e3c7381cf4d1ea4b2bea for a reason, I don't know why though.

MarcClown:
Seems like there was a conflict and the fix did not make it into the branch.
A similar issue ( #5037 ) states that it is fixed by this merge - but unfortunately does not seem to be in because of that conflict.
Would be great if this could be fixed. Any ETA ?

kennethreitz:
the current hook we have is the only necessary one, i determined. hence, the other ones were removed during the big refactor.

PR #5047, opened 2019-04-08
fix dependency resolution conflict and CI/CD

tonybaloney:
Removes the call to <4.1 pytest. Pytest 4.4 was released this week breaking CI/CD

Rebuild graph and Pipfile.lock.

Also removed `detox` as it's now built into tox :-)

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [dev-packages]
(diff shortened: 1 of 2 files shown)


tonybaloney:
I underestimated this fix! Python 2 failing because of a dependency on pathlib2 from Pytest.. 
Python3 failing because 1 of the tests depends on an old API

kennethreitz:
oh man :(

PR #5049, opened 2019-04-08
Support pytest 4

hroncok:
Fixes https://github.com/kennethreitz/requests/issues/5048

See https://docs.pytest.org/en/latest/deprecations.html#marks-in-pytest-mark-parametrize

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ test_requirements = [
     'pytest-mock',
     'pytest-xdist',
     'PySocks>=1.5.6, !=1.5.7',
-    'pytest>=2.8.0'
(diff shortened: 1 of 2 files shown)


hroncok:
Not sure about the CI here, it seems to be failing on master as well.
I've also tried to update the Pipfile.lock, but I have no idea how to do it properly without regenerating the entire Pipfile.lock. Tried some magic.

luckydenis:
> Fix some issues to improve adherence to best practices

Facepalm, not the and not there wrote. Sorry.

hroncok:
I've dropped the pipfile changes, they conflicted.

Issue #4945, opened 2019-01-21
API reference doesn't include Response.raw


pfalcon:
Look under http://docs.python-requests.org/en/master/api/#requests.Response - `raw` attribute is not listed, even though it's described in http://docs.python-requests.org/en/master/user/quickstart/#raw-response-content , etc.

## Expected Result

The API reference should be complete. 

## Actual Result

See above.

## Reproduction Steps

See above.

## System Information

The model.py has it like this (f37daf25006a4b475b1bea90fecdd035146dbc3b):

~~~
        #: File-like object representation of response (for advanced usage).
        #: Use of ``raw`` requires that ``stream=True`` be set on the request.
        # This requirement does not apply for use internally to Requests.
        self.raw = None
~~~

Note that last comment line doesn't start with `#:`. Maybe that causes the entire comment block to **not** be treated as doc-comment and the field is excluded from docs.

pfalcon:
Ping?

PR #5076, opened 2019-04-30
Possible fix for #4945?

myrheimb:
Could the missing colon on line 612 in models.py be the reason for the missing reference in the docs as described in issue #4945?


kennethreitz:
beautiful.

PR #5090, opened 2019-05-15
http://pipenv.org > https://github.com/pypa/pipenv

jb-reefer:
It looks like a few of the resources on https://docs.pipenv.org/en/latest/ are broken, like the gif under "A short animation of pipenv at work" and the installation instructions are a bit behind what's on the GitHub repo. Maybe we should link to the repo directly? 

First time contributing, feel free to reject this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ PyPy.
 Installation
 ------------
 
-To install Requests, simply use [pipenv](http://pipenv.org/) (or pip, of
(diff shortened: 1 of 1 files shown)


kennethreitz:
let's fix pipenv.org instead :)

the website can't be a one man operation! ;D

luksurious:
`pipenv.org` seems dead, so the link goes into Nirvana now. Can this be reopened?

PR #5104, opened 2019-05-29
[models] Fix '_encode_params'/'_encode_files'

wy-z:
Look at https://github.com/kennethreitz/requests/blob/master/requests/models.py#L100 or
https://github.com/kennethreitz/requests/blob/master/requests/models.py#L131
```
for v in vs:
    if v is not None:
        ...
```
When the type of `vs` is `dict`, `v` will be the field key, but not field value, the body of the request will be puzzling.
It would be better to prevent it.

Signed-off-by: weiyang <weiyang.ones@gmail.com>

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -119,6 +119,9 @@ class RequestEncodingMixin:
         elif hasattr(data, "__iter__"):
             result = []
             for k, vs in to_key_val_list(data):
+                if isinstance(vs, Mapping):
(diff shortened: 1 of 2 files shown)


wy-z:
https://github.com/kennethreitz/requests/issues/5058

martimlobao:
Why was this PR deleted? This was a good fix imo.

wy-z:
I'll reopen this if necessary. @martimlobao

recepsirin (approved):

PR #5105, opened 2019-05-29
Update sessions.py

alanyee:
-Made removing a key-value pair more modular
-Remove unneeeded return statement


kennethreitz:
great improvement, thanks!

PR #5110, opened 2019-06-04
Remove an obsolete attribute from Session.__attrs__

nedbat:
The extra attribute in `Session.__attrs__` makes an unpickled session have one more attribute than the original.  Removing it makes the original and the unpickled be identical.


nateprewitt:
Great catch, @nedbat! That's only been lingering for 8 years . Would you mind rebasing if you get a moment?

nedbat:
Rebased, thanks!

nedbat:
Hmm, rebasing and force-pushing apparently also "dismisses stale review"?  That certainly wasn't my intent....!

Issue #5113, opened 2019-06-05
get_encoding_from_headers methods return value


bainiwaj:
Summary.
The problem with this method is that for certain languages the ISO-8859 does not work.
## Expected Result
I would like all HTTP header text to be returned for any language and not just english.
What you expected.
Certain languages such as french http header text does not work.
## Actual Result
Encoding error with non english
What happened instead.
encoding error
## Reproduction Steps

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).

nateprewitt:
I'm going to resolve this with the same reasoning Kenneth used on the PR. This is currently the best interpretation we have over the available RFCs and reference clients. You can always set `encoding` on the Response object yourself before trying to read its contents, and it should work as expected. That would be the advised path forward in cases where encoding isn't specified.

PR #5114, opened 2019-06-05
bug fix 5113 by update to utils.py

bainiwaj:
fix to bug #5113

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -490,7 +490,10 @@ def get_encoding_from_headers(headers):
         return params['charset'].strip("'\"")
 
     if 'text' in content_type:
-        return 'ISO-8859-1'
+        return 'utf-8'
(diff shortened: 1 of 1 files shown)


requests/utils.py
```diff
@@ -490,7 +490,10 @@ def get_encoding_from_headers(headers):
         return params['charset'].strip("'\"")
 
     if 'text' in content_type:
-        return 'ISO-8859-1'
+        return 'utf-8'
+    else
+        return 'None'
+
```
kobayashi, line 496: Need to replace `else` with `else:`. Or  just remove Line 494:495

kennethreitz:
this is not the appropriate solution. change your code to accommodate, please. we are honoring RFCs here, and honoring expected behavior.

PR #5119, opened 2019-06-24
Update sessions.py

alanyee:
Use `.format()` over `%` given PEP 3101

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -162,7 +162,7 @@ class SessionRedirectMixin(object):
                 resp.raw.read(decode_content=False)
 
             if len(resp.history) >= self.max_redirects:
-                raise TooManyRedirects('Exceeded %s redirects.' % self.max_redirects, response=resp)
+                raise TooManyRedirects('Exceeded {} redirects.'.format(self.max_redirects), response=resp)
 
             # Release the connection back into the pool.
             resp.close()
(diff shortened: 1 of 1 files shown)


kobayashi:
This is great for python3. @kenneth-reitz, I just want to make sure [requests](https://github.com/kennethreitz/requests) is continuing to spport python2.7. right?

alanyee:
`.format()` should work for Python 2.7.

requests/sessions.py
```diff
@@ -728,7 +728,7 @@ def get_adapter(self, url):
                 return adapter
 
         # Nothing matches :-/
-        raise InvalidSchema("No connection adapters were found for '%s'" % url)
+        raise InvalidSchema("No connection adapters were found for {!r}".format(url))
```
bisguzar, line 731: What **!r** stands for? Python 2.7 supporting .format as well as you tell. Its ok but I couldn't get first one.

alanyee: `{!r}` in `.format()` is the equivalent to `%r` in  `%`.

bisguzar: The original line has 's' flag, why you used 'r'. I don't have enough knowledge about format flags

alanyee: `url` should always be a string. So far `url = 'https://example.com/`, `"No connection adapters were found for '{!s}'".format(url)` is the same as `"No connection adapters were found for {!r}".format(url)`, as both output `"No connection adapters were found for 'https://example.com/'`,  for the raw form of a string includes its quotes. `{!r}` or `repr()` tends to be faster than `{!s}` or `str()`.

kennethreitz:
interesting fact: % is much faster (at least in my tests, years ago) than `.format`. I don't think we need to be concerned about that here though ;)

Issue #5131, opened 2019-07-03
requests.utils function accepts invalid ip/cidr input


disconnect3d:
TLDR: On some libc implementations (e.g. glibc) the `socket.inet_aton` function parses IP strings trailed with whitespace and garbage.

An example below:
![image](https://user-images.githubusercontent.com/10009354/60627678-1f563a00-9def-11e9-88d3-3d37397164e5.png)

I have already filled an issue about this in `socket.inet_aton` but it is yet going to be discussed whether it will be fixed in Python.

## Expected Result

The `requests.utils` functions `address_in_network`, `is_ipv4_address` and `is_valid_cidr` should fail with invalid input.

## Actual Result

Invalid input is passed.

## Reproduction Steps

```python
import requests

print(requests.utils.address_in_network('1.1.1.1 wtf', '1.1.1.1/24'))
print(requests.utils.is_ipv4_address('1.1.1.1 disconnect3d was here...'))
print(requests.utils.is_valid_cidr('1.1.1.1 obviously not but yes/24'))
```

## System Information

The issue is related to libc implementation and has been tested on glibc 2.27 and 2.29.

It also occurs on MacOS which I am not sure if it is based on glibc.

HJones82493:
PR #5135 
I feel as though the call to the socket library is actually unnecessary here.

disconnect3d:
Ping @kennethreitz @Lukasa @sigmavirus24 @nateprewitt @slingamn can we finally get some love here and get this fixed?

This could probably also be marked as a security issue since if someone uses this in a security relevant context, they can get in trouble. (Although I doubt and hope that nobody does use those functions in such a way).

sigmavirus24:
These are not intended for use by others. The fact the language has no way of making things as private or internal means that we can't stop people from giving a function not intended for public use garbage or getting garbage out

sigmavirus24:
Also, @disconnect3d rather than demand free labor of others, you're welcome to submit a fix for this

disconnect3d:
> These are not intended for use by others. The fact the language has no way of making things as private or internal means that we can't stop people from giving a function not intended for public use garbage or getting garbage out

The typical way to mark this is to prefix the name with underscore or two ;).

sigmavirus24:
Yeah and people still use it from this library so we stopped bothering. I've been maintaining this library for almost a decade now. Writing python for longer. I know what the convention is and I know people don't care

HJones82493:
[...]

PR #5135, opened 2019-07-10
Workaround for ipv4 address validation

HJones82493:
This fixes the is_ipv4_address function. Initially intended to be a workaround for Issue #5131 it is now a removal of an external dependency. Since the check is a simple one, there is no need to rely on external sources.


requests/utils.py
```diff
@@ -616,7 +616,8 @@ def address_in_network(ip, net):
 
     :rtype: bool
     """
-    ipaddr = struct.unpack('=L', socket.inet_aton(ip))[0]
+    if not is_ipv4_address(ip):
```
disconnect3d, line 624: The code below uses `ipaddr` local variable which has been removed.

disconnect3d: I assume that we don't care about 3-1 octet IP addresses that the `socket.inet_aton` can parse e.g.:
```python
In [11]: socket.inet_aton('0x7f000001')
Out[11]: b'\x7f\x00\x00\x01'

In [12]: socket.inet_aton('127.0.0')
Out[12]: b'\x7f\x00\x00\x00'

In [13]: socket.inet_aton('127.0')
Out[13]: b'\x7f\x00\x00\x00'
```

If so, we can just use the builtin `ipaddress` module to validate if IP is in proper format:
```python
In [2]: ipaddress.IPv4Address('192.168.0.1')
Out[2]: IPv4Address('192.168.0.1')

In [3]: ipaddress.IPv4Address('192')
AddressValueError: Expected 4 octets in '192'
```

and then use the `ipaddr = struct.unpack('=L', socket.inet_aton(ip))[0]` line to parse it.

Note that the same problem stands for `net` argument - i.e. it also uses `sockeet.inet_aton` so we need to validate it's format first.

disconnect3d:
Just for a reference, the `socket.inet_aton` issue is being discussed on Python's bug tracker: https://bugs.python.org/issue37495

requests/utils.py
HJones82493: I didn't plan on handling 1-3 octet IPs unless we also consider those valid? I didn't think we would consider those as valid IPs.

We already have a check for valid ip that works (is_valid_ipv4) that I changed to remove the socket dependency in it. I think the best solution would be to leave the socket dependencies in the functions that use them and just put a check first, that way we don't allow it to pass IPs like '1.1.1.1 wtf' into the broken Python functions.

requests/utils.py
```diff
@@ -645,11 +645,12 @@ def is_ipv4_address(string_ip):
         return False
     for octet in ip_split:
         try:
-            int(octet)
+            tmp = int(octet)
```
disconnect3d, line 662: > We already have a check for valid ip that works (is_valid_ipv4)
Sure, but it's better to use standard methods that work properly instead of reinventing the wheel.

The current version of  `is_ipv4_address` still doesn't handle all the edge cases - for example, it returns `True` for an ip of `'08.08.08.08'`.

The `ipaddress` module has functions to check given IPs and networks. This could probably be used for both IP and CIDR checking.

pss904: The ipaddress module is also fallible, `ipaddress.IPv4Address('01.01.01.01')` returns successfully.

HJones82493: The ipaddress module doesn't exist for Python 2.7

requests/utils.py
```diff
@@ -644,9 +644,16 @@ def is_ipv4_address(string_ip):
...
-                return False
+            if octet_size > 1 and int(octet[0]) == 0:
+                if octet_size == 2 and int(octet[1]) > 7:
+                    return False
+                if octet_size == 3:
+                    if int(octet[1]) == 0 and int(octet[2]) > 7:
+                        return False
+                    if int(octet[1]) > 0:
+                        return False
             tmp = int(octet)
```
recepsirin, line 662: Here seems to me that it need to rewrite :/

requests/utils.py
```diff
@@ -638,10 +640,26 @@ def is_ipv4_address(string_ip):
...
+        try:
+            if octet_size > 1 and int(octet[0]) == 0:
+                if octet_size == 2 and int(octet[1]) > 7:
+                    return False
+                if octet_size == 3:
+                    if int(octet[1]) == 0 and int(octet[2]) > 7:
+                        return False
+                    if int(octet[1]) > 0:
+                        return False
+            tmp = int(octet)
```
recepsirin, line 662: Here seems to me that it need to rewrite

recepsirin: between 646 and 657

nateprewitt:
Now that we've dropped support for Python 2.7-3.6, I think we're more likely to rely on `ipaddress` than try to maintain this ourselves. I'm going to close this out in favor of that approach.

PR #5146, opened 2019-07-28
Fix typo in api.requests() docstring formatting

dharmab:
Fix a formatting typo which caused the documentation for `api.request()`'s `method` arg to be corrupted when rendered:

2019-07-28-090608_764x154_scrot


dharmab:
Looks like this is a dupe of #5145

PR #5147, opened 2019-07-28
Remove break-line

myslak71:
Remove unnecessary break-line - it just looks better now :)


nateprewitt:
Hi @myslak71, thanks for contributing to Requests! Unfortunately, we don’t typically accept PRs with stylistic/formatting changes. We’re working on getting linting set back up which should catch these in bulk going forward.

PR #5174, opened 2019-08-24
Fix build - use python 2.7 compatible super() call.

asherf:
Issue introduced here: https://github.com/psf/requests/pull/4983/files#diff-28e67177469c0d36b068d68d9f6043bfR425

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -419,7 +419,7 @@ class Session(SessionRedirectMixin):
         self.mount('https://', HTTPAdapter())
         self.mount('http://', HTTPAdapter())
 
-        super().__init__()
+        super(Session, self).__init__()
 
     def __enter__(self):
         return self


asherf:
@nateprewitt @kennethreitz friendly ping 🙏

epenet:
@asherf, I think the first parameter should be the base class
```python
    super(SessionRedirectMixin, self).__init__()
```
not the current class
```python
    #Invalid
    super(Session, self).__init__()
```

Refs #5175

asherf:
> @asherf, I think the first parameter should be the base class
> 
> ```python
>     super(SessionRedirectMixin, self).__init__()
> ```
> 
> not the current class
> 
> ```python
>     #Invalid
>     super(Session, self).__init__()
> ```
> 
> Refs #5175

@epenet I looked in several places in the code:
https://github.com/psf/requests/blob/master/requests/adapters.py#L59
https://github.com/psf/requests/blob/master/requests/adapters.py#L123
https://github.com/psf/requests/blob/master/requests/exceptions.py#L25
the pattern that you suggest doesn't seem the one that is being used....

also if you look at the [python docs](https://docs.python.org/2/library/functions.html#super), the example there uses the 'current' class:
```python
class C(B):
    def method(self, arg):
        super(C, self).method(arg)
```

epenet:
Thanks @asherf, there's a lot of contradictory information on super() out there, so I'm happy with your change.

It would be good for @nateprewitt or @kennethreitz to merge it as it's currently upsettingTravis CI on all other PR.

nateprewitt:
Hi @asherf, thanks for the PR. I’m not convinced we want the code that was merged in the referenced commit in the first place. I need to take a look through, see what testing would need to be added and what security implications it’s going to have.

I’ll leave this open until a decision is made but it’s likely we’ll revert the other commit.

nateprewitt:
I've pushed 09fd857 to resolve the issue. Resolving in favor of that.

Issue #5178, opened 2019-08-27
HTTPDigestAuth fails from secondary thread


epenet:
Hi,

I am looking into this [Home Assistant issue 13524](https://github.com/home-assistant/home-assistant/issues/13524) where a request is prepared once at startup (on an initialisation thread), and then updated at regular interval from a thread pool. This is working fine with HTTPBasicAuth, but failing with HTTPDigestAuth.

Is this a bug? or is this by design? If it is by design, can you help me find a workaround for Home Assistant?

## Expected Result
```
{
  "authenticated": true,
  "user": "user"
}
```

## Actual Result
```
Traceback (most recent call last):
  File ".../python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File ".../python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "test.py", line 8, in do_update
    request
  File ".../python3.7/site-packages/requests/sessions.py", line 653, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File ".../python3.7/site-packages/requests/hooks.py", line 31, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File ".../python3.7/site-packages/requests/auth.py", line 247, in handle_401
    if self._thread_local.pos is not None:
AttributeError: '_thread._local' object has no attribute 'pos'
```

## Reproduction Steps

```python
import requests
import threading

def do_update(name, request):
    print("do_update %s: starting", name)
    with requests.Session() as sess:
        response = sess.send(
            request
        )
    print(response.text)
    print("do_update %s: finishing", name)
    
method = 'GET'
resource = 'http://httpbin.org/digest-auth/auth/user/passwd'
auth = requests.auth.HTTPDigestAuth('user', 'passwd')
request = requests.Request(
    method, resource, auth=auth
).prepare()

#Works on main thread
do_update("main thread", request)

#Doesn't work on secondary thread
x = threading.Thread(target=do_update, args=("other thread",request,))
x.start()
x.join()

```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.4"
  },
  "platform": {
    "release": "4.19.57+",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
[...]

PR #5179, opened 2019-08-28
HTTPDigestAuth thread fix

epenet:
Ensure thread-state is initialised on handle_401
Refs #5178


epenet:
5 yrs open - I give up

PR #5200, opened 2019-09-19
Fixed a tiny typo

stuartmaxwell:
certianly -> certainly


nateprewitt:
Hi @stuartmaxwell, thanks for contributing to Requests. It looks like this is a duplicate of #5197 so I'm goin resolve this in favor of the original. Cheers!

PR #5210, opened 2019-09-23
Session instance `verify` precedence over REQUESTS_CA_BUNDLE

### LINKED ISSUE #5209: REQUESTS_CA_BUNDLE overrides request.Session verify property
When `REQUESTS_CA_BUNDLE` is defined, the CA certificate passed to `requests.Session` constructor is overridden when `requests.Session.request` is called.

Guilty method/lines:
https://github.com/psf/requests/blob/master/requests/sessions.py#L694-L710

## Expected Result

I would expect for the priority to be:

1. `requests.Session.request` verify kwarg
2. `requests.Session` verify property
3. `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE`

## Actual Result

As `requests.Session.request` verify kwarg was not set, it took the value from `REQUESTS_CA_BUNDLE` and only then `requests.Session.verify` was checked, but ignored, resulting in an invalid CA used (for that Session).

Related issue (in docker-py project) https://github.com/docker/docker-py/issues/2433

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": "2.7"
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.4"
  },
  "platform": {
    "release": "18.6.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "1010103f",
    "version": "19.0.0"
  },
  "requests": {
    "version": "2.22.0"
  },
  "system_ssl": {
    "version": "1000213f"
  },
  "urllib3": {
    "version": "1.25.3"
  },
  "using_pyopenssl": true
}
```

fopina:
To fix #5209 (more details in the issue)


Patazerty:
Any reason to not merge this simple fix ?

sigmavirus24:
@Patazerty 

1. Doesn't look like it would merge cleanly
1. Is documented as the behaviour that's unfortunate but not fixable until a major version bump due to a strict backwards compatibility policy

fopina:
@sigmavirus24 I think that if a PR from 10 months ago would merge cleanly it'd mean the project was quite inactive 😃 I can update it if its to be considered. Or is 1 related to something else?

About 2, I couldn't find it as documented behavior anywhere and it hardly looks like expected behavior for anyone. If verify is not defined for Session, it keeps the same behavior

Crocmagnon:
I'm also very surprised by this behavior which cost me a couple of hours of debugging until I noticed that when unsetting `REQUESTS_CA_BUNDLE` the issue goes away.

IMO env vars should never have priority over specific code instructions.

sigmavirus24:
> IMO env vars should never have priority over specific code instructions.

Right. It's been acknowledged as a bug, just not one that can be easily fixed in a 2.x release while maintaining the guarantees about minor version upgrades.

PR #5216, opened 2019-09-29
documentation link fixed

Leetcore:
websites is down, link to requests.readthedocs.io


ewdurbin:
Thank you @Leetcore, until we can sort out what happened with python-requests.org this looks good.

chmreid:
@ewdurbin also see #5236

Issue #5162, opened 2019-08-14
requests.ConnectionError inherits only from OSError


Q-back:
`requests.ConnectionError` doesn't inherit from ConnectionError from std, which is confusing.

## Expected Result

```
(Pdb++) inspect.getmro(ConnectionError)
(<class 'requests.exceptions.ConnectionError'>, <class 'requests.exceptions.RequestException'>, <class 'ConnectionError'>, <class 'OSError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>)
```

## Actual Result

```
(Pdb++) inspect.getmro(ConnectionError)
(<class 'requests.exceptions.ConnectionError'>, <class 'requests.exceptions.RequestException'>, <class 'OSError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>)
```

## Reproduction Steps

```python
import requests
inspect.getmro(requests.ConnectionError)
issubclass(requests.ConnectionError, ConnectionError)

```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.2"
  },
  "platform": {
    "release": "18.2.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.22.0"
  },
  "system_ssl": {
    "version": "1000212f"
  },
  "urllib3": {
    "version": "1.25.3"
  },
  "using_pyopenssl": false
}
```

nateprewitt:
The standard library ConnectionError serves a distinctly different purpose than Requests' ConnectionError. Subclassing the error binds us to its socket semantics which we don't want. It's also a breaking change to remap it.

PR #5219, opened 2019-10-01
Change requests.ConnectionError to inherit standard ConnectionError

Overv:
This PR changes the requests version of `ConnectionError` to also inherit from the built-in `ConnectionError`, see #5162

### DIFF
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -72,3 +72,5 @@ elif is_py3:
     basestring = (str, bytes)
     numeric_types = (int, float)
     integer_types = (int,)
+
+    ConnectionError = ConnectionError
--- a/requests/compat.py
+++ b/requests/compat.py
@@ -54,6 +54,8 @@ if is_py2:
     numeric_types = (int, long, float)
     integer_types = (int, long)
 
+    class ConnectionError(OSError): pass
+
 elif is_py3:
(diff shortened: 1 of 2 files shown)


YPCrumble:
This PR doesn't solve the issue that `requests.exceptions.ConnectionError` redefines the builtin `ConnectionError`. What about renaming this to `RequestsConnectionError` or another name altogether?

Also - does this actually produce the result you're expecting? You expected:

(<class 'requests.exceptions.ConnectionError'>, <class 'requests.exceptions.RequestException'>, <class 'ConnectionError'>, <class 'OSError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>)

☝️ but the PR doesn't inherit from `ConnectionError` - it inherits from `OSError`. I suspect you'd find this PR creates a slightly different inheritance like this:

(<class 'requests.exceptions.ConnectionError'>, <class 'requests.exceptions.RequestException'>, <class 'OSError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>)

Note how this only shows the redefined builtin, not the actual builtin per the issue.

nateprewitt:
I don't believe this is an approach we can accept for #5162. We can't overwrite the standard lib ConnectionError.

PR #5220, opened 2019-10-02
Fix connection pool managers not taking varying TLS parameters into account

### LINKED ISSUE #5140: Client Side Certs with session
Summary.

Trying to get client side certificates to work with session but I'm seeing a difference between the results when making the call with a session and the helper get method.

## Expected Result
Calling get with a Session returns the same result as calling `requests.get`.

## Actual Result
Request fails as if the certificate was not passed.

## Reproduction Steps

Screen Shot 2019-07-18 at 11 42 00 AM

## System Information

    $ python -m requests.help

```
(panopta_env) ➜  ~ python -m requests.help
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "2.7.16"
  },
  "platform": {
    "release": "18.6.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.22.0"
  },
  "system_ssl": {
    "version": "1000212f"
  },
  "urllib3": {
    "version": "1.25.3"
  },
  "using_pyopenssl": false
}
```

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).

Overv:
This PR fixes #5140 and #4325 by returning different connection and proxy pools based on the connection parameters for the current request.

I tested it with the following code:

```python
import requests

session = requests.Session()

print(session.get('https://client.badssl.com/'))
print(session.get('https://client.badssl.com/', cert=('client.crt', 'client.key')))
print(session.get('https://client.badssl.com/', cert=('client.crt', 'client.key')))
print(session.get('https://client.badssl.com/'))
print(session.get('https://client.badssl.com/'))
```

Which correctly outputs:

```
<Response [400]>
<Response [200]>
<Response [200]>
<Response [400]>
<Response [400]>
```

Instead of the previous output:

```
<Response [400]>
<Response [400]>
<Response [400]>
<Response [400]>
<Response [400]>
```

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -119,14 +119,13 @@ class HTTPAdapter(BaseAdapter):
             self.max_retries = Retry.from_int(max_retries)
         self.config = {}
         self.proxy_manager = {}
+        self.pool_manager = {}
 
         super(HTTPAdapter, self).__init__()
 
         self._pool_connections = pool_connections
         self._pool_maxsize = pool_maxsize
         self._pool_block = pool_block
-
-        self.init_poolmanager(pool_connections, pool_maxsize, block=pool_block)
 
     def __getstate__(self):
         return {attr: getattr(self, attr, None) for attr in self.__attrs__}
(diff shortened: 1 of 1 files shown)


LukasHeimann:
Would love to see this land, but there is another complexity to it: for `cert`, its not just the file _names_  that might change with every request, but also the file _contents_. Ideally, this would also check the files' `mtime`.

Given that this is open for a while, should I try to file a new issue for this?

PR #5221, opened 2019-10-02
Change raise_for_status message to hide password in URL

### LINKED ISSUE #5021: raise_for_status leaks HTTP basic auth
Calling `raise_for_status()` output (leaks) the HTTP basic auth password when raising an exception.

## Expected Result

```
>>> import requests
>>> r = requests.get("http://user:somepassw@httpbin.org/status/401")
>>> r.raise_for_status()
  File "/usr/lib/python3.7/site-packages/requests/models.py", line 940, in raise_for_status
requests.exceptions.HTTPError: 401 Client Error: UNAUTHORIZED for url: http://user:<redacted>@httpbin.org/status/401
```
## Actual Result

Here we can see the HTTP password in the resulting exception (`somepassw` in our case)

```
>>> import requests
>>> r = requests.get("http://user:somepassw@httpbin.org/status/401")
>>> r.raise_for_status()
  File "/usr/lib/python3.7/site-packages/requests/models.py", line 940, in raise_for_status
requests.exceptions.HTTPError: 401 Client Error: UNAUTHORIZED for url: http://user:somepassw@httpbin.org/status/401
```
## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.2"
  },
  "platform": {
    "release": "4.20.8-arch1-1-ARCH",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.21.0"
  },
  "system_ssl": {
    "version": "1010101f"
  },
  "urllib3": {
    "version": "1.24.1"
  },
  "using_pyopenssl": false
}
```

Overv:
This PR fixes #5021 by replacing a password in the URL with asterisks.

Example:

```python
import requests

r = requests.get("http://foo:secret@httpbin.org/status/401")
r.raise_for_status()
```

Output:

```
  File "5021.py", line 4, in <module>
    r.raise_for_status()
  File "***/python-requests/requests/requests/models.py", line 968, in raise_for_status
requests.exceptions.HTTPError: 401 Client Error: UNAUTHORIZED for url: http://foo:***@httpbin.org/status/401
```


Overv:
@jackyzha0 @barroca I'll have time to get back to this next weekend.

barroca:
@Overv I opened a similar PR for this that might help you to write a test #5244 let me know if you need help.

Overv:
I've added a unit test.

tests/test_requests.py
```diff
@@ -789,6 +789,15 @@ def test_status_raising(self, httpbin):
...
         assert not r.ok
 
+    def test_status_raising_hides_password(self, httpbin):
+        host = urlparse(httpbin()).netloc
+
+        r = requests.get('http://user:pass@' + host + '/status/404')
+        with pytest.raises(requests.exceptions.HTTPError) as e:
+            r.raise_for_status()
+
+        assert 'pass' not in str(e)
```
barroca: 💡 I'd rather assert e is equals with the string with pass obfuscated, and also add a test that there doesn't obfuscate things that wasn't suppose to, but I'm happy to approve it.

Overv: I've made the test more specific now. How would you implement a test that checks if other things are not obfuscated?

barroca: You can create a test like this:
```
    def test_status_raising_doesnt_hides_other_stuff(self, httpbin):
        host = urlparse(httpbin()).netloc

        r = requests.get('http://shouldntchange@' + host + '/status/404')
        with pytest.raises(requests.exceptions.HTTPError) as e:
            r.raise_for_status()

        host_without_port = host.split(':')[0]

        assert str(e.value) == '404 Client Error: NOT FOUND for url: http://shouldntchange@' + host_without_port + '/status/404'
```

This is just a validation that the match doesn't put wrong information if the password is not supplied

Overv: Adding that test actually made me find two bugs!

* The host in the exception message no longer included the port, which shouldn't happen.
* Asterisks were included even if there was no password specified.

Thanks for the suggestion.

mpolena:
Running into this same issue for requests with `client_id` and `client_secret` parameters in the url.   Those values would get exposed despite the proposed fix here.  Seems like a more general solution is needed.

barroca (approved):

ZenithClown:
Running into this same issue for requests with `access_key` and `apikey` which exposes the API key.

PR #5225, opened 2019-10-10
Give the body of the response when raising an exception for a 400 error.

RubenGarcia:
This makes debugging the reason for the error much easier, e.g. in keycloak / OpenID.

Before the patch:
400 Client Error: Bad Request for url: <url>

After the patch:
400 Client Error: Bad Request for url: <url>, response body is {"error":"unauthorized_client","error_description":"Invalid client secret"}


nateprewitt:
Hi @RubenGarcia,

We've reviewed this proposal in the past but there are a number of cases, both unintentional and intentional, where an unreasonable amount of data is sent back through a connection. In that case you may not be able to do anything useful with this information, or more severe negative implications may happen with your program. For that reason, I don't think we'll accept this at this time.

RubenGarcia:
Would something more intelligent, like sending the first 128 bytes of the body, be acceptable? It would be enough for APIs, and common cases, without being overly verbose in misconfigured servers.

sigmavirus24:
No.

Issue #3341, opened 2016-06-17
Setting session.timeout doesn't do anything


ikus060:
I was using an old version of requests (I don't know the exact version). When settings the `session.timeout` it was affecting the timeout value during a call to `session.get()`. With recent version of requests, this feature doesn't seams to work.

I'm currently using version 2.10.0

Basically, I'm expecting this:
```
session = requests.session()
session.timeout = 0.0001
session.get('http://localhost/')
```
To have the same behavior as:
```
requests.get('http://localhost/', timeout=0.0001)
```

Lukasa:
This behaviour was deliberately changed: see #2856, and #2011 which have discussion on them.

sigmavirus24:
And to be clear, that version must have been prior to 1.0 when this behaviour was removed.

ikus060:
After reading both issues, I still don't understand why it's not supported. What is the rational behind this decision ?

ikus060:
Also, whats it the alternative to avoid setting the timeout value for each and every get() request ?

Lukasa:
@ikus The rationale is that timeouts are not of-a-kind with the things that are set on `Sessions`. They represent a property of the transport layer and don't semantically apply to a _session_: how does one "time out" a session? In general they are a per-request property, and defaulting them is a property of the transport layer, which is where [Transport Adapters](http://docs.python-requests.org/en/master/user/advanced/#transport-adapters) are used to store configuration.

[This comment](https://github.com/kennethreitz/requests/issues/2011#issuecomment-64440818) provides an option for defaulting the value.

ikus060:
I will not argue to much about this decision. But I think it goes against the "Python HTTP Requests for Humans". In order to make sure every GET requests does have a good timeout: (a) I need to review all the code to add `timeout` as an argument to each call of `get()` or (b) I need to monkey patch the default `HTTPAdapter` and provide a default timeout of my own. 

Defining the timeout value at the session level would be more human even if the property is not strictly related to the session it self.

Lukasa:
@ikus060 You do not need to monkeypatch the default HTTPAdapter: just install the correct adapters in at the point where you construct the session.

ikus060:
If we are talking of this code:

``` python
class MyHTTPAdapter(requests.adapters.HTTPAdapter):
    def __init__(self, timeout=None, *args, **kwargs):
        self.timeout = timeout
        super(MyHTTPAdapter, self).__init__(*args, **kwargs)

[...]

PR #5230, opened 2019-10-16
#3070 Defaults on session level

FilipMalczak:
See #3070 and [this](https://stackoverflow.com/questions/41295142/is-there-a-way-to-globally-override-requests-timeout-setting). Also, compare #3341 

TODO:
- [ ] docs for the session instance attributes
- [ ] tests for this attributes' usage
- [ ] decision on default value for timeout

----------------------------

At first commits I've added session-level `timeout=None` and `allow_redirects=True` attributes that are now used in `x or self.x` in `request(...)` method. I've also removed `allow_redirects` defaults in `get(...)` and friends.

Default values are a subject for discussion, but let's start at least drafting this.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -528,7 +528,7 @@ class Session(SessionRedirectMixin):
         # Send the request.
         send_kwargs = {
             'timeout': timeout or self.timeout,
-            'allow_redirects': allow_redirects or self.allow_redirects,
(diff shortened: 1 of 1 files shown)


FilipMalczak:
Missing:
- [ ] docs for the session instance attributes
- [ ] tests for this attributes' usage

nateprewitt:
I don't believe we'll be able to accept this. `timeout` was intentionally omitted from the Session class in Requests 2.x and we don't intend to add it within this major version.

PR #5232, opened 2019-10-17
Add timeout as a default attribute for Session objects

ofek:
We encountered a bug today wherein we mistakenly assumed it was already supported as an attribute like all the other kwargs


ofek:
@Lukasa @kennethreitz @nateprewitt Hello there! After opening this I came across https://github.com/psf/requests/issues/3341

Does that decision (from way back then) still remain true today? As it is, the behavior is not very user-friendly and is entirely unexpected, hence the reason so many people keep opening the same PR 😄

PR #5233, opened 2019-10-17
Fix broken URLS in readme

chmreid:
[This commit](https://github.com/psf/requests/commit/74b72ce4265583db0430080ab67d3d5e0c4b44b2) removed several references to the URL 2.python-requests.org but left several behind. This PR cleans those up and replaces with requests.readthedocs.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 <span align="center">
 
 <pre>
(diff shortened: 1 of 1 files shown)


chmreid:
@ssbarnea looks like you were much more thorough in replacing the old URLs...

Admins - approve #5236 instead!

PR #5239, opened 2019-10-21
Catch AttributeError in utils.super_len

### LINKED ISSUE #5229: requests.utils.super_len() can't cope with tarfile members
## Expected Result

POSTing a large file that I have extracted from a tarfile.  I expect that `requests` should be able to determine the length of that file, set the `Content-Length` header, and upload it efficiently.

## Actual Result

`requests` fails to determine the file length, goes with `Transfer-Encoding: chunked`, and the upload is much slower than desired.

## Reproduction Steps

Start with a small tarfile, say

```bash
echo foo > file1
tar cf tarfile.tar file1
```

Then use `requests.utils.super_len` to try and get the length of `file1` in the tarfile:

```python
#!/usr/bin/env python
import tarfile
import requests

tar = tarfile.open("tarfile.tar")
member = tar.extractfile("file1")
print(f"super_length of member is {requests.utils.super_len(member)}")
```

result:

```
  File "./demo.py", line 7, in <module>
    print(f"super_length of member is {requests.utils.super_len(member)}")
  File "/data/developer/.virtualenvs/req-tar/lib/python3.6/site-packages/requests/utils.py", line 119, in super_len
    fileno = o.fileno()
AttributeError: '_FileInFile' object has no attribute 'fileno'
```

The behavior of the member is quite strange here: `hasattr(member, fileno)` is true, but trying to call `member.fileno()` results in an `AttributeError`.

I expect that a sensible fix might be for [this exception](https://github.com/psf/requests/blob/fab1fd10d0b115e635b9ef1364f8444089725000/requests/utils.py#L121) to catch a broader range of errors.  

One can get the length of the member by seeking, so I expect that `super_len()` should then succeed.

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.6.8"
  },
  "platform": {
    "release": "3.10.0-1062.1.1.el7.x86_64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.22.0"
  },
  "system_ssl": {
    "version": "100020bf"
  },
  "urllib3": {
    "version": "1.25.6"
  },
  "using_pyopenssl": false
}
```

dimbleby:
This allows it to handle files obtained via `Tarfile.extractfile()`.

Fixes #5229.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -11,6 +11,9 @@ dev
   backwards compatible as it inherits from previously thrown exceptions.
   Can be caught from `requests.exceptions.RequestException` as well.
 
+- Catch `AttributeError` when calculating length of files obtained by
+  `Tarfile.extractfile()`
+
 2.26.0 (2021-07-13)
 -------------------
 
@@ -1702,7 +1705,7 @@ This is not a backwards compatible change.
 -   Automatic Authentication API Change
 -   Smarter Query URL Parameterization
 -   Allow file uploads and POST data together
--   
+-
 
     New Authentication Manager System
 
@@ -1721,7 +1724,7 @@ This is not a backwards compatible change.
 0.2.3 (2011-02-15)
 ------------------
 
--   
+-
 
     New HTTPHandling Methods
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -124,7 +124,10 @@ def super_len(o):
     elif hasattr(o, 'fileno'):
         try:
             fileno = o.fileno()
-        except io.UnsupportedOperation:
+        except (io.UnsupportedOperation, AttributeError):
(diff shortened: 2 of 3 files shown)


sethmlarson (approved):
The change here looks good to me, can you add a note in `HISTORY.md` about this fix?

dimbleby:
Wow, I had forgotten all about this.  Still a good fix though, so I have rebased and added an entry to HISTORY.md

nateprewitt (approved):
Thanks @dimbleby!

PR #5240, opened 2019-10-21
fix: typo

chrisguox:


nateprewitt:
Hi @chrisguox, thanks for contributing to Requests! I think the current usages probably make more sense than the new proposals, so we'll pass on this one.

PR #5244, opened 2019-10-23
Add option to hide url from raise_for_status

barroca:
This PR adds an option on raise_for_status function that hides the url from the log message so it is possible to protect the sensitive data from the url without having to do complex parsing of it.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -935,8 +935,11 @@ class Response(object):
 
         return l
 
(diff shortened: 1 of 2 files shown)


sethmlarson:
This seems application specific. If the error message is being displayed to the user the URL can be masked in that section of application code or alternatively you can re-raise the exception with an edited message.

Issue #5245, opened 2019-10-23
Incorrect json() docstring


alekna:
What `json()` actually does - it deserializes JSON into a Python object.

So the docstring below must be wrong:
https://github.com/psf/requests/blob/fab1fd10d0b115e635b9ef1364f8444089725000/requests/models.py#L894-L918

albendz:
Do you mind if I take this task as an intro issue to the project? I'm looking to be a new contributor

lousyd:
This has tripped me up as well. I'm new to Python. I expected response.json() to return JSON, but it doesn't. I don't know if I'm expected to do something further to turn it into actual JSON (and if so, what), or if this is an oversight in the requests code.

toothstone:
I also ran into issues with this - besides the fact that it indeed returns a decoded Python representation of the content, I'm confused by the "if any" clause - there is nothing that checks the existence of content, nor if the response actually has Content-Type application/json set, it simply dumps `self.text` into a json loader.

So a better way to phrase this might be along the lines of:
"Tries to return a decoded representation of the response's content as if it were JSON"

kougazhang:
Plus one.
Why not fix it after passed so many years?

nateprewitt:
This was fixed in #6529.

PR #5246, opened 2019-10-24
Update documentation for response.json()

ellie:
The current documentation says that the function is encoding - actually, it is decoding!

I'm also wondering if it might be worth putting in some type annotations at some point

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -892,7 +892,7 @@ class Response(object):
         return content
 
     def json(self, **kwargs):
-        r"""Returns the json-encoded content of a response, if any.
+        r"""Decodes and returns the JSON response body, if any.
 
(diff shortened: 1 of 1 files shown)


alekna:
It doesn't return JSON body. It returns Python object.

Fixes #5245

toothstone:
As long as this is open - I would like to add that the "if any" clause is very unclear, from the code I can't tell what is meant to say.

From what I can see the function tries to return the decoded representation *as if* the content were JSON - there are no checks if it actually is, is declared by the response to be, or even exists at all.

jimmysisonlucas (commented):
Thanks for informing me

sethmlarson:
I think we'll keep the current docstring, thanks!

PR #5250, opened 2019-10-30
Fix link to requests-threads in community/recommended doc

MasterOdin:
The link for `Requests-Threads` was missing a trailing `_` causing it to be italicized instead of being a proper link.


xmo-odoo:
Might be worth adding doc building to the CI and either disabling `default_role` (such that using it is an error) or switching it out for `any` (though I don't know that that would find explicit link references).

Issue #5027, opened 2019-03-20
In [21]: import requests /usr/lib/python3/dist-packages/cryptography/x509/base.py:16: RuntimeWarning: coroutine 'wait' was never awaited   from cryptography.x509.extensions import Extension, ExtensionType RuntimeWarning: Enable tracemalloc to get the object allocation traceback


Ro0tk1t:
Summary.

## Expected Result

What you expected.

## Actual Result

What happened instead.

## Reproduction Steps

```python
import requests

```

```
/usr/lib/python3/dist-packages/cryptography/x509/base.py:16: RuntimeWarning: coroutine 'wait' was never awaited
  from cryptography.x509.extensions import Extension, ExtensionType
RuntimeWarning: Enable tracemalloc to get the object allocation traceback

```

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).

timofurrer:
Can you please fill out the template: describe what your expected and actual result was and how to reproduce.
And please paste the traceback (if valuable) in to the post body and no the title. Please use a descriptive title instead.

Thank you 🎉

nateprewitt:
There isn't enough information here to evaluate what is happening. Given this is an issue in cryptography and not Requests directly, I'm going to resolve. I'd recommend checking the issue tracker with cryptography if you're still able to reproduce the issue.

PR #5251, opened 2019-11-01
pytest 5

jayvdb:


jayvdb:
This may not be all of the changes needed to work on pytest 5, but it was enough for the openSUSE build to pass https://build.opensuse.org/request/show/744656 - it is using `-k "not (TestTimeout or connect or test_https_warnings)"`, so is skipping 12 and deselecting 16.

nateprewitt:
This should be resolved in #5027

PR #5252, opened 2019-11-04
RFC8288 support for link relations

dirkjonker:
Comply to RFC8288 for relation type (section 3.3)
https://tools.ietf.org/html/rfc8288#section-3.3

Changes:
- Only parse first "rel" parameter
- Allow multiple link relation types, i.e.
  rel="start http://example.net/relation/other"


sethmlarson:

PR #5259, opened 2019-11-12
docs - clarify json() error exception for Python 3

wesinator:
python 3 exception is `JSONDecodeError:  Expecting value: line 1 column 1 (char 0)`

https://github.com/psf/requests/issues/4908

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -152,7 +152,7 @@ There's also a builtin JSON decoder, in case you're dealing with JSON data::
 
 In case the JSON decoding fails, ``r.json()`` raises an exception. For example, if
 the response gets a 204 (No Content), or if the response contains invalid JSON,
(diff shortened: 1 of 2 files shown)


kennethreitz:
This should absolutely not be merged.

wesinator:
> This should absolutely not be merged.

why ?

nateprewitt:
Hi @wesinator, I think we’ll pass on this one. If you take a look at Python 3, you’ll find that JSONDecodeError to be a ValueError. Including the simplejson stack trace doesn’t provide information specific to Requests, just general stacktrace handling.

wesinator:
> 
> 
> Hi @wesinator, I think we’ll pass on this one. If you take a look at Python 3, you’ll find that JSONDecodeError to be a ValueError. Including the simplejson stack trace doesn’t provide information specific to Requests, just general stacktrace handling.

ok, fair. I know it's general stacktrace

thanks,

PR #5261, opened 2019-11-13
Doctest Updated

SKAUL05:
Added Response 404 Doctest


nateprewitt:
Hi @SKAUL05, thanks for this PR. Unfortunately, I don’t think we’ll be able to merge this. The doctstring is meant to be a simple example, so we won’t enumerate the potential status codes here.

Issue #5279, opened 2019-11-28
Test case, test_total_timeout_connect failed


templetontsai:
Summary.
When getting the latest release, v2.20, and run the test case it failed in the case test_total_timeout_connect. The test case is expecting, ConnectTimeout exception but instead it gets ConnectionError. The error message is as follows:

```
requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```
## Expected Result
I would expect all the test cases should be passed without any modification.

## Actual Result
4 test case fail.
Error message is consistent with the following:

```
requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```
## Reproduction Steps
It seems a symptom only happens when using WiFi. You need to make sure to use WiFi not landline.

Run the following test case:
pytest tests/test_requests.py::TestTimeout::test_total_timeout_connect

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": ""
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.5"
  },
  "platform": {
    "release": "4.15.0-70-generic",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.22.0"
  },
  "system_ssl": {
    "version": "1010100f"
  },
  "urllib3": {
    "version": "1.21.1"
  },
  "using_pyopenssl": false
}

```

templetontsai:
Upon investigation, it is a mismatch of exception is thrown and the expected exception of test case.

Exception thrown:  requests.exceptions.ConnectionError
Expected exception in the test case: ConnectTimeout

```
    def test_total_timeout_connect(self, timeout):
        try:
            requests.get(TARPIT, timeout=timeout)
            pytest.fail('The connect() request should time out.')
        except ConnectTimeout:
            pass
```

Furthermore, ConnectionTimeout is child class of ConnectionError.

```
class ConnectTimeout(ConnectionError, Timeout): 
```

This is the root cause. A fix can be throwing ConnectTimeout specifically as follow:

```
        except (ProtocolError, socket.error) as err:
            raise ConnectTimeout(err, request=request)
```

nateprewitt:
[...]

PR #5280, opened 2019-11-28
Fix test case, test_total_timeout_connect_failed #5279

templetontsai:
This pr is to fix the test case fail regarding ConnecTimeout, #5279.


nateprewitt:
Hi @templetontsai,

As I said in #5279, the exception here is what we're expecting. I'd validate your local network setup to resolve the issue you're seeing.

PR #5281, opened 2019-12-02
Unified check version compatibility

Mellcap:
I have noticed that there is a `# noqa: F811` comment at `__init__.py` when we check the version compatibility of urllib3.

And there exist duplicate code to check the compatibility of different packages.

So I write a method to handle it.


nateprewitt:
Hi @Mellcap,

Thanks for contributing to Requests! This is an internal implementation detail that we're using to validate a limited number of dependencies. While I think this is a reasonable abstraction, we don't intend to increase the number of dependencies we use. I'm a bit worried this may adds too many constraints on us modifying things going forward. I think we'll pass on this for now.

PR #5285, opened 2019-12-06
Using 'quote' instead 'quote_plus' 

### LINKED ISSUE #5170: How to parse blank to %20, not + ?
If http param contains blank, how to parse blank to %20, not + ?
```
import requests

param = {
    "sql": "project = 'foo' and name = 'bar'"
}

res = requests.get(
    url='http://xxxx',
    params=param
)

print(res.url)
# actual    http://xxxx/?sql=project+%3D+%27foo%27+and+name+%3D+%27bar%27
# expect    http://xxxx/?sql=project%20%3D%20%27foo%27%20and%20name%20%3D%20%27bar%27

```

shivareddyiirs:
Modified call to urlencode to change its default behavior. 
In place of using 'quote_plus' it uses 'quote' now.
fixes #5170

edit: for python 2 no change


shivareddyiirs:
Works fine in python 3 but not in python 2.

shivareddyiirs:
For python 2 its behavior is intact.

shivareddyiirs:
@nateprewitt ,
Hi can you review it , It has very minimal changes.

requests/models.py
```diff
@@ -102,7 +102,10 @@ def _encode_params(data):
                         result.append(
                             (k.encode('utf-8') if isinstance(k, str) else k,
                              v.encode('utf-8') if isinstance(v, str) else v))
-            return urlencode(result, doseq=True)
+            if is_py2:
+                return urlencode(result, doseq=True)
+            else:
+                return urlencode(result, doseq=True,quote_via=quote)
```
BoboTiG: There is a missing space before the comma.

shivareddyiirs: Thanks for comments corrected

shivareddyiirs:
> Thanks for the patch :)
> Do you think you could you add a simple test for that part of the code?

I have never written test earlier.
But surely can try to learn. Can you help me?

BoboTiG:
Perhaps could you try something like that (not tested, you will need to adapt):
```python
def test_params_quote_via():
    # Verifiy parameters are parsed using quote() (spaces are seen as "%20", not "+", for instance)
    params = {'sql': "project = 'foo' and name = 'bar'"}
    p = PreparedRequest()
    p.prepare(
        method='GET',
        url='https://www.example.com/',
        params=params
    )
    expected = 'https://www.example.com/?sql=project%20%3D%20%27foo%27%20and%20name%20%3D%20%27bar%27'
    assert p.url == expected
```
You could put it after that test: https://github.com/psf/requests/blob/7b565d886c852609a849e79e4ad8f3f8fa8d8c23/tests/test_requests.py#L2255

shivareddyiirs:
Thank you very much.
I would commit the test by   tomorrow evening (IST)
Thanks & Regards
Shiva Reddy K.
Scientist/Engineer 'SD'
Indian Institute of Remote Sensing,
Indian Space Research Organisation
Department of Space
4-Kalidas Road
Dehradun
mobile: 0135-2524126

tests/test_requests.py
```diff
@@ -2264,6 +2264,25 @@ def test_json_encodes_as_bytes():
...
+def test_params_encodes_quote():
+    # in python 3 params are encoded using quote in place of quote_plus (Spaces are encoded as %20 in place of '+')
+    params = {"key": "value1 value2"}
+    p = PreparedRequest()
+    p.prepare(
+        method='GET',
+        url='https://www.example.com/',
+        params=params
+    )
+    if is_py2:
```
BoboTiG:
```python
expected = 'https://www.example.com/?key=value1%20value2'
if is_py2:
     expected = 'https://www.example.com/?key=value1+value2'
assert p.url == expected
```
Please check you code locally first.

BoboTiG: And could you use the same example as shown on the original issue?

shivareddyiirs: I have modified and used same text as shown in the issue.

shivareddyiirs: Today , due to some work related constraint, I am not able to check it locally.

tests/test_requests.py
```diff
@@ -2264,6 +2264,25 @@ def test_json_encodes_as_bytes():
...
+    params = {"sql": "project = 'foo' and name = 'bar'"}
+    p = PreparedRequest()
+    p.prepare(
+        method='GET',
+        url='https://www.example.com/',
+        params=params
+    )
+    if is_py2:
+        #checking that there is no change in case of python 2
+        expected='https://www.example.com/?sql=project+%3D+%27foo%27+and+name+%3D+%27bar%27'
```
BoboTiG:
```suggestion
        expected = 'https://www.example.com/?sql=project+%3D+%27foo%27+and+name+%3D+%27bar%27'
```

tests/test_requests.py
```diff
@@ -2264,6 +2264,25 @@ def test_json_encodes_as_bytes():
...
+    # in python 3 params are encoded using quote in place of quote_plus (Spaces are encoded as %20 in place of '+')
+    params = {"sql": "project = 'foo' and name = 'bar'"}
+    p = PreparedRequest()
+    p.prepare(
+        method='GET',
+        url='https://www.example.com/',
+        params=params
+    )
+    if is_py2:
+        #checking that there is no change in case of python 2
```
BoboTiG:
```suggestion
        # checking that there is no change in case of python 2
```

tests/test_requests.py
```diff
@@ -2264,6 +2264,25 @@ def test_json_encodes_as_bytes():
...
+        expected='https://www.example.com/?sql=project+%3D+%27foo%27+and+name+%3D+%27bar%27'
+        assert p.url == expected
+    else:
+        #checking that there is change in behaviour in case of python 3
```
BoboTiG:
```suggestion
        # checking that there is change in behaviour in case of python 3
```

tests/test_requests.py
```diff
@@ -2264,6 +2264,25 @@ def test_json_encodes_as_bytes():
...
+        expected='https://www.example.com/?sql=project%20%3D%20%27foo%27%20and%20name%20%3D%20%27bar%27'
```
BoboTiG:
```suggestion
        expected = 'https://www.example.com/?sql=project%20%3D%20%27foo%27%20and%20name%20%3D%20%27bar%27'
```

shivareddyiirs:
Yes it seems confusing at first instance.
In  python 2  requests was not  behaving as per normal expectation of user.
But in python 3 it is possible , with least efforts. (due to change in python 3 API).

nateprewitt:
Hi @shivareddyiirs,

Thanks for the work you've put in here. If we were to change this behaviour in Requests it would need to be uniform across all versions of Python, otherwise users will inevitably stumble on it. This would also be a breaking change which means it couldn't be merged into master until we were ready for a Requests 3.0 release.

That said, the behaviour you're seeing with params is intended and documented in the [`Request` docstring](https://github.com/psf/requests/blob/master/requests/models.py#L211-L213). The params passed as dictionaries or lists of tuples will be [form-encoded](https://en.wikipedia.org/wiki/Percent-encoding#The_application/x-www-form-urlencoded_type) which uses + instead of %20. If you want to fiddle with the query string, you're still able to pass a string into the `params` argument.

I don't think we have intention to change this presently since it's easy to work around and has been operating under the same semantics since Requests' inception. Moving away from the idea of form-encoding by default would be a larger discussion for the project.

shivareddyiirs:
> 
> 
> Hi @shivareddyiirs,
> 
> Thanks for the work you've put in here. If we were to change this behaviour in Requests it would need to be uniform across all versions of Python, otherwise users will inevitably stumble on it. This would also be a breaking change which means it couldn't be merged into master until we were ready for a Requests 3.0 release.
> 
> That said, the behaviour you're seeing with params is intended and documented in the [`Request` docstring](https://github.com/psf/requests/blob/master/requests/models.py#L211-L213). The params passed as dictionaries or lists of tuples will be [form-encoded](https://en.wikipedia.org/wiki/Percent-encoding#The_application/x-www-form-urlencoded_type) which uses + instead of %20. If you want to fiddle with the query string, you're still able to pass a string into the `params` argument.
> 
> I don't think we have intention to change this presently since it's easy to work around and has been operating under the same semantics since Requests' inception. Moving away from the idea of form-encoding by default would be a larger discussion for the project.

Thanks,  I have made further changes , now it can work without braking the API.

nateprewitt:
Hey @shivareddyiirs,

It looks like the latest revision doesn't work for Python 2 and still has varied behaviour in the tests. I think we're going to maintain the existing behaviour and maintain the current advice to pass a string if you need non-form-encoded parameters. Thanks again for your work!

shivareddyiirs:
Sure.

PR #5286, opened 2019-12-06
If they do't have same length, there is no need to create two dict to compare.

sunsongpy:


nateprewitt:
Hi @sunsongpy, checks like [this](https://github.com/python/cpython/blob/109fc2792a490ee5cd8a423e17d415fbdedec5c8/Objects/dictobject.c#L2751) are baked into the base objects, so there’s no need for us to reimplement it here. Thanks for the suggestion!

nateprewitt:
Apologies, I glanced over the PR too quickly. If you can provide benchmarks for this change showing a significant improvement, I think we can look at it. Otherwise, I think we’ll pass for now.

sunsongpy:
> Apologies, I glanced over the PR too quickly. If you can provide benchmarks for this change showing a significant improvement, I think we can look at it. Otherwise, I think we’ll pass for now.

Sorry, I know this change won't bring much of a significant improvement, but  it certainly makes the code more rigorous. If the length is different, it's definitely False (two dict comparison of Cpython source code is the first step is to compare the length, why we don't do like that first).

sunsongpy:
> Yes this can improve performance .

thanks

shivareddyiirs:
> 
> 
> > Apologies, I glanced over the PR too quickly. If you can provide benchmarks for this change showing a significant improvement, I think we can look at it. Otherwise, I think we’ll pass for now.
> 
> Sorry, I know this change won't bring much of a significant improvement, but it certainly makes the code more rigorous. If the length is different, it's definitely False (two dict comparison of Cpython source code is the first step is to compare the length, why we don't do like that first).

After seeing the @nateprewit link . I agree with him. Adding extra lines of code just to do what is already being done, is not justified.

PR #5293, opened 2019-12-18
fix raise_for_status docstring

graingert:
the exception isn't stored it's created new each call


nateprewitt:
Hey @graingert, would you mind rebasing onto master. Then I think we should be set.

PR #5302, opened 2019-12-31
Change <Response [XXX]> to self.text

JonPizza:
From talking to people and personal experiences, I always feel this is more intuitive. Only downside is inside a REPL, but I think there is a way to detect that?

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -668,7 +668,7 @@ class Response(object):
         setattr(self, 'raw', None)
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


sethmlarson:
This would spam up a console if it were a long response body and also be a breaking change. It's unlikely that this will be merged.

nateprewitt:
Hi @JonPizza,

Seth is right here. This would add a minor short cut to `text` but risks printing very long bodies into the console unexpectedly.

I think we’ll pass on this for now, thanks for the suggestion!

Issue #5304, opened 2020-01-02
pytest.raises with a string as the second argument removed in pytest 5


hroncok:
See http://doc.pytest.org/en/latest/deprecations.html#raises-warns-with-a-string-as-the-second-argument

The non-comntextmanager form of `pytest.raises` was removed in pytest 5. It is used in:

https://github.com/psf/requests/blob/eedd67462819f8dbf8c1c32e77f9070606605231/tests/test_requests.py#L774-L778

The error is:

```
self = <tests.test_requests.TestRequests object at 0x7f68030882b0>
httpbin = <function prepare_url.<locals>.inner at 0x7f68034c8f70>

    def test_conflicting_post_params(self, httpbin):
        url = httpbin('post')
        with open('Pipfile') as f:
>           pytest.raises(ValueError, "requests.post(url, data='[{\"some\": \"data\"}]', files={'some': f})")
E           TypeError: 'requests.post(url, data=\'[{"some": "data"}]\', files={\'some\': f})' object (type: <class 'str'>) must be callable

tests/test_requests.py:777: TypeError
```

The code was added in https://github.com/psf/requests/pull/1503 [to support Python < 2.7](https://github.com/psf/requests/pull/1503#issuecomment-22333666):

> You can't use the context manager form of assertRaises() on Pythons earlier than 2.7. You'll need to use one of these two forms...

Since 2.6 is not supported anymore, this could be converted to:

```python
        with open('Pipfile') as f:
            with pytest.raises(ValueError):
                requests.post(url, data='[{"some": "data"}]', files={'some': f})
            with pytest.raises(ValueError):
                requests.post(url, data=u('[{"some": "data"}]'), files={'some': f})
```

## Expected Result

Test pass with pytest 5.

## Actual Result

Test `TestRequests.test_conflicting_post_params` fails with pytest 5.

## Reproduction Steps

```console
$ rm .tox -rf
$ tox -e py38
```

(I've actually needed to install pygments into the environment, otherwise test collection has failed.)

## System Information

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.8.1"
  },
  "platform": {
    "release": "5.3.16-300.fc31.x86_64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
[...]

PR #5305, opened 2020-01-02
Fix test_conflicting_post_params to work on pytest 5

hroncok:
The non-contextmanager form of pytest.raises was removed in pytest 5.
http://doc.pytest.org/en/latest/deprecations.html#raises-warns-with-a-string-as-the-second-argument

It was used here to support Python < 2.7, but that is no longer needed.
https://github.com/psf/requests/pull/1503#issuecomment-22333666

Fixes https://github.com/psf/requests/issues/5304


hroncok:
What can I do to help getting this merged? I consider the change fairly trivial, self contained and easy to grasp. But the more this remains open, the chances of diverging are getting higher.

If there is anything else required from my side, please do let me know.

nateprewitt:
Thanks @hroncok!

hroncok:
Thanks for merging @nateprewitt.

PR #5309, opened 2020-01-08
Update models.py

alanyee:
Remove needless list()


requests/models.py
```diff
@@ -235,7 +235,7 @@ def __init__(self,
         hooks = {} if hooks is None else hooks
 
         self.hooks = default_hooks()
-        for (k, v) in list(hooks.items()):
+        for (k, v) in hooks.items():
```
sukurcf: I don't think we need parenthesis around **(**k, v**)** either.

nateprewitt:
Thanks for the PR, @alanyee! We typically don't take stylistic update PRs as it sets a precedent producing a lot of additional maintenance churn. We do appreciate it though.

PR #5323, opened 2020-01-23
Add response context to errors raised by iter_content

millerdev:
Errors raised by `Response.iter_content` did not have request/response context.

I couldn't figure out how to simulate `StreamConsumedError` or `ConnectionError` as raised from that context, and I could not find any other tests that simulate those errors either. However, the additions are so trivial and other tests _were_ updated to test the very same thing, so I'm confident in the changes.

Except for `StreamConsumedError` ~this introduces a circular reference between the `Response` and error objects because of `<Response>._error`. One way I thought of solving that issue is to break the cycle with a weakref, but that seems annoying since it would likely involve changing the implementation of `RequestException` in a way that may make it lose references.~ Another option would be to pass `request=self.request` instead of `response=self`. I added https://github.com/psf/requests/pull/5323/commits/8bb881fe425da8e2b5043da73462e6128d728506 to do that instead since that seems like a reasonable idea.


sethmlarson:

PR #5334, opened 2020-02-05
Refactored check_compatibility function

Hayashi-Yudai:
In `check_compatibility` function in `requests/__init__.py` some codes are redundant. This function distribute three elements of `urllib3_version` and `chardet_version` and after that transform them to integer values. It can be written with `map` function.


nateprewitt:
Hi @Hayashi-Yudai, thanks for the PR. We typically don't accept minor cosmetic changes as it sets a precedent increasing general maintenance churn for the library. We appreciate it though.

PR #5336, opened 2020-02-13
Remove explicit unicode literal in README.md and doc

aless10:
As the end of python 2 has come, I open this PR in order to get rid of the explicit unicode literal in the README file and the documentation.
I know that this could be against the PEP 414 https://www.python.org/dev/peps/pep-0414/ but it should be harmless for the project.


docs/user/quickstart.rst
```diff
@@ -148,7 +148,7 @@ There's also a builtin JSON decoder, in case you're dealing with JSON data::
 
     >>> r = requests.get('https://api.github.com/events')
     >>> r.json()
-    [{u'repository': {u'open_issues': 0, u'url': 'https://github.com/...
+    [{'repository': {'open_issues': 0, u'url': 'https://github.com/...
```
sigmavirus24: Missed one here

aless10: Thanks! I'll fix it as soon as possible.

aless10: Fixed!

PR #5338, opened 2020-02-17
Removed pass from the warnings and fixed the docstrings

rednafi:
* Removed multiple redundant `pass` from the `Warnings` in `requests/requests/exceptions.py`. This maintains consistency with the previously defined exceptions.

* Added missing `.` to the docstring for keeping consistency


rednafi:
Thanks. Noted.

Issue #5341, opened 2020-02-17
idna requirement causing problems now idna 2.9 has been released.


cjw296:
requrests pins idna to <2.9, which causes problems for pip now that 2.9 has been released.

## Expected Result

projects that require requests to install cleanly when new second-point idna releases are made.

## Actual Result

```
pip install .[dev]
...
Collecting idna>=2.0.0; extra == "secure"
...

ERROR: requests 2.22.0 has requirement idna<2.9,>=2.5, but you'll have idna 2.9 which is incompatible.
...
```

Then, when you try and use a setuptools entrypoint script:

```
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (idna 2.9 (.../site-packages), Requirement.parse('idna<2.9,>=2.5'), {'requests'})
```

## Reproduction Steps

See above.

## System Information

```
Python 3.6
setuptools 45.2.0
pip 20.0.2
```

cjw296:
See also https://github.com/psf/requests/issues/5337 and https://github.com/psf/requests/pull/5226

jacobtomlinson:
I am also experiencing this.

cjw296:
@sethmlarson  - any chance we could keep this issue open until the release goes out? Saves people reporting dupes...

dralley:
Yes please. A timeline on the new release would also be appreciated.

nateprewitt:
@dralley The release just went up onto PyPI.

marcmaxson:
I resolved this by upgrading requests:
`pip install requests --upgrade`
from version 0.22 to 0.23, then dependency is `idna<3,>=2.5`

tadejp:
> I resolved this by upgrading requests:
> `pip install requests --upgrade`
> from version 0.22 to 0.23, then dependency is `idna<3,>=2.5`

this actually worked, thanks

Issue #5337, opened 2020-02-16
Stop downgrading dependencies


KOLANICH:
`requests` has restricted dependencies versions with `<` conditions. This causes dependencies downgrades.

## Expected Result

It must never have been done.

## Actual Result

It has been done.

## Reproduction Steps

```bash
git clone --depth 1 https://github.com/chardet/chardet
cd chardet
python3 ./setup.py bdist_wheel
pip install --upgrade ./dist/*.whl
```

`Successfully installed chardet-4.0.0`

```bash
git clone --depth 1 https://github.com/psf/requests
cd requests
python3 ./setup.py bdist_wheel
pip install --no-index ./dist/*.whl
```

`No matching distribution found for chardet<3.1.0,>=3.0.2 (from requests==2.22.0)`

nateprewitt:
Resolving in favor of #5341.

PR #5226, opened 2019-10-10
Fix duplicated idna requirement

Jamim:
Hello everyone,

The `idna` is in the `install_requires` already, so it makes sense to remove it from extra requirements.

Also, `poetry` has been confused by this duplicated requirement.
https://github.com/sdispater/poetry/issues/1449

Best regards!

PR #5342, opened 2020-02-17
Limit to major releases rather than second point.

cjw296:
requests should trust dependent packages to do semver rather than artificially limiting version compatibility, which causes problems for pip.

Fixes https://github.com/psf/requests/issues/5341, https://github.com/psf/requests/issues/5337 and supercedes https://github.com/psf/requests/pull/5226.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,7 @@ setup(
     cmdclass={'test': PyTest},
     tests_require=test_requirements,
     extras_require={
-        'security': ['pyOpenSSL >= 0.14', 'cryptography>=1.3.4', 'idna>=2.0.0'],
+        'security': ['pyOpenSSL >= 0.14', 'cryptography>=1.3.4'],
         'socks': ['PySocks>=1.5.6, !=1.5.7'],
         'socks:sys_platform == "win32" and python_version == "2.7"': ['win_inet_pton'],
     },
--- a/setup.py
+++ b/setup.py
@@ -42,8 +42,8 @@ if sys.argv[-1] == 'publish':
 packages = ['requests']
 
 requires = [
-    'chardet>=3.0.2,<3.1.0',
-    'idna>=2.5,<2.9',
+    'chardet>=3.0.2,<4',
+    'idna>=2.5,<3',
     'urllib3>=1.21.1,<1.26,!=1.25.0,!=1.25.1',
     'certifi>=2017.4.17'
 


setup.py
```diff
@@ -42,9 +42,9 @@ def run_tests(self):
 packages = ['requests']
 
 requires = [
-    'chardet>=3.0.2,<3.1.0',
-    'idna>=2.5,<2.9',
-    'urllib3>=1.21.1,<1.26,!=1.25.0,!=1.25.1',
+    'chardet>=3.0.2,<4',
+    'idna>=2.5,<3',
+    'urllib3>=1.21.1,<2,!=1.25.0,!=1.25.1',
```
sethmlarson: I agree with the other dependencies, however probably best to leave the `urllib3` version locks. Requests and urllib3 coordinate releases usually so the version locks are safe. :)

cjw296: If releases are coordinated, surely there's no need for a lockstep version pin?

nateprewitt: I agree with Seth that the maintainers should be managing the `urllib3` version bump.

Requests digs into the internals of urllib3 in a handful of places. While urllib3 may release a non-breaking change for external APIs, those changes have broken Requests in the past. There have also been cases where a release of urllib3 introduces a non-trivial bug that requires a few weeks to fix. For that reason we usually wait a few days to a week to do the official bump.

nateprewitt:
Thanks for the patch, @cjw296! I'm a bit torn on this one. I agree that logistically this is a painful process for all parties involved. A minor version bump *should* be backwards compatible, but inevitably, that rule will be broken. The question comes down to how should a library used in a number of pieces of critical infrastructure handle that case?

We've had a lot of discussion on this in #4673 (Reasoning: [1](https://github.com/psf/requests/issues/4673#issuecomment-395069592), [2](https://github.com/psf/requests/issues/4673#issuecomment-395449858)). The specifics of that ticket are for urllib3, but it applies similarly to chardet and idna. The short of it being we'd rather be explicit on what we support in the event issues arise. The way pip handles conflicting dependencies makes this case more complicated, but that's unfortunately the environment we're currently in.

I'm happy to have the discussion again if we've got opinions from Seth or others on the Requests team.

setup.py
cjw296: Okay, any reason not to just specify `urllib3>=1.25,<1.26`?

cjw296:
@sethmlarson seems in agreement with what I've suggested, and I didn't realise how intertwined the `urllib3` usage was, so what you've both said makes sense. Would you be amenable to merge and release if I rebase, and change the `urllib3` spec to the one I suggest above?

setup.py
nateprewitt: The != declarations were made because the first couple releases of urllib3 1.25 had breaking changes. We'll clean those up once the next minor version is out and we bump the pin to < 1.27.

cjw296: Sure, but can you think of a reason why anyone would need, or indeed want, to stick to urllib3 1.21.1 to 1.24.x? (I ask, because as we've learned, pip's "solver" doesn't like complicated version requirements, so anything we can do to simplify, the better).

nateprewitt:
chardet is pretty stable and dormant now, so I'm not so concerned about that pin. idna is kind of an unknown. It's probably safe, I'm just not excited on having to do an emergency release to pin it in the future after any potential damage has already been done.

If Seth wants to approve and merge, I'm ok with it once the urllib3 pins are reverted.

setup.py
nateprewitt: We’ve left the wider gap open since individuals may not have complete control over their ecosystem and what version of urllib3 that’s being vended.

Eventually we’ll need to start pulling the tail up, but we haven’t really talked about when to do that yet. I’d say we’d want to support at least 3 trailing minor versions. Let’s just leave it for now.

cjw296:
To be clear: I've had to do emergency releases of at least three packages as a result of this pin and the idna release. It's a risk call for you, the requests maintainers: do you want to do emergency releases of requests on every new release of a dependency, or do you want to do an emergency release under the, hopefully rare, circumstances that a dependency doesn't do semver right.

cjw296:
@sethmlarson - over to you :-)

jacobtomlinson:
Just to add weight to @cjw296 I have also had to take emergency action as a result of the idna release and requests pinning it to a minor version.

From my perspective the risk of a minor release being breaking is valid but hopefully low. Though I completely acknowledge that it must be frustrating for the requests maintainers to have to resolve it with an emergency release when it inevitably happens. 

However debugging the issues caused today by the idna release, because requests has pinned a minor version, has been unpleasant. One specific example is that a package I maintain does not depend on requests, but a dependency does. Another dependency has idna as a dependency but is pinned on major versions. It seems pip was unable to resolve this dependency graph correctly and pulled in 2.9, breaking requests, the dependency of my package and therefore the whole package. Adding requests as a dependency to my package and pinning to `<2.9` has resolved it and resulted in an emergency release. 

This has been frustrating to untangle and has fallen on those that depend on requests to resolve.

nateprewitt:
Yep, I'll work on getting a cleaned up release out today or tomorrow.

PR #5343, opened 2020-02-17
Moved some metadata from setup.py into setup.cfg.

KOLANICH:
And fixed metadata extraction from files.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 # Learn more: https://github.com/kennethreitz/setup.py
+import ast
 import os
 import re
 import sys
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -39,75 +40,36 @@ def run_tests(self):
...
-about = {}
-with open(os.path.join(here, 'requests', '__version__.py'), 'r', 'utf-8') as f:
-    exec(f.read(), about)
-
-with open('README.md', 'r', 'utf-8') as f:
-    readme = f.read()
-with open('HISTORY.md', 'r', 'utf-8') as f:
-    history = f.read()
+
+def extract_global_vars(t):
```
nateprewitt, line 44: While you can work out what this doing, it's not immediately obvious. We try to keep the common packaging files in Requests as accessible as possible, so new community members have something simple to reference. We may want a slightly less black magic solution for this.

nateprewitt:
Thanks for getting the ball rolling on this @KOLANICH. We know the PyPA packages have started to move to this layout for their setup.{py,cfg} files. We're watching adoption right now before moving the Requests layout over. We can probably revisit this in another few months. I'm going to close this for now though.

setup.py
KOLANICH: We can move everything into `setup.cfg`. Version can be set from a git tag using `setuptools_scm` and written to a `.py` file.

KOLANICH:
>We know the PyPA packages have started to move to this layout for their setup.{py,cfg} files. 

Metadata in `setup.cfg` is available since 2016-12.

>We're watching adoption right now before moving

IMHO looking to other packages is completely unneeded. You know, changing it is a piece of work with almost no benefits for package owners. And sometimes with drawbacks. Using `setup.cfg` usually benefits not developer, but other people, since the format is machine-readable, machine-editable (for example, as I have said, I deeply dislike versions pinnings you use, so I feel like I have to write a script removing pinnings, unfortunately for `setup.py` it is nearly impossible task because its structure can be arbitrary and cannot be parsed in general) and non-Turing-complete (so in future it can be possible to get rid of `setup.py` as a possible place for malware, also it is easier to audit).

nateprewitt:
> Metadata in setup.cfg is available since 2016-12.

 Right, but moving things like the trove classifiers and other setup config has only become common in the last 6ish months and is predominantly limited to packages managed by the PyPA.

>  You know, changing it is a piece of work with almost no benefits for package owners. And sometimes with drawbacks. Using setup.cfg usually benefits not developer, but other people, since the format is machine-readable, machine-editable [...]

Yep, we're not doing this for our sake. We'd like to be cognizant about all of our users needs but need to do what's best for the common user. Like I said, I'm not against this but I'd like to see how it shakes out in the rest of the ecosystem before we "endorse" this as the format going forward. I'm happy to revisit the layout maybe mid-year.

KOLANICH:
>moving things like the trove classifiers and other setup config has only become common in the last 6ish months

I don't want to check, but it feels like I use it at least for 2 years. I have started using setup.cfg almost immediately it has landed into release, I remember that for the first time it required explicit reading of it, but somwhen it has changed. In 2018, if I remember right, I was already using `setup.cfg` without explicit reading.

PR #5344, opened 2020-02-17
Added .editorconfig according to PEP 8

KOLANICH:


nateprewitt:
Thanks @KOLANICH, we typically won't include non-packaging dot files in Requests. I think we'll pass on this one.

PR #5347, opened 2020-02-18
Bump maximum version of idna

jacobtomlinson:
Looks like `idna` 2.9 was released yesterday. Bumping maximum version.


jacobtomlinson:
I've just seen #5342 which would be a more preferable solution.

sethmlarson:
Closed in favor of #5342.

Issue #5351, opened 2020-02-18
Document description about getting cookies.


skokado:
I am very grateful for this project

The documentation seems to be incorrect about getting cookies of request.
You must disable redirects to get a cookie for a request to an endpoint.

I have confirmed this behavior locally like below.

```python
>>> url = 'http://httpbin.org/cookies/set/hoge/fuga'  # It is expected that a cookie "hoge = fuga" can be getted.
# Case: Does not work
>>> res = requests.get(url)
>>> res.cookies
# => <RequestsCookieJar[]>

# Case: Does work well
>>> res = requests.get(url, allow_redirects=False)
>>> res.cookies
# => <RequestsCookieJar[Cookie(version=0, name='hoge', value='hogehoge', port=None, port_specified=False, domain='httpbin.org', domain_specified=False, domain_initial_dot=False, path='/', path_specified=True, secure=False, expires=None, discard=True, comment=None, comment_url=None, rest={}, rfc2109=False)]>
```

This is because the response to the URL after the redirect is automatically replaced by allow_redirects at the time of the request.

I believe the documentation should be correct as it is reasonable that the default is True,
so I think that it is user-friendly to correct the document showing the use case.

nateprewitt:
Hi @skokado,

I think what you may want here is the `history` attribute on your response. It will keep the trail of the requests, along with the state of each request hop. That way you can go back to the first response (e.g. `res.history[0].cookies`).

You should be able to find [more documentation](https://requests.readthedocs.io/en/master/user/quickstart/#redirection-and-history) on this immediately under the cookie section in the docs.

skokado:
Hi @nateprewitt 

Thank you for reply and great tips.
I didn't know that... and I'm convined your answer :)

nateprewitt:
Great to hear! I’m going to resolve this and the PR since we have this easier mechanism to access the redirects. If you have further suggestions for the docs, please let us know!

PR #5352, opened 2020-02-18
Fix document about getting cookies in Document

skokado:
relevant to #5351


nateprewitt:
Resolving with clarification in #5351.

PR #5370, opened 2020-02-28
Also update GitHub's Security Disclosure document

sethmlarson:
Forgot to update this document with new disclosure info.


sethmlarson:
Don't think there is a nice way except maybe pointing to one from the other? Maybe the documentation link can point to the GitHub disclosure workflow?

PR #5373, opened 2020-03-01
Update structures.py

AkDavies:
Changed lower_items() to use iterable unpacking to clarify the intent of the function. Changed copy() and __eq__() to retrieve the class via inspection using type() -- this way behavior will be preserved in classes that inherit from CaseInsensitiveDict.

### DIFF
--- a/requests/structures.py
+++ b/requests/structures.py
@@ -65,14 +65,14 @@ class CaseInsensitiveDict(MutableMapping):
     def lower_items(self):
         """Like iteritems(), but with all lowercase keys."""
         return (
-            (lowerkey, keyval[1])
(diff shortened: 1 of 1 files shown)


sigmavirus24:
>  this way behavior will be preserved in classes that inherit from CaseInsensitiveDict.

I don't believe this is meant to be reused in consumers of the library and using `type(self)` makes this less understandable than it was previously

nateprewitt:
I agree with Sigmavirus24, this makes the code a bit harder to read. The structures are implementation details of Requests and aren’t really intended for extension beyond the library. I think we’ll pass on this but thanks for the contribution!

PR #5375, opened 2020-03-02
Added a warning when stripping authentication from requests on a redirect

kmaork:
Hello!
After seeing this [question](https://stackoverflow.com/questions/60358216/python-requests-post-request-dropping-authorization-header), I thought a warning should be added when authentication is stripped from redirected requests, in order to save users a significant amount of debugging time.
Apparently the matter has already been discussed in [this issue](https://github.com/psf/requests/issues/2949), and it has been agreed upon that *some* message should be passed to the user, like a debug log or a warning. But requests still does not use logs, and it seems people are still confused by this.
Maybe it is time to add a warning?

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -10,6 +10,7 @@ requests (cookies, auth, proxies).
 import os
 import sys
 import time
+import warnings
 from datetime import timedelta
 from collections import OrderedDict
 
@@ -22,7 +23,9 @@ from .hooks import default_hooks, dispatch_hook
 from ._internal_utils import to_native_string
 from .utils import to_key_val_list, default_headers, DEFAULT_PORTS
 from .exceptions import (
-    TooManyRedirects, InvalidSchema, ChunkedEncodingError, ContentDecodingError)
+    TooManyRedirects, InvalidSchema, ChunkedEncodingError, ContentDecodingError,
+    RequestsWarning
+)
 
 from .structures import CaseInsensitiveDict
 from .adapters import HTTPAdapter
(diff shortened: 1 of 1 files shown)


kmaork:
Does the PR need another review?

kmaork:
Can we make progress with this? 🙂 @yanz-androidify

Ezhvsalate:
Great PR, lost an hour today trying to understand why my service was broken after moved to OpenShift (it internally redirected requests). This warning could have helped me a lot.

SaritNike:
This PR would come really handy @yanz-androidify @kmaork why does it still say that the merging is blocked? How can we get someone with write access to review and approve this PR? It's waiting since May.  
The current state of requests makes it nearly unusable for me when the host is on some cloud provider's server which does automatic redirects on ingress.

scorchio:
I've lost a good 2 hours trying to find out what's wrong when I was playing around with PyGithub, which is using requests under the hood.

@nateprewitt, could you please take a look and provide some feedback? It would be great if requests could be a bit more helpful in such scenarios.

yannicschroeer:
Please add this. The current behavior is annoyingly implicit!

kmaork:
Hey @nateprewitt , @sigmavirus24 , @sethmlarson - this PR was approved ages ago by @yanz-androidify but seems to have been forgotten since. Any chance of it getting reviewed?

sigmavirus24:
@kmaork I don't see an approval from that account nor are they a maintainer (nor does that account seem to exist any longer).

As it stands, yes this can be annoyingly secure and hidden from the user but this usage of warnings is inappropriate for the solution. It's not the right mechanism and it will result in more issues generated here than users helped.

PR #5380, opened 2020-03-09
Added a better way to assign default values to variables

sukurcf:


nateprewitt:
Hey @sukurcf,

Thanks for contributing! Unfortunately, the suggestion isn’t directly equivalent to the logic that’s currently in place. You’re now defaulting for any “false-y” value instead of explicitly None. This would be considered a breaking change, so we can’t accept it at this time.

PR #5391, opened 2020-03-17
Send only one Host header in chunked request

### LINKED ISSUE #5274: Two Host header are being sent when doing chunk transfer with Session
Two Host header are being sent when doing chunk transfer with Session
---------------
>>> import requests
>>> requests.__version__
'2.22.0'
```
import requests
import time
from requests.packages import urllib3

#Disable https insecure warnings for verify=False
urllib3.disable_warnings()

with open('1.ts','rb') as fh:
    file_data = fh.read()

def chunk_trans():
    for i in range(0, 3):
        time.sleep(0.1)
        yield file_data

req = requests.Request('POST', 'http://192.168.123.151/1.ts', data=chunk_trans(), headers={'Host':'a.b.c.d'})
prep = req.prepare()
s = requests.Session()
s.send(prep, timeout=2, verify=False, allow_redirects=False)

```

Actual Output: Two host headers are seen on server side.
```
Host: 192.168.123.151
Host: a.b.c.d
Transfer-Encoding: chunked
```
-------------

only "a.b.c.d" host header should come.

hodbn:
Send only one Host header when specifying a custom Host header in a chunked request.

Before this fix, the default Host header was sent along with the custom Host header.

Closes #5274

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -458,10 +458,11 @@ class HTTPAdapter(BaseAdapter):
                 low_conn = conn._get_conn(timeout=DEFAULT_POOL_TIMEOUT)
 
                 try:
+                    skip_host = 'Host' in request.headers
                     low_conn.putrequest(request.method,
                                         url,
                                         skip_accept_encoding=True,
-                                        skip_host='Host' in request.headers)
+                                        skip_host=skip_host)
 
                     for header, value in request.headers.items():
                         low_conn.putheader(header, value)
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -460,7 +460,8 @@ class HTTPAdapter(BaseAdapter):
                 try:
                     low_conn.putrequest(request.method,
                                         url,
-                                        skip_accept_encoding=True)
+                                        skip_accept_encoding=True,
+                                        skip_host='Host' in request.headers)
 
                     for header, value in request.headers.items():
                         low_conn.putheader(header, value)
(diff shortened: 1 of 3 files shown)


hodbn:
Hi @AaronRobson :)
Can you take another look at this?

requests/adapters.py
```diff
@@ -459,7 +459,8 @@ def send(self, request, stream=False, timeout=None, verify=True, cert=None, prox
                 try:
                     low_conn.putrequest(request.method,
                                         url,
-                                        skip_accept_encoding=True)
+                                        skip_accept_encoding=True,
+                                        skip_host='Host' in request.headers)
```
AaronRobson: That's good, this appears to be the same way that the same situation is handled in the urllib3 code itself: https://github.com/urllib3/urllib3/blob/74d6be1ab66cef44c0f479c24b0fc1756a8fe4e9/src/urllib3/connection.py#L243-L246

AaronRobson (approved):

nateprewitt (approved):
Thanks for digging in a finding a simple solution here, @hodbn! This seems like a straight forward win that alleviates the risks of other patches we've had for #5274. Apologies this has been sitting so long.

I've added some minor test refactoring to make these easier to expand later and rebased your changes onto our `main` branch. I think this should be set!

sethmlarson:
@nateprewitt What's the state of this PR?

nateprewitt:
It's ready, it just needs to be rebased to rerun tests and we can merge. It's on my list for Monday now that 2.27.0 is unblocked.

nateprewitt:
Thanks again @hodbn!

PR #5404, opened 2020-03-29
typo

ndanil:
typo in options?


docs/user/quickstart.rst
```diff
@@ -47,7 +47,7 @@ OPTIONS? These are all just as simple::
     >>> r = requests.put('https://httpbin.org/put', data = {'key':'value'})
     >>> r = requests.delete('https://httpbin.org/delete')
     >>> r = requests.head('https://httpbin.org/get')
-    >>> r = requests.options('https://httpbin.org/get')
+    >>> r = requests.options('https://httpbin.org/options')
```
dtrodrigues, line 50: /options isn't a valid path. The original example looks to be correct.

nateprewitt:
@dtrodrigues is right, this is the correct endpoint in the current docs. Thanks for checking!

PR #5406, opened 2020-03-29
Port of PR #4173 from v3 to v2 to fix issue #4391

### LINKED ISSUE #4391: requests.session ignores second `verify` argument?
Am doing some testing for HTTPS certificates, and am seeing some unexpected behaviour from the requests "session" object:

```
import requests
# www.ottg.eu has a known invalid https certificate:
requests.session().get('https://www.ottg.eu', verify=True)
# raises error as expected

session = requests.session()
session.get('https://www.ottg.eu', verify=False)
# as expected does not raise, prints InsecureRequestWarning, 

session.get('https://www.ottg.eu', verify=True)
# unexpectedly, this seems to ignore the verify parameter, and it does not raise
# it also still prints the InsecureRequestWarning
```

That doesn't seem right?

tested against requests 2.18.4 on ubuntu 17.10

atzannes:
Fixes #4391

# Problem
This code was failing to verify the SSL:
```python
    def test_session_verify_not_stateful(self):
        insecure_url = "https://expired.badssl.com/"
        s = requests.Session()
        assert s.verify is True

        with pytest.raises(Exception):
            s.get(insecure_url, verify=True)

        s.get(insecure_url, verify=False)
        assert s.verify is True

        with pytest.raises(Exception):
            s.get(insecure_url, verify=True)  # <- This line was not raising bc we got a cached unverfied connection
```
The problem was that we were modifying the connection object returned to us by the poolmanager, which is illegal, causing us to get unverified connections when requesting verified ones.

# Approach
I had come up with the same solution as PR #4173 independently, not having found the issue before trying to fix it, but my unit-tests were not as thurough, so I added those from PR #4173

The issue was closed in 2018 because it was fixed in v3 (which is still not released), and presumably because it was not fixable in v2, but perhaps this fix works for v2 (new urllib3 perhaps that allows it?)

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -260,9 +260,12 @@ class HTTPAdapter(BaseAdapter):
         return manager
 
     def cert_verify(self, conn, url, verify, cert):
-        """Verify a SSL certificate. This method should not be called from user
-        code, and is only exposed for use when subclassing the
+        """This call has been deprecated but kept for API compatibility.
+
+        This method should not be called from user code, and is only
+        exposed for use when subclassing the
         :class:`HTTPAdapter <requests.adapters.HTTPAdapter>`.
+
         :param conn: The urllib3 connection object associated with the cert.
         :param url: The requested URL.
         :param verify: Either a boolean, in which case it controls whether we verify
@@ -270,45 +273,7 @@ class HTTPAdapter(BaseAdapter):
             to a CA bundle to use
         :param cert: The SSL certificate to verify.
         """
-        if url.lower().startswith('https') and verify:
-
-            cert_loc = None
-
-            # Allow self-specified cert location.
-            if verify is not True:
-                cert_loc = verify
-
-            if not cert_loc:
-                cert_loc = extract_zipped_paths(DEFAULT_CA_BUNDLE_PATH)
-
-            if not cert_loc or not os.path.exists(cert_loc):
-                raise IOError("Could not find a suitable TLS CA certificate bundle, "
-                              "invalid path: {}".format(cert_loc))
-
-            conn.cert_reqs = 'CERT_REQUIRED'
-
-            if not os.path.isdir(cert_loc):
-                conn.ca_certs = cert_loc
-            else:
-                conn.ca_cert_dir = cert_loc
-        else:
-            conn.cert_reqs = 'CERT_NONE'
-            conn.ca_certs = None
-            conn.ca_cert_dir = None
-
-        if cert:
-            if not isinstance(cert, basestring):
-                conn.cert_file = cert[0]
-                conn.key_file = cert[1]
-            else:
-                conn.cert_file = cert
-                conn.key_file = None
-            if conn.cert_file and not os.path.exists(conn.cert_file):
-                raise IOError("Could not find the TLS certificate file, "
-                              "invalid path: {}".format(conn.cert_file))
-            if conn.key_file and not os.path.exists(conn.key_file):
-                raise IOError("Could not find the TLS key file, "
-                              "invalid path: {}".format(conn.key_file))
+        pass
 
     def build_response(self, req, resp):
         """Builds a :class:`Response <requests.Response>` object from a urllib3
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -52,6 +52,65 @@ DEFAULT_RETRIES = 0
 DEFAULT_POOL_TIMEOUT = None
 
 
+def _build_conn_kwargs(url, verify, cert):
+    """Extract the arguments needed to create an appropriate connection.
+
+    This method should not be called from user code, and is only
+    exposed for use when subclassing the
+    :class:`HTTPAdapter <requests.adapters.HTTPAdapter>`.
+
+    :param url: The requested URL.
+    :param verify: Either a boolean, in which case it controls whether we verify
+        the server's TLS certificate, or a string, in which case it must be a path
+        to a CA bundle to use
+    :param cert: The SSL certificate to verify.
+    :returns: optional kwargs for building connection, relating to SSL verification
+    :rtype: dict
+    """
+    conn_kwargs = {}
+
+    if url.lower().startswith('https') and verify:
+
+        cert_loc = None
+
+        # Allow self-specified cert location.
+        if verify is not True:
+            cert_loc = verify
+
+        if not cert_loc:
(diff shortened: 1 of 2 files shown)


jeremycline:
The issue was that fixing it involved breaking an API in v2. I suppose a feature flag could be added, something like ``really_actually_do_verify`` which defaults to False but is documented as something you  _probably_ want, and the old behavior and API are available with that flag set to False. That'd make it easier to work around.

atzannes:
Hi @jeremycline thanks for the feedback. Which API is broken in v2 by this change? Are you refering to `get_gonnection` in adapters.py which now takes two additional optional arguments, or the removal of `cert_verify` which was illegally modifying the connection object?

jeremycline:
The removal of ``cert_verify`` which is a public method. If you kept that and switched behavior based on a feature flag it'd probably be more palatable.

atzannes:
ok, great! I wasn't sure whether `cert_verify` was treated as public given the comment saying "This method should not be called from user code."

Having played around with it a bit, I have a suspicion that keeping that function in, even without a feature flag, will do the right thing in combination with building and passing the `conn_kwargs` to the poolmanager. I'll test that hypothesis and get a update up soon. Again, thanks for the quick feedback @jeremycline !

jeremycline:
Oh, also, just to be clear I'm not a maintainer here, my feedback has no influence on the acceptance of such a pull request, is provided "as is", etc. :smile:

atzannes:
@jeremycline I know you contributed the PR where I got most of the unit-tests from for this PR. And I was happy to see that the solution I had come up with was effectively identical with the one you had. By reading your PR, I inferred that you're not a maintainer here. But your feedback is very welcome and valuable nonetheless.

Ok, so I force-pushed and now there are two commits:
- the first where I've only added the `_build_conn_kwargs` function and where `cert_verify` (the member function and its call in `send`) are untouched. Tests pass even despite its shenanigans.
- the second where `cert_verify` has been deprecated (its body is replaced by `pass`), but we're still calling it from `send`, so that any code that has derived from HTTPAdapter and has not overriden `send` but has overriden `cert_verify` would continue working presumably.

I prefer having `cert_verify` deprecated because it was doing something illegal. 

Thoughts? Concerns?

requests/adapters.py
```diff
@@ -201,8 +260,10 @@ def proxy_manager_for(self, proxy, **proxy_kwargs):
         return manager
 
     def cert_verify(self, conn, url, verify, cert):
```
jeremycline, line 262: I think you'll want to document the method as deprecated, but leave it as-is. Yes, it doesn't work the way most people want, but leaving the signature there with a behavior change is an API breakage, too.

What seems less invasive to me is to instead is to call this method in the default implementation conditionally based on a feature flag which defaults to `False`. That way existing users see no behavior change.

atzannes: I see, I'll revert to the original the body of `cert_verify`. Do we want a warning (like `urllib3.connection` does) notifying that the method has been deprecated if it's called, or just a comment in the docstring.

If I understand your suggestion for a feature flag, the constructor for `Session` objects would take an optional boolean that defaults to False. When True, it would cause to skip the call to `cert_verify` from within `HTTPAdapter.send`. Given that the tests pass without the feature flag (and with the original `cert_verify` code), are there advantages to having this feature-flag? One advantage I can see in the case of having a warning produced by `cert_verify` being called, is that it could get the attention of developers to switch to using the feature-flag. But then, wouldn't we be stuck with having that feature-flag? I haven't written software that's as widely deployed as requests, so I don't have lots of experience about which types of changes are preferable when fixing a bug, and what are the consequences downstream. For example, I would have though it preferable for code to break than to make an unintended unverified request.

jeremycline: Well, if you're not breaking any APIs no flag is necessary, but I'm a bit surprised it's not breaking something. I can't say, though, without thoroughly re-acquainting myself with urllib3 and requests code. 

Sorry I can't be more help, but I'm no longer using either one in any particularly complex situations so it's hard to gather the energy to briefly dive back into the codebases for a review. Maybe in the next couple weeks I'll feel inspired to do some digging, but I don't know.

atzannes: So what was happening before this change was that in `adapter.HTTPAdaper.send` we were first calling `get_connection`, then calling `cert_verify`. The `get_connection` call was getting to the right `PoolManager` and returning a `ConnectionPool` and then the `cert_verify` was (illegally) modifying that `ConnectionPool` object. It appears that the first time you call `cert_verify` on a connection pool, it behaves like a call to `HTTPSConnectionPool.set_cert`. According to the docs "this method should only be called once, before the connection is used". 

The reason I think this change is working without breaking anything is that now `get_connection` uses `verify` and `cert` to call `_build_conn_kwargs` which is passed to the `PoolManager` to return the right `ConnectionPool`. Keep in mind that `_build_conn_kwargs` uses the same logic as `cert_verify` except it builds a kwarg dict instead of modifying a `ConnectionPool` in place. So, by using these connection kwargs, the `PoolManager` is returning to us the "correct" `ConnectionPool` and subsequently, when we call `cert_verify`, it sets some attributes on the connection, but those attributes should already exist and have the correct values (as per the `conn_kwargs`). So, in effect `cert_verify` should be a no-op because we used the same values in `conn_kwargs` when we got the `ConnectionPool` from the `PoolManager`.

Did this make sense? I'm trying to describe at a high-ish level what I think is happening. Like you, I am not intimately familiar with the inner workings of urllib3, but these were my conclusions from running the failing test I added through `pdb` and looking at what we got back from urllib3.

atzannes: Also, I understand you're not a maintainer here, and understanding why this works would require some non-trivial digging. Do you think I should ping the maintainers ?

atzannes:
@nateprewitt @Lukasa gentle ping. This is an API-compatible backport of @jeremycline's great fix for issue #4391. 

- Added `_build_conn_kwargs` which mirrors what `cert_verify` does, but instead of modifying the `conn` argument in-place, it builds kwargs to be passed to the `PoolManager` so that it returns the right connection. The subsequent call to `cert_verify` should be a no-op if `PoolManager` honored the kwargs we passed.
- API change: `get_connection` in HTTPAdaper takes two new optional arguments: `verify`, and `cert` because it needs to call `_build_conn_kwargs`.

atzannes:
[Is there anybody out there?](https://www.youtube.com/watch?v=fNLhxKpfCnA)

sethmlarson:

PR #5413, opened 2020-04-05
Add text_with_headers to Response model

JonPizza:
This property will return the entire HTTP response including the headers. There doesn't seem to be a `test_models.py` file in the tests folder so I'm not sure if I should make a test for this, but if so where would I make it?

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -870,6 +870,15 @@ class Response(object):
             content = str(self.content, errors='replace')
 
         return content
+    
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @JonPizza,

Requests is currently under a feature freeze. This will probably be useful to a subset of users but is pretty straight forward to implement if needed. I think we'll pass on this for now. Thanks!

PR #3666, opened 2016-11-03
documentation line added for consistency with the post method

hootnot:
Docs PR regarding #3664 and #2979

PR #5418, opened 2020-04-10
Remove extraneous paramater from doctring for put() method

gregpoulos:
There's an unused `json` parameter in the docstring from the put() method.


nateprewitt:
Hi @gregpoulos, thanks for the contribution! This was added in #3666 because there was some confusion on whether verbs other than POST supported the json argument. The documentation change is intended to make it clear despite it not being an explicit arg. I think we'll probably leave this in place for now.

gregpoulos:
@nateprewitt Understood. Thanks for the quick response!

PR #5424, opened 2020-04-14
WIP: Allow use of system CA certs

jflemer-ndp:
This allows setting `REQUESTS_CA_BUNDLE=SYSTEM` to make Requests behaves like the base python standard library with respect to CA cert trust.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -212,7 +212,10 @@ class HTTPAdapter(BaseAdapter):
             to a CA bundle to use
         :param cert: The SSL certificate to verify.
         """
-        if url.lower().startswith('https') and verify:
+        if url.lower().startswith('https') and verify == 'SYSTEM':
+          conn.cert_reqs = 'CERT_REQUIRED'
+
+        elif url.lower().startswith('https') and verify:
 
             cert_loc = None
 


sigmavirus24:
Based on a comment on the original issue (https://github.com/psf/requests/issues/2966#issuecomment-613577111) this is not trustworthy (i.e., we can't guarantee that the systems will remotely do the right thing) enough of a fix to accept.

jflemer-ndp:
This commit doesn't change default behavior.

sigmavirus24:
I don't see what your comment has to do with my point that anyone attempting to use this value would have no visibility into whether it's remotely doing the right thing because this library and it's dependencies can't guarantee that.

jflemer-ndp:
Agree, it doesn't guarantee anything other than that it will cause Requests to use the underlying python ssl behavior. And users have to explicitly enable it, therefor assume responsibility for use.

sigmavirus24:
Right. And users can already do that without overloading the usage of this environment bundle with a transport adapter. For example, the [toolbelt](/requests/toolbelt) has a SSL Adapter which allows for modification of the defaults and makes it easy to search for usage of non-default TLS settings. A similar approach could be taken by a user to force the default certificate locations to be used instead of certifi including doing something like:

```
REQUESTS_CA_BUNDLE="$(python -c 'import ssl; print(ssl.get_default_verify_paths().cafile)')"
```

PR #5426, opened 2020-04-16
word correction

cciprian5:


sigmavirus24:
This isn't a correction. `sans` means `without`.

cciprian5:
Mea culpa... I did not know that this word exists in english. Thought that
it's a leftover from french translation.

sigmavirus24:
It's one of those French words that actually gets used in (at least American) English with relative frequency. I'm not against replacing it with "without" but it's not a faux pas in this case

PR #5431, opened 2020-04-19
Add more details in error reponse of "Bad Request"

bitsapien:
When there is an exception raised on a 400 "Bad Request" from the server. The information provided on the exception is not enough to debug problems. Adding "content" attribute from the response to understand the error from the server's perspective.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -932,7 +932,7 @@ class Response(object):
             reason = self.reason
 
         if 400 <= self.status_code < 500:
-            http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This doesn't scale. `response.content` isn't guaranteed to be pre-populated and in the event of a malicious server, this causes us to hang before we even raise the exception because a server could send an endless stream of bytes and requests has no mechanism to stop it.

sigmavirus24:
Also, there's already a response attribute. If you want more information  while handling this, log the `error.response.content` value yourself and at your own risk

nateprewitt:
Hi @bitsapien,

Sigmavirus24 already laid out the reasoning it looks like. We already have cases of the scenario he describes happening in the wild, so for that reason, I think we'll pass on this. The workaround provided should get you to where you want.

PR #5437, opened 2020-04-25
docs: make danger of verify=False clearer

elithrar:
This PR makes the danger of setting `verify=False` clearer to users of requests. In many cases, users may set this to `False` when running into problems locally with self-signed certificates, but not realize the risk they are opening themselves up to.

As an example of how the dangers of not verifying TLS (SSL) certificates is communicated, see Go's `InsecureSkipVerify` docs: https://golang.org/pkg/crypto/tls/

```
    // InsecureSkipVerify controls whether a client verifies the
    // server's certificate chain and host name.
    // If InsecureSkipVerify is true, TLS accepts any certificate
    // presented by the server and any host name in that certificate.
    // In this mode, TLS is susceptible to man-in-the-middle attacks.
    // This should be used only for testing.
```

Happy to take feedback on the wording / doc-style - this is my first PR to requests but I care a lot about 'safe defaults' in OSS!

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -387,6 +387,13 @@ class Session(SessionRedirectMixin):
         self.stream = False
 
         #: SSL Verification default.
+        #: Defaults to `True`, requiring requests to verify the TLS certificate at the
+        #: remote end.
+        #: If verify is set to `False`, requests will accept any TLS certificate
+        #: presented by the server, and will ignore hostname mismatches and/or
+        #: expired certificates, which will make your application vulnerable to
+        #: man-in-the-middle (MitM) attacks.
+        #: Only set this to `False` for testing.
         self.verify = True
 
         #: SSL client certificate default, if String, path to ssl client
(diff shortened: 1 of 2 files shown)


elithrar:
Fixed some RST formatting / consistency.

requests/sessions.py
```diff
@@ -387,6 +387,13 @@ def __init__(self):
         self.stream = False
 
         #: SSL Verification default.
+        #: Defaults to `True`, requiring requests to verify the TLS certificate at the
+        #: remote end.
+        #: If verify is set to `False`, requests will accept any TLS certificate
+        #: presented by the server, and will ignore hostname mismatches and/or
+        #: expired certificates, which will make your application vulnerable to
+        #: man-in-the-middle (MitM) attacks.
+        #: Only set this to `False` for testing.
```
sigmavirus24, line 396: There are (rarely) good cases for setting this to False. I'd remove this line as it's likely to cause a fire-storm for the already overtaxed maintainers of this project

elithrar: I would agree! The reason that I mentioned this was to get ahead of questions about "why is this configurable in the first place?" - which is a reasonable question for someone to ask about an option like this. I'm not sure I grok the 'firestorm' part, though.

Happy to elide if there's strong feelings about it, but "when would I set this to False" is a fair question for a newcomer to ask.

sigmavirus24: There are people who jump on things like this as soon as they find it and then they post about it in places to get attention and trigger a pile-on. The project has had issues like this in the past (for example, when it kept it's own certificate bundle copy from Mozilla instead of keeping that in certifi) and so given how sensitive people can be around this particular flag and topic, I'd rather leave this as an uncontroversial item. I think the "Only" in this will be controversial enough that people will get spitting mad and try to rile others up as well

elithrar: That’s unfortunate, and hopefully we can point folks to prior art in other libraries/PLs. I have seen pile-ons the other way - e.g where the defaults are _unsafe_. But I digress.

To make sure I understand your ask concretely: you want me to remove the “Only set this...” sentence from the comment (only) - ?

sigmavirus24: Or phrase it as `"You might consider settings to False during local development or testing."` as I think that's the idea you're conveying

elithrar: Done.

elithrar:
Gentle ping - does this need anything else to be merged?

PR #5440, opened 2020-04-26
fix failed testcase in test_requests

sgy123456:
when I run pytest in test_requests, an error occurred:

def test_conflicting_post_params(self, httpbin):
        url = httpbin('post')
        with open('Pipfile') as f:
>           pytest.raises(ValueError, "requests.post(url, data='[{\"some\": \"data\"}]', files={'some': f})")
E           TypeError: 'requests.post(url, data=\'[{"some": "data"}]\', files={\'some\': f})' object (type: <class 'str'>) must be callable

tests/test_requests.py:777: TypeError

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -774,8 +774,10 @@ class TestRequests:
     def test_conflicting_post_params(self, httpbin):
         url = httpbin('post')
         with open('Pipfile') as f:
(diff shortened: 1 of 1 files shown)


nateprewitt:
This looks like a dupe of #5305. I'm going to resolve in favor of that. Thanks for the contribution though @sgy123456!

PR #5451, opened 2020-05-09
Update supported python version in doc

jacty:
I think python 3.8 is supported now after I checked some other places in this rep.


nateprewitt:
I think this ended up getting resolved in a more generic way with #5567. Resolving since the issue is addressed. Thanks for the PR @jacty!

PR #5453, opened 2020-05-12
Update Travis syntax

edumco:
### This pull request  brings the following changes:

- Ordered the commands to follow Travis CI execution order: <https://docs.travis-ci.com/user/job-lifecycle/#the-job-lifecycle> 

- Adds explicit os and dist commands protecting from changes on Travis default distro

- Forces the build to use the latest minor version of each python release so build become fully reproducible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,6 @@ os: linux
 
 dist: xenial
 
-# command to install dependencies
-install: "make"
-# command to run tests
-script:
-  - make test-readme
-  - make ci
-cache: pip
 jobs:
   include:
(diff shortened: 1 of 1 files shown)


sigmavirus24:
We don't want to pin to a minor version. Travis is supposed to auto-upgrade those for us

edumco:
> supposed to auto-upgrade those for us

I cant revert that, what do you think about the other changes?

sigmavirus24:
They're churn, functionally unnecessary, and in some cases useless (see apt caching)

Issue #3759, opened 2016-12-08
cookies.py why replace escaped quote with nothing?


saft:
https://github.com/kennethreitz/requests/blob/master/requests/cookies.py#L346

Why do this? An escaped quote can be sent in the case of sending json in the cookie and this code mangles that json. I'm raising this because the http library seems to think that formatting the json this way on the server is the right way to do it. See http/cookies.py function _quote line 225 in my 3.4 version.

sigmavirus24:
Can you provide an example of a server sending JSON in the cookie header? The RFCs ([6265](https://tools.ietf.org/html/rfc6265)) for Cookies are explicit and JSON really doesn't fit into the cookie header appropriately.

saft:
The following library does it quite explicitly. I'm not sure how stable it is but it is referenced in the aiohttp documentation.
https://github.com/aio-libs/aiohttp-session/blob/master/aiohttp_session/__init__.py#L219

sigmavirus24:
> I'm raising this because the http library seems to think that formatting the json this way on the server is the right way to do it.

So you seemed to mention that there was a server sending JSON in the set-cookie header. Now you're pointing at another library? I don't see how that library is storing JSON in a cookie at all. Do you mean that it allows for it?

saft:
Maybe this line helps to illustrate where aiohttp_session puts json into the cookie:
https://github.com/aio-libs/aiohttp-session/blob/master/aiohttp_session/__init__.py#L245

aiohttp_session uses aiohttp which uses http/cookies.py. The server is my own.

To look at it from the other perspective, why remove all escaped quotes from cookie data - 6265 doesn't disallow that?

Lukasa:
As always, the easiest thing to do when you're trying to work out the answer to a problem like this is to just look at the Git logs. If you do, they direct you to #1440.

The TL;DR is that unescaping all quotes was probably unnecessary. The goal of the original patch was simply to remove enclosing quote marks, but we took a (probably a bit too aggressive) shortcut. I'm open to a patch that changes that behaviour.

saft:
From #1440 I can't figure out what the *end* intent of it was. Perhaps to be out-of-the-box compatible with Django? Certainly now set_cookie actually keeps the enclosing quote marks.

On further analysis, RFC6265 does seem to say that backslash and double quotes are invalid characters. However what set_cookie does is to just remove all occurrences of \\".

I would love to try to improve that bit of code but I'm not sure what I should be trying to improve it towards.

[...]

PR #5459, opened 2020-05-15
Revert PR 1440, do not modify cookie value

yanxurui:
Revert PR  #1440 as discussed in issue #3759 
It does not make sense to modify the cookie value arbitrarily.


sigmavirus24:
At this point the behaviour is too ingrained to be reverted arbitrarily and would need to wait for a major version release which is not planned to the best of my knowledge

nateprewitt:
This functionality has been in the library for 7 years now, so we can't pull it out without breaking users. If we were to ever put another major version up, this could possibly be looked at, but I don't think we can merge this in 2.x. Thanks for the suggestion!

PR #5464, opened 2020-05-19
Allow semicolon in the href of Link header

ibnesayeed:
Semicolons are valid characters in URIs, splitting them while parsing the `Link` header results in unexpected token and wrong `url` attribute (notice `;oldid=934259284` portion the below example).

```
<https://en.wikipedia.org/w/index.php?title=COVID-19_pandemic&amp;oldid=934259284>; rel="original", <https://web.archive.org/web/timemap/link/https://en.wikipedia.org/w/index.php?title=COVID-19_pandemic&amp;oldid=934259284>; rel="timemap"; type="application/link-format"
```

This PR attempts to fix current parsing issue and also renames a variable to better reflect the purpose of it.

I would note here that this parser still is far from perfect and can be better implemented using state machine.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -826,9 +826,9 @@ def parse_header_links(value):
 
     links = []
 
-    replace_chars = ' \'";'
+    strip_chars = ' \'";'
 
-    value = value.strip(replace_chars)
+    value = value.strip(strip_chars)
     if not value:
         return links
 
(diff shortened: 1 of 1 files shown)


ibnesayeed:
@nateprewitt is there anything we can to do to get it merged?

recepsirin (approved):

recepsirin (approved):

PR #5477, opened 2020-05-29
Drop monkeypatching with pyopenssl

WhyNotHugo:
Monkey patching url3 with pyopenssl 3 is no longer a good idea nor even necessary, and actually has security drawbacks, as described in #5267.

This also undoes the changes I initially made in #1347, so I guess it means my previous contribution has completed its life cycle.


nateprewitt:
Hi @WhyNotHugo,

This was already addressed in #5443 and left patching as a potential fallback for now for older versions of Python 2.7. This patch has some additional removals from our debugging helper function that I don't think we want to remove. I think we'll pass on this for now.

WhyNotHugo:
Oh, I was under the impression that Python2 was being dropped anyway.

PR #5480, opened 2020-06-06
Using range method to check status code

nikolasvargas:
More readable

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -918,7 +918,7 @@ class Response(object):
     def raise_for_status(self):
         """Raises :class:`HTTPError`, if one occurred."""
 
-        http_error_msg = ''
+        http_error_msg, reason = '', ''
         if isinstance(self.reason, bytes):
             # We attempt to decode utf-8 first because some servers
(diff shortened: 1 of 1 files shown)


sigmavirus24:
But it's not

nikolasvargas:
ok

requests/models.py
```diff
@@ -918,7 +918,7 @@ def links(self):
     def raise_for_status(self):
         """Raises :class:`HTTPError`, if one occurred."""
 
-        http_error_msg = ''
+        http_error_msg, reason = '', ''
```
AaronRobson, line 921: This change appears to be unrelated to the range change.
Was it suggested by a linting tool?

nateprewitt:
Hi @nikolasvargas,

Thanks for the contribution! I don't know if I agree this is more readable though. It has the additional negative that we generate a bunch of additional numbers in the range we have to iterate but will never possibly use. The current implementation is pretty direct and common usage for comparison operators in Python.

I think we'll leave this as is for now.

PR #5488, opened 2020-06-11
Allow further configuration of HTTPAdapter's PoolManager by passing pool_kwargs dict

dtheodor:
There are various useful configurations of underlying urllib3 PoolManager not supported by the existing  HTTPAdapter, such as a global timeout, source_address, and many more https://github.com/urllib3/urllib3/blob/29b214a129883301f91ae4a74fd7ef2958cbf7b0/src/urllib3/poolmanager.py#L49-L75

It is trivial to support them directly. The construction of the HTTPAdapter's poolmanager at `HTTPAdapter.init_poolmanager(..., **pool_kwargs)` already contains this provision as the arbitrary `pool_kwargs` dict that is passed to the PoolManager constructor. All it takes is to also add this to the HTTPAdapter constructor, and pass it.

It also appears that this was the original intention that was somehow forgotten; this `**pool_kwargs` argument is not being passed by any caller of `init_poolmanager()`.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -99,6 +99,7 @@ class HTTPAdapter(BaseAdapter):
         which we retry a request, import urllib3's ``Retry`` class and pass
         that instead.
     :param pool_block: Whether the connection pool should block for connections.
+    :param pool_kwargs: Extra keyword arguments used to initialize the urllib3 Pool Manager.
 
     Usage::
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This project isn't accepting new features. You can easily accomplish this by calling `init_poolmanager()` after creating your `HTTPAdapter` instance. It will replace the previous poolmanger with your newly configured poolmanager. Is it extra work? Yes. Does it get you where you need this pull request to merge to achieve your deepest darkest pool configuration dreams? No.

nateprewitt:
I'm inclined to agree with Sigmavirus24 on this one. This is increasing surface area on a more or less frozen API for the adapters and is pretty trivial to configure with `init_poolmanager`. I think we'll pass on this, but thanks for the proposal!

PR #5491, opened 2020-06-13
Prefer Transfer-Encoding=chunked even if both Transfer-Encoding and Content-Length are set

ikalnytskyi:
RFC 7230 says (see section [3.3.3 Message Body Length](https://tools.ietf.org/html/rfc7230#page-32)):

> If a message is received with both a `Transfer-Encoding` and a `Content-Length` header field, the `Transfer-Encoding` overrides the `Content-Length`.

Since HTTP servers MUST ignore `Content-Length` if `Transfer-Encoding` is sent, **python-requests** MUST follow the same logic and send payload using `chunked` approach if both are detected in among request headers.

Normally, `Content-Length` MUST NOT be sent in the same request with `Trasnfer-Encoding`, however, it may happened both intentionally (e.g. malicious intent) and accidentally (e.g. proxying a request downstream).

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -417,7 +417,7 @@ class HTTPAdapter(BaseAdapter):
         url = self.request_url(request, proxies)
         self.add_headers(request, stream=stream, timeout=timeout, verify=verify, cert=cert, proxies=proxies)
 
-        chunked = not (request.body is None or 'Content-Length' in request.headers)
+        chunked = request.body is not None and request.headers.get('Transfer-Encoding') == 'chunked'
 
         if isinstance(timeout, tuple):
             try:
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -24,6 +24,36 @@ def test_chunked_upload():
     assert r.request.headers['Transfer-Encoding'] == 'chunked'
 
 
+def test_chunked_upload_with_content_length():
+    """can safely send generators"""
+
+    close_server = threading.Event()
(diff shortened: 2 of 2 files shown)


nateprewitt:
Hi @ikalnytskyi, this was actually addressed in our 3.0.0 proposal branch with #3897. Instead of guessing, we would have explicitly prohibited the user from sending both. They need to make an explicit decision before the client will perform the request.

That said, it's unlikely this code will end up released anytime in the near future if ever. We're unable to make the change in a 2.x release due to it being breaking for existing users. In the event we ever release 3.0.0, we'll likely stick with the already merged PR. Thanks for the contribution though.

ikalnytskyi:
Hi @nateprewitt,

Thanks for stopping by and checking this.

> We're unable to make the change in a 2.x release due to it being breaking for existing users.

Can you please elaborate how it will break existing users? Whoever hit this bug is already broken. Receiving part does not understand what `requests` sends. That's how I found this in the first place: neither nginx nor uwsgi nor gunicorn are able to understand what was sent since once they see `Transfer-Encoding: chunked` they expect chunked data.

I still think something should be done. If you don't like respecting `Transfer-Encoding: chunked` supplied by user, can we consider resetting its value before sending out a request? 

> In the event we ever release 3.0.0, we'll likely stick with the already merged PR. 

I agree that explicit approach is better then *guessing*. It's just we already have *guessing* in 2.x, and the *guess* condition is wrong.

PR #5492, opened 2020-06-15
Remove what appears to be an un-needed call to self.resolve_redirects when not following redirects as part of creating a request's history.

dbaxa:


sigmavirus24:
At one point, `resolve_redirects` used to bail early if we weren't following redirects but I guess that changed. Now if there's a redirect to follow, we make the request but we don't use that response which is pretty bad.

Good catch. @nateprewitt @sethmlarson this PR looks good

dbaxa:
@sigmavirus24 let me know if you want me to fix up the source branch name and or  commit message :-) .

sigmavirus24:
I'm not a maintainer so not my call. 🙂

PR #5497, opened 2020-06-16
Call url.lower() once inside get_adapter

dbaxa:


nateprewitt:
Hi @dbaxa,

edit: Redacting my first statement, I read over the PR too quickly. I think this is a semi-reasonable optimization, but it's probably not entirely necessary given the number of adapters typically loaded. Given this isn't a high touch point, I'm not sure it's necessary right now. Thanks for the suggestion though!

dbaxa:
@nateprewitt can we apply this PR then ...?

ghost:
Re: [psf/requests] Call url.lower() once inside get_adapter (#5497)

PR #5509, opened 2020-06-20
Docs: list CURL_CA_BUNDLE as supported

bbodenmiller:
https://github.com/psf/requests/blob/2b3436e0e7831676044b57f6f2cc9eb7c188293e/requests/sessions.py#L707 supports `CURL_CA_BUNDLE` too so lets list it in docs.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -237,7 +237,8 @@ or persistent::
 .. note:: If ``verify`` is set to a path to a directory, the directory must have been processed using
   the c_rehash utility supplied with OpenSSL.
 
-This list of trusted CAs can also be specified through the ``REQUESTS_CA_BUNDLE`` or ``CURL_CA_BUNDLE`` environment variables.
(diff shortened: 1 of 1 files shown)


docs/user/advanced.rst
```diff
@@ -237,7 +237,7 @@ or persistent::
 .. note:: If ``verify`` is set to a path to a directory, the directory must have been processed using
   the c_rehash utility supplied with OpenSSL.
 
-This list of trusted CAs can also be specified through the ``REQUESTS_CA_BUNDLE`` environment variable.
+This list of trusted CAs can also be specified through the ``REQUESTS_CA_BUNDLE`` or ``CURL_CA_BUNDLE`` environment variables.
```
nateprewitt: This should probably be clearer that `CURL_CA_BUNDLE` is the fallback to `REQUESTS_CA_BUNDLE`. Could we add a few extra words or a small sentence to that effect?

bbodenmiller: @nateprewitt made an attempt. LMK if you have better idea.

PR #5511, opened 2020-06-22
connection_from_url already parses and handles lower casing scheme, it also lower cases the host, for us so there is no need to perform mutations on a url when a proxy is not in use inside of HTTPAdapter.get_connection.

dbaxa:


requests/adapters.py
```diff
@@ -309,9 +309,7 @@ def get_connection(self, url, proxies=None):
             proxy_manager = self.proxy_manager_for(proxy)
             conn = proxy_manager.connection_from_url(url)
         else:
-            # Only scheme should be lower case
-            parsed = urlparse(url)
-            url = parsed.geturl()
+            # connection_from_url handles case conversions.
```
dbaxa, line 312: As a result of this change the code is arguably more consistent with the `proxy` code path.

dbaxa:
cc @sigmavirus24

ghost:
Re: [psf/requests] connection_from_url already parses and handles lower
casing scheme, it also lower cases the host, for us so there is no need to
perform mutations on a url when a proxy is not in use inside of
HTTPAdapter.get_connection. (#5511)

---------- Forwarded message ---------
From: David Black ***@***.***>
Date: Thu, Jun 17, 2021, 4:50 PM
Subject: Re: [psf/requests] connection_from_url already parses and handles
lower casing scheme, it also lower cases the host, for us so there is no
need to perform mutations on a url when a proxy is not in use inside of
HTTPAdapter.get_connection. (#5511)
To: psf/requests ***@***.***>
Cc: Subscribed ***@***.***>


cc @sigmavirus24 <https://github.com/sigmavirus24>

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/psf/requests/pull/5511#issuecomment-863631506>, or
unsubscribe

dbaxa:
cc @sigmavirus24

nateprewitt:
This change has reasonable potential to cause issues for users subclassing this adapter. I'm hesitant to make a change like this because it doesn't provide clear value that outweighs the risk.

PR #5514, opened 2020-06-23
add handle for qop="auth-int" in auth.py

maxmokry:
add handle for qop="auth-int"


nateprewitt:
I think this was covered originally in https://github.com/psf/requests/issues/2408. We don't intend to support auth-int at this time. Resolving as a non-feature for now.

PR #5520, opened 2020-06-26
Removing extra documentation parameter json when undefined

tuxskar:
On the functions `put` and `patch` the parameter `json` is not defined as function parameter but it is on the docstring.
This commit is to remove the extra lines from the docstring. 
The json parameter can be passed on the kwargs but never as positional parameter as it would be following the current docstring.

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -125,7 +125,6 @@ def put(url, data=None, **kwargs):
     :param url: URL for the new :class:`Request` object.
     :param data: (optional) Dictionary, list of tuples, bytes, or file-like
         object to send in the body of the :class:`Request`.
-    :param json: (optional) json data to send in the body of the :class:`Request`.
     :param \*\*kwargs: Optional arguments that ``request`` takes.
     :return: :class:`Response <Response>` object
     :rtype: requests.Response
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @tuxskar,

Thanks for the contribution! This is a reversion of #3666 where these were specifically added. They're intended to make the docstrings consistent and clarify that all of the APIs support the json parameter. I think we'll leave it in place for now.

tuxskar:
I see, thanks for the response. 

If the idea is to have consistency with the `post` method it could be added the `json` parameter on `patch` and `put` then, to avoid having calls like:
`requests.patch('<url>', None, 'jsondata')`
raising errors but in reality are documented that could be used like this.

nateprewitt:
@tuxskar, just to clarify, `requests.patch('<url>', json=jsondata)` works as intended and that's what's documented here. If there's some confusion about how keyword args work, there's a number of more general [resources](https://realpython.com/python-kwargs-and-args/) detailing them.

tuxskar:
@nateprewitt Thanks for the answer.

The example `requests.patch('<url>', json=jsondata)`works as intended because the key `json` is added as a key of kwargs when making the call to the function but on the docstring says that there are 3 paramaters defined using `:param` which means that json can be used passed as key-value (and will be added to kwargs) and **as a positional argument** (added to the args if it is defined, or using the positional arguments).

Because it is defined as a positional argument on the docstring it would allow to do calls like:
`requests.patch('<url>', None, 'jsondata')`
But in reality the 3rd parameter is not defined on the function and this call will raise an error.

In order to have the json parameter properly defined and not been used as other kwarg argument it must be added as a parameter of the function like:
`def patch(url, data=None, json=None, **kwargs):`
And then the calls:
`requests.patch('<url>', None, 'jsondata')`
and
`requests.patch('<url>', json='jsondata')`
Will work properly with no errors and it will be aligned with the docstring.

Issue #4740, opened 2018-07-17
Cant't get correct response (post multipart/form-data including file)


wangyujieitcast:
URL=“http://recognition.image.myqcloud.com/ocr/general”

When I used postman(Uploading screenshot file failed multiple times):
headers
    Authorization:my_key
form-data:
    appid:1257121697
    image:local_file(File Type)
It's working! But when I used requests:
headers = {
    "Host": 'recognition.image.myqcloud.com',
    "Content-type": 'multipart/form-data',
    "Authorization": my_key ,
}

data = {
    "appid": "1257121697",
}

files = {'image': (img_file, open(img_file, 'rb'))}

response = requests.post(URL, data=data, headers=headers, files=files)
response status_code is 400.

Issue #4589, opened 2018-04-14
Making a request with multipart/form-data (file=) should throw an error if the barrier defined in the headers does not match


SM-Wistful:
I am unable to make a pull request at this time, but I think this should be added. The server will not understand the request if the content-type header's barrier= is not the same as the one generated by requests. At the very least there should be a warning.

sigmavirus24:
Please provide us with the output of

```
python -m requests.help
```

If that is unavailable on your version of Requests please provide some basic information about your system (Python version, operating system, etc).

Additionally, this appears to be a duplicate of several issues. In the future, please search **closed and** open issues before creating new ones that are duplicates.

Issue #4554, opened 2018-03-25
Custom multipart/form-data Content-Type header silently discards file upload


wlerin:
Summary.

## Expected Result

```
{
  "args": {},
  "data": "",
  "files": {
    "logo": "data:image/jpeg;base64,<base64 image data>"
  },
  "form": {},
  "headers": {
    "Accept": "application/json",
    "Accept-Encoding": "gzip, deflate",
    "Connection": "close",
    "Content-Length": "154247",
    "Content-Type": "multipart/form-data; boundary=------------------------<some boundary>",
    "Host": "httpbin.org",
    "User-Agent": "python-requests/2.18.4"
  },
  "json": null,
  "url": "https://httpbin.org/post"
}
```

## Actual Result
```
{
  "args": {},
  "data": "",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "application/json",
    "Accept-Encoding": "gzip, deflate",
    "Connection": "close",
    "Content-Length": "154247",
    "Content-Type": "multipart/form-data",
    "Host": "httpbin.org",
    "User-Agent": "python-requests/2.18.4"
  },
  "json": null,
  "url": "https://httpbin.org/post"
}
```

## Reproduction Steps

```python
import requests

headers = {
  'Content-Type': 'multipart/form-data',
  'Accept': 'application/json'
}

file = open('path_to_file.jpg', 'rb')
files = {"logo": file}

endpoint = "https://httpbin.org/post"
r = requests.post(endpoint, files=files, headers=headers)
print(r.json())
```

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.6.0"
  },
  "platform": {
    "release": "4.4.0-103-generic",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.18.4"
  },
  "system_ssl": {
    "version": "1000207f"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": false
}
```
---------------------------

[...]

Issue #3744, opened 2016-12-02
python requests post mp3 and data error


467754239:
#### coding

```python
url = ""
data = {'uuid':'-jx-1', 'alarmType':1, 'timeDuration':10}
headers = {'Content-type': 'multipart/form-data'}

mp3_fd = open('/home/ec2-user/voice/123.mp3', 'rb')
files = {'messageFile': mp3_fd}

req = requests.post(url, files=files, json=data, headers=headers)
print req.status_code
print req.content
```

### error
```
500
<html><head><title>Apache Tomcat/6.0.43 - Error report</title><style><!--H1 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:22px;} H2 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:16px;} H3 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:14px;} BODY {font-family:Tahoma,Arial,sans-serif;color:black;background-color:white;} B {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;} P {font-family:Tahoma,Arial,sans-serif;background:white;color:black;font-size:12px;}A {color : black;}A.name {color : black;}HR {color : #525D76;}--></style> </head><body><h1>HTTP Status 500 - Request processing failed; nested exception is org.springframework.web.multipart.MultipartException: Could not parse multipart servlet request; nested exception is org.apache.commons.fileupload.FileUploadException: the request was rejected because no multipart boundary was found</h1><HR size="1" noshade="noshade"><p><b>type</b> Exception report</p><p><b>message</b> <u>Request processing failed; nested exception is org.springframework.web.multipart.MultipartException: Could not parse multipart servlet request; nested exception is org.apache.commons.fileupload.FileUploadException: the request was rejected because no multipart boundary was found</u></p><p><b>description</b> <u>The server encountered an internal error that prevented it from fulfilling this request.</u></p><p><b>exception</b> <pre>org.springframework.web.util.NestedServletException: Request processing failed; nested exception is org.springframework.web.multipart.MultipartException: Could not parse multipart servlet request; nested exception is org.apache.commons.fileupload.FileUploadException: the request was rejected because no multipart boundary was found
	org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:979)
[...]

PR #5521, opened 2020-06-29
Support for multipart/form-data boundary

virajkanwade:
If custom boundary is set in content-type header, use it.

Relates to issues:
#1997
#3744
#621
#4740
#4589
#4554

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -34,7 +34,8 @@ from ._internal_utils import to_native_string, unicode_is_ascii
 from .utils import (
     guess_filename, get_auth_from_url, requote_uri,
     stream_decode_response_unicode, to_key_val_list, parse_header_links,
-    iter_slices, guess_json_utf, super_len, check_header_validity)
+    iter_slices, guess_json_utf, super_len, check_header_validity,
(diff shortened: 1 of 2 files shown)


mavwolverine:
Shouldn't need adding another library for something which could easily be achieved in the base lib

sigmavirus24:
The library is in feature-freeze and this is the worst way to implement this. So many users copy headers out of their Chrome inspector without knowing what they mean and this has wildly undefined behaviour with the mix of other parameters a user might actually pass.

You're also breaking the guarantee of ordering with this PR for something that a third-party library does _better_.

mavwolverine:
@sigmavirus24 
> this has wildly undefined behaviour with the mix of other parameters a user might actually pass.

- it is utilizing a function very similar to a function which already exists in the lib: get_encoding_from_headers
- it is looking for specifically only 'boundary' param passed in the content-type header, nothing else.

Can you please help me understand why would this have wildly undefined behavior?

nateprewitt:
I don't believe we intend to accept a PR for this at this time, resolving as it's currently unactionable.

PR #5533, opened 2020-07-15
Update Python Version in Shell Example in README

wyz23x2:
Python 3.8.4 is now released, but the version in README is still 3.7.4. Let's update it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
     
     <div align="left">
     <p></p>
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Why? This isn't necessary to illustrate the code in the console

nateprewitt:
Hi @wyz23x2, thanks for contributing to Requests! Unfortunately, we don't usually accept cosmetic changes to documentation and I'm not sure this provides new info for the reader. I think we'll pass on this for now.

PR #5542, opened 2020-07-29
Fix documentation for proxies, missing an "s" for https

jonathan-s:
It looks like the documentation regarding how proxies should be defined is using the wrong scheme for https. I assume that this is a mistake, otherwise it would be good to explain why this is the case in the documentation.


sethmlarson:
This is intentional. Requests doesn't support HTTPS proxies currently.

PR #5546, opened 2020-08-02
Fix Reitz's GitHub handle in docs

jack1142:
Currently it links to some account created in April 21 2020, so I'm taking a wild guess that it isn't intentional.


nateprewitt:
Hi @jack1142, thanks for contributing! We actually already have a branch up with this change that will get merged in the near future. I think we'll keep the doc changes consolidated to there for now.

PR #5584, opened 2020-09-11
Update kens github

Visgean:
His website points to https://github.com/ken-reitz, kennethreitz seems to be trolling account (there is Islamophobic profile picture )

image


sethmlarson (approved):
Thanks!

PR #5594, opened 2020-09-23
Update advanced.rst

icfly2:
replaced reference to requests-async with httpx the replacement project


nateprewitt (approved):
This looks good, thanks @icfly2!

PR #5595, opened 2020-09-24
removed unused import from setup.py

vivekvashist:
Removed import re which looks like its unused in setup.py


recepsirin (approved):

recepsirin (approved):

dimbleby:
#5780

nateprewitt:
Yep, looks like this got lost in the shuffle. Apologies @vivekvashist and thanks for the contribution. Resolving since this is no longer needed.

PR #5611, opened 2020-10-01
Create codeql-analysis.yml

sigmavirus24:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
```
sethmlarson: Can we hard-code `python` here and not use the matrix feature?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master, proposed/3.0.0]
```
sethmlarson: Is there value in checking the v3 branch if it's likely not going to be released?

sethmlarson (changes requested):
Interesting service! Some of the comments in the yml are a little much, up to you if you want to prune them down a bit.

Also, if this makes security vulnerabilities public have we tried running it locally and fixed issues before making this automated?

sigmavirus24:
To all of your questions: I don't know. I saw the announcement and thought it wouldn't hurt here. I'll look more closely later

.github/workflows/codeql-analysis.yml
sigmavirus24: Better question: Is there value in keeping it around if it's not going to be released?

.github/workflows/codeql-analysis.yml
sigmavirus24: Done

.github/workflows/codeql-analysis.yml
sigmavirus24: Removed it in the second commit as well

nateprewitt: I think there's a lot of valuable work on the proposed/3.0.0 branch that *could* be released at some point. I was keeping it up to date until we had some of the crazier commit history changes in 2018. I think we can get it back in sync but it's going to take a pretty tedious merge. Alternatively, we could hand pick the relevant feature commits, but that's also a big time investment.

That said, I don't think there's any value in checking the current proposal branch here.

sigmavirus24: Frankly, I think there's a lot of internal refactoring that could make a 3.0 easier to release and maintain long-term. It's something I've thought about for over a year

sethmlarson (approved):
LGTM

nateprewitt (approved):

PR #5643, opened 2020-10-30
Respect the NETRC environment variable

tarmath:
Resurecting https://github.com/psf/requests/pull/4339 with incorporated review/feedback

Fixes https://github.com/psf/requests/issues/4318

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -169,14 +169,20 @@ def super_len(o):
 def get_netrc_auth(url, raise_errors=False):
     """Returns the Requests tuple auth for a given url from netrc."""
 
+    netrc_file = os.environ.get('NETRC')
+    if netrc_file is not None:
+        netrc_locations = (netrc_file,)
+    else:
+        netrc_locations = ('~/{}'.format(f) for f in NETRC_FILES)
+
     try:
         from netrc import netrc, NetrcParseError
 
         netrc_path = None
 
-        for f in NETRC_FILES:
+        for f in netrc_locations:
             try:
-                loc = os.path.expanduser('~/{}'.format(f))
+                loc = os.path.expanduser(f)
             except KeyError:
                 # os.path.expanduser can fail when $HOME is undefined and
                 # getpwuid fails. See https://bugs.python.org/issue20164 &
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -213,7 +213,8 @@ Note: Custom headers are given less precedence than more specific sources of inf
 
 * Authorization headers set with `headers=` will be overridden if credentials
   are specified in ``.netrc``, which in turn will be overridden by the  ``auth=``
-  parameter.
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -13,6 +13,8 @@ requests.egg-info/
 *.pyc
 *.swp
 *.egg
+.eggs/
+.tox/
 env/
 .venv/
 .eggs/
```
sigmavirus24: Why are you duplicating `.eggs` here? And `.tox` below?

tarmath: ah jeez, let me fix that...

tarmath: hrm, I amended the commit and force-pushed on my fork, but this change is still showing up...

tarmath: ok here goes.

sigmavirus24: GitHub does that to avoid confusing folks with "I left feedback here! They force-pushed and now it's disappeared!"

sigmavirus24 (dismissed):

requests/utils.py
```diff
@@ -169,12 +169,15 @@ def super_len(o):
...
+
     try:
         from netrc import netrc, NetrcParseError
 
         netrc_path = None
 
-        for f in NETRC_FILES:
+        for f in netrc_locations:
             try:
                 loc = os.path.expanduser('~/{}'.format(f))
```
nateprewitt: This doesn't look quite right. Are all `NETRC` paths relative to the users home directory? I think we only want this formatting for the `NETRC_FILES`.

nateprewitt (commented):
Yeah, I think I'm fine with adding support. Left a quick question about the current implementation.

requests/utils.py
nateprewitt: Actually, it looks like the last PR has [similar feedback](https://github.com/psf/requests/pull/4339#discussion_r144723124). If we can get that or something similar incorporated, I think we're set.

tarmath: very true... how's this then?

tarmath: I can also do it exactly as suggested in previous PR... let me know

requests/utils.py
```diff
@@ -35,7 +35,7 @@
 from .exceptions import (
     InvalidURL, InvalidHeader, FileModeWarning, UnrewindableBodyError)
 
-NETRC_FILES = ('.netrc', '_netrc')
+NETRC_FILES = ('~/.netrc', '~/_netrc')
```
nateprewitt: I'm a little hesitant to mutate this directly with how long it's been around. Let's keep the formatting inside `get_netrc_auth` if possible.

tarmath: ah makes total sense, wouldn't have thought of that... changing then!

tarmath: done.

sigmavirus24 (approved):

nateprewitt (approved):
Great, this looks good to me. Thanks @tarmath!

rgevaert:
@tarmath thanks for this!

PR #5649, opened 2020-11-05
Document how to turn off Keep-Alive in 2.x

### LINKED ISSUE #5646: Unable to turn off Keep-Alive
Documentation doesn't specify how to turn off Keep-Alive
https://requests.readthedocs.io/en/latest/user/advanced/#keep-alive

abitrolly:
Ref. #5646

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -344,6 +344,11 @@ Note that connections are only released back to the pool for reuse once all body
 data has been read; be sure to either set ``stream`` to ``False`` or read the
 ``content`` property of the ``Response`` object.
 
+To turn off Keep-Alive, set ``Connection`` header to ``close``::
+
+    s = requests.Session()
+    s.headers['Connection'] = 'close'
+
 .. _streaming-uploads:
 
 Streaming Uploads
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This documentation isn't entirely correct and that issue is already clsoed

abitrolly:
@sigmavirus24 I don't know how to improve it. What is not correct here?

abitrolly:
I edited the description to mention issue without closing it (for me).

sigmavirus24:
> What is not correct here?

It's not the only way to disable keep-alive. Sessions are not the only way people use this and the solution can be more generic without being specific because people who use this library generally understand the concept of headers and what the Keep-Alive setting in the Connection header is, which is why this doesn't need to be documented as desperately as you think

abitrolly:
@sigmavirus24 it is needs to be documented, because in the past there were multiple ways to do this, which don't work now https://stackoverflow.com/questions/21008953/python-requests-1-how-to-disable-keep-alive

abitrolly:
Reworded and moved to the paragraph about sessions.

sigmavirus24:
> @sigmavirus24 it is needs to be documented, because in the past there were multiple ways to do this, which don't work now https://stackoverflow.com/questions/21008953/python-requests-1-how-to-disable-keep-alive

Information will always be outdated on the internet. Your initial commit was wrong immediately and is still overly specific to _your_ usecase. Beyond which, you seem to have (potentially deliberately) misunderstood my characterization of the urgency with the necessity

nateprewitt:
I'm going to resolve this as there hasn't been any movement in the last year.

abitrolly:
Killing keep alive literally.

PR #5652, opened 2020-11-10
Move CI to Github Actions

nateprewitt:
Moving us to GH Actions since our Travis configs are in relative disarray and we haven't been running Mac/Windows builds for years at this point.

### Other changes in this PR:

**Pipenv**
Pipenv has caused a lot of friction with users trying to submit PR updates that need dependency changes, so this removes it. I'm thinking we can leave the Pipfile in case anyone is using it but may remove it in a future release. I've readded a `requirements-dev.txt` to store dev dependencies.

**Removal of pytest-xdist**
`pytest-xdist` causes threading issues on macOS for some versions of Python [StackOverflow issue](https://stackoverflow.com/questions/50168647/multiprocessing-causes-python-to-crash-and-gives-an-error-may-have-been-in-progr/52230415). The performance gain on tests is negligible, so I've removed it for config simplicity. If there's strong demand, we could attempt the bash updates or split out the macOS test command in the action definition.

**Adding Python 3.9**
Adding 3.9 to tests so we can update our trove classifiers before releasing 2.25.


nateprewitt:
TODO: We'll need to remove the Travis integration in Settings after this is merged.

Makefile
```diff
@@ -1,21 +1,21 @@
...
-	pip install pipenv --upgrade
-	pipenv install --dev
+	pip install -e .[socks]
+	pip install -r requirements-dev.txt
 test:
 	# This runs all of the tests, on both Python 2 and Python 3.
 	detox
 ci:
-	pipenv run py.test -n 8 --boxed --junitxml=report.xml
+	pytest tests --junitxml=report.xml
```
sigmavirus24, line 9: Since we're (rightfully) dropping pipenv, should we get rid of the rest of its cruft (Pipenv, Pipenv.lock)?

sigmavirus24:
> TODO: We'll need to remove the Travis integration in Settings after this is merged.

I used to be able to help with that but no longer

Makefile
nateprewitt: ~I'd been thinking we'd do that in the next minor release. Pipenv has \~1M downloads a month, but I have no idea if anyone even uses the Requests' Pipfile. I'm not tied to waiting if you or Seth feel strongly.~
I tested pipenv when we remove the Pipfile and it actually degrades pretty gracefully. Let's just remove it now.

nateprewitt:
> I used to be able to help with that but no longer

It looks like it's just a toggle in settings we can shut off/remove after merge. The maintainers group should have updated permissions now so we can do that when we're ready.

sigmavirus24 (approved):

sethmlarson (approved):
Woo, excited to see the Pipfile go, that means we don't have to maintain our own dependencies on the urllib3 downstream tests!

PR #5658, opened 2020-11-15
Improve the Session.request() documentation

pzahemszky:
This PR improves the documentation of the `Session.request()` method. Its `stream` parameter's description used to have remnants of the old `prefetch` parameter (from PR #1283), which are now rephrased slightly.

The new sentence is the same as that in the [Body Content Worklflow](https://requests.readthedocs.io/en/latest/user/advanced/#body-content-workflow) section of the documentation.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -498,8 +498,9 @@ class Session(SessionRedirectMixin):
         :type allow_redirects: bool
         :param proxies: (optional) Dictionary mapping protocol or protocol and
             hostname to the URL of the proxy.
-        :param stream: (optional) whether to immediately download the response
-            content. Defaults to ``False``.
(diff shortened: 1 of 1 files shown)


pzahemszky:
This is my first PR in this repo, so please let me know if I missed something.

abitrolly:
Repeating documentation doesn't clarify it enough. I would say "open a stream to iterate over content manually"

nateprewitt:
Thanks for the PR, @pzahemszky. I believe we'll pass on this as it unfortunately doesn't improve clarity and lacks some of the nuance from the [full docs](https://docs.python-requests.org/en/master/user/advanced/#body-content-workflow). To keep doc strings succinct, we'll keep the current wording.

Issue #5662, opened 2020-11-17
GHA: Tests fail with ubuntu-20.04


hugovk:
GitHub Actions fails when using the Ubuntu 20.04 image, and pass with 18.04.

## Expected Result

Tests pass on GitHub Actions with `ubuntu-20.04`

## Actual Result

Tests fail on GitHub Actions with `ubuntu-20.04` for CPython, pass on PyPy.

## Reproduction Steps

I noticed this annotation on [this recent build](https://github.com/hugovk/requests/actions/runs/368989395):

> Ubuntu-latest workflows will use Ubuntu-20.04 soon. For more details, see https://github.com/actions/virtual-environments/issues/1816

I ran a quick test with explicit `ubuntu-20.04` instead of `ubuntu-latest`:

* https://github.com/hugovk/requests/commit/43e1c9ee709d2efd18c9294b941babd816193f3d
* https://github.com/hugovk/requests/actions/runs/369014648

I've not looked into the failures in detail, one is `requests/adapters.py:514: SSLError`.

The linked https://github.com/actions/virtual-environments/issues/1816 says:

> **This change will be rolled out over a period of several weeks beginning on November, 30.** (December, 22 - January, 4 will be a migration break due to holidays)

It's also possible to explicitly pin to `ubuntu-18.04`, which passes:

* https://github.com/hugovk/requests/commit/69e5c04372eb6f6dc2d14bb8617d47df1ba1c199
* https://github.com/hugovk/requests/actions/runs/369033603

nateprewitt:
Good catch, @hugovk! It looks like our local httpbin server for tests is having trouble either with the newer version of openssl or configuration being vended in 20.04. I think the best bet is to pin us to ubuntu-18.04 for now. We'll gladly accept a PR for that or I'll work on getting it resolved by the end of the week.

PR #5663, opened 2020-11-18
Pin GHA to Ubuntu 18.04

hugovk:
Fixes https://github.com/psf/requests/issues/5662. 

`ubuntu-latest` points to 18.04 and will soon flip to 20.04.

However tests currently fail on 20.04, so let's explicitly pin to 18.04 for now.


nateprewitt (approved):
Thanks, @hugovk!

PR #5128, opened 2019-07-02
Use urllib for chunked requests

lmvlmv:
Resubmit of [https://github.com/kennethreitz/requests/pull/4958](https://github.com/kennethreitz/requests/pull/4958 ) with up to date merge. Also addresses https://github.com/kennethreitz/requests/pull/4179. 

Specifically, in my use case, chunked requests when a proxy is in use results in failure to evaluate the proxy configuration and the request being sent direct. It appears that urllib3 supports chunking as an option so this fork to lower level urllib3 code seems to no longer be required.

PR #4179, opened 2017-06-27
Fix chunked encoding with SSL via proxy

jamespic:
This is a patch applying @bpitman's proposed fix for #3844.

Assuming a proxy is running on port 8875 (`docker run --rm --name='tinyproxy' -p 8875:8888 dannydirect/tinyproxy:latest ANY` works for me), the issue can be reproduced with

```
requests.put('https://www.google.com', data=iter('Hello World'), proxies={'https': 'http://localhost:8875'})
```

Issue #4392, opened 2017-11-20
Streamed request with preset Host header sends two Host headers


mjpieters:
The following code sends **two** host headers:

```python
import requests

def gen(): yield b'binary data'

requests.post(
    'http://localhost:3000/', 
    data=gen(),
    headers={'Host': 'example.com'}
)
```

(I tested with `nc -vvl 3000` on OS X).

Headers received:

```
POST / HTTP/1.1
Host: localhost:3000
User-Agent: python-requests/2.18.1
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Host: example.com
Transfer-Encoding: chunked
```

This is because when chunking a request, we don't tell the `httplib` / `http.client` library to skip the host header if one is already set. From the [`adapters.HTTPAdapter.send()` implementation](https://github.com/requests/requests/blob/9a8a826f226e6973d72914b4a4fc806e0b5036f4/requests/adapters.py#L456-L458):

```python
                    low_conn.putrequest(request.method,
                                        url,
                                        skip_accept_encoding=True)
```

which leaves the [`skip_host` parameter](https://docs.python.org/3/library/http.client.html#http.client.HTTPConnection.putrequest) set to `False`. It probably should use `skip_host='host' in request.headers` here.

## System Information

```
>>> requests.help.main()
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": "2.0.3"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.6.3"
  },
  "platform": {
    "release": "16.7.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "1010006f",
    "version": "17.3.0"
  },
  "requests": {
    "version": "2.18.1"
  },
  "system_ssl": {
    "version": "100020cf"
  },
  "urllib3": {
    "version": "1.22"
  },
  "using_pyopenssl": true
}
```

nateprewitt:
Yep, this looks like a miss from #3637. Since we're not using urllib3's chunked request interface, we never got the change integrated in Requests.

The quick path here is to carry the host dedupe logic from urrlib3  into that try block. The better but likely more involved solution would be to try to use urllib3's chunked logic in Requests. This is something we've discussed before but probably out of scope for this issue.

@requests/core I'm assuming the former option is the advisable one, but do you have different thoughts?

Lukasa:
[...]

PR #5664, opened 2020-11-18
Use urllib for chunked requests (Resubmit)

lmvlmv:
This is a resubmit of Use urllib for chunked requests #5128 which was merged and then pulled without warning.

"Resubmit of #4958 with up to date merge. Also addresses #4179.

Specifically, in my use case, chunked requests when a proxy is in use results in failure to evaluate the proxy configuration and the request being sent direct. It appears that urllib3 supports chunking as an option so this fork to lower level urllib3 code seems to no longer be required."

[](https://github.com/psf/requests/pull/5128)

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -456,6 +456,10 @@ class HTTPAdapter(BaseAdapter):
 
                 low_conn = conn._get_conn(timeout=DEFAULT_POOL_TIMEOUT)
 
+                if hasattr(conn, 'proxy'):
+                    if conn.proxy is not None and not getattr(low_conn, 'sock', None):
+                        conn._prepare_proxy(low_conn)
+
                 try:
                     low_conn.putrequest(request.method,
                                         url,
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -435,68 +435,19 @@ class HTTPAdapter(BaseAdapter):
             timeout = TimeoutSauce(connect=timeout, read=timeout)
 
         try:
-            if not chunked:
-                resp = conn.urlopen(
-                    method=request.method,
-                    url=url,
-                    body=request.body,
-                    headers=request.headers,
-                    redirect=False,
-                    assert_same_host=False,
-                    preload_content=False,
-                    decode_content=False,
-                    retries=self.max_retries,
-                    timeout=timeout
-                )
-
-            # Send the request.
-            else:
-                if hasattr(conn, 'proxy_pool'):
-                    conn = conn.proxy_pool
-
-                low_conn = conn._get_conn(timeout=DEFAULT_POOL_TIMEOUT)
-
-                if hasattr(conn, 'proxy'):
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @lmvlmv,

Thanks for reraising the issue. We're aware there's still demand for simplifying this. The reason the first PR was reverted was a lack of review, and failing to meet the original criteria we'd set to make this change.

When we first discussed this in #4392, we'd wanted a pretty rigorous testing to ensure we're not introducing regressions. We have the start of some of that but we'll need to expand it more before we're comfortable merging.

jbgomond:
Thanks a lot :)

dmazin:
> Hi @lmvlmv,
> 
> Thanks for reraising the issue. We're aware there's still demand for simplifying this. The reason the first PR was reverted was a lack of review, and failing to meet the original criteria we'd set to make this change.
> 
> When we first discussed this in #4392, we'd wanted a pretty rigorous testing to ensure we're not introducing regressions. We have the start of some of that but we'll need to expand it more before we're comfortable merging.

@nateprewitt Is anyone currently working on said unit tests? I can take a crack at them otherwise.

theGOTOguy:
I know this PR is quite stale by now, but it would be really nice if requests would finally use urllib3's built-in Retry functionality for chunked requests.

However lacking the review of the original PR might have been, thousands of users of requests are independently inventing their own way to deal with this type of error:  either dealing with the sporadic failure in another part of their system or rolling their own method of retrying, potentially requiring additional code paths that themselves have a maintenance and testing cost.  I would argue that not having this solution in requests is doing more damage to code health downstream than accepting it until such a time that a regression is reported.

nateprewitt:
Hi @theGOTOguy, I think we all agree this would be a useful enhancement to Requests. The reason the original PR was reverted before release is the change wasn't validated in any way. I wrote up a number of tests for this previously to try to get this into the 2.26 release but switching this functionality in a non-major version will be breaking. We have exceptions that change, the actual chunking is slightly different, and there are likely other subtle changes that weren't caught in the initial check.

This is towards the top of the list of easy improvement we can make to Requests, but it's currently stuck with a dozen other changes that were originally slated for a "3.0" release. Given 3.0 is unlikely to see fruition, we have a larger issue to solve of how/if we introduce breaking changes into Requests, before we can consider merging this. We've briefly discussed following pip into a CalVer style system to allow for more regular breakages, but there are pretty limited resources to get that going right now.

sigmavirus24:
Also we only ever get "Shame on you for not having done this sooner with the ~5m of combined time the two of you have in any give month" or "I'd like to help!" with 0 follow-up. And then we spend most of that ~5m explaining (once again) that we want to make this change but need time because people don't engage properly with the project

shadycuz:
@nateprewitt @lmvlmv This PR can be closed. It's been updated and merged in #6226.

PR #5665, opened 2020-11-19
Add support for HTTPS proxies.

jalopezsilva:
Starting with 1.26 urllib3 now supports HTTPS proxies. To enable the support two changes are needed:

* An additional proxy_kwargs argument that can be passed from client config. This dictionary will be used to pass any arguments needed to the underlying httpsession.
* The use_forwarding_for_https mode requires us to send the absolute URI when enabled.

The resulting API is very similar except it takes an additional parameter when needed. An example:

```
    session = requests.Session()
    proxies = {"http": "https://proxy.example", "https": "https://proxy.example"}

    proxies_kwargs = {
        "proxy_ssl_context": proxy_ssl_context(),
    }

    response = session.get(
       "https://www.google.com", proxies=proxies, proxies_kwargs=proxies_kwargs
    )
```

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -59,7 +59,7 @@ class BaseAdapter(object):
         super(BaseAdapter, self).__init__()
 
     def send(self, request, stream=False, timeout=None, verify=True,
-             cert=None, proxies=None):
+             cert=None, proxies=None, proxies_kwargs=None):
         """Sends PreparedRequest object. Returns Response object.
 
         :param request: The :class:`PreparedRequest <PreparedRequest>` being sent.
@@ -73,6 +73,7 @@ class BaseAdapter(object):
             to a CA bundle to use
         :param cert: (optional) Any user-provided SSL certificate to be trusted.
         :param proxies: (optional) The proxies dictionary to apply to the request.
+        :param proxies_kwargs: (optional) Additional arguments to pass to urllib3 for proxies.
         """
         raise NotImplementedError
 
@@ -289,7 +290,7 @@ class HTTPAdapter(BaseAdapter):
 
         return response
 
-    def get_connection(self, url, proxies=None):
+    def get_connection(self, url, proxies=None, proxies_kwargs=None):
         """Returns a urllib3 connection for the given URL. This should not be
         called from user code, and is only exposed for use when subclassing the
         :class:`HTTPAdapter <requests.adapters.HTTPAdapter>`.
@@ -300,13 +301,16 @@ class HTTPAdapter(BaseAdapter):
         """
         proxy = select_proxy(url, proxies)
 
+        if not proxies_kwargs:
+            proxies_kwargs = {}
+
         if proxy:
             proxy = prepend_scheme_if_needed(proxy, 'http')
             proxy_url = parse_url(proxy)
             if not proxy_url.host:
                 raise InvalidProxyURL("Please check proxy URL. It is malformed"
                                       " and could be missing the host.")
-            proxy_manager = self.proxy_manager_for(proxy)
+            proxy_manager = self.proxy_manager_for(proxy, **proxies_kwargs)
             conn = proxy_manager.connection_from_url(url)
         else:
             # Only scheme should be lower case
@@ -326,7 +330,7 @@ class HTTPAdapter(BaseAdapter):
         for proxy in self.proxy_manager.values():
             proxy.clear()
 
-    def request_url(self, request, proxies):
+    def request_url(self, request, proxies, proxies_kwargs=None):
         """Obtain the url to use when making the final request.
 
         If the message is being sent through a HTTP proxy, the full URL has to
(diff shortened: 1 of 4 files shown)


requests/sessions.py
```diff
@@ -470,7 +470,8 @@ def prepare_request(self, request):
     def request(self, method, url,
             params=None, data=None, headers=None, cookies=None, files=None,
             auth=None, timeout=None, allow_redirects=True, proxies=None,
-            hooks=None, stream=None, verify=None, cert=None, json=None):
+                proxies_kwargs=None, hooks=None, stream=None, verify=None,
```
nateprewitt, line 473: I'm a little worried about putting this in the middle of the function signature. People *shouldn't* be passing these as positional args, but I'm sure they are.

sigmavirus24:
Requests is under feature freeze which means we will not add any additional keyword arguments to the public API like this.

jalopezsilva:
@sigmavirus24 thanks for the feedback! Is adding an additional keyword argument to the public API the only concern? I've been reviewing the options available and there's no easy way around this unless we modify the ``proxy_manager_for`` method in ``HTTPAdapter``

Could we add a proxy_kwargs attribute to HTTPAdapter? Rather than passing the ``proxy_kwargs`` dictionary through multiple top level APIs we could expose it like this:

```
s = requests.Session()
s.adapters['https://'].proxy_kwargs = { ... }
```
That minimizes some of the changes needed. Would that work?

Facebook needs these changes upstreamed so I would love it if we can find a path forward here. I can create our own ``HTTPAdapter`` internally but t it leaves the ``requests`` community without being able to benefit from the work I put in to expand the [HTTPS proxy support in urllib3.](https://github.com/urllib3/urllib3/pull/1923)

sethmlarson:
@nateprewitt Are we still planning on doing this for 2.26.0? It doesn't look like it's moved since being added to the milestone.

nateprewitt:
I popped it out of 2.26.0 so we can get a release unblocked. I'll follow up with @jalopezsilva on getting this in the next minor release.

jalopezsilva:
That would be amazing @nateprewitt. Let me know which changes are required.

requests/adapters.py
```diff
@@ -344,13 +348,20 @@ def request_url(self, request, proxies):
...
+        is_using_https_forwarding = (
+            proxies_kwargs and proxies_kwargs.get('use_forwarding_for_https', False)
+        )
+
+        is_using_https_proxy = False
         using_socks_proxy = False
         if proxy:
             proxy_scheme = urlparse(proxy).scheme.lower()
             using_socks_proxy = proxy_scheme.startswith('socks')
+            is_using_https_proxy = proxy_scheme.startswith('https')
```
nateprewitt, line 360: We support a number of possible socks* schemes which is the reason for using `startswith`. I'm wondering if we want to be more explicit here by checking for exactly `https`.

requests/adapters.py
```diff
@@ -344,13 +348,20 @@ def request_url(self, request, proxies):
...
 
         url = request.path_url
-        if is_proxied_http_request and not using_socks_proxy:
+        if (is_proxied_http_request and not using_socks_proxy or
+                (is_using_https_proxy and is_using_https_forwarding)):
```
nateprewitt, line 364: nit; this conditional is getting a kind of hard to parse at first glance. Perhaps we could define a couple variables like `is_forwarded_https_proxy` and `is_nonsocks_http_proxy` to simplify the concepts.

e.g.
```python
if is_nonsocks_http_proxy or is_forwarded_https_proxy:
```

requests/adapters.py
```diff
@@ -59,7 +59,7 @@ def __init__(self):
         super(BaseAdapter, self).__init__()
 
     def send(self, request, stream=False, timeout=None, verify=True,
-             cert=None, proxies=None):
+             cert=None, proxies=None, proxies_kwargs=None):
```
nateprewitt, line 62: This is the probably the more controversial part of this change. Requests has considered its top level APIs more-or-less frozen since 2.0. Adding this is a deviation from that philosophy.

While we should continue to uphold that stance, I don't see another way for us to expose this urllib3 functionality. Given we have some legitimately broken use cases, I think a rare exception may be warranted.

requests/sessions.py
sigmavirus24: Because we can't prevent them from passing them positionally, I agree that this should be tacked onto the end.

requests/adapters.py
sigmavirus24: Yeah, I agree this could be more simply `proxy_scheme == 'https'` but I'm wondering if there are cases I'm unaware of where someone might use `https+<somethingelse>://proxy` or `<somethingelse>+https://proxy`.  It shouldn't be too much trouble to add support for those if necessary though and on the chance those aren't easily supportable with this work, then we won't accidentally opt them in

sigmavirus24 (commented):
Yeah, I think there are a lot of public functions whose signatures are being altered here in a way that is going to cause _significant_ pain for users.

Adapters are one point of extension that many users take advantage of. Some subclass the default `HTTPAdapter` in order to customize certain things and may have added some custom logic to `get_connection` or even `request_url` (although I'd hope not).

If they are, however, we're unconditionally sending that parameter and we're not doing it as a keyword argument below in `send` which means that people then need to add `proxies_kwargs` before upgrading or they may break things desperately.

While many subclass `HTTPAdapter`, an order of magnitude more subclass `Session` and many override `request` and `send`. I'm very hesitant to accept these changes as is

PR #5673, opened 2020-11-29
updated `get_encoding_from_headers` to return utf-8 if the content type is set to application/json

### LINKED ISSUE #5667: r.text and r.json() return different results *in some cases*
This might be an interesting one... I found that `r.text` and `r.json()` _can_ return different results in some specific cases. I don't understand why the difference between the two cases is changing the result of `r.text`.

Maybe `r.text` should always default to using `utf-8` as the decoding if `application/json` is set as the response's content type, following https://www.ietf.org/rfc/rfc4627.txt (ctrl+f for `JSON text SHALL be encoded in Unicode.`).

I'm using the latest version of requests.

## Expected Result

I would expect `r.text` and `r.json()` to return ~ the same thing. More specifically, I would expect `json.loads(r.text)` and `r.json()` to return the same thing, but the issue seems to be with `r.text`'s decoding specifically.

## Actual Result

In the following code, I am making a sample request and replacing the request's response with custom content so we have full control over it. The custom content is utf-8 encoded. In the next version of this code, you'll see the `name` change  when it shouldn't.
```python
import requests
import json

r = requests.get("https://api.covidtracking.com/v1/us/current.json")
r._content = b'{"name":"rd\xce\xba"}'  # This is utf-8
r.headers = {
    "Content-Type": "application/json",
}
print(r.json())
print(json.loads(r.text))
```

The above code prints:
```
{'name': 'rdκ'}
{'name': 'rdκ'}
```
which is fantastic.

Replacing the request's content with `b'{"name":"rd\xce\xba","uuid":"1234"}'`, which simply adds a `uuid` field to the JSON, and running the code again prints:
```
{'name': 'rdκ', 'uuid': '1234'}
{'name': 'rdÎº', 'uuid': '1234'}
```

The `name` is different even though it did not change at all! The existence of `"uuid":"1234"` in the response's contents somehow changes the decoding. I have no clue why.

## Reproduction Steps

Run this code:
```python
import requests
import json

r = requests.get("https://api.covidtracking.com/v1/us/current.json")
r._content = b'{"name":"rd\xce\xba","uuid":"1234"}'
r.headers = {
    "Content-Type": "application/json",
}
print(r.json())
print(json.loads(r.text))
```

The issue should be fixed when the two print statements match... I think.

## System Information

    $ python -m requests.help

```
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": "2.7"
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.4"
  },
  "platform": {
    "release": "10",
    "system": "Windows"
  },
  "pyOpenSSL": {
    "openssl_version": "1010103f",
    "version": "19.0.0"
  },
  "requests": {
    "version": "2.25.0"
  },
  "system_ssl": {
    "version": "1010104f"
  },
  "urllib3": {
    "version": "1.24.2"
  },
  "using_pyopenssl": true
}

jjmaldonis:
This PR fixes #5667 by adding a small bit of code to `get_encoding_from_headers` that returns `utf-8` if the content type is set to `application/json` and `charset` is not set. This follows [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt) (ctrl+f for `shall be`) which reads: "JSON text SHALL be encoded in Unicode.  The default encoding is UTF-8."

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -503,6 +503,10 @@ def get_encoding_from_headers(headers):
     if 'text' in content_type:
         return 'ISO-8859-1'
 
+    if 'application/json' in content_type:
+        # Assume UTF-8 based on RFC 4627: https://www.ietf.org/rfc/rfc4627.txt since the charset was unset
+        return 'utf-8'
+
 
 def stream_decode_response_unicode(iterator, r):
     """Stream decodes a iterator."""


sigmavirus24 (approved):

jjmaldonis:
Awesome, thanks for the fast responses here and for merging!

PR #5681, opened 2020-12-08
5677: Rebuild proxies on Session#send

mateusduboli:
- Adds some test cases to highlight the issue, some of those seem excessive, I would like suggestions on how to combine them.
- Rebuilds the proxy configuration on `send`, making it behave like `requests.Session#request` that already did that.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -633,7 +633,7 @@ class Session(SessionRedirectMixin):
         kwargs.setdefault('stream', self.stream)
         kwargs.setdefault('verify', self.verify)
         kwargs.setdefault('cert', self.cert)
-        kwargs.setdefault('proxies', self.proxies)
+        kwargs.setdefault('proxies', self.rebuild_proxies(request, self.proxies))
 
         # It's possible that users might accidentally send a Request object.
         # Guard against that specific failure case.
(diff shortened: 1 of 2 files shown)


requests/sessions.py
```diff
@@ -633,7 +633,7 @@ def send(self, request, **kwargs):
         kwargs.setdefault('stream', self.stream)
         kwargs.setdefault('verify', self.verify)
         kwargs.setdefault('cert', self.cert)
-        kwargs.setdefault('proxies', self.proxies)
+        kwargs.setdefault('proxies', self.rebuild_proxies(request, self.proxies))
```
sigmavirus24, line 636: I'm 99% certain this isn't actually correct. This does work regardless of whether it's used or not. We should do:

```py
if kwargs['proxies']:
    kwargs['proxies'] = self.rebuild_proxies(request, kwargs['proxies'])
```

After this line though

mateusduboli: What do you mean by "this"?

The problem or the proposed solution?

For the problem, I've added a fell test cases, which when I remove this line change those were not passing.

As for the solution.

Should the `proxies` parameter be overwritten by the environment?

If we `rebuild_proxies` there, it will load the environment variables, instead of using what is being passed on the arguments, my solution intentionally considers the arguments as the highest priority.

mateusduboli: @sigmavirus24 

Does my answer help you understand the problem in any way?

I'm facing this problem in production and the tests do show the inconsistent use of proxy environment variables by `requests.session`.

sigmavirus24: I don't recall 😆 but your logic makes sense.

sigmavirus24 (approved):

Issue #5671, opened 2020-11-28
Urllib3 >1.26.x | Introduces default User-Agent header


prilly-dev:
With pyrequest pullrequest https://github.com/psf/requests/commit/03957eb1c2b9a1e5e6d61f5e930d7c5ed7cfe853 v2.25.0 urllib3 is updated to v1.26.0. with this version bump urllib3 has changed default behavior in regards to header "User-Agent" : https://pypi.org/project/urllib3/ 

urllib3 1.26.0
```
Added default User-Agent header to every request (Pull #1750)
Added urllib3.util.SKIP_HEADER for skipping User-Agent, Accept-Encoding, and Host headers from being automatically emitted with requests (Pull #2018)
```
Urllib3 now introduces default User-Agent header when no user-agent header is set. Urlib3 also introduces the function "urllib3.util.SKIP_HEADER" as a means to deactivate default user-agent header.

## Expected Result

Pyrequest should adopt easy means of deactivating "User-Agent" headers by utilizing the "urllib3.util.SKIP_HEADER" function

## Actual Result

Default "User-agent" header is enforced when no header= is specified.
by using headers={"User-Agent": none} no longer deactivate default "user-agent" while using headers={"User-Agent": ""} will deactivate "user-agent" header

## Reproduction Steps

```python
import requests

req = Request('POST', url=BASE_URL + endpoint, data=params).prepare()  ---> Deafult "user-agent" is enforced "User-Agent: python-urllib3/1.26.2"

req = Request('POST', url=BASE_URL + endpoint, headers={"User-Agent": None}, data=params).prepare() ---> Does not deactivate "User-Agent"

req = Request('POST', url=BASE_URL + endpoint, headers={"User-Agent": ""}, data=params).prepare()  ---> Deactivates "User-Agent" succesfully


```

## System Information

    $ python -m requests.help

```
<paste here>
```

This command is only available on Requests v2.16.4 and greater. Otherwise,
please provide some basic information about your system (Python version,
operating system, &c).

Python 3.8, Docker

nateprewitt:
Hi @prilly-dev, thanks for bringing this to our attention. I was actually surprised to see Requests wasn't emitting a header here, but it appears this is specific to the Session workflow.

We'll probably need something that is tolerant to urllib3 < 1.26 not being installed.  I'm not immediately sure if that means conditionally relying on the urllib3 SKIP_HEADER util in compat or if there are other options. We'll take a look at getting a fix together.

[...]

PR #5693, opened 2020-12-12
Pass urllib3.SKIP_HEADER when headers should be unset

sigmavirus24:
urllib3 introduced some default headers and a way to skip them if
desired. Let's use that sentinel value to pass along information about
Requests' users desire to skip those headers as well.

Closes gh-5671

---
I'm pretty sure the tests will fail to start with, but wanted to get a sketch of something up for early review


requests/models.py
```diff
@@ -447,9 +448,14 @@ def prepare_headers(self, headers):
         self.headers = CaseInsensitiveDict()
         if headers:
             for header in headers.items():
+                name, value = header
+                if value is None:
+                    if hame.lower() in {"user-agent", "accept-encoding", "host"}:
```
sethmlarson: Can use `urllib3.util.SKIPPABLE_HEADERS` here. Also typo `hame` -> `name`

sethmlarson (changes requested):
One comment, otherwise looks good

requests/models.py
sigmavirus24: Thanks for the pointer on the constant and a silly typo

requests/models.py
```diff
@@ -447,9 +448,14 @@ def prepare_headers(self, headers):
         self.headers = CaseInsensitiveDict()
         if headers:
             for header in headers.items():
+                name, value = header
+                if value is None:
+                    if name.lower() in urllib3.util.SKIPPABLE_HEADERS:
+                        value = urllib3.SKIP_HEADER
```
nateprewitt: This is going to fail for older versions of urllib3, right? What do you think about doing something in compat like:

```Python
try:
    import urllib3
    SKIP_HEADER = urllib3.SKIP_HEADER
    SKIPPABLE_HEADERS = urllib3.util.SKIPPABLE_HEADERS
except AttributeError:
    SKIP_HEADER = None
    SKIPPABLE_HEADERS = []
```

sethmlarson: Also this is giving an attribute error, should be `urllib3.util.SKIP_HEADER`

sigmavirus24: Done and done. Good call @nateprewitt

nateprewitt (commented):
I think the changes look good for what we originally discussed. There is this new oddity the failing test is pointing out though. We're now creating a PreparedRequest with these odd placeholders for "None" when inspected. That's gonna be a bit jarring from a ux standpoint, and I'm wondering if it's going to create issues for anyone relying on validating/modifying their PreparedRequest before sending.

I'm not immediately sure how we get around that. Do we need to keep a list of headers that need this value and then inject the placeholder just before calling out to urllib3?

PR #5696, opened 2020-12-19
Document that the timeout can be an int

JustAnotherArchivist:
Requests simply delegates to `urllib3.util.Timeout`, and urllib3 has of course supported ints for a very long time – at least Sept 2013 per the docstring changes, but likely much longer than that. The [timeout section in the advanced docs](https://github.com/psf/requests/blob/c2b307dbefe21177af03f9feb37181a89a799fcc/docs/user/advanced.rst#timeouts) actually has an example that uses an int, but the API docs (via function docstrings) don't mention this so far.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -65,9 +65,9 @@ class BaseAdapter(object):
         :param request: The :class:`PreparedRequest <PreparedRequest>` being sent.
         :param stream: (optional) Whether to stream the request content.
         :param timeout: (optional) How long to wait for the server to send
-            data before giving up, as a float, or a :ref:`(connect timeout,
+            data before giving up, as an int, a float, or a :ref:`(connect timeout,
             read timeout) <timeouts>` tuple.
-        :type timeout: float or tuple
+        :type timeout: int or float or tuple
         :param verify: (optional) Either a boolean, in which case it controls whether we verify
             the server's TLS certificate, or a string, in which case it must be a path
             to a CA bundle to use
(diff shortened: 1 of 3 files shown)


sigmavirus24:
`int`s are just converted to `float`s under the covers and really floats are the better thing to be using. It'd probably be better to convert the timeout section in our docs to using floats.

JustAnotherArchivist:
I disagree. Despite the internal conversion, from the perspective of the user code, the best data type to represent an integer is obviously an int, and it has always been supported by the code. There's no reason not to document this and instead give the appearance of requiring a float for such timeouts (adding unnecessary `.0` clutter to the code), especially since integer timeouts are far from rare and likely already very widely used.

There is a point to be made that technically, any object with a `__float__` method (or an `__index__` method since Python 3.8) is also supported, but that would be a quite obscure use case and is not documented in urllib3's documentation. Another little detail is that due to the conversion to a float, very large ints would not be treated accurately, but that's irrelevant for real-world usage.

sethmlarson:
Going to agree with @sigmavirus24, int is a subclass of float so is covered. Closing this PR.

JustAnotherArchivist:
@sethmlarson Uhm, at least in Python 3, `int` is definitely not a subclass of `float`. They both inherit directly from `object`. I'm pretty sure this was true on Python 2 as well, but I've long purged that from my systems, so can't test.

```
>>> isinstance(42, float)
False
>>> int.mro()
[<class 'int'>, <class 'object'>]
>>> float.mro()
[<class 'float'>, <class 'object'>]
```

PR #5699, opened 2020-12-20
Don't perform authentication if no credential provided

shelld3v:
## The old behavior:

```python
>>> requests.get('http://127.0.0.1:8080', auth=(None, None))
```

```http
GET / HTTP/1.1
Host: 127.0.0.1:8080
User-Agent: python-requests/2.25.0
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Authorization: Basic Tm9uZTpOb25l (None:None)
```

The same thing will happen with:

```python
>>> requests.get('http://127.0.0.1:8080', auth=('None', 'None'))
```

This is weird, right!

## Now:
```python
>>> requests.get('http://127.0.0.1:8080', auth=(None, None))
```

```http
GET / HTTP/1.1
Host: 127.0.0.1:8080
User-Agent: python-requests/2.25.0
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
```
(No authentication)

This will be easier if the application allows auth username and password from arguments:

```python
response = self.session.request(
                    self.httpmethod,
                    url,
                    data=self.data,
                    proxies=proxy,
                    verify=False,
                    allow_redirects=self.redirect,
                    headers=dict(self.headers),
                    timeout=self.timeout,
                    auth=(self.user, self.pass)
                )
```

If the user didn't provide credentials, which means they don't need to authenticate, then `self.user == None` and `self.pass == None`. And in this situation, `requests` with my update won't perform authentication!

In the old version (without my update), it will be like:

```python
if self.user and self.pass:
    response = self.session.request(
                        self.httpmethod,
                        url,
                        data=self.data,
                        proxies=proxy,
                        verify=False,
                        allow_redirects=self.redirect,
                        headers=dict(self.headers),
                        timeout=self.timeout,
                        auth=(self.user, self.pass)
                    )
else:
    response = self.session.request(
                        self.httpmethod,
                        url,
                        data=self.data,
                        proxies=proxy,
                        verify=False,
                        allow_redirects=self.redirect,
                        headers=dict(self.headers),
                        timeout=self.timeout
                    )
```

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -101,7 +101,7 @@ class HTTPBasicAuth(AuthBase):
         # Only perform authentication if username or password was provided, or both
         if self.username is not None or self.password is not None:
             r.headers['Authorization'] = _basic_auth_str(self.username, self.password)
-            return r
+        return r
 
 
 class HTTPProxyAuth(HTTPBasicAuth):
@@ -111,7 +111,7 @@ class HTTPProxyAuth(HTTPBasicAuth):
         # Only perform authentication if username or password was provided, or both
         if self.username is not None or self.password is not None:
             r.headers['Proxy-Authorization'] = _basic_auth_str(self.username, self.password)
-            return r
+        return r
 
 
 class HTTPDigestAuth(AuthBase):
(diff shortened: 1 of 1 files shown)


shelld3v:
Also, `None` username or password will become an empty username/password

PR #5700, opened 2020-12-20
Auto-add labels to issues

shelld3v:


sigmavirus24:
Many folks open bugs and feature requests that are neither. Auto-labelling would then require more effort to remove those labels on the part of the maintainers

PR #5705, opened 2020-12-23
Update README.md

xcodz-dot:
A liitle bit of stats update 😄


sigmavirus24:
Hi @xcodz-dot, thanks for sending us a PR. 500,000+ by default includes 750,000. I don't think keeping these "up-to-date" is particularly important. If it were, we should be uploading the downloads per month count and lots of other things here. As it stands, these statistics aren't the most important thing to manage or update

PR #5715, opened 2021-01-08
Add json arguments to .put and .patch

mdelcambre:
This PR adds the `json` argument to both the `put` and `patch` function definitions. This is to bring them into agreement with the docstring of each function and to mirror the definition of `post`.
This should not change any functionally as the `json` arg would be captured by the `**kwargs` expansion. 

Obviously, if this difference was intentional, feel free to close without merging. Alternatively, if you prefer I can quickly put together a pr to update the docstring. It was just something that briefly tripped me up and I figured it was quick and easy to change.

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -119,7 +119,7 @@ def post(url, data=None, json=None, **kwargs):
     return request('post', url, data=data, json=json, **kwargs)
 
 
-def put(url, data=None, **kwargs):
+def put(url, data=None, json=None, **kwargs):
     r"""Sends a PUT request.
 
     :param url: URL for the new :class:`Request` object.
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Duplicate of https://github.com/psf/requests/pull/3664

mdelcambre:
Swear I searched before submitting but must have only looked in issues not closed PRs, sorry about that.

Ironically, it was the PR from that (#3666) that introduced the issue I was trying to solve. Currently, the documentation specifically calls out JSON as a parameter, but it is not actually a parameter in the function definition (obviously it is just grabbed by **kwargs). It creates seemingly broken documentation as pictured below.

Screen Shot 2021-01-08 at 4 37 59 PM


However, perhaps the current state is the lesser of all evils. Happy to drop it and not take up any more time.

PR #5717, opened 2021-01-10
Make raise_for_status() chainable

sambragg:
Frequently, I find myself doing the following:
```python
r = requests.get(url)
r.raise_for_status()
json = r.json()
```

I thought it may be useful to be able to do the following instead, where `raise_for_status()` is chainable:
```python
r = requests.get(url)
json = r.raise_for_status().json()
```

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -941,6 +941,8 @@ class Response(object):
 
         if http_error_msg:
             raise HTTPError(http_error_msg, response=self)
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Unfortunately, Requests is under a feature-freeze. As a result, we won't be expanding the behaviour of `raise_for_status()`

PR #5722, opened 2021-01-19
To work just like normal browsers, default headers in  requests.api.get(), requests.api.request(), requests.models.Request.__init__() and requests.sessions.Session.request() are set

shreyanavigyan:
The headers argument in `requests.api.get()`, `requests.api.request()`, `requests.models.Request.__init__()` and `requests.sessions.Session.request()` are set to a default dictionary value rather than NoneType Data Type to handle errors like 503 server error though the URL is correct. Some websites block webscrapers. To avoid being blocked, this default headers is very useful. This works in Mac OS X, Linux and Windows as of January 2020.

For example, https://www.amazon.com blocks Web-scrapers nowadays. Before this change, the returned response object's status code was always 503 though the website existed. 

before

After this change, the returned response object's status code is always 200.

after

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -57,10 +57,28 @@ def request(method, url, **kwargs):
     # Use the default headers
     try:
         if headers is None:
-            headers = {
-                'User-Agent':
-                'Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2228.0 Safari/537.36'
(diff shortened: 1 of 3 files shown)


nateprewitt:
Hi @shreyanavigyan,

I don't think it makes sense to add these user agents to Requests. They aren't actually accurate and the goal of Requests isn't to impersonate a browser. Users are free to add headers as they see fit, but we're not trying to violate protective restrictions services have put in place by default.

Thanks for the suggestion though.

PR #5736, opened 2021-01-29
Make the response class in adapter easier to override

sonthonaxrk:
I find myself pinning for this quite a bit.


sigmavirus24:
We're not accepting new features at this time

sonthonaxrk:
@sigmavirus24 
What's the process to add a feature? Do you tender roadmaps? Or is the internal API of requests just utterly frozen?

Issue #5746, opened 2021-02-10
content-type: text/csv does not apply UTF-8-decoding by default (RFC 7111 violation?)


jgehrcke:
Created a test/repro/MWE sheet here with a cell containing the letter `ö`: https://docs.google.com/spreadsheets/d/1q02F0AjDfCo_XlgFtT7HHU96emOcyf0PRQGKGoQTXCE/edit?usp=sharing

When you HTTP GET `https://docs.google.com/spreadsheets/d/1q02F0AjDfCo_XlgFtT7HHU96emOcyf0PRQGKGoQTXCE/export?format=csv` Google generates and sends a CSV document with UTF-8-encoded text in the response body.

Fetching this URL with `requests` and accessing the response content with the `text` attribute reveals that `requests` does _not_ UTF-8-decode the response body bytes:

```text
$ python -c 'import requests; r=requests.get("https://docs.google.com/spreadsheets/d/1q02F0AjDfCo_XlgFtT7HHU96emOcyf0PRQGKGoQTXCE/export?format=csv"); print(r.text)'
foo,bar,umlaut: Ã¶Ã¶
```

Library version:
```
$ pip list | grep requests
requests                 2.25.1
```

Response headers (from a `curl -v ...`):

```
> GET /<snip> HTTP/2
> Host: doc-14-2s-sheets.googleusercontent.com
> user-agent: curl/7.69.1
> accept: */*
> 
{ [5 bytes data]
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
{ [264 bytes data]
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
{ [264 bytes data]
* old SSL session ID is stale, removing
{ [5 bytes data]
* Connection state changed (MAX_CONCURRENT_STREAMS == 100)!
} [5 bytes data]
< HTTP/2 200 
< content-type: text/csv
< x-robots-tag: noindex, nofollow, nosnippet
< cache-control: no-cache, no-store, max-age=0, must-revalidate
< pragma: no-cache
< expires: Mon, 01 Jan 1990 00:00:00 GMT
< date: Wed, 10 Feb 2021 10:12:52 GMT
< content-disposition: attachment; filename="<snip>.csv"; filename*=UTF-8''<snip>.csv
< access-control-allow-origin: *
[...]

PR #5747, opened 2021-02-10
utils.py: default to utf-8 for text/csv (#5746)

jgehrcke:
This is a super quick attempt to address https://github.com/psf/requests/issues/5746 -- an invitation for feedback. Would appreciate your opinions. Thanks!

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -507,6 +507,9 @@ def get_encoding_from_headers(headers):
         # Assume UTF-8 based on RFC 4627: https://www.ietf.org/rfc/rfc4627.txt since the charset was unset
         return 'utf-8'
 
+    if 'text/csv' in content_type:
+        # Assume UTF-8 based on RFC 7111: https://tools.ietf.org/html/rfc7111 since the charset was unset
+        return 'utf-8'
 
 def stream_decode_response_unicode(iterator, r):
     """Stream decodes a iterator."""


jgehrcke:
(happy to add tests, changelog entry, etc -- brief instructions appreciated)

sigmavirus24:
I believe the last change we accepted along these lines was contentious and broke things. Plus 7111 is likely not widely implemented by folks distributing `text/csv` files at this point given it's recency (and given that 16+ year old RFCs aren't widely implemented either). I'm mildly -1 on accepting this

jgehrcke:
@sigmavirus24 thanks for the feedback. I share the conservative approach.

Do you see a different method that fixes the specific example given in #5746 (Google Sheets CSV export) while minimizing risk for breakage?

Just hypothetically.. what kind of breakage and unintended change in behavior would we be talking about? UTF-8-decoding byte sequences that would otherwise be... how-decoded? :)

sigmavirus24:
So this function will override logic if we could otherwise detect it from the `Content-Type` header. This means if someone is using UTF-16, UTF-32 or some other encoding that's not decode-able by UTF-8 then the Response.encoding value which was previously detected correctly will now always be `utf-8` which will fail if they try to access `text`.

jgehrcke:
> if we could otherwise detect it *from the Content-Type header*.

(emphasis mine)

Just to be sure: you agree that the new code path would not trigger when `charset` is _not_ set in the `Content-Type` header, right? This takes precedence:

```python
    if 'charset' in params:
        return params['charset'].strip("'\"")
```

How would the library, right now, automatically detect for example `UTF-16` in the absence of `charset`?

I have just noticed that the patch as proposed would not change behavior for `text/csv`, because this condition takes precedence:

```
    if 'text' in content_type:
        return 'ISO-8859-1'
```

That is, when `charset` is _not_ present in the `Content-Type` header, but when the substring `text` is found in the `content_type` string (it is an object of type `str`, and not a different kind of iterable, right?) then ISO-8859-1 is chosen for decoding for decoding the byte sequence.

The ` if 'text/csv' in content_type` would need to be added before that so that it can itself take precedence.

Is it fair to say that currently in all cases where the `Content-Type` header contains the word "text" but does _not_ specify a `charset` the library picks ISO-8859-1?

If the answer is a definite 'yes' that would be good news, because ISO-8859-1 is a subset of UTF-8.

sigmavirus24:
> Is it fair to say that currently in all cases where the `Content-Type` header contains the word "text" but does _not_ specify a `charset` the library picks ISO-8859-1?

So that's only one part of it. The other half of this coin is that people probably shouldn't be using [`Response.text`](https://github.com/psf/requests/blob/8c211a96cdbe9fe320d63d9e1ae15c5c07e179f8/requests/models.py#L852-L874) because it forces to a string however possible and is likely lossy.

But yes, latin-1 would be the default if `charset` isn't specified which is the right approach for most servers to take.

requests/utils.py
```diff
@@ -507,6 +507,9 @@ def get_encoding_from_headers(headers):
         # Assume UTF-8 based on RFC 4627: https://www.ietf.org/rfc/rfc4627.txt since the charset was unset
         return 'utf-8'
 
+    if 'text/csv' in content_type:
+        # Assume UTF-8 based on RFC 7111: https://tools.ietf.org/html/rfc7111 since the charset was unset
+        return 'utf-8'
```
dilyanpalauzov, line 513: Considering that some lines to the top there is `if 'text' in content_type:`, is the herein amended code ever executed?

nateprewitt:
I agree with Sigmavirus24 that this isn't the right approach. The risk being introduced doesn't outweigh the problem it solves. If this does eventually solidify as widely implemented, we can consider it for a major release but it won't be accepted for Requests 2.x.

Issue #4254, opened 2017-08-25
Git Clone Fails


evanshauser:
The repository does not clone, due to a bad commit. Let me know if I can do anything more to help!

## Expected Result

The repository should be cloned successfully with the command:

```
git clone git@github.com:requests/requests.git
```

## Actual Result

I get this message:

```
Cloning into 'requests'...
remote: Counting objects: 20967, done.
remote: Compressing objects: 100% (25/25), done.
error: object 5e6ecdad9f69b1ff789a17733b8edc6fd7091bd8: badTimezone: invalid author/committer line - bad time zone
fatal: Error in object
fatal: index-pack failed
```

## Reproduction Steps

run:

```
git clone git@github.com:requests/requests.git
```

Lukasa:
Thanks for the report. This is a known problem: see #3088, #3008, and #2690 for previous discussions. The problem can be easily resolved on your end either by performing a shallow clone (if you don't need the history) or turning off the fsck timestamp check. This cannot be fixed without rewriting our history and causing *way* more problems that way than by just letting it sit.

sigmavirus24:
Please, search open **and** closed issues before opening new ones in the future, @evanshauser

Issue #3805, opened 2017-01-10
Cannot clone repo due to bad timezone error


nyuszika7h:
```
nyuszika7h@cadoth ~/src > git clone https://github.com/kennethreitz/requests.git
Cloning into 'requests'...
remote: Counting objects: 18867, done.
remote: Compressing objects: 100% (11/11), done.
error: object 5e6ecdad9f69b1ff789a17733b8edc6fd7091bd8: badTimezone: invalid author/committer line - bad time zone
fatal: Error in object
fatal: index-pack failed
```

Lukasa:
Thanks for raising this issue! In future, please search the issue history before opening new issues on projects, as you'll often find that your issue has been reported before. In this case, this is a duplicate of #3088, #3008, and #2690.

ghost:
Sorry. I did search, but didn't get the right terms.

Lukasa:
If you have an error message, it's usually sufficient just to try to search for that. In this case, I searched for "fatal: index-pack failed" and #3088 popped right up.

ghost:
Yeah, well, I searched for "timezone" because I didn't realize there's a space in it in the error message. (Also, it's strange, I don't remember enabling fsck, but I found it in my git config.)

Lukasa:
fsck is on-by-default on new installs of git, so unless you're persisting a gitconfig from an older version of git you'll always find it on. =)

ghost:
I do have a custom gitconfig, so that's strange. This kind of issue could have been avoided if they enabled fsck by default in the first place, but oh well. You can't rewrite 6 years of history now, that would just mess things up even more. I do wonder how that weird timezone ended up in a commit though.

Lukasa:
Yeah, it's really annoying because it causes legitimate problems for lots of people, especially those who use git in automated scripts, but we really can't do anything about it now. =(

vermeeren:
Even though it hurts to do this, you may want to consider rewriting the history to fix this problem once and for all. For Breathe I ended up doing this and I added a notice to the top of the README, in the biggest heading possible, for a while.

Projects using this as a submodule will have to update the SHA once. But currently many setups encounter a clone failure meaning existing submodules will not work at all and cloning this repo directly won't either.

Ref: https://github.com/michaeljones/breathe/issues/340#issuecomment-394396673

Issue #3088, opened 2016-04-11
badTimezone: cannot clone with git 2.8.1


rillian:
I can't clone this repo. It complains:

```
$ git clone https://github.com/kennethreitz/requests
Cloning into 'requests'...
remote: Counting objects: 17048, done.
remote: Compressing objects: 100% (39/39), done.
error: object 5e6ecdad9f69b1ff789a17733b8edc6fd7091bd8: badTimezone: invalid author/committer line - bad time zone
fatal: Error in object
fatal: index-pack failed
```

Lukasa:
This is a known problem: see #3008 and #2690 for previous discussions. The problem can be easily resolved on your end either by performing a shallow clone (if you don't need the history) or turning off the fsck timestamp check.

rillian:
Thanks for the quick response.

kennethreitz:
@rillian did you enable verification yourself or did another tool do that?

rillian:
You're right, I have those options set in my ~/.gitconfig; it's nothing to do with the git version. I don't know if I did it manually or if one of our audit tools added it, because how is that not the default? :/

``` ini
[transfer]
          fsckobjects = true
[fetch]
          fsckobjects = true
[receive]
          fsckObjects = true 
```

I've been trying to figure out how to get YCM to work around this, but it's tricky with submodules because an initial shallow clone generally fails. There doesn't seem to be a `fetch.fsckBadTimezone`. I couldn't get your `$ git config fsck.badTimezone ignore` suggestion to work, just disabling `fetch.fsckObjects` entirely.

rogueresistor:
```git config --global fetch.fsck.badTimezone ignore```

Works for me when cloning submodules, this just disables it for fetch.

Issue #3008, opened 2016-02-12
invalid git history


thestinger:
The Git history for this repository is invalid. You'll run into this if you enable verification of the history in your configuration (which should really be the default):

```
[transfer]
	fsckobjects = true

[fetch]
	fsckobjects = true

[receive]
	fsckobjects = true
```

```
% git clone https://github.com/kennethreitz/requests.git
Cloning into 'requests'...
remote: Counting objects: 16580, done.
error: object 5e6ecdad9f69b1ff789a17733b8edc6fd7091bd8: badTimezone: invalid author/committer line - bad time zone
fatal: Error in object
```

If you look at the commit with verification disabled, you can see the problem:

```
commit 5e6ecdad9f69b1ff789a17733b8edc6fd7091bd8
Author: Shrikant Sharat Kandula <shrikantsharat.k@gmail.com>
Date:   Thu Sep 8 02:38:50 2011 +51800

    Typo in documentation
    
    The kwarg is named `headers`, not `header`. Also, its a dict, not a set.
```

I don't have a suggestion on how to approach this. It's quite unfortunate. The history would probably have to be rewritten, and that would be awful. I do think it's worth noting though... verification by default may become the default on various distributions or even upstream.

untitaker:
> verification by default may become the default on various distributions

I'd assume there are many more large/popular repositories that contain at least broken timestamps. IMO it'd be impudent if a distro patches git in a way that would break cloning repos like this. Given requests' history with downstream patches, I guess others would agree.

> or even upstream.

Upstream should decide, yes.

---

EDIT: From the larger Python repos, it seems that [Werkzeug](https://github.com/mitsuhiko/werkzeug) is also affected by this.

thestinger:
If verification is off, there aren't even checks to make sure that the hashes are correct. It would make sense if there was a loose mode accepting some invalid history, like invalid timestamps.

kennethreitz:
Well, one solution is to `$ git config fsck.badTimezone ignore`

kennethreitz:
OK, here's my position on this.
- This is nothing for us to be concerned with. The only people effected by this are people who signed up to be effected by this by automatically running `git fsck` when performing normal Git operations. I can definitely see why this could be considered a good idea, but since almost no one does this, this is not something Requests needs be concerned with.
[...]

PR #5755, opened 2021-02-17
Provide directions for cloning repository despite malformed commits

pdmccormick:
Issues #2690, #3008, #3088, #3805 #4254 all refer to the issue with commit 5e6ecdad9f69b1ff789a17733b8edc6fd7091bd8 that cause Git to error out when cloning the repository. As originally discussed in #2690, this issue cannot be avoided without rewriting history, so why don't we instead include some helpful directions at the bottom of the README?


sigmavirus24 (approved):
Thanks so much!

Issue #5773, opened 2021-03-15
Connection timeout (not Read, Wall or Total) is consistently taking twice as long


MestreLion:
I'm aware that several issues related to timeout were opened (and closed) before, so I'm trying to narrow this report down to a very specific scope: **connection** timeout is behaving in a consistent, wrong way: it times out at _precisely_ **twice** the requested time.

Results below are _so_ consistent we must acknowledge there is _something_ going on here! I **beg** you guys not to dismiss this report before taking a look at it!

What this report is **not** about:
- Total/Wall timeout:
That would be a nice feature, but I'm fully aware this is currently outside the scope of Requests. I'm focusing on _connection_ timeout only.

- Read timeout:
All my tests were using http://google.com:81, which fails to even _connect_. There's no read involved, the server exists but never responds, not even to refuse the connection. No data is ever transmitted. No HTTP connection is ever established. This is **not** about `ReadTimeoutError`, this is about `ConnectTimeoutError`.

- Accurate timings / network fluctuations:
Not asking for millisecond precision. I don't even care about _whole seconds_ imprecision. But, surprisingly, `requests` is being incredibly accurate... to _twice_ the time.

## Expected Result

`requests.get('http://google.com:81', timeout=(4, 1))` should take _approximately_ 4 seconds to timeout.

## Actual Result

It _consistently_ takes about 8.0 seconds to raise `requests.ConnectTimeout`. It always takes twice the time, for timeouts ranging from 1 to 100. Exception message clearly says in the end: `Connection to google.com timed out. (connect timeout=4)`, a very distinct message from read timeouts.

## Reproduction Steps

```python
import requests, time, os, sys

# Using a know URL to test connection timeout
def test_timeout(timeout, url='http://google.com:81'):
    start = time.time()
    try:
        requests.get(url, timeout=timeout)
        print("OK!")  # will never reach this...
    except requests.ConnectTimeout:  # any other exception will bubble out
        print('{}: {:.1f}'.format(timeout, time.time()-start))

print("\n1 to 10, simple numeric timeout")
for i in range(1, 11):
    test_timeout(i)

print("\n1 to 10, (x, 1) timeout tuple")
for i in range(1, 11):
    test_timeout((i, 1))

print("\n1 to 10, (x, 10) timeout tuple")
for i in range(1, 11):
    test_timeout((i, 1))

print("\nLarge timeouts")
for i in (20, 30, 50, 100):
    test_timeout((i, 1))
```

Results:
```
[...]

PR #5774, opened 2021-03-15
#5773: Fixed the number of seconds specified for timeout being doubled

cpyberry:
## issue

#5773

## changes

Create a subclass that divides the connect argument of TimeoutSauce by 2.


MestreLion:
IMHO blindly halving the specified timeout is not a good strategy, specially if the underlying issue is still unknown.

Besides, there are a few issues with that implementation, from not handling a `None` value to shadowing the imported name

cpyberry:
Thank you for looking at my code.

cpyberry:
I forgot about handling None.

PR #5554, opened 2020-08-09
WIP: Make Brotli support enabled by default, if appropriate urllib3 version is available

gdubicki:
This is still a work in progress - I have to make sure that the tests are actually running, not skipped every time 😅 but [I would like to get feedback if the direction is right](https://github.com/psf/requests/blob/master/docs/dev/contributing.rst#get-early-feedback).

I had to update some deps to make the tests pass after a lot of packages being updated after I added `brotli` to dev-packages. But I assume that this should be done anyway. Especially as there were discrepancies in the versions - f.e. `pytest` version in `setup.py` commit message declared v. 4 compatibility while the version in `Pipenv` was set to be at most v. 3.10.1.

Speaking of dependencies I have a doubt how to declare required urllib3 version. Technically the first with brotli support is 1.25.1, but this is a version not supported by Requests... Should I update all the references to require urllib3 >= 1.25.2 then instead of >= 1.25.1?

PR #5783, opened 2021-03-29
Add support for brotli decoding

dilyanpalauzov:
When the `brotli` or `brotlicffi` packages are installed, `urllib3.util.make_headers()` inserts ',br' in the Accept-Encoding header and decodes br from the answers.

### DIFF
--- a/docs/community/faq.rst
+++ b/docs/community/faq.rst
@@ -11,6 +11,9 @@ Encoded Data?
 Requests automatically decompresses gzip-encoded responses, and does
 its best to decode response content to unicode when possible.
 
+When either the `brotli <https://pypi.org/project/Brotli/>`_ or `brotlicffi <https://pypi.org/project/brotlicffi/>`_
+package is installed, requests also decodes Brotli-encoded responses.
+
 You can get direct access to the raw response (and even the socket),
 if needed as well.
 
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -42,7 +42,10 @@ DEFAULT_CA_BUNDLE_PATH = certs.where()
 
 DEFAULT_PORTS = {'http': 80, 'https': 443}
 
-DEFAULT_ACCEPT_ENCODING = make_headers(accept_encoding=True)["accept-encoding"]
+# Ensure that ', ' is used to preserve previous delimiter behavior.
+DEFAULT_ACCEPT_ENCODING = ", ".join(
+    re.split(r",\s*", make_headers(accept_encoding=True)["accept-encoding"])
+)
 
 
 if sys.platform == 'win32':
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -20,6 +20,7 @@ import tempfile
 import warnings
 import zipfile
 from collections import OrderedDict
+from urllib3.util import make_headers
 
 from .__version__ import __version__
 from . import certs
@@ -820,7 +821,7 @@ def default_headers():
     """
     return CaseInsensitiveDict({
         'User-Agent': default_user_agent(),
-        'Accept-Encoding': ', '.join(('gzip', 'deflate')),
+        'Accept-Encoding': make_headers(accept_encoding=True)["accept-encoding"],
         'Accept': '*/*',
         'Connection': 'keep-alive',
     })
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -42,6 +42,8 @@ DEFAULT_CA_BUNDLE_PATH = certs.where()
 
 DEFAULT_PORTS = {'http': 80, 'https': 443}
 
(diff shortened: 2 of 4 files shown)


VeNoMouS:
Should that not be something more along the lines of...

```python
make_headers(accept_encoding=True).get('accept-encoding', ', '.join(('gzip', 'deflate'))),
```

as a 'fail safe'

dilyanpalauzov:
You can change it as you want, as long Requests sends transparently `Accept-Encoding: br`.

.join is an overkiller to generate a constant string.

gdubicki:
My 2c: I like the approach of reusing code from urllib3 more than mine from #5554. :)

However the [minimal supported by requests version of urllib3 is now 1.21.1](https://github.com/psf/requests/blob/master/setup.py#L46), which [does not support brotli yet](https://github.com/urllib3/urllib3/blame/1.21.1/urllib3/util/request.py). The [minimal version that has this support is 1.25.1](https://github.com/urllib3/urllib3/commit/9eecb6bd0fab0a4546dd35c7ee42e8cd9c59274f). So I think that it should be mentioned in docs update here, along with info about required `brotli` or `brotlicffi` packages.

gdubicki:
Such fail-safe is not needed, @VeNoMouS , as even the oldest supported by requests urllib3 has `make_headers()` that will return a dict with `Accept-Encoding` key, when called with `accept_encoding=True`.

See https://github.com/urllib3/urllib3/blame/1.21.1/urllib3/util/request.py#L7-L55 .

docs/community/faq.rst
```diff
@@ -11,6 +11,9 @@ Encoded Data?
 Requests automatically decompresses gzip-encoded responses, and does
 its best to decode response content to unicode when possible.
 
+When either the brotli or brotlicffi package is installed, requests
```
sethmlarson: Can we link to the individual packages here?

docs/community/faq.rst
```diff
@@ -11,6 +11,9 @@ Encoded Data?
...
+also decodes brotli-encoded responses.
```
sethmlarson: nit: Brotli is a proper name so should be capitalized here and in the changelog

requests/utils.py
```diff
@@ -820,7 +821,7 @@ def default_headers():
     """
     return CaseInsensitiveDict({
         'User-Agent': default_user_agent(),
-        'Accept-Encoding': ', '.join(('gzip', 'deflate')),
+        'Accept-Encoding': make_headers(accept_encoding=True)["accept-encoding"],
```
sethmlarson: Should we generate this header once and store the result so it doesn't need to be called every time? The value is determined by imports so shouldn't be changing often.

sethmlarson (changes requested):
This looks mostly good, few comments for you

docs/community/faq.rst
dilyanpalauzov: I have capitalized Brotli-decompression on both places.

requests/utils.py
dilyanpalauzov: This optimization makes sense.  I kindly ask you to implement it in the way, you think fits best.

dilyanpalauzov:
> However the [minimal supported by requests version of urllib3 is now 1.21.1](https://github.com/psf/requests/blob/master/setup.py#L46), which [does not support brotli yet](https://github.com/urllib3/urllib3/blame/1.21.1/urllib3/util/request.py). The [minimal version that has this support is 1.25.1](https://github.com/urllib3/urllib3/commit/9eecb6bd0fab0a4546dd35c7ee42e8cd9c59274f). So I think that it should be mentioned in docs update here, along with info about required `brotli` or `brotlicffi` packages.

Please propose wording and where to update the text.  Currently the brotli-texts are updated on three places - faq, history and quickstart.  I consider it better, if the documentation does not repeat the same thing on different places.

naorlivne:
I apologize If this might come off as a bit rude but this PR seems to be the only thing blocking 2.26.0 milestone (the other changes in it have no pending change requested), and it's been a month without any progress, if for more time is needed that's of course understandable but would it be possible to bump this for a release after the next one so as not to have it block other changes from being released?

ghost:
ap

On Mon, Jun 14, 2021, 1:46 AM Naor Livne ***@***.***> wrote:
nah..

ghost:
Just bump 2.2.1 for the update ... majority needs to redilimbruim it's
depth not making it by lying ..add support!

sethmlarson (approved):
I'm now happy with my changes but I'd like another reviewer to take a quick peek now that I'm the author of most of this diff. cc @nateprewitt @sigmavirus24

sigmavirus24 (approved):

PR #5784, opened 2021-03-31
fix typo

DavidKimDY:
fix typo

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,7 +3,7 @@
 # Requests documentation build configuration file, created by
 # sphinx-quickstart on Fri Feb 19 00:05:47 2016.
 #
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -3,7 +3,7 @@
 # Requests documentation build configuration file, created by
 # sphinx-quickstart on Fri Feb 19 00:05:47 2016.
 #
-# This file is execfile()d with the current directory set to its
+# This file is execfile() with the current directory set to its
```
sigmavirus24, line 6: This is:

1. Intentional
2. Auto-generated for us

sigmavirus24:
Thank you for your contribution! Unfortunately, we didn't write the content of that comment and don't maintain the comments in that file.

Issue #5785, opened 2021-04-04
Grammar error in GitHub description


saltedcoffii:
The GitHub description reads as follows: "A simple, yet elegant HTTP library."

It should be: "A simple, yet elegant, HTTP library."

This is also a problem on the pypi page under the project description.

PR #5791, opened 2021-04-10
Grammar fix in description

cocobennett:
Fixes https://github.com/psf/requests/issues/5785

Could not find the second error as described `This is also a problem on the pypi page under the project description.`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Requests
 
-**Requests** is a simple, yet elegant HTTP library.
+**Requests** is a simple, yet elegant, HTTP library.
 
 ```python
 >>> import requests


sigmavirus24 (approved):

sigmavirus24:
Thank you!

kumiDa:
@cocobennett,
To answer this-
> Could not find the second error as described This is also a problem on the pypi page under the project description.

This would essentially be fixed by your patch as the `long_description` picked up by PyPI page to populate the description section is set to `readme` in `setup.py` as seen in this line: https://github.com/psf/requests/blob/c45a4dfe6bfc6017d4ea7e9f051d6cc30972b310/setup.py#L70

PR #5792, opened 2021-04-11
Short-circuit raise_for_status code for statuses that will not raise …

ptmcg:
…an exception

This short-circuit code will skip the steps for decoding response.reason if the status code is not an exception-raising status code.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -191,3 +191,4 @@ Patches and Suggestions
 - "Dull Bananas" <dull.bananas0@gmail.com> (`@dullbananas <https://github.com/dullbananas>`_)
 - Alessio Izzo (`@aless10 <https://github.com/aless10>`_)
 - Sylvain Marié (`@smarie <https://github.com/smarie>`_)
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Decoding the `reason` certainly isn't that expensive in reality. Why add this complexity here? What does this gain anybody?

ptmcg:
Early exit from an exception-raising routine, when it can be quickly determined that no exception will be raised, would be considered by some to be less complexity, not more. At run-time it saves several additional if-tests for all successful statuses, and for those reading the code, it clearly shows that only status codes >= 400 are really of any interest when raising an exception.

(I could understand if there were some possible restructuring of HTTP status codes in the future, but this status code organization is pretty cast in stone now.)

nateprewitt:
Hi @ptmcg, thanks for the initiative here! Unfortunately, I think I'm in agreement with @sigmavirus24. This is an optimization that adds extra code clutter without providing a tangible user benefit.

I did some quick benchmarking and performance difference is measured in fractions of a _microsecond_. There's never going to be a situation where we accept latency involved with sending a request across a network but are concerned about sub-microsecond precision in post processing.

PR #5795, opened 2021-04-16
Replace getattr(...) is not None with hasattr(...)

takos22:
Replace `getattr(body, 'tell', None) is not None` with `hasattr(body, 'tell')`


nateprewitt:
Hi @takos22,

This syntax is intentional. You can read the reasoning [here](https://hynek.me/articles/hasattr/).

takos22:
Ok my bad, didn't know about that issue with ``hasattr``. Thanks for the info and for the fast response.

PR #5800, opened 2021-04-25
Custom HTTP adapters via Session constructor

rupertcw:
I often find myself having to do the following when using `Session` with custom retrying logic:
```
retry = Retry(3, backoff_factor=1.)
http_adapter = requests.adapters.HTTPAdapter(pool_connections=4, max_retries=retry)
session = requests.Session()
session.mount('https://', http_adapter)
session.mount('http://', http_adapter)
```

Having to mount the adapter myself for such a simple use-case seems superfluous, error-prone and not very user-friendly so I'm suggesting we open up the Session constructor a little bit:

```
retry = Retry(3, backoff_factor=1.)
http_adapter = requests.adapters.HTTPAdapter(pool_connections=4, max_retries=retry)
session = requests.Session(http_adapter=http_adapter)
```

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -359,7 +359,7 @@ class Session(SessionRedirectMixin):
         'max_redirects',
     ]
 
-    def __init__(self):
+    def __init__(self, http_adapter=None):
 
         #: A case-insensitive dictionary of headers to be sent on each
         #: :class:`Request <Request>` sent from this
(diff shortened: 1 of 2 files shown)


sigmavirus24:
This project is under a feature-freeze which means we're not accepting any changes to the API at this time.

Thanks for submitting this but we will not be accepting it

PR #5808, opened 2021-05-01
Make use of the 'yield from' keyword whenever possible for a better pythonic look!

Harmouch101:
Signed-off-by: Harmouch101 <mahmoudddharmouchhh@gmail.com>


sigmavirus24:
We're not accepting cosmetic changes or changes which break compatibility with Python 2.7

wiseaidev:
Oh! I didn't notice it supports python 2.7. However, I think the project needs to be updated to support only versions > 3.5  of python since other versions have reached the [end of life](https://endoflife.date/python) in terms of security support.

PR #5810, opened 2021-05-04
Throw value error when serializing JSON object with NaN value

### LINKED ISSUE #5767: Invalid JSON serialization of data with NaNs when using `json` request argument
When passing data to `json` argument in a `requests.post()` (or comparable) request and data contains "nan" floats (`float("nan")`), the data payload will be invalid JSON (in strict sense). Example:

    >>> import requests
    >>> requests.__version__
    '2.25.1'
    >>> data = {"foo": float("nan")}
    >>> resp = requests.post("https://httpbin.org/post", json=data)
    >>> resp.json()["data"]
    '{"foo": NaN}'

While the `NaN` in this result (the JSON serialized data payload that is sent) is valid JavaScript, it is not valid JSON.

The reason is that stdlib's `json.dumps` has `allow_nan` option which is `True` by default and consequently generates invalid JSON by default:

    >>> json.dumps(data)
    '{"foo": NaN}'

## Expected Result

A ValueError should be raised (at least by default). E.g.:

    >>> json.dumps(data, allow_nan=False)
    ...
    ValueError: Out of range float values are not JSON compliant

## Actual Result

library silently sends invalid JSON

tallalnparis4ev:
I am attempting to fix #5767 with this PR. Feedback is really appreciated.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -466,7 +466,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             # urllib3 requires a bytes-like body. Python 2's json.dumps
             # provides this natively, but Python 3 gives a Unicode string.
             content_type = 'application/json'
-            body = complexjson.dumps(json)
+            body = complexjson.dumps(json, allow_nan=False)
             if not isinstance(body, bytes):
                 body = body.encode('utf-8')
 
--- a/requests/models.py
+++ b/requests/models.py
@@ -29,7 +29,7 @@ from .auth import HTTPBasicAuth
 from .cookies import cookiejar_from_dict, get_cookie_header, _copy_cookie_jar
 from .exceptions import (
     HTTPError, MissingSchema, InvalidURL, ChunkedEncodingError,
-    ContentDecodingError, ConnectionError, StreamConsumedError)
+    ContentDecodingError, ConnectionError, StreamConsumedError, InvalidJSONError)
 from ._internal_utils import to_native_string, unicode_is_ascii
 from .utils import (
     guess_filename, get_auth_from_url, requote_uri,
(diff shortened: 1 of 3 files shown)


sigmavirus24 (changes requested):
I don't believe a `request()` method will throw a ValueError ever and so this constitutes a backwards incompatible change. We should create a new exception, however, that inherits from `RequestsException` to cover this case and do something like

```py
try:
    body = complexjson.dumps(json, allow_nan=False)
except ValueError as ve:
    raise InvalidJSONError(json, original_exc=ve)
```

requests/models.py
```diff
@@ -466,7 +466,12 @@ def prepare_body(self, data, files, json=None):
             # urllib3 requires a bytes-like body. Python 2's json.dumps
             # provides this natively, but Python 3 gives a Unicode string.
             content_type = 'application/json'
-            body = complexjson.dumps(json, allow_nan=False)
+
+            try:
+              body = complexjson.dumps(json, allow_nan=False)
+            except ValueError as ve:
+              raise InvalidJSONError(ve, request=self)
```
tallalnparis4ev, line 473: @sigmavirus24 is this okay? Couldn't do it the exact way you specified, but `request` contains the JSON info.

sigmavirus24: I have an instinct of "No" but that's because people do not read the docs and sometimes pass open files and such _in addition to_ the `json=` keyword and keeping a reference to those might not be _the best_ but that's a "The user has shot themselves in the foot" situation only for high traffic (lots of requests doing the wrong thing explicitly that has undefined behaviour), poorly designed software (because it's potentially not releasing/decref'ing the exception which doesn't decref the request which doesn't decref the file obj).

tl;dr - This should be fine

sigmavirus24 (approved):

Issue #3829, opened 2017-01-24
Session.verify=False ignored when REQUESTS_CA_BUNDLE environment variable is set


intgr:
One would expect that when the caller explicitly asks to make unverified requests, then the `REQUESTS_CA_BUNDLE` environment variable doesn't affect it. The reality is different, however.

```python
import os
import requests

os.environ['REQUESTS_CA_BUNDLE'] = 'asd.pem'  # Must an existing file

r = requests.get('https://self-signed.badssl.com/', verify=False)
print(r)
# Prints: <Response [200]>

session = requests.Session()
session.verify = False

r = session.get('https://self-signed.badssl.com/')
print(r)
# Fails: requests.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:749)
```

Lukasa:
Thanks for raising this issue! This is a related issue to #2018: specifically, we prefer the environment to the `Session` value assuming the environment specifies an auth value. The fix is easy, but we've unfortunately ossified around this value, so we can't fix it until v3.

Given that we fail-closed here (that is, it's not possible to use this arrangement to force us not to verify when we should), this isn't a security vulnerability, so there is no way we can justify bringing it forward to before v3.

intgr:
OK, thanks for the quick response. Is 3.0.0 coming some time soon or is it just a plan for now?

Lukasa:
"soon" is a bit strong. However, there's a branch that code can be committed to, and there is active work being done on urllib3 v2, which once done will be the catalyst for us to actually ship requests v3.

intgr:
For whoever else is struggling with this problem, I created a wrapper class as workaround:

```python
class WrappedSession(requests.Session):
    """A wrapper for requests.Session to override 'verify' property, ignoring REQUESTS_CA_BUNDLE environment variable.

    This is a workaround for https://github.com/kennethreitz/requests/issues/3829 (will be fixed in requests 3.0.0)
    """
    def merge_environment_settings(self, url, proxies, stream, verify, *args, **kwargs):
        if self.verify is False:
            verify = False

        return super(WrappedSession, self).merge_environment_settings(url, proxies, stream, verify, *args, **kwargs)
```

megahall:
This issue hit me today. I had to debug a good amount of code to track it down.

Where are we now relative to when it was identified back in January?

Lukasa:
No change.

Justsoos:
[...]

PR #5816, opened 2021-05-14
Warn user if environment variables are used

Akasurde:
When session.verify=False, session.trust_env=True and
REQUESTS_CA_BUNDLE or CURL_CA_BUNDLE is defined as environment
variables then, notify user that requests will use environment variables
rather than silently failing.

partially fixes: #3829

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>


Akasurde:
I spend hours debugging this error. I am aware that this odd behavior is fixed in 3.* release.

requests/sessions.py
```diff
@@ -715,6 +716,10 @@ def merge_environment_settings(self, url, proxies, stream, verify, cert):
             # Look for requests environment configuration and be compatible
             # with cURL.
             if verify is True or verify is None:
+                if os.environ.get('REQUESTS_CA_BUNDLE'):
+                    warnings.warn('Setting verify value to environment value of REQUESTS_CA_BUNDLE')
+                if os.environ.get('CURL_CA_BUNDLE'):
```
sigmavirus24: Assuming we wanted to get more harassment from our users over warning messages... this should be an `elif`. If both are set we're emitting both warnings for the same request which will confuse users because we're going to prefer `REQUESTS_CA_BUNDLE` over `CURL_CA_BUNDLE` and so should only warn once with the real informatoin

Akasurde: Oh, yes. Sorry my bad.

hardenchant:
Really nice pr, I checked it – warnings works well, can save a tons of devs time.
It would be nice to have more verbose warning text, for those who are not familiar with this problem.

Akasurde:
> It would be nice to have more verbose warning text, for those who are not familiar with this problem.

Could you please suggest something? 😂

sigmavirus24:
Closing/re-opening since I can't trigger CI to run here

intgr:
What's the status of this PR?

intgr:
> I am aware that this odd behavior is fixed in 3.* release.

From #6050 looks like 3.x isn't going to happen any time soon, and commitments to fix things in 3.x don't necessarily hold.

Akasurde:
> What's the status of this PR?

I would love to see this PR getting merged. Let me know if you need something from myside.

PR #5821, opened 2021-05-18
Update advanced.rst

Akay7:


ujjwaltwitx (approved):

docs/user/advanced.rst
```diff
@@ -446,7 +446,7 @@ argument.
     def print_url(r, *args, **kwargs):
         print(r.url)
 
-If an error occurs while executing your callback, a warning is given.
+If an error occurs while executing your callback, an exception is going to bubble up.
```
sigmavirus24: "going to bubble up" is a confusing euphemism. Let's rewrite this altogether:

```suggestion
Your callback function must handle its own exceptions. Any unhandled exception will not be handled by Requests and thus must be handled by the code calling Requests.
```

Akay7: @sigmavirus24  Thanks for the suggestion, I'm slightly changing this.

sigmavirus24: @Akay7 your changes are grammatically incorrect and still relying on colloquialisms which may not be understood by all. Please use the suggestion I wrote.

Akay7: @sigmavirus24 
I'm sorry, but I have a different opinion. But if you would like to update docs on your own feel free to do it.

NONVME (approved):

sigmavirus24 (approved):

PR #5835, opened 2021-06-01
Fix GitHub links

fmarier:
All of these links now redirect to a repo under a different GitHub user account.

Not sure whether you want to update the changelog, but I'm happy to amend this PR if you'd rather not change `HISTORY.md`.

### DIFF
--- a/docs/community/recommended.rst
+++ b/docs/community/recommended.rst
@@ -59,4 +59,4 @@ Betamax
 `Betamax`_ records your HTTP interactions so the NSA does not have to.
 A VCR imitation designed only for Python-Requests.
 
-.. _betamax: https://github.com/sigmavirus24/betamax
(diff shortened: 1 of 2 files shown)


nateprewitt:
Thanks for the PR, @fmarier! Let’s leave the change log alone for now. I’m fine fixing the other references in the docs though.

fmarier:
I've removed the changelog from this PR.

nateprewitt (approved):
Cool, thanks @fmarier!

nateprewitt:
Hmm, tests are failing on Python 3.6+ but it's obviously not related to this PR. Looks like something has broken in the last couple weeks. The only change I can see if urllib3 releasing a new version. We'll track that in a separate issue. Thanks again!

PR #5857, opened 2021-07-05
Fix calling get_connection() with a positional argument when it is optional

BoboTiG:


sigmavirus24:
This says it's fixing something, but it's not strictly necessary and there's no context for why you feel this to be necessary. Can you elaborate?

BoboTiG:
Oh sorry for the lack of details, it seems I closed my computer without having filled the description...

I subclassed the `HTTPAdapter` class, here is the revelant code: https://github.com/httpie/httpie-snapdsocket/blob/aa0d58beca2475b461ba9eac420fff5583e67bed/httpie_snapdsocket.py#L21
It works like a charm.

But then, I wanted to replace 
```python
def get_connection(self, url, proxies=None):
    # ...
    return super().get_connection(url, proxies=proxies)
```
with 
```python
def get_connection(self, url, **kwargs):
    # ...
    return super().get_connection(url, **kwargs)
```
To be resilient to future API updates on the `requests` side.

It will not work:
```python
  File "httpie-snapdsocket/tests.py", line 17, in test_simple
    res = session.get('snapd:///v2/find?name=' + package).json()
  File ".../site-packages/requests/sessions.py", line 555, in get
    return self.request('GET', url, **kwargs)
  File ".../requests/sessions.py", line 542, in request
  File ".../site-packages/requests/sessions.py", line 655, in send
  File ".../site-packages/requests/adapters.py", line 412, in send
    conn = self.get_connection(request.url, proxies)
TypeError: get_connection() takes 2 positional arguments but 3 were given
```

I can live with it though.

sigmavirus24:
> ```python
> def get_connection(self, url, **kwargs):
>     # ...
>     return super().get_connection(url, **kwargs)
> ```

Why not

```python
def get_connection(self, url, proxies=None, **kwargs):
    # ...
    kwargs.setdefault("proxies", proxies)
    return super().get_connection(url, **kwargs)
```

That achieves what you want without making a unnecessary change here

BoboTiG:
The goal was to not care about `kwargs` at all. So that when/if the `get_connection()` signature changes, my code will still work.

sigmavirus24:
It's incredibly unlikely to change

PR #5862, opened 2021-07-08
General improvements

machinexa2:
Fstrings are faster than %s and .format(). .format() speed is nearly equal to %s


sethmlarson:
We still support Python versions which don't have f-strings so can't accept this PR.

PR #5864, opened 2021-07-08
More Options To Run Example Code

jameskaggwa:
#### Making the example code sample available as a copy and paste and then run in any exiting python file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,27 @@
 >>> r.json()
 {'disk_usage': 368627, 'private_gists': 484, ...}
 ```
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @jameskaggwa, thanks for contributing to Requests! Unfortunately, I don't think we'll accept this PR at this time. We expect a basic level of Python knowledge with the docs, so understanding writing code in a file vs in the interpreter is assumed. Otherwise, the docs become significantly bloated and more difficult to find relevant information.

PR #5867, opened 2021-07-09
Disable requests[security] and remove 3.5 support references

nateprewitt:
## requests[security]

This PR will remove functionality from the `requests[security]` extra. We initially removed default support for PyOpenSSL in Requests 2.24.0 (#5443) as it is now considered less secure. Deprecation of the `extras_require` was announced in [Requests 2.25.0](https://github.com/psf/requests/blob/master/HISTORY.md#2250-2020-11-11) and we're officially removing the `extras_require` functionality in Requests 2.26.0. 

Projects currently using `requests[security]` after this change will continue to operate as if performing a standard requests installation (secure by default).

While not recommended, if PyOpenSSL support is still a requirement in your project, you may install `urllib3[secure]` and add `urllib3.contrib.pyopenssl.inject_into_urllib3()` to your code to re-enable the functionality.

## Python 3.5 Support

We will also be removing support for Python 3.5 in Requests 2.26.0 as announced in the [2.25.0 release notes](https://github.com/psf/requests/blob/master/HISTORY.md#2250-2020-11-11).

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [2.7, 3.6, 3.7, 3.8, 3.9]
         os: [ubuntu-18.04, macOS-latest, windows-latest]
         include:
           # pypy3 on Mac OS currently fails trying to compile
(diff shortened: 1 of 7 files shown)


sethmlarson (changes requested):
Looks good! Only thing I found was we should update `python_requires` to add `!=3.5.*`

sethmlarson (approved):
LGTM

PR #5874, opened 2021-07-15
Remove legacy version check for cryptography

akx:
It's unlikely that someone would be running [a version of cryptography from before June 2016](https://pypi.org/project/cryptography/1.3.3/) these days.

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -84,17 +84,6 @@ def check_compatibility(urllib3_version, chardet_version, charset_normalizer_ver
     else:
         raise Exception("You need either charset_normalizer or chardet installed")
 
-def _check_cryptography(cryptography_version):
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Given that CentOS 7 was released in 2014 and many companies use that instead of Ubuntu or RHEL, I'm not entirely certain we can make that assumption

nateprewitt:
I think I'm also not inclined to remove this yet. People can still manually reenable PyOpenSSL in requests via urllib3, and this check would still be needed/useful in that case.

sethmlarson:
I agree with @sigmavirus24 and @nateprewitt and that makes a complete set. Thanks for opening this though, this may be something we can drop in the future when urllib3 drops pyOpenSSL/cryptography but that's a ways off.

PR #5878, opened 2021-07-16
Unify interface definition

weddige:
With this pull request I am trying to get rid of some differences in the interface definition and doc strings of `requests.api` and `requests.sessions.Session`.

In not all cases there was an obvious right choice and some discussion might be necessary.

### DIFF
--- a/requests/api.py
+++ b/requests/api.py
@@ -15,7 +15,6 @@ from . import sessions
 
 def request(method, url, **kwargs):
     """Constructs a :class:`Request <Request>`, prepares it and sends it. 
-    Returns :class:`Response <Response>` object.
 
     :param method: method for the new :class:`Request` object.
     :param url: URL for the new :class:`Request` object.
@@ -54,6 +53,8 @@ def request(method, url, **kwargs):
         development or testing.
     :param cert: (optional) if String, path to ssl client cert file (.pem).
         If Tuple, ('cert', 'key') pair.
+    :return: :class:`Response <Response>` object
+    :rtype: requests.Response
 
     Usage::
 
@@ -71,7 +72,7 @@ def request(method, url, **kwargs):
 
 
 def get(url, params=None, **kwargs):
-    r"""Sends a GET request. Returns :class:`Response` object.
+    r"""Sends a GET request.
 
     :param url: URL for the new :class:`Request` object.
     :param params: (optional) Dictionary or bytes to be sent in the query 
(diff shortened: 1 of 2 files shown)


requests/sessions.py
```diff
@@ -471,17 +471,17 @@ def request(self, method, url,
             params=None, data=None, headers=None, cookies=None, files=None,
             auth=None, timeout=None, allow_redirects=True, proxies=None,
             hooks=None, stream=None, verify=None, cert=None, json=None):
```
weddige, line 473: The documentation has `json` after `data` and at least in `requests.api.request` there are no positional arguments except `url`.

As it is a breaking change to move `json` to the front of the list, I didn't change it. However, I think it might be something to think about:

```python
    def request(self, method, url, *,
            params=None, data=None, json=None, headers=None, cookies=None,
            files=None, auth=None, timeout=None, allow_redirects=True,
            proxies=None, hooks=None, stream=None, verify=None, cert=None):
```

requests/sessions.py
```diff
@@ -544,80 +552,90 @@ def request(self, method, url,
...
 
         :param url: URL for the new :class:`Request` object.
+        :param params: (optional) Dictionary or bytes to be sent in the query 
+            string for the :class:`Request`.
         :param \*\*kwargs: Optional arguments that ``request`` takes.
+        :return: :class:`Response <Response>` object
         :rtype: requests.Response
         """
 
-        kwargs.setdefault('allow_redirects', True)
```
weddige, line 554: This line is what motivated me to this pull request. I removed this line here and in `options`, as its already the default.

requests/sessions.py
```diff
@@ -543,81 +551,91 @@ def request(self, method, url,
 
         return resp
 
-    def get(self, url, **kwargs):
-        r"""Sends a GET request. Returns :class:`Response` object.
+    def get(self, url, params=None, **kwargs):
```
weddige, line 554: I decided to add the positional parameter here instead of deleting it in `request.api`.

requests/api.py
```diff
@@ -123,7 +141,6 @@ def put(url, data=None, **kwargs):
     :param url: URL for the new :class:`Request` object.
     :param data: (optional) Dictionary, list of tuples, bytes, or file-like
         object to send in the body of the :class:`Request`.
-    :param json: (optional) json data to send in the body of the :class:`Request`.
```
weddige, line 126: The alternative would have been to introduce `json` as positional argument, but as it is currently neither in `requests.api` nor `requests.sessions` I just updated the doc string.

sigmavirus24:
Changing function argument definitions to attempt to match things is not something we're going to accept anytime soon, especially not before the next major version. Further, attempting to try to synchronize these docstrings is going to be a futile effort unfortunately as they're going to drift by nature of them being separate strings.

PR #5881, opened 2021-07-19
docs: Update the Red Hat vulnerabilities contact

jeremycline:
I am departing Red Hat so it would be best to contact the Python
maintenance team there instead of me if a vulnerability is discovered.

Cc: @hroncok


hroncok (approved):
I am a member of that team and I approve :)

nateprewitt (approved):
Thanks for letting us know, @jeremycline. Best of luck with whatever's next!

PR #5890, opened 2021-07-28
Fix creating non-listening sockets in tests on some platforms

mgorny:
Fix the listen() invocation for the test server not to pass a backlog
value of zero.  The value of zero means no backlog which effectively
means that the socket can not accept any connections.  This does not
matter for the majority of platforms since the value is only advisory
and the platform tends to go with a bigger backlog anyway.  However,
a few platforms (e.g. alpha or riscv Linux) do take the value literally,
and therefore the tests fail since they are unable to connect to
the server.  Let Python use a 'default reasonable value' instead.

### DIFF
--- a/tests/testserver/server.py
+++ b/tests/testserver/server.py
@@ -78,7 +78,9 @@ class Server(threading.Thread):
     def _create_socket_and_bind(self):
         sock = socket.socket()
         sock.bind((self.host, self.port))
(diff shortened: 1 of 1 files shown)


mgorny:
I'm sorry about the failure, it didn't occur to me to test on Python 2.

sigmavirus24 (approved):

Issue #5892, opened 2021-07-28
Non-standard HTTP status codes?


jenstroeger:
According to [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status), the following status codes are not defined yet available through `requests`

https://github.com/psf/requests/blob/b227e3cb82c8d42ea27790d615ecafe12528d3bc/requests/status_codes.py#L32
https://github.com/psf/requests/blob/b227e3cb82c8d42ea27790d615ecafe12528d3bc/requests/status_codes.py#L84-L87
https://github.com/psf/requests/blob/b227e3cb82c8d42ea27790d615ecafe12528d3bc/requests/status_codes.py#L89
https://github.com/psf/requests/blob/b227e3cb82c8d42ea27790d615ecafe12528d3bc/requests/status_codes.py#L100

The above `444` and `499` seem to be from the [unofficial nginx errors](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#nginx)?

In addition, [`508` Loop Detected](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/508) is missing from that list.

I think it would help to document this in some more detail.

sigmavirus24:
> I think it would help to document this in some more detail.

What detail do you feel is required? Would it be simpler to ignore non-official but fairly widely used status codes here and remove them if this is so confusing?

steveberdy:
In addition to some documentation explaining the source of the extra codes, perhaps we can have other dictionaries of codes, such as an `_nginx_codes` dictionary that can be, by default, added to the `_codes` dictionary at [init](https://github.com/psf/requests/blob/b227e3cb82c8d42ea27790d615ecafe12528d3bc/requests/status_codes.py#L107). If users don't want nginx codes, they can set an environment variable or use some method to set a preference at runtime.
Would this make sense?
It would make the error codes available more modular, and wouldn't break backwards compatibility, as the full list of codes will be used by default.

jenstroeger:
Makes a lot of sense 👍🏼 

Want a PR for that?

nateprewitt:
[...]

PR #5895, opened 2021-07-30
Clean up HTTP status code definitions

jenstroeger:
Follow-up to the conversation in issue https://github.com/psf/requests/issues/5892.

As is, this change is not a breaking change and produces the exact same `_codes` dictionary. However, I’ve separated out the official HTTP status codes from third-party status codes and added references to relevant documentation.

It would probably make sense to at least complete the list of [nginx codes](https://www.nginx.com/resources/wiki/extending/api/http/); [IIS codes](https://docs.microsoft.com/en-us/troubleshoot/iis/http-status-code) have non-integer sub-codes which may be problematic to represent here.

### DIFF
--- a/requests/status_codes.py
+++ b/requests/status_codes.py
@@ -22,14 +22,16 @@ the names are allowed. For example, ``codes.ok``, ``codes.OK``, and
 
 from .structures import LookupDict
 
-_codes = {
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
+_official = {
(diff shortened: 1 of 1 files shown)


nateprewitt:
I think the feedback here is the same as in the issue. We don't have an interest in expanding on what we already feel was a poor decision. It encourages users to continue (or start) building on a non-core feature, which will create more user pain if it is removed/consolidated in the future.

PR #5908, opened 2021-08-22
fix typo

hongkuancn:
byte string is `b'test'`, while unicode string is `'test'`

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -514,8 +514,8 @@ class TestRequests:
         assert p.headers['Authorization'] == _basic_auth_str(username, password)
 
     def test_basicauth_encodes_byte_strings(self):
-        """Ensure b'test' formats as the byte string "test" rather
-        than the unicode string "b'test'" in Python 3.
+        """Ensure b'test' formats as the byte string "b'test'" rather
+        than the unicode string "test" in Python 3.
         """
         auth = (b'\xc5\xafsername', b'test\xc6\xb6')
         r = requests.Request('GET', 'http://localhost', auth=auth)


tests/test_requests.py
```diff
@@ -514,8 +514,8 @@ def test_set_basicauth(self, httpbin, username, password):
         assert p.headers['Authorization'] == _basic_auth_str(username, password)
 
     def test_basicauth_encodes_byte_strings(self):
-        """Ensure b'test' formats as the byte string "test" rather
-        than the unicode string "b'test'" in Python 3.
+        """Ensure b'test' formats as the byte string "b'test'" rather
+        than the unicode string "test" in Python 3.
```
sigmavirus24, line 518: This is entirely correct. 

```
Python 3.6.12 (default, Dec 24 2020, 11:04:11)
[GCC 10.2.1 20201125 (Red Hat 10.2.1-9)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import base64
>>> base64.b64decode('xa9zZXJuYW1lOnRlc3TGtg==')
b'\xc5\xafsername:test\xc6\xb6'
>>> _.decode()
'ůsername:testƶ'
>>>
```

Note that the encoded username and password is

```
xa9zZXJuYW1lOnRlc3TGtg==
```

If what you were saying was correct it should be 

```
>>> auth = tuple(str(a) for a in auth)
>>> auth
("b'\\xc5\\xafsername'", "b'test\\xc6\\xb6'")
>>> base64.b64encode(":".join(auth).encode())
b'YidceGM1XHhhZnNlcm5hbWUnOmIndGVzdFx4YzZceGI2Jw=='
```

hongkuancn: Thanks for the reply, but I think you misunderstood my point. I am not saying the code is wrong. I just modified the comment. The original comment means `"test"` is byte string and `b'test'` is unicode string. That is not correct. It should be the other way around.

sigmavirus24: No, you misunderstand the docstring. `b'test'` in the start of the sentence is the Python object. When you call `str(b'test')` that will become `"b'test'"` because Python 3 will not auto-decode the byte string whereas what we want to ensure happened was `b'test'.decode('utf-8')` which will result in `"test"`. This is the exact scenario the docstring is describing. You're the one who is confused

PR #5909, opened 2021-08-25
Add thread locker workflow

sigmavirus24:


.github/workflows/lock-issues.yml
```diff
@@ -0,0 +1,15 @@
+me: 'Lock Threads'
```
nateprewitt: It looks like this got clipped during copy. `me`->`name`

nateprewitt (changes requested):
Just a minor fix needed and then I think this looks good. Thanks, @sigmavirus24!

.github/workflows/lock-issues.yml
```diff
@@ -0,0 +1,15 @@
...
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
```
nateprewitt, line 15: Actually one more thought while we're here. It looks like some other projects have [shortened the window](https://github.com/pyca/pyopenssl/blob/main/.github/workflows/lock.yml#L13-L14) from 365 days to 90. That may be more reasonable to prevent unrelated pings resetting such a long timer.

nateprewitt (approved):
🚀

.github/workflows/lock-issues.yml
sigmavirus24: Oh, I missed this. Hold on. #5911

PR #5917, opened 2021-08-27
Move from urlparse to parse_url for prepending schemes

nateprewitt:
This addresses the base problem raised in #5855 with the `urlparse` changes in Python 3.9 and potential issues in future 3.7/3.8 releases. We've avoided changing other uses of `urlparse` for now due to issues with `parse_url`'s strict validation requirements when performing parsing.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -21,6 +21,7 @@ import warnings
 import zipfile
 from collections import OrderedDict
 from urllib3.util import make_headers
+from urllib3.util import parse_url
 
 from .__version__ import __version__
 from . import certs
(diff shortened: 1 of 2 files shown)


geofft:
Thanks for this! I've pulled in this patch into our internal monorepo and some preliminary testing shows that it's working (I can `export https_proxy=some-proxy.example.com:3128`, like a lot of our scripts do, and things work again under Python 3.9). I'll hopefully get it deployed today and can report on whether our users can successfully switch to 3.9.

geofft:
We've rolled this out and it works in prod.

nateprewitt:
Thanks for checking, @geofft! I think we just need another maintainer to give this a glance and we can look at moving the review forward.

requests/utils.py
```diff
@@ -932,15 +933,22 @@ def prepend_scheme_if_needed(url, new_scheme):
...
+    parsed = parse_url(url)
+    scheme, auth, host, port, path, query, fragment = parsed
 
-    # urlparse is a finicky beast, and sometimes decides that there isn't a
-    # netloc present. Assume that it's being over-cautious, and switch netloc
-    # and path if urlparse decided there was no netloc.
+    # urlparse and parse_url determine that there isn't a netloc present in some
+    # urls. We've chosen to assume parsing is being over-cautious, and switch
+    # the netloc and path. This is maintained for backwards compatibility.
+    netloc = parsed.netloc
```
sethmlarson, line 974: Uhhh, what? This doesn't make any sense to me, is this something that's only needed for `urlparse`?

nateprewitt: Yeah, this is only a quirk for `urlparse`. I'd thought I repro'd with `parse_url` initially, but I'm unable to now.

I'll reword the whole comment for clarity. I'm not confident the test suite covers all of the initial issue though. I think I'm still in favor of leaving the swap in place as a fallback for anything we're missing.

nateprewitt: @sethmlarson I've reworded things to be a bit clearer.

sigmavirus24 (approved):

PR #5923, opened 2021-09-02
General doc updates

nateprewitt:
General fixes for documentation including:
* Adding alt tags to images
* Fixing doc compilation errors
* Updating broken links
* Updating info in README.md
* Updating target branch for repo URLs
* Adding logo back to docs sidebar per Kenneth's request


sigmavirus24 (approved):
I highly doubt anything is actually binding and as far as logos go, we would be served by not having one that's offensive to people. Let's add it back for now and figure out what to do later about making this an actually inclusive project

PR #5924, opened 2021-09-02
Refactor rebuild_proxies to separate resolution and auth handling

nateprewitt:
This is a proposal for handling #5888 that's come out of discussion in an alternative PR (#5893). Given that we've stalled out on progress there for a few weeks, this PR is to help drive a conclusion. The approach taken here is to mitigate the incorrect behavior of stripping `Proxy-Authorization` headers off all requests sent with `Session.send`. This will not address the performance concerns from #5891 as that's a more significant problem without a clear answer.

The goal of driving this separately from #5891 is that one is an unintended breakage in basic behavior of the library and the second is an unfortunate performance regression. I'd like to address the former more quickly if possible instead of letting the problem fester.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -29,7 +29,7 @@ from .adapters import HTTPAdapter
 
 from .utils import (
     requote_uri, get_environ_proxies, get_netrc_auth, should_bypass_proxies,
-    get_auth_from_url, rewind_body
+    get_auth_from_url, rewind_body, resolve_proxies
 )
 
 from .status_codes import codes
@@ -269,7 +269,6 @@ class SessionRedirectMixin(object):
         if new_auth is not None:
             prepared_request.prepare_auth(new_auth)
 
-
     def rebuild_proxies(self, prepared_request, proxies):
         """This method re-evaluates the proxy configuration by considering the
         environment variables. If we are redirected to a URL covered by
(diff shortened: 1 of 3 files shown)


requests/sessions.py
```diff
@@ -633,7 +620,10 @@ def send(self, request, **kwargs):
         kwargs.setdefault('stream', self.stream)
         kwargs.setdefault('verify', self.verify)
         kwargs.setdefault('cert', self.cert)
-        kwargs.setdefault('proxies', self.rebuild_proxies(request, self.proxies))
+        if 'proxies' not in kwargs:
```
omermizr, line 623: This change should actually fix the performance regression in most cases (`Session.request` - which I assume is the most common flow - always sets `proxies` on `kwargs`).

requests/sessions.py
```diff
@@ -633,7 +620,10 @@ def send(self, request, **kwargs):
...
+            kwargs['proxies'] = resolve_proxies(
```
omermizr, line 624: So now we're:
1. No longer stripping the proxy-auth header. Sounds right to me.
2. No longer setting the proxy-auth header at all. Is this the desired behavior? What if someone uses a proxy and passes the username + password in the url? (though I'm not sure if that flow is supported or not)

requests/utils.py
```diff
@@ -830,6 +830,34 @@ def select_proxy(url, proxies):
...
+
+    :rtype: dict
+    """
+    proxies = proxies if proxies is not None else {}
+    url = request.url
+    scheme = urlparse(url).scheme
+    no_proxy = proxies.get('no_proxy')
+    new_proxies = proxies.copy()
+
+    bypass_proxy = should_bypass_proxies(url, no_proxy=no_proxy)
```
omermizr, line 850: Didn't we want to unify 850 and 851 so we don't call `should_bypass_proxies` if `trust_env` is `False`?

omermizr:
Hey @nateprewitt, thanks for this!
I added a few comments following up on our discussions.

omermizr (commented):
Hey @nateprewitt, thanks for this!
I added a few comments following up on our discussions.

requests/sessions.py
nateprewitt: For 2.) we were never doing that before this change and I don't think that was an intended byproduct. We should still support our normal proxy auth flows that were available prior to #5681.

requests/utils.py
nateprewitt: We do, I was going to have @dbaxa rebase their change onto whatever we merge so they can still have a commit in the repo.

requests/sessions.py
nateprewitt: Yep, I think `proxies` will be the escape hatch when performance is a concern here :)

Tabbbbs (changes requested):
[](https://github.com/psf/requests/pull/5924/files)

requests/utils.py
dbaxa: Sounds good to me.

dbaxa (approved):

dbaxa:
When can we merge this @nateprewitt ?

nateprewitt:
I believe we're waiting on either some consensus with this proposal or feedback for another direction. @sigmavirus24 or @sethmlarson, do you have any thoughts on how we proceed with #5888?

sigmavirus24 (approved):

PR #5928, opened 2021-09-03
Add support for 3.10

nateprewitt:
This will make sure we don't introduce any regressions into the code base prior to the official 3.10 release. At that point, we'll promote this to the normal test runner on all platforms.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [2.7, 3.6, 3.7, 3.8, 3.9]
+        python-version: [2.7, 3.6, 3.7, 3.8, 3.9, 3.10-dev]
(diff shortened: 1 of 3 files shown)


.github/workflows/run-tests.yml
```diff
@@ -16,6 +16,8 @@ jobs:
           # brotlipy. Moving pypy3 to only test linux.
           - python-version: pypy3
             os: ubuntu-latest
+          - python-version: 3.10.0-beta.4
```
sethmlarson: Would setting this to `3.10-dev` mean we get release candidates (or beyond)?

nateprewitt: That's a great question! I'd just pulled this out of the urllib3 without much research, let me take a look.

nateprewitt: `3.10-dev` it is! Thanks for the callout, I just moved it into the main runner list for all platforms since it's already passing. 🎉

sethmlarson (approved):

PR #5933, opened 2021-09-15
ci: optimize ci workflow

willin:
ignore paths that do not need to be tested.


sigmavirus24:
Thank you for sending this but we'll likely not accept this. We used to have CI jobs that verified docs built properly. We'll likely reinstate that in the future

PR #5938, opened 2021-09-18
Update README Logos

FayasNoushad:
- Updated logos to center


sigmavirus24:
Hi there, thanks for sending a pull request! As a rule we don't accept minor cosmetic updates like this.

fayasnoushad:
> Hi there, thanks for sending a pull request! As a rule we don't accept minor cosmetic updates like this.

Ok, No problem.

PR #5940, opened 2021-09-21
Don't mention pip as the reason for supporting py2

verhovsky:
since pip no longer supports Python 2

### DIFF
--- a/docs/community/faq.rst
+++ b/docs/community/faq.rst
@@ -63,9 +63,8 @@ Python 2 Support?
 Yes! We do not have immediate plans to `sunset
 <https://www.python.org/doc/sunset-python-2/>`_ our support for Python
 2.7. We understand that we have a large user base with varying needs,
-and intend to maintain Python 2.7 support within Requests until `pip
-stops supporting Python 2.7 (there's no estimated date on that yet)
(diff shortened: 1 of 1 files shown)


verhovsky:
I don't know if that is actually the official plan, but I get the impression that that's what's going to happen from this comment https://github.com/psf/requests/pull/5660#issuecomment-729236559

nateprewitt:
Hi @verhovsky, thanks for the PR. I'm not sure that comment was related to Python 2.7's deprecation. We had initially kept it as a requirement for `pip` but now that that's gone, we don't have a strong reason to keep it. It's just a matter of figuring out when we decide to formally stop Python 2.7 support. That likely won't happen before a major version bump of Requests.

If you'd like to replace the reference to `pip` with something more general like "and intend to maintain Python 2.7 support within Requests until otherwise announced.", I'm fine merging that.

verhovsky:
@nateprewitt how about just removing that part? "We do not have immediate plans" and "[we] intend to maintain Python 2.7 support within Requests until otherwise announced" are basically saying the same thing twice.

nateprewitt (approved):
:+1: Great, I think this is fine for the time being. Thanks!

PR #5942, opened 2021-09-23
Pass `chunk_size` to `Response.iter_*` methods in a consistent way

ofek:
Use case: we would like to set a default `chunk_size` as shown in the new test.

Current behavior errors (as shown in the first test-only commit):

```
>               self._content = b''.join(self.iter_content(CONTENT_CHUNK_SIZE)) or b''
E               TypeError: iter_content() got multiple values for argument 'chunk_size'

.tox\default\lib\site-packages\requests\models.py:838: TypeError
```

The alternative would be adding a new instance attribute to serve as a fallback but I doubt new features are being considered before the next major release and we are currently blocked on this https://github.com/DataDog/integrations-core/pull/10183

cc @nateprewitt WDYT?

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -697,7 +697,7 @@ class Response(object):
 
     def __iter__(self):
         """Allows you to use a response as an iterator."""
-        return self.iter_content(128)
+        return self.iter_content(chunk_size=128)
 
     @property
     def ok(self):
@@ -835,7 +835,7 @@ class Response(object):
             if self.status_code == 0 or self.raw is None:
                 self._content = None
             else:
-                self._content = b''.join(self.iter_content(CONTENT_CHUNK_SIZE)) or b''
+                self._content = b''.join(self.iter_content(chunk_size=CONTENT_CHUNK_SIZE)) or b''
 
         self._content_consumed = True
         # don't need to release the connection; that's been handled by urllib3
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -10,6 +10,7 @@ import collections
 import contextlib
 import warnings
 import re
+import functools
 
 import io
 import requests
@@ -1984,6 +1985,12 @@ class TestRequests:
         assert not r.history[1].is_redirect
         assert r.url == urls_test[2]
 
+    def test_default_streaming_chunk_size_allows_full_read(self, httpbin):
+        r = requests.get(httpbin('bytes/5'), stream=True)
+        r.iter_content = functools.partial(r.iter_content, chunk_size=3)
+
+        assert len(r.content) == 5
+
 
 class TestCaseInsensitiveDict:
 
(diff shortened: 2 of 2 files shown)


nateprewitt (approved):
Hey @ofek,

Thanks for the PR! I think this is pretty reasonable solution, I'm in favor of being more explicit here. I'd like to make sure @sigmavirus24 is in agreement, and then I'm fine merging.

sigmavirus24:
I'm actually quite hesitant to accept this. In the past when we've changed how we're invoking a method on a class that people interact a lot with it always breaks someone who is doing something quite like what's happening in the associated PR. It's happened countless times in the past to the point we've rejected PRs like this out of hand without further discussion or elaboration. (We take backwards compatibility a bit more seriously given how much of an impact even the slightest change can have on the wider ecosystem.)

People assume the best way to get their custom behavior is to rely on the underlying library to almost never change and we do need to make changes from time to time.

As a rule, I prefer using named arguments but I don't think accepting this change will go well and there's nothing to prevent it from regressing that explains why we want to keep this behaviour this making it an easy test to delete.

Datadog is already providing a wrapper around requests today. I'd argue that folks that want to change the default chunk size should be providing a wrapper around responses too. Think collaboration over inheritance or monkey-patching.

ofek:
Would the addition of an instance attribute that serves as a fallback be accepted? If not, is there an existing mechanism for requests to return custom response objects?

sigmavirus24:
Chunk size isn't a session-level property. It's a per-response property.

As I said, a Response wrapper that takes care of this for you is the right way to go. Think collaboration over inheritance.

ofek:
> Chunk size isn't a session-level property. It's a per-response property.

Yes I mean e.g.:

```python
class Response(object):
    def __init__(self):
        ...
        self.default_chunk_size = None

    def iter_lines(self, chunk_size=None, decode_unicode=False, delimiter=None):
        if chunk_size is None:
            if self.default_chunk_size is None:
                chunk_size = ITER_CHUNK_SIZE
            else:
                chunk_size = self.default_chunk_size

        for chunk in self.iter_content(chunk_size=chunk_size, decode_unicode=decode_unicode):
            ...
```

> As I said, a Response wrapper that takes care of this for you is the right way to go. Think collaboration over inheritance.

Do you mean composition like this?

```python
class CustomResponse:
    def __init__(self, response):
        self.__response = response

    def __getattr__(self, name):
        attr = getattr(self.__response, name)
        setattr(self, name, attr)
        return attr
```

sigmavirus24:
> Do you mean composition like this?

Yes. You can then override `iter_content` with logic to wrap the `__response` logic. That will also be your most reliable solution across versions of Requests that DataDog is supporting as well

ofek:
I'm not a huge fan of adding a dependency (the aforementioned naïve approach was insufficient) but it does work: https://github.com/DataDog/integrations-core/pull/10183/commits/b57857c6c88592d245471f91fd780134c2f95de0

Thank you both for your time!

PR #5953, opened 2021-10-08
Add ipv6 support to should_bypass_proxies

derekhiggins:
Add support to should_bypass_proxies to support
IPv6 ipaddresses and CIDRs in no_proxy. Includes
adding IPv6 support to various other helper functions.

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -723,18 +723,43 @@ def requote_uri(uri: str) -> str:
         return quote(uri, safe=safe_without_percent)
 
 
+def _get_mask_bits(mask: int, totalbits: int = 32) -> int:
+    """Converts a mask from /xx format to a int to be used as a mask for IP's in int format
+
+    Example: if mask is 24 function returns 0xFFFFFF00
+             if mask is 24 and totalbits=128 function
+                returns 0xFFFFFF00000000000000000000000000
+    """
+    bits = ((1 << mask) - 1) << (totalbits - mask)
+    return bits
+
+
 def address_in_network(ip: str, net: str) -> bool:
     """This function allows you to check if an IP belongs to a network subnet
 
     Example: returns True if ip = 192.168.1.1 and net = 192.168.1.0/24
              returns False if ip = 192.168.1.1 and net = 192.168.100.0/24
+             returns True if ip = 1:2:3:4::1 and net = 1:2:3:4::/64
 
     :rtype: bool
     """
-    ipaddr = struct.unpack("=L", socket.inet_aton(ip))[0]
     netaddr, bits = net.split("/")
-    netmask = struct.unpack("=L", socket.inet_aton(dotted_netmask(int(bits))))[0]
-    network = struct.unpack("=L", socket.inet_aton(netaddr))[0] & netmask
+    if is_ipv4_address(ip) and is_ipv4_address(netaddr):
+        ipaddr = struct.unpack(">L", socket.inet_aton(ip))[0]
+        netmask = _get_mask_bits(int(bits))
+        network = struct.unpack(">L", socket.inet_aton(netaddr))[0]
+    elif is_ipv6_address(ip) and is_ipv6_address(netaddr):
+        ipaddr_msb, ipaddr_lsb = struct.unpack(
+            ">QQ", socket.inet_pton(socket.AF_INET6, ip)
+        )
+        ipaddr = (ipaddr_msb << 64) ^ ipaddr_lsb
+        netmask = _get_mask_bits(int(bits), 128)
+        network_msb, network_lsb = struct.unpack(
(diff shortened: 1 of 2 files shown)


requests/utils.py
```diff
@@ -679,31 +703,58 @@ def is_ipv4_address(string_ip):
...
+    """
+    Compare 2 IP's, uses socket.inet_pton to normalize IPv6 IPs
+
+    :rtype: bool
+    """
+    if a == b:
+        return True
+    try:
+        return socket.inet_pton(socket.AF_INET6, a) == socket.inet_pton(socket.AF_INET6, b)
+    except:
```
dtantsur: nit: use `socket.error`?

dtantsur (approved):

requests/utils.py
derekhiggins: done, thanks

requests/utils.py
```diff
@@ -648,18 +648,42 @@ def requote_uri(uri):
...
+             returns True if ip = 1:2:3:4::1 and net = 1:2:3:4::/64
 
     :rtype: bool
     """
-    ipaddr = struct.unpack('=L', socket.inet_aton(ip))[0]
     netaddr, bits = net.split('/')
-    netmask = struct.unpack('=L', socket.inet_aton(dotted_netmask(int(bits))))[0]
-    network = struct.unpack('=L', socket.inet_aton(netaddr))[0] & netmask
+    if is_ipv4_address(ip) and is_ipv4_address(netaddr):
+        ipaddr = struct.unpack('>L', socket.inet_aton(ip))[0]
```
derekhiggins: I've tested this switch to 'big-endian' from 'native' on both big and little endian systems. Unit tests passed on both.

derekhiggins:
Hi @nateprewitt @sethmlarson , is this something you'd consider merging ?

melewitz:
I'm facing this same issue.  @nateprewitt @sethmlarson Any update on your thoughts?

novacain1:
Running into the same problem with this, it sure would be nice to have the ability to specify IPv6 CIDRs in no_proxy.  What's holding this up?

derekhiggins:
@nateprewitt @sigmavirus24 @sethmlarson would you have a chance to look at this please, its been hanging around for over 2 years and still causing me problems.

frenzymadness:
It'd be nice to have this implemented. Is there any way I can help?

derekhiggins:
Hi @frenzymadness , if you can review or confirm that this works for your use case then that would be appreciated but
ultimately this PR has been waiting on reviews  from one of the maintainers I'm not sure how best to get it some traction besides pinging them on here.

src/requests/utils.py
```diff
@@ -674,18 +674,46 @@ def requote_uri(uri):
...
+            ">QQ", socket.inet_pton(socket.AF_INET6, ip)
+        )
+        ipaddr = (ipaddr_msb << 64) ^ ipaddr_lsb
+        netmask = _get_mask_bits(int(bits), 128)
+        network_msb, network_lsb = struct.unpack(
+            ">QQ", socket.inet_pton(socket.AF_INET6, netaddr)
+        )
+        network = (network_msb << 64) ^ network_lsb
+    else:
+        return False
```
frenzymadness, line 762: Is returning `False` the correct thing to do when this function is called with IPv4 IP and IPv6 network or vice versa? What about raising an exception here?

frenzymadness (commented):
I checked the code and it looks good to me. I'm looking for more ways to test it.

src/requests/utils.py
derekhiggins: I think this is fine as is (i.e. returning false), the function is used to test if a IP address is a member of any of the CIDR's listed in no_proxy, having a no_proxy with mixed IPv4 and IPv6 CIDR's would be valid, I don't think I'd expect an exception to the raised in this instance.

frenzymadness:
We are going to test this change internally to get some confidence.

frenzymadness:
Derek tested the package in OpenShift 4.15 and the functionality proposed here works as expected. Is there any chance to merge this?

frenzymadness:
During additional testing, my colleague raised an interesting question: Are `::/0` and `0.0.0.0/0` valid values for `no_proxy`? The current implementation for IPv4 does not allow masks to be lower than 1 and the new implementation for IPv6 does the same.

I see a use-case for `::/0` - let's say I have a proxy running on both protocols but I want it to be used only for IPv4 - then it might make sense to use `NO_PROXY` for all IPv6 addresses. What do you think?

frenzymadness:
What can I do to get this merged? We've reviewed the change and tested it in a real environment with real use cases. Is there anything else we can do?

derekhiggins:
Rebased onto current main.

Also worth noting - this patch has been carried in Fedora's python3-requests package since at least 2024.
It would be good to get it merged upstream rather than continued as a downstream patch.

Issue #4322, opened 2017-10-02
Reuse ssl.SSLContext across all SSL connections within a requests.Sesssion


alex:
requests should pass `ssl_context` to `urllib3.PoolManager`: https://github.com/requests/requests/blob/2a4dd64fb4c072c646277c9d057737b05351b86f/requests/adapters.py#L161-L162

Otherwise a new SSLContext is allocated for all requests. SSLContexsts can be huge, like 1MB, so doing this will be a performance and memory usage win.

Stevoisiak:
It looks like Requests doesn't define `ssl_context` anywhere. Where should it be created?

wagnerluis1982:
How is the status of this issue?

echo4eva:
Hello, me and my group mates are going to be working on this issue.

Echo-Lima-India:
@alex Could you tell me where you have an instance of ssl_context within the code? Thank you.

nateprewitt:
This should be largely addressed in 9a40d1277807f0a4f26c9a37eea8ec90faa8aadc.

PR #5971, opened 2021-11-03
Reusing SSLContext in a session

perilEcho:
A pull request to take a hit at fixing issue #4322 

Worked with my team and we made it so that HTTPAdapter has a default SSLContext initialized when the adapter is constructed. SSLContext is made by using urllib3's function to create context. The context is passed into the poolmanager as pool_kwargs which can take in SSLContext.

Tested the reuse of SSLContext by using jmcackle's test functions they've made for adapters.


nateprewitt:
I believe this ended up addressed in 9a40d1277807f0a4f26c9a37eea8ec90faa8aadc.

PR #5982, opened 2021-11-23
Update sessions.py

junqfisica:
This change will make the hierarchy request **kwargs -> Session args -> environment to be respected. This also solves problems with pip as discussed [here](https://github.com/pypa/pip/issues/9691#issuecomment-791608247)  and  #5735

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -527,7 +527,7 @@ class Session(SessionRedirectMixin):
         )
         prep = self.prepare_request(req)
 
-        proxies = proxies or {}
+        proxies = proxies or self.proxies
 
         settings = self.merge_environment_settings(
             prep.url, proxies, stream, verify, cert


sigmavirus24:
Read #5735 more closely please. There are certainly people who are unknowingly relying on this behaviour in 2.x and we cannot break them suddenly. It's worse because even if we document this they will still be surprised as they likely have no understanding that they're affected

junqfisica:
@sigmavirus24 Thank you for the comment but I disagree this is a wanted behavior. The current status of Session breaks the usage of pip when behind a proxy and the change I proposed is not the same as in #5735. My change keeps the hierarchy the way it should be as already mentioned in the documentation.

The biggest problem right now is that when you use 

 `pip install package --proxy=http\\:proxy_url` 

The way they inject proxy to request is by creating a session object and setting the proprety proxies to it:
```
from requests import Request, Session

proxies = { "http"  : http_proxy, "https" : https_proxy}
session = Session()
session.proxies = proxies
session.request(url)
````

The piece of code above doesn't work as expected in many proxy environments, because Request is actually overriding the given proxies by the system proxies and due to changes in urlib3 it doesn't work if you are doing an HTTPS request using an HTTP proxy server. Many developers can't directly change environment proxy. Thus, due to this problem newer versions of pip (> 20.2.x) can't install packages using the proxy. 

I agree that pip could solve that by giving the proxy directly to request like: 

```
from requests import Request, Session

proxies = { "http"  : http_proxy, "https" : https_proxy}
session = Session()
session.proxies = proxies
session.request(url, proxies)
````

Although, since you allow to set proxies like: `session.proxies = proxies` that SHOULD be respected. 
Saying that I don't see how my proposed change would break anything, since I pinpoint the problem of overriding the proxy. I would please ask for careful consideration on this change.

sigmavirus24:
> Thank you for the comment but I disagree this is a wanted behavior.

Never said wanted. I said unknowingly relied upon. An application is working today that very well may break in confusing and unexpected ways after releasing this. The behaviour is wrong, we intend to fix it in a deliberately breaking release when we can dedicate time towards such a development effort. It's not something we can fix in the 2.x release series.

This change has been carefully considered for as long as the original PR has been open.

PR #5983, opened 2021-11-23
Update docs with for passing an array as a param

yashshah1:
When passing an array for a `GET` request, some servers expect the key to be suffixed with a `[]`
to parse the URL correctly. Adding a note in the documentation for this case.

Further, Is it a good idea to make this behavior a default one? Concatenate a `[]` to the end of all keys
whose value is an array. Would love to hear your thoughts!


sigmavirus24:
> When passing an array for a `GET` request, some servers expect the key to be suffixed with a `[]`
> to parse the URL correctly. Adding a note in the documentation for this case.

We can't and won't document every non-standard edge case that exists. We expect users to understand the services they're talking to and cater to them directly.



> Further, Is it a good idea to make this behavior a default one?

No. It's not a good idea to change behaviour in completely backwards incompatible ways

PR #5990, opened 2021-11-27
Skip the test requiring trustme when it is not available

mgorny:
Modify the nosan_server fixture to cause test_https_warnings to be
skipped when trustme is not installed on the system, rather than causing
the whole test suite to fail unconditionally.  This makes it possible
to run all the remaining tests on systems where trustme cannot be
installed due to its dependencies.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,7 +13,6 @@ import threading
 
 import pytest
 from requests.compat import urljoin
-import trustme
 
 
 def prepare_url(value):
@@ -38,6 +37,10 @@ def httpbin_secure(httpbin_secure):
 
 @pytest.fixture
 def nosan_server(tmp_path_factory):
+    # delay importing until the fixture in order to make it possible
+    # to deselect the test via command-line when trustme is not available
+    import trustme
+
     tmpdir = tmp_path_factory.mktemp("certs")
     ca = trustme.CA()
     # only commonName, no subjectAltName


sigmavirus24:
If one cannot install `idna` or `cryptography`, I suspect there are bigger issues that will prevent requests from working as securely or reliably as users expect. I'm not inclined to skip this test and give those users a false sense of security.

mgorny:
Don't requests work well with the Python stdlib OpenSSL support? My grep-foo may be weak here but I don't really see any actual code referencing cryptography (besides some version check and info), and idna doesn't seem to require cryptography.

In other words, I can't find anything requiring it but this one test.

sigmavirus24:
_Right now_ we only use the standard library. That isn't a guarantee going into the future. Generally speaking, architectures that can't install cryptography seem to have limited security support in general. That is a good indicator that use of relatively modern cipher suites won't continue to work and our tests passing on those platforms and architectures are going to end up being a lie. 

Also, this project doesn't claim to support those architectures or platforms meaning we don't need to skip tests on them because we can't even test easily on them. `alpha, hppa, ia64, m68k, s390 (not -x)` are highly niche architectures, we don't skip tests for downstreams because they attempt to support more than we can reasonably support.

mgorny:
Would it work for you if I changed the PR to just move the `import` inside the fixture, so that we would be able to skip that test via command-line without changing the default test suite behavior?

sigmavirus24:
@mgorny that would work

mgorny:
Ok, done! I've added a comment to explain why it's not in global scope.

sethmlarson (approved):
Seems good to me, thanks for this!

nateprewitt (approved):
Fine by me as well 🚀

sigmavirus24 (approved):

PR #5993, opened 2021-11-30
Run tox in parallel using -p instead of detox

davisagli:
This updates `make test` to run tests in parallel using `tox -p` instead of `detox`. This has been supported since tox 3.7.0 and detox's PyPI page recommends switching to it.


sigmavirus24:
@davisagli is this the fork you wanted to target?

davisagli:
@sigmavirus24 for this one, yes

sigmavirus24 (approved):

PR #5995, opened 2021-12-01
Add Access Token Authorization class

kleszcz:
Added class for attaching access token to request object as an alternative for basic auth. I find it useful for services where OAuth is not an option.


sigmavirus24:
We're not accepting new features at this time

PR #5998, opened 2021-12-05
Update README.md

vashisthabhinav:
Grammatical errors corrected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 {'disk_usage': 368627, 'private_gists': 484, ...}
 ```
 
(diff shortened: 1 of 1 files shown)


FabianSt305 (changes requested):
Requests is a name.

Requests *is* available on PyPI.
Just like: Five Guys *is* a fast food chain.

FabianSt305:
"Allows" -> "Allow" has the same issue. 

Only the "package" error was actually an error I think.

nateprewitt (approved):
Thanks @vashisthabhinav

PR #6003, opened 2021-12-09
Rejecting URLs with Unverified SSL Certs

Mr1Vgy:
Fixes an issue with unverified SSL certificates not being rejected


sigmavirus24:
Non-TLS URLs will not send certificates to verify. Why do you think this library should reject those URLs no matter what?

Spoiler: We will *not* be accepting this PR

nateprewitt:
Thanks for the proposal @Mr1Vgy, but unfortunately this breaks all non-https URIs, which is not something we intend to change. I'm resolving since we won't be able to accept this.

PR #6011, opened 2021-12-19
Add success attribute for Response objects

Chrovo:
Add an attribute that returns True when a status code is a success status code(in between 200 and 299 inclusive).

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -933,6 +933,11 @@ class Response(object):
                 l[key] = link
 
         return l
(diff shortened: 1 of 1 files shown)


sethmlarson:
Thanks for opening. I think we're going to decline this PR unfortunately. If users want to make decisions based on the status code they should write this logic themselves instead of using a property that requires future  readers to learn  its implementation. Explicit is better than implicit!

PR #6012, opened 2021-12-20
Update the `ok` property

Chrovo:
Simplifying the code, I'm not sure if there was a specific reason for this design but my implementation should perform the same thing unless I'm misunderstanding.


sigmavirus24:
This allows folks to define their own response subclass that defines it's own raise_for_status and get an ok that conforms with that for free

PR #6036, opened 2022-01-07
Limit size of payload in JSONDecodeError

chyzzqo2:
2.27 introduced a change in behavior where now the exception raised by parsing invalid data as json contains the full body of the invalid response. This gets included it's string representation. This can cause problems when the response is very large. This PR tries to limit the size of the response that we store this way, to what might be around the reported error position. But we could also just return to first n bytes or remove the response altogether and let users fetch it, if needed from the error.response object.

### DIFF
--- a/requests/exceptions.py
+++ b/requests/exceptions.py
@@ -34,6 +34,16 @@ class InvalidJSONError(RequestException):
 class JSONDecodeError(InvalidJSONError, CompatJSONDecodeError):
     """Couldn't decode the text into json"""
 
+    def __init__(self, *args, **kwargs):
+        """
+        Construct the JSONDecodeError instance first with all
+        args. Then use it's args to construct the IOError so that
+        the json specific args aren't used as IOError specific args
+        and the error message from JSONDecodeError is preserved.
+        """
+        CompatJSONDecodeError.__init__(self, *args)
+        InvalidJSONError.__init__(self, *self.args, **kwargs)
+
 
 class HTTPError(RequestException):
     """An HTTP error occurred."""
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -23,14 +23,14 @@ from requests.cookies import (
     cookiejar_from_dict, morsel_to_cookie)
 from requests.exceptions import (
     ConnectionError, ConnectTimeout, InvalidSchema, InvalidURL,
-    MissingSchema, ReadTimeout, Timeout, RetryError, TooManyRedirects,
+    MissingSchema, ReadTimeout, Timeout, RetryError, RequestException, TooManyRedirects,
     ProxyError, InvalidHeader, UnrewindableBodyError, SSLError, InvalidProxyURL, InvalidJSONError)
 from requests.models import PreparedRequest
 from requests.structures import CaseInsensitiveDict
 from requests.sessions import SessionRedirectMixin
 from requests.models import urlencode
 from requests.hooks import default_hooks
-from requests.compat import MutableMapping
+from requests.compat import JSONDecodeError, is_py3, MutableMapping
 
 from .compat import StringIO, u
 from .utils import override_environ
@@ -2585,5 +2585,15 @@ class TestPreparingURLs(object):
 
     def test_json_decode_compatibility(self, httpbin):
         r = requests.get(httpbin('bytes/20'))
-        with pytest.raises(requests.exceptions.JSONDecodeError):
+        with pytest.raises(requests.exceptions.JSONDecodeError) as excinfo:
             r.json()
+        assert isinstance(excinfo.value, RequestException)
+        assert isinstance(excinfo.value, JSONDecodeError)
+        assert r.text not in str(excinfo.value)
+
+    @pytest.mark.skipif(not is_py3, reason="doc attribute is only present on py3")
+    def test_json_decode_persists_doc_attr(self, httpbin):
+        r = requests.get(httpbin('bytes/20'))
+        with pytest.raises(requests.exceptions.JSONDecodeError) as excinfo:
+            r.json()
+        assert excinfo.value.doc == r.text


sigmavirus24:
Is the problem that it gets included in the string representation or that it's included? If it's the former, I'm far more inclined to just not show it there.

chyzzqo2:
Mostly the fact that it is in the string representation. We could just override that in the class. Or we could make the doc a kwarg that we pop like response. The actual args for IOError are `(errno, strerror[, filename[, winerror[, filename2]]])`. I don't think that doc fits very cleanly into any of those.

chyzzqo2:
Here's an alternative the just reverses the init order so that the message comes from the base json decode error rather than the ioerror

nateprewitt:
Thanks @chyzzqo2, let's see how the tests run. I think this is closer to the right approach. `doc` has always been included as a str in Python 3 for `JSONDecodeError` but it looks like we're passing it as the wrong argument in the new exception.

nateprewitt:
Alright tests look good and I did some quick spot checking with the script below. This seems to be relatively in-line with previous behavior (2.26.0). We're no longer mis-populating fields.

We do have a number of new unpopulated fields from the multi-inheritance but I don't think that's a deal breaker for now. We'll consider this change a bugfix. If there's enough demand, we may choose to write something to appropriately populate all of the exception metadata later based on which `json` module is used. We can do this as an enhancement though.

```python
import platform
import requests

print(requests.__version__)
print(platform.python_version())

r = requests.get('https://www.python.org/psf/')
try:
    r.json()
except Exception as e:
    attributes = [i for i in dir(e) if not i.startswith('_')]
    print('Attributes: {}'.format(attributes))
    if hasattr(e, 'doc'):
        err = "Error Type: {}, Doc type: {}, Doc len: {}"
        print(err.format(type(e), type(e.doc), len(e.doc)))
    else:
        print("No doc attribute present.")

    print(e)
```

## Requests 2.26.0
#### Python 3.6 (stdlib)
```
2.26.0
3.6.15
Attributes: ['args', 'colno', 'doc', 'lineno', 'msg', 'pos', 'with_traceback']
Error Type: <class 'json.decoder.JSONDecodeError'>, Doc type: <class 'str'>, Doc len: 35177
Expecting value: line 3 column 1 (char 2)
```

#### Python 3.6 (simplejson)
```
2.26.0
3.6.15
Attributes: ['args', 'colno', 'doc', 'end', 'endcolno', 'endlineno', 'lineno', 'msg', 'pos', 'with_traceback']
Error Type: <class 'simplejson.errors.JSONDecodeError'>, Doc type: <class 'str'>, Doc len: 35177
Expecting value: line 3 column 1 (char 2)
```

#### Python 3.10 (std lib)
```
2.26.0
3.10.1
Attributes: ['args', 'colno', 'doc', 'lineno', 'msg', 'pos', 'with_traceback']
Error Type: <class 'json.decoder.JSONDecodeError'>, Doc type: <class 'str'>, Doc len: 35177
Expecting value: line 3 column 1 (char 2)
```

### Python 3.10 (simplejson)
```
2.26.0
3.10.1
Attributes: ['args', 'colno', 'doc', 'end', 'endcolno', 'endlineno', 'lineno', 'msg', 'pos', 'with_traceback']
Error Type: <class 'simplejson.errors.JSONDecodeError'>, Doc type: <class 'str'>, Doc len: 35177
Expecting value: line 3 column 1 (char 2)
```

## Requests (PR #6036)
#### Python 3.6 (std lib)
```
2.27.1
3.6.15
Attributes: ['args', 'characters_written', 'colno', 'doc', 'errno', 'filename', 'filename2', 'lineno', 'msg', 'pos', 'request', 'response', 'strerror', 'with_traceback']
Error Type: <class 'requests.exceptions.JSONDecodeError'>, Doc type: <class 'str'>, Doc len: 35177
Expecting value: line 3 column 1 (char 2)
```

#### Python 3.6 (simplejson)
```
2.27.1
3.6.15
Attributes: ['args', 'characters_written', 'colno', 'doc', 'end', 'endcolno', 'endlineno', 'errno', 'filename', 'filename2', 'lineno', 'msg', 'pos', 'request', 'response', 'strerror', 'with_traceback']
Error Type: <class 'requests.exceptions.JSONDecodeError'>, Doc type: <class 'str'>, Doc len: 35177
Expecting value: line 3 column 1 (char 2)
```

#### Python 3.10 (std lib)
```
2.27.1
3.10.1
Attributes: ['args', 'characters_written', 'colno', 'doc', 'errno', 'filename', 'filename2', 'lineno', 'msg', 'pos', 'request', 'response', 'strerror', 'with_traceback']
Error Type: <class 'requests.exceptions.JSONDecodeError'>, Doc type: <class 'str'>, Doc len: 35177
Expecting value: line 3 column 1 (char 2)
```

#### Python 3.10 (simplejson)
```
2.27.1
3.10.1
Attributes: ['args', 'characters_written', 'colno', 'doc', 'end', 'endcolno', 'endlineno', 'errno', 'filename', 'filename2', 'lineno', 'msg', 'pos', 'request', 'response', 'strerror', 'with_traceback']
Error Type: <class 'requests.exceptions.JSONDecodeError'>, Doc type: <class 'str'>, Doc len: 35177
Expecting value: line 3 column 1 (char 2)
```

nateprewitt (approved):

requests/exceptions.py
```diff
@@ -33,6 +33,9 @@ class InvalidJSONError(RequestException):
 
 class JSONDecodeError(InvalidJSONError, CompatJSONDecodeError):
     """Couldn't decode the text into json"""
+    def __init__(self, *args, **kwargs):
+        CompatJSONDecodeError.__init__(self, *args)
+        InvalidJSONError.__init__(self, *self.args, **kwargs)
```
sigmavirus24, line 45: Can we add comments here explaining why:

1. We're using super old Python 2.5 methods of calling parent `__init__` methods
2. We're calling the `__init__`s in this particular order - presuming the order matters
3. This fixes the issue of `.doc` ending up in the message/default `__repr__`

requests/exceptions.py
```diff
@@ -33,6 +33,9 @@ class InvalidJSONError(RequestException):
 
 class JSONDecodeError(InvalidJSONError, CompatJSONDecodeError):
     """Couldn't decode the text into json"""
+    def __init__(self, *args, **kwargs):
```
sigmavirus24, line 37: Can we add an empty line prior to the `__init__` here? That's already the style of this sub-module if not the entirety of `requests`

sigmavirus24 (changes requested):
Are there any tests we can add to verify this doesn't revert back to containing the entirety of `doc` in the `__repr__`? It's also unclear how this prevents that from happening on a visual inspection

requests/exceptions.py
```diff
@@ -34,6 +34,16 @@ class InvalidJSONError(RequestException):
 class JSONDecodeError(InvalidJSONError, CompatJSONDecodeError):
     """Couldn't decode the text into json"""
 
+    def __init__(self, *args, **kwargs):
+        """
+        Construct the JSONDecodeError instance first with all
+        args. Then use it's args to construct the IOError so that
+        the json specific args aren't used as IOError specific args
+        and the error message that JSONDecodeError builds is preserved
```
nateprewitt: nit; This might be phrased "[...] the error message from JSONDecodeError is preserved."

Also note the missing period in the current sentence.

tests/test_requests.py
```diff
@@ -2585,5 +2585,11 @@ def test_post_json_nan(self, httpbin):
 
     def test_json_decode_compatibility(self, httpbin):
         r = requests.get(httpbin('bytes/20'))
-        with pytest.raises(requests.exceptions.JSONDecodeError):
+        data = r.text
+        with pytest.raises(requests.exceptions.JSONDecodeError) as excinfo:
             r.json()
+        assert isinstance(excinfo.value, RequestException)
```
nateprewitt, line 2590: Do we need to use `.value` in these assertions? Just `excinfo` should be sufficient and that's how the typical user will probably interact with them.

tests/test_requests.py
```diff
@@ -2585,5 +2585,11 @@ def test_post_json_nan(self, httpbin):
...
+        assert isinstance(excinfo.value, JSONDecodeError)
+        assert data not in str(excinfo.value)
+        if is_py3:
```
nateprewitt: Also somewhat of a nit. It may be better to just make this its own test with a `@unittest.skipIf(not is_py3)` at the top.

e.g.

```python
@unittest.skipIf(not is_py3)
def test_json_decode_persists_doc_attr(self, httpbin):
    r = requests.get(httpbin('bytes/20'))
    with pytest.raises(requests.exceptions.JSONDecodeError) as e:
        r.json()
    assert e.doc == r.text
```

nateprewitt (changes requested):
Left a few small notes but I think we're close to wrapping this up if @sigmavirus24 doesn't have more concerns.

tests/test_requests.py
chyzzqo2: don't we want to test that the py2 object is also what we expect? Or if we don't can i can make it py3 only?

tests/test_requests.py
chyzzqo2: excinfo is a pytest thing that wraps the actual exception, i think we have to look at value to get the original object

requests/exceptions.py
chyzzqo2: updated

tests/test_requests.py
nateprewitt: The original test doesn't assert anything about Python 2 either. So we don't lose anything with this change. I suppose we could assert doc _isn't_ on the Python 2 exception but I'm not sure how meaningful that is.

Edit: to be clear I'm proposing two tests. The one you have and moving the Python 3 specifics to its own test.

chyzzqo2: ok, i think i updated it with what you had in mind.

requests/exceptions.py
```diff
@@ -34,6 +34,16 @@ class InvalidJSONError(RequestException):
 class JSONDecodeError(InvalidJSONError, CompatJSONDecodeError):
     """Couldn't decode the text into json"""
 
+    def __init__(self, *args, **kwargs):
+        """
+        Construct the JSONDecodeError instance first with all
+        args. Then use it's args to construct the IOError so that
+        the json specific args aren't used as IOError specific args
+        and the error message from JSONDecodeError is preserved
```
nateprewitt: I think the last sentence still needs a period at the end.

tests/test_requests.py
```diff
@@ -2585,5 +2585,18 @@ def test_post_json_nan(self, httpbin):
...
+        assert isinstance(excinfo.value, JSONDecodeError)
+        assert data not in str(excinfo.value)
+
+    @pytest.mark.skipif(not is_py3, reason="doc attribute is only present on py3")
+    def test_json_decode_persists_doc_attr(self, httpbin):
+        r = requests.get(httpbin('bytes/20'))
+        data = r.text
+        with pytest.raises(requests.exceptions.JSONDecodeError) as excinfo:
+            r.json()
+        assert data not in str(excinfo.value)
```
nateprewitt: We probably don't need to reassert this since it's already handled in the test above.

tests/test_requests.py
```diff
@@ -2585,5 +2585,18 @@ def test_post_json_nan(self, httpbin):
...
+        with pytest.raises(requests.exceptions.JSONDecodeError) as excinfo:
             r.json()
+        assert isinstance(excinfo.value, RequestException)
+        assert isinstance(excinfo.value, JSONDecodeError)
+        assert data not in str(excinfo.value)
+
+    @pytest.mark.skipif(not is_py3, reason="doc attribute is only present on py3")
+    def test_json_decode_persists_doc_attr(self, httpbin):
+        r = requests.get(httpbin('bytes/20'))
+        data = r.text
```
nateprewitt: nit; since we're only using this once we can probably just use r.text in the final assertion as shown in the original proposal.

nateprewitt (changes requested):
Great, thanks for the updates @chyzzqo2! I think there were a couple more things we discussed in the last review I left comments on. Let me know if you have any questions.

tests/test_requests.py
chyzzqo2: I liked having both in the same test, assures us that we are testing with the right string. but i'll take it out.

nateprewitt (approved):
Looks good to me, thanks @chyzzqo2!

sethmlarson (approved):
Seems good to me.

sigmavirus24 (commented):
I'm still not convinced this is maintainable or correct but I also understand that it's probably a tonne of work to get compatibility on python 2 and 3 that isn't a scary nightmare

mykolasolodukha:
Hello! I'm facing an issue now (after upgrading to `requests==2.28.0`) with this update.

Basically, I want to test the `try/except requests.JSONDecodeError` paths in my application using `pytest`. At some point in tests I do `mock_request.return_value.json.side_effect = requests.JSONDecodeError()` which is now unable to run because 
```
self = JSONDecodeError(), args = (), kwargs = {}

    def __init__(self, *args, **kwargs):
        """
        Construct the JSONDecodeError instance first with all
        args. Then use it's args to construct the IOError so that
        the json specific args aren't used as IOError specific args
        and the error message from JSONDecodeError is preserved.
        """
>       CompatJSONDecodeError.__init__(self, *args)
E       TypeError: JSONDecodeError.__init__() missing 3 required positional arguments: 'msg', 'doc', and 'pos'
```

Anything we should be concerned about at this point, after the PR is released? Or is there something I do wrong?

nateprewitt:
Hi @mykolasolodukha,

Being able to create an empty JSONDecodeError wasn't originally intended when implementing this error. The function signature was kept ambiguous due to differences between Python2/3 and these exceptions generally aren't intended to be reused externally. We may be able to make the signature more explicit now but in the interim, you can instantiate the error similar to below:
```
requests.JSONDecodeError('','',0)
```

The stdlib JSONDecodeError takes argument types `str, str, int` for msg, doc, pos respectively.

mykolasolodukha:
Thanks a lot @nateprewitt !

PR #6039, opened 2022-01-09
Allow HTTPResponse subclassing

cclecle:
- defining a new static variable in  HTTPAdapter : _clsHTTPResponse
- using this variable instead of direct HTTPAdapter object creation


sigmavirus24:
We're under a feature freeze and this is adding an undocumented, private attribute that users might be expected to override. We will not be accepting this

cclecle:
Maybe it need to be discussed ... 
If it is only a documentation problem, tell me where to add it.

Reason of this pull-request: 
https://stackoverflow.com/questions/16903528/how-to-get-response-ssl-certificate-from-requests-in-python/70642124#70642124

Issue #6008, opened 2021-12-16
Unicode surrogates in POST data lead to exception


pcworld:
In requests 2.26.0 with Python 3, when passing unicode strings with surrogate characters as POST data (either directly as a `str` or as part of a `dict`), sending the request crashes.

## Expected Result

The library should not crash, even when fed with unvalidated user input.

## Actual Result / Reproduction Steps

```
>>> import requests
>>> requests.__version__
'2.26.0'

>>> requests.post('https://example.invalid', data={'name': 'test\udced\udcb3\udc83.pdf'})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/.local/lib/python3.7/site-packages/requests/api.py", line 117, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/home/user/.local/lib/python3.7/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/user/.local/lib/python3.7/site-packages/requests/sessions.py", line 528, in request
    prep = self.prepare_request(req)
  File "/home/user/.local/lib/python3.7/site-packages/requests/sessions.py", line 466, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
  File "/home/user/.local/lib/python3.7/site-packages/requests/models.py", line 319, in prepare
    self.prepare_body(data, files, json)
  File "/home/user/.local/lib/python3.7/site-packages/requests/models.py", line 515, in prepare_body
    body = self._encode_params(data)
  File "/home/user/.local/lib/python3.7/site-packages/requests/models.py", line 104, in _encode_params
    v.encode('utf-8') if isinstance(v, str) else v))
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 4-6: surrogates not allowed

>>> requests.post('https://example.com/', 'test\udced\udcb3\udc83.pdf')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/requests/api.py", line 116, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/usr/lib/python3/dist-packages/requests/api.py", line 60, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 646, in send
[...]

PR #6048, opened 2022-01-25
Unicode surrogates in POST data lead to exception

mftb:
Trying my hand at fixing #6008


mftb:
@sigmavirus24 @nateprewitt PTAL

PR #6054, opened 2022-01-30
Add hooks `prepared` and `request`

Speedlulu:
* `prepared` is dispatched when `Session.prepare` is called, it receives the `PreparedRequest` that was just prepared.
* `request` is dispatched when `Session.request` is called, it receives the `Request` that was just created.


sigmavirus24:
Requests is not accepting new features or API expansions at this point in time.

Issue #6056, opened 2022-02-04
Catch `urllib3.exceptions.SSLError` in `Response.iter_content`


anadius:
I believe requests should catch that and raise `requests.exceptions.SSLError`:

```
Traceback (most recent call last):
  File "urllib3\response.py", line 438, in _error_catcher
  File "urllib3\response.py", line 519, in read
  File "http\client.py", line 465, in read
  File "socket.py", line 705, in readinto
  File "ssl.py", line 1273, in recv_into
  File "ssl.py", line 1129, in read
ssl.SSLError: [SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC] decryption failed or bad record mac (_ssl.c:2536)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "tkinter\__init__.py", line 1921, in __call__
  File "tkinter\__init__.py", line 839, in callit
  File "updater_gui.py", line 421, in metadata_cached
  File "concurrent\futures\_base.py", line 438, in result
  File "concurrent\futures\_base.py", line 390, in __get_result
  File "concurrent\futures\thread.py", line 58, in run
  File "updater\updater.py", line 314, in get_metadata
  File "pzd.py", line 256, in get_info
  File "pzd.py", line 216, in _get_central_dir
  File "pzd.py", line 176, in _get_chunk
  File "pzd.py", line 170, in _get_range
  File "requests\models.py", line 836, in content
  File "requests\models.py", line 758, in generate
  File "urllib3\response.py", line 576, in stream
  File "urllib3\response.py", line 512, in read
  File "contextlib.py", line 153, in __exit__
  File "urllib3\response.py", line 449, in _error_catcher
urllib3.exceptions.SSLError: [SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC] decryption failed or bad record mac (_ssl.c:2536)
```

This is requests 2.26.0.

That's some edge case and I've seen that happen only once (this is automatic error report from my program so I don't know any details, it's anonymous). I'm not trying to solve the problem, I just want to be able to catch `requests.exceptions.SSLError` instead of `urllib3.exceptions.SSLError`.

nateprewitt:
Thanks for reporting the issue @anadius. We've definitely had a few cases where either new or rarely raised exceptions from urllib3 leak through. I've added #6057 as a proposal to handle this.

nateprewitt:
We've merged #6057, so I'd expect this change to be available in the next release.

PR #6057, opened 2022-02-05
Wrap urllib3's SSLError as requests' SSLError

nateprewitt:
This should handle the issue raised in #6056 by wrapping the `SSLError` from urllib3 (`urllib3.exception.SSLError`) with ours (`requests.exception.SSLError`). Test was added to ensure we're doing the expected exception translation for everything we remap on streamed responses.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -14,6 +14,7 @@ import re
 import io
 import requests
 import pytest
+import urllib3
 from requests.adapters import HTTPAdapter
 from requests.auth import HTTPDigestAuth, _basic_auth_str
 from requests.compat import (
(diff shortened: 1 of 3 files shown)


tests/test_requests.py
```diff
@@ -910,7 +927,7 @@ def test_certificate_failure(self, httpbin_secure):
         """
         When underlying SSL problems occur, an SSLError is raised.
         """
-        with pytest.raises(SSLError):
+        with pytest.raises(RequestsSSLError):
```
sethmlarson, line 930: Does this mean that previously Requests wasn't wrapping SSLErrors? If this one is being raised on connect instead of read I wonder if this would be a big break for users?

sethmlarson (approved):
LGTM

Issue #3197, opened 2016-05-16
session proxies not honored during request


mbondfusion:
Using a session object and setting the proxies dictionary does not affect the proxies used during a request.

ex: Still tries to use system proxy.
```
s = requests.Session()
s.proxies = {'http': None, 'https': None}
s.post(url='http://10.0.1.1', json={'test': 'data'})
```

Using proxies during each request works, but it would be great to set them at the session level. Specifically if you potentially have dozens or ".post()" or ".get()" calls throughout your script.

Any thoughts?

Lukasa:
Thanks for this! This is a duplicate of #2018.

PR #6058, opened 2022-02-05
fix session proxy issue

achapkowski:
Fix session proxy issue where Session.proxies is ignore on self.request().

#### Reproducible Code

```python
import requests
s = requests.Session()
s.proxies = proxy
s.get("https://google.com").text
```

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -514,7 +514,7 @@ class Session(SessionRedirectMixin):
         )
         prep = self.prepare_request(req)
 
-        proxies = proxies or {}
+        proxies = proxies or self.proxies or {}
 
         settings = self.merge_environment_settings(
             prep.url, proxies, stream, verify, cert


achapkowski:
resolves: https://github.com/psf/requests/issues/6059

achapkowski:
Hey @nateprewitt or @sigmavirus24 how do you get a PR reviewed for a future release?  This is a simple fix that allows Session objects to honor proxy settings.

sigmavirus24:
Patience. We have busy lives being open source. Also I suspect this to be a duplicate of a well documented backwards incompatible PR that we have discussed ad nauseum.

achapkowski:
@sigmavirus24 thanks, I just wanted to ensure I have everything that is needed to assist getting this in or rejected.  Hopefully it will be added.

nateprewitt:
Session proxies are generally a mess right now and changing them has unexpected transitive effects.

I think this is slightly different than some of our long running issues but it's going to take some dedicated effort to map out impact.

achapkowski:
@nateprewitt I understand that it might have an effect, but it seems ```self.proxies``` does nothing.  This just adds it back into the chain if set.  I think what is a bigger issue is the fact the the `Session.proxies` are not used at all.  All this does is ensure if set, it's acknowledged in the proxy workflow. 

### Original
```
proxies = proxies or {}
```

### Change

```
proxies = proxies or self.proxies or {}
```

It seems when `proxies` was set, it was never pulled in from the class level.

nateprewitt:
Hi @achapkowski,

We understand the problem you're seeing but I think this PR is jumping to incorrect conclusions. It's important we understand what is actually happening rather than what's being observed before making changes.

If you read the next line, you'll see a call to our [resolver](https://github.com/psf/requests/blob/main/requests/sessions.py#L519-L521) for session settings. We do in fact [use `self.proxies`](https://github.com/psf/requests/blob/95f456733656ed93645ff0250bfa54f6d256f6fe/requests/sessions.py#L712). This function takes in to account several factors though for determining the proxies used by a session. Something in that flow isn't choosing your Session level proxy when considering changes. The proposed solution here is likely overly simplistic for the underlying problem though.

dmich2:
This fix is wrong if you have tried to set proxies in both your request & session. Then it still ignores your Session.proxies.

The problem is in [merge_environment_settings()](https://github.com/psf/requests/blob/95f456733656ed93645ff0250bfa54f6d256f6fe/requests/sessions.py#L691)

If self.trust_env is True it sets the environment proxies on top of the request proxies (if they aren't already set). Then it later does `proxies = merge_setting(proxies, self.proxies)` which unilaterally clobbers any session.proxies.

What really needs to be done is NOT to change the initial request proxies argument but keep the env_proxies separate.
Then update the env_proxies with the session proxies, and THEN update that dict with the incoming requests proxies.

Something like:

        proxies_env = merge_setting({}, env_proxies)
        proxies_env_session = merge_setting(self.proxies, proxies_env)
        proxies = merge_setting(original_proxies, proxies_env_session)

Where I wasn't sure about just doing dict.copy() in this case or what happens if I update env_proxies directly.

achapkowski:
I didn't know what proxy order took priority.

nateprewitt:
This is definitely a dupe of #3197 and #2018 then. Thanks for taking a look @dmich2. We’ll close this and the issue and track things in the original. This already has a patch but isn’t mergeable until a major version bump is made.

achapkowski:
@nateprewitt I don't think closing it is the right thing.  Is there now a time schedule for the next major release?, if not, this issue will just be a major issue for people.

sigmavirus24:
@achapkowski this is the wrong solution and a duplicate. This has been an issue for 8 years. Ignoring this context is not conducive to fixing this. Based on past interactions I've witnessed with you and experiences of my own, I'm locking this.

PR #6065, opened 2022-02-13
prepare url if http or ws

heyarnold1:
websockets can also include params and so should be prepared as well


sigmavirus24:
Requests doesn't support websockets

Issue #6071, opened 2022-02-23
CURL_CA_BUNDLE= disables certificate verification


owtaylor:
I'm not the first to notice this, see:
 
 https://stackoverflow.com/questions/48391750/disable-python-requests-ssl-validation-for-an-imported-module

Which implies people have even relied on the current behavior as a hack ... but I think it's pretty clear that the current behavior is an accidental bug, which should be fixed (for requests 3?)

Vaguely related to #3829

## Expected Result

An empty-string CURL_CA_BUNDLE should use default system verification, the same way as:

 * An unset CURL_CA_BUNDLE
 * An empty-string or unset REQUESTS_CA_BUNDLE
 * Behavior of curl/libcurl with an empty-string or unset CURL_CA_BUNDLE

## Actual Result

Empty CURL_CA_BUNDLE disables certificate verification

## Reproduction Steps

 * Set CURL_CA_BUNDLE to an empty value, try to fetch a self-signed or invalid HTTPS endpoint => success

nateprewitt:
Hi @owtaylor,

Thanks for bringing this to our attention! This was definitely not intended behavior and we agree it's a bug. We've pushed up a PR (#6074) to fix the issue in the next minor release of Requests. While this may be breaking for the workflow discussed in Stackoverflow, that was never a supported method for disabling verification.

Please let us know if you have any other concerns or comments on the PR. Thanks!

moshekaplan:
@nateprewitt : It pains me to request this, but given how popular the hack is, do you think it might make sense to create an environment variable for globally disabling SSL validation in requests?

sigmavirus24:
> given how popular the hack is,

Can you give evidence of this being popular?

moshekaplan:
> > given how popular the hack is,
> 
> Can you give evidence of this being popular?

https://stackoverflow.com/questions/48391750/disable-python-requests-ssl-validation-for-an-imported-module has approximately 50 upvotes and 29k views.

sigmavirus24:
> > > given how popular the hack is,
> > 
> > Can you give evidence of this being popular?
> 
> https://stackoverflow.com/questions/48391750/disable-python-requests-ssl-validation-for-an-imported-module has approximately 50 upvotes and 29k views.

Yeah, we're not interested in giving people a global hammer to misuse based off of view numbers that don't reflect how widely used that was. I have viewed stack overflow answers that were useless because a search engine thought them relevant. You're qualification for popular is deeply flawed

[...]

PR #6074, opened 2022-02-25
Fix environment CA Bundle resolution

nateprewitt:
This is a proposal to address #6071. Rather than settling on the final value from `CURL_CA_BUNDLE` we'll end the conditional change with `verify`. This ensures we don't accidentally take a "falsey" value and disable verification for misconfigured environments.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -702,11 +702,14 @@ class Session(SessionRedirectMixin):
             for (k, v) in env_proxies.items():
                 proxies.setdefault(k, v)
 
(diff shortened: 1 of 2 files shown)


sigmavirus24 (approved):

RafalSkolasinski:
So... we considered this a feature and not a bug. How can we control `verify=False` through environmental variables now?

RafalSkolasinski:
Use case when call of `requests` api is buried deep in the dependency stack and you **need to force** `verify=False`

Issue #6023, opened 2021-12-29
Requests: Python 2.7 Support and Deprecation


nateprewitt:
# Overview

As we enter our third year of Python 2.7 reaching end-of-life, Requests has decided it's time to start deprecating our support. While we have yet to confirm a date, we want to provide early notice that this is coming at some point in 2022. This follows as the majority of our critical downstream consumers have finished their deprecation campaigns. While we understand this may be viewed as an inconvenience for those still using Python 2.7, I'll try to layout the steps we're taking to ensure a smooth transition.

In January, we will be releasing Requests 2.27.0. This is intended to be the last minor release with guaranteed support, including all 2.27.x hotfix releases. We've taken steps to ensure this release series includes fixes for some of the longer standing pain points and provide as much long term stability as we're able. Users who intend to continue Python 2.7 use should review the `Mitigation Steps` section of this issue.

Going forward, we're still working through what a Python 3-only Requests looks like. We'd like to preface now that it's unlikely to contain many of the more exciting features that may have previously been promised by former maintainers. In future releases, we'll look to add quality-of-life improvements as they make sense while maintaining the primary focus of stability in the library. We'll provide more on that as information is available.

# Mitigation Steps

For Requests users that intend to remain on Python 2.7 through 2022, we recommend taking the following mitigation steps.

1. For users of `pip` 9.0+, we'll also be releasing an update to the `python_requires` metadata to `>=3.7` in Requests. This will prevent unintended installation on unsupported Python versions.
2. For users who want to codify this requirement across all environments, ensure that you've explicitly pinned Requests in your workspace or project to `requests<2.28`. This will prevent unintended interruptions with future releases.
3. For users of LTS operating systems providing Python 2.7 support through 2023, you'll need to work with their support teams for  bug or security issues for any releases after 2.27.

dulmandakh:
It's good to hear that requests decided to drop Python 2.7. I would propose to release version 3 without much refactoring but mainly dropping Python 2.7 compatibility, so that Python 3 users can benefit from less dependency and less code to run. Then we can do refactors or non-breaking changes in minor releases. Also it's possible to have another major release with more breaking changes. Otherwise it'll take much longer to drop Python 2.7 and do refactoring at the same time.

nateprewitt:
[...]

PR #6075, opened 2022-02-27
Drop EOL Python 3.6

ofek:
See:

- https://endoflife.date/python
- https://www.python.org/dev/peps/pep-0494/#lifespan


sigmavirus24:
We haven't even dropped 2.7. Please don't spam this project

ofek:
@sigmavirus24 I know we haven't dropped Python 2, that won't happen until v3 AFAIU. Are you saying dropping EOL 3.x is blocked on dropping 2.7? I saw commits dropping [3.4](https://github.com/psf/requests/commit/68e581f1b5b4425a0ee950a2433fc3c92e671d0f) and [3.5](https://github.com/psf/requests/commit/f6c0619d892a41dcf84933810ffda89e9f6b10d4) which is why I thought it was alright to open this PR.

Also, I don't think it's quite right to characterize a contribution as spam. I came along and saw something that might need doing so I tried to help out (was going push a final commit to update docs files later today).

I understand there's a lot going on in the world right now though, so no worries.

nateprewitt:
Hi @ofek,

We are planning to drop [Python 2.7](https://github.com/psf/requests/issues/6023) and 3.6 in the next minor release. We're currently keeping `main` clean for any other hotfix issues that may come in before that release. Thanks for contribution and we'll look at trying to pull some of this in when it's time.

ofek:
@nateprewitt Oh that's awesome, thank you!

PR #6091, opened 2022-03-23
Remove support for Python 2.7, 3.6 and Pypy3

nateprewitt:
This PR is a proposal to officially remove support for Python 2.7 as discussed in #6023. The next minor release (Requests 2.28.0) will no longer provide support for any of the end of life Python runtimes. Users are encouraged to upgrade to Python 3.8 or later. If you're unable to do so, we'd recommend pinning to `requests<2.28`.

This will be the first of a few PRs that will be doing some general code clean up now that we're moving to a Python 3-only code base. I'll post the others as drafts shortly.


requests/sessions.py
```diff
@@ -39,10 +39,7 @@
 
 # Preferred clock, based on which one is more accurate on a given system.
 if sys.platform == 'win32':
```
sethmlarson, line 41: Since we're now Python 3.7+ I wonder if we can start using `time.perf_counter` everywhere? Might need to do more research into why Windows was given only perf_counter/clock instead of `time.time()`

setup.py
```diff
@@ -108,7 +102,6 @@ def run_tests(self):
     extras_require={
         'security': [],
         'socks': ['PySocks>=1.5.6, !=1.5.7'],
-        'socks:sys_platform == "win32" and python_version == "2.7"': ['win_inet_pton'],
```
sethmlarson, line 111: [Finally free...](https://pypi.org/project/win-inet-pton) :smiling_face_with_tear:

tests/compat.py
```diff
@@ -13,9 +10,5 @@
 except ImportError:
     cStringIO = None
 
-if is_py3:
-    def u(s):
-        return s
-else:
-    def u(s):
-        return s.decode('unicode-escape')
+def u(s):
```
sethmlarson, line 15: Should we go through and remove all uses of `u(...)` in the tests to simplify them?

tests/test_requests.py
```diff
@@ -844,7 +843,7 @@ def test_conflicting_post_params(self, httpbin):
             with pytest.raises(ValueError):
                 requests.post(url, data='[{"some": "data"}]', files={'some': f})
             with pytest.raises(ValueError):
-                requests.post(url, data=u('[{"some": "data"}]'), files={'some': f})
+                requests.post(url, data=u'[{"some": "data"}]', files={'some': f})
```
sethmlarson: Do we need `u'...'`?

sethmlarson (commented):
This is so exciting! Here's a handful of comments/questions:

tests/test_requests.py
nateprewitt: No we don't, I have a separate linting PR that removes these lines entirely since they're redundant. I can move that into this one.

tests/compat.py
nateprewitt: I left the definition in case it's in use externally, but yes, it should be removed from all of our tests. I'll do a second pass if I missed some.

requests/sessions.py
nateprewitt: It was a performance issue for Windows where the clock wasn't granular enough for CI to pass on Appveyor (#3988). I can take a look at standardizing everything onto perf_counter but that's probably a separate deep dive.

tests/test_requests.py
hugovk: `pyupgrade **/*.py --py37-plus` will deal with a bunch of `u''`s and other stuff.

nateprewitt: Yep, that's what's in the linting commit currently. Unfortunately, it just makes the two declarations identical so it will still require manual removal in one of the PRs.

setup.py
```diff
@@ -90,14 +84,14 @@ def run_tests(self):
         'Natural Language :: English',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.6',
```
hugovk:
```suggestion
```

hugovk (commented):
A few other bits to clean up:

`tox.ini`: add `310`, remove `27` and `36`

Several files: clean up `use_chardet_on_py3`

`requests/utils.py`: `replacer = os.rename if sys.version_info[0] == 2 else os.replace`

`tests/testserver/server.py`:

```python
        # NB: when Python 2.7 is no longer supported, the argument
        # can be removed to use a default backlog size
        sock.listen(5)
```

tests/compat.py
hugovk: Perhaps add a deprecation warning for any external use?

nateprewitt:
Thanks @hugovk, those are all great call outs! `use_chardet_on_py3` is an unfortunate misnomer because it just forcing use of `chardet`. The only reason it references py3 is because chardet was the default for Python 2. I don't think we can change the extra name at this point.

I addressed the rest of @sethmlarson's comments too and added a hard failure in the setup.py for anyone installing on `pip<9`.

setup.py
```diff
@@ -8,7 +7,32 @@
...
+if CURRENT_PYTHON < REQUIRED_PYTHON:
+    sys.stderr.write(
+        """
+==========================
+Unsupported Python version
+==========================
+This version of Requests requires Python {}.{}, but you're trying to
+install it on Python {}.{}.
+This may be because you are using a version of pip that doesn't
+understand the python_requires classifier. Make sure you
```
sethmlarson: We discussed this in the urllib3 Discord with @nateprewitt and @graingert and figured out that this series of steps isn't always going to solve the problem because pip *also* dropped Python versions over time.

There were a few solutions proposed:
- Keep the current message
- Always recommend installing pip<21 (last version that supported Python 2 and 3.5)
- Recommend installing [pip-with-requires-python](https://pypi.org/project/pip-with-requires-python)
- Not give a recommendation on *how* to upgrade pip and setuptools?

I also raised the point of recommending users upgrade pip/setuptools would likely break OS installations.

nateprewitt: I think `pip-with-requires-python` is likely the safest option. Alternatively, we'd vend the requires values from the package in our setup.py and compose the install message from those values as @graingert suggested. That avoids any risk for Python 3.4 users who still comprise a significant number of downloads.

sethmlarson: @pradyunsg Noted that new Python versions come with a "good" pip version out of the box. Maybe our strategy should be to instruct the user to either pin to <2.28 or upgrade to at least Python 3.7 (since they'll need that version to install Requests anyways) and then we don't need to mention pip/setuptools at all.

nateprewitt: I went ahead and [ripped out the whole of touching pip versions](https://github.com/psf/requests/pull/6091/commits/febcd5199c278d99232124e22b3ab2d912b7cc16) after further discussion. If users want to upgrade their pip, that's something they can manage, otherwise the simplest option is to upgrade their Python version which will provide this automatically.

Other users are free to pin to older supported versions as discussed in the original deprecation announcement.

setup.py
```diff
@@ -17,17 +17,13 @@
...
-version of Python. If you can't upgrade your pip (or Python), request
-an older version of Requests:
-    $ python -m pip install "requests<2.28"
+This version of Requests requires at least Python {}.{}, but
+you're trying to install it on Python {}.{}. To resolve this,
+consider upgrading to a supported Python version.
+
+If you can't upgrade your Python version, you'll need to
+pin to an older version of Requests:
+    python -m pip install "requests<2.28"
```
pradyunsg:
```suggestion
pin to an older version of Requests (<2.28).
```

Let's just go all the way here. :P

nateprewitt: Fair enough :D

sethmlarson (approved):
LGTM! :shipit:

PR #6096, opened 2022-03-25
Add support for non-latin1 basic auth encodings (such as utf-8)

timbedard:


nateprewitt:
Hi @timbedard, thanks for the PR. Unfortunately, I don't believe we'll be able to accept it. The high level Requests APIs have been frozen for the majority of 2.x and we don't intend to change those interfaces.

I believe the general consensus has also been to favor bytes in most cases going forward. It removes any guessing (which is impossible to do right consistently) and avoid us leaking an `encoding` argument on to every interface in Requests.

qbedard:
Oh hey, no worries. It was a draft for a reason. I came to the same conclusion shortly after this and just came back to close it. Thanks for the fast response and keep up the good work!

PR #6095, opened 2022-03-23
Add automatic code formatting to Requests

nateprewitt:
### Overview
This is a draft PR for adding standardized code formatting to Requests. Now that we intend to drop Python 2 support, we have the opportunity to clean up multiple conflicting code styles that have crept into Requests over the years.

This PR will introduce a few standard changes:

* Integration of `black` for code formatting, bringing Requests in-line with what has largely become the standard.
* Integration of `isort` to ensure our imports are automatically grouped and cleaned.
* Integration of `pyupgrade` which will automatically pull most of the code base to Python 3.7 syntax.
* Addition of a `.pre-commit-config.yaml` to orchestrate the above tools.

### TODO
- [x] Rebase on to #6091 before merge
- [x] Resolve formatting issues with server response line wrapping in `tests/test_lowlevel.py`
- [ ] Add .git-blame-ignore-revs file
- [x] Fix inconsistent f-string upgrades
- [x] Enable and fix flake8 issues

PR #6116, opened 2022-04-29
Add .git-blame-ignore-revs

nateprewitt:
Follow up to #6095 to add a .git-blame-ignore-revs file. This allows us to strip large cosmetic commits from the git blame history to help keep the information relevant to actual logic changes.


sethmlarson (approved):
LGTM

Issue #6120, opened 2022-05-04
New exception when redirect url is invalid


setla:
Currently when redirect response with invalid redirect URL is returned, we receive the same error (`InvalidURL, InvalidSchema, LocationParseError`) as when passing wrong URL that is used for first request in chain. It would be good to have those cases separated (e.g. to try fix it in custom way)

PR #6121, opened 2022-05-04
Add and use new exception when redirect url is invalid

setla:
Feature described in issue https://github.com/psf/requests/issues/6120

new `InvalidRedirectURL` inherits `InvalidURL, InvalidSchema, LocationParseError` for backward compatibility


sigmavirus24 (changes requested):
I'm fairly against this. It's not a necessary exception to add and expand the API which we are explicitly not doing

PR #6125, opened 2022-05-09
Use proxy from string

el1s7:
I find myself making a dict everytime I want to test some proxy, which is unnecessary. Simply create the proxies dict from the string, in case a string supplied.


nateprewitt:
Hi @el1s7, thanks for the PR. The dictionary is our established interface for proxies though, and we aren't making any modifications to the top-level APIs at this time. I think we'll have to pass on this PR.

PR #6154, opened 2022-06-08
Add validation for header name

nateprewitt:
Following up on #6083, this refactors our header validation function from 2.11.0 to consider the header name. We'd originally avoided adding header name validation because we wanted to limit the change scope to header splitting with new lines. Since then the standard library has made similar changes to ours and now raises a ValueError in `http.client`. This gives us inconsistent errors depending on which portion of the header you provide a bad value.

```python
>>> requests.get("https://httpbin.org/get", headers={":bad": "header"})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/adapters.py", line 489, in send
    resp = conn.urlopen(
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/connectionpool.py", line 727, in urlopen
    httplib_response = self._make_request(
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/connectionpool.py", line 433, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/connection.py", line 309, in request
    super().request(method, url, body=body, headers=headers)
  File "/Users/nateprewitt/.pyenv/versions/3.10.4/lib/python3.10/http/client.py", line 1282, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/Users/nateprewitt/.pyenv/versions/3.10.4/lib/python3.10/http/client.py", line 1323, in _send_request
    self.putheader(hdr, value)
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/connection.py", line 274, in putheader
    super().putheader(header, *values)
  File "/Users/nateprewitt/.pyenv/versions/3.10.4/lib/python3.10/http/client.py", line 1250, in putheader
[...]

PR #6157, opened 2022-06-08
Add GitHub action to automate linting

nateprewitt:
This was intended to be a follow up to #6095 but got lost in the shuffle post PyCon. It's currently failing on a single line change that's addressed in #6154. Once that's merged this should be ready to merge and ensure we don't introduce any future inconsistencies.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
...
+  push:
+  pull_request:
+
+jobs:
+  lint:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.10
```
sethmlarson: Thoughts on following pre-commit's example config and use setup-python without specifying a Python  version?

sethmlarson (approved):
Looks good, one thought so you won't need to modify config in the future.

.github/workflows/lint.yml
nateprewitt: Fine by me, I'd brought this over from other projects but we can switch to `actions/setup-python@v3` per their Readme.

Issue #6139, opened 2022-05-21
Misusing of system-wide proxy on Windows / HTTPS connections over system-wide proxy on Windows do not work


ValdikSS:
On Windows 10 and 11, if a system-wide proxy is configured in the OS settings, Requests won't load HTTPS websites.
Instead of the regular CONNECT header, Requests sends TLS request to the proxy server (ip and port). The TLS ClientHello does not have Server Name Indication field, that's why I assume that Requests misinterprets the proxy as a Secure Web Proxy (a regular proxy with TLS encapsulation on top), which Windows does not support.

## Expected Result

If system-wide proxy is configured in Windows, Requests sends CONNECT header to the proxy server when HTTPS website is requested.

## Actual Result

If system-wide proxy is configured in Windows, Requests sends TLS request to the proxy address when HTTPS website is requested.

## Reproduction Steps

1. Configure system-wide proxy on Windows
2. 
```python
import requests
requests.get("https://w3.org/")
```

Result:
My proxy does not respond to the packets it does not understand, and I receive the following error:
`HTTPSConnectionPool(host='w3.org, port=443): Max retries exceeded with url: / (Caused by ProxyError('Cannot connect to proxy.', timeout('_ssl.c:1114: The handshake operation timed out')))`

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "2.0.12"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.3"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.10.4"
  },
  "platform": {
    "release": "10",
    "system": "Windows"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.27.1"
  },
  "system_ssl": {
    "version": "101010ef"
  },
  "urllib3": {
    "version": "1.26.9"
  },
  "using_charset_normalizer": true,
  "using_pyopenssl": false
}
```

ValdikSS:
This is a Python's urllib bug, which is triggered by newer urllib3 version, thus affecting requests.

https://bugs.python.org/issue42627
https://github.com/python/cpython/issues/86793
https://github.com/python/cpython/commit/9743524fc14db39e2f309cf3d372c7e8e18fc93a

[...]

PR #6162, opened 2022-06-11
Fix HTTPS websites with system-wide HTTP proxy on Windows

ValdikSS:
Due to urllib bug, requests fails to open any HTTPS websites on Windows
if there's system-wide HTTP proxy configured. This is because urllib
incorrectly prepends the protocol to the proxy host and port, as in:

{'http':  'http://host:port',
 'https': 'https://host:port',
 'ftp':   'ftp://host:port'}

Such configuration forces urllib3 to use HTTPS proxy (Secure Web Proxy,
also known as TLS Proxy) for HTTPS URLs, which the configured proxy
most likely does not support.

Detect incorrect behavior and rewrite the protocol to http.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -816,6 +816,25 @@ def should_bypass_proxies(url, no_proxy):
     return False
 
 
+def getproxies_fix_system_proxies_windows():
+    """Fix for urllib getproxies() bug on Windows with Python < 3.10.5
+    See: https://bugs.python.org/issue42627
+
+    :rtype: dict
+    """
+    proxy_hosts = []
+    protocols = {"http", "https", "ftp"}
+    proxies = getproxies()
+    if proxies.keys() != protocols:
+        return proxies
+    for protocol in proxies:
+        proxy_hosts.append(proxies[protocol].lstrip(protocol + "://"))
+    proxy_hosts = list(set(proxy_hosts))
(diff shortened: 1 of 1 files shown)


ValdikSS:
Issue #6139

requests/utils.py
```diff
@@ -816,6 +816,25 @@ def get_proxy(key):
     return False
 
 
+def getproxies_fix_system_proxies_windows():
+    """Fix for urllib getproxies() bug on Windows with Python < 3.10.5
+    See: https://bugs.python.org/issue42627
```
sigmavirus24, line 821: Nothing precludes this from running on the versions unaffected according to you. This comment is aside from whether this is the correct fix

sigmavirus24 (changes requested):

requests/utils.py
ValdikSS: This is a no-op for a fixed Python/urllib version. This commit checks whether getproxies() returns the same proxy host and port for three protocols, each prefixed with the protocol, and rewrites it to "http://". The fixed urllib already returns proxies with "http://", so this commit does nothing in this case.

ValdikSS: @sigmavirus24, do you think the code needs more checks?

ValdikSS:
Could anyone take a look at this merge request please?

PR #6163, opened 2022-06-14
Update README.md (some better wording)

ajsharda17:


nateprewitt:
Hi @ajsharda17, thanks for the PR! Unfortunately, we don't accept PRs with stylistic changes to documentation due to their subjective nature. They often result in additional churn because there is rarely consensus on what constitutes "better". We do appreciate the initiative though!

PR #6169, opened 2022-06-20
Allow charset normalizer >=2 and <3

frenzymadness:
The current definition `charset_normalizer~=2.0.0` does not allow charset-normalizer 2.1.0 which has been released a couple of hours ago. The newer version seems to be compatible with requests and [the compatibility check](https://github.com/psf/requests/blob/da9996fe4dc63356e9467d0a5e10df3d89a8528e/requests/__init__.py#L80-L84) suggests that requests should allow versions up to 3.0.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ provides-extra =
     use_chardet_on_py3
 requires-dist =
     certifi>=2017.4.17
-    charset_normalizer~=2.0.0
+    charset_normalizer>=2,<3
     idna>=2.5,<4
     urllib3>=1.21.1,<1.27
 
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ if sys.argv[-1] == "publish":
     sys.exit()
 
 requires = [
-    "charset_normalizer~=2.0.0",
+    "charset_normalizer>=2,<3",
     "idna>=2.5,<4",
     "urllib3>=1.21.1,<1.27",
     "certifi>=2017.4.17",


frenzymadness:
All tox environments passed locally.

hswong3i:
Similar and duplicated fix from https://github.com/psf/requests/pull/6175#issuecomment-1165150374:

> In between I give a quick dirty workaround for my packaging scripts, and it is now working with CentOS 8 + molecule 4.0.0 + charset_normalizer 2.1.0 + requests 2.28.0 (patched):
> * https://github.com/alvistack/psf-requests/commit/4910bb75c0122efd77e11274de2793c95e2f630b
> * https://build.opensuse.org/package/show/home:alvistack/psf-requests-2.28.0

setup.cfg
```diff
@@ -5,7 +5,7 @@ provides-extra =
     use_chardet_on_py3
 requires-dist =
     certifi>=2017.4.17
-    charset_normalizer~=2.0.0
+    charset_normalizer~=2.0
```
sethmlarson: Instead of `~` let's use `>=2,<3` in line with other dependencies.

setup.py
```diff
@@ -59,7 +59,7 @@ def run_tests(self):
     sys.exit()
 
 requires = [
-    "charset_normalizer~=2.0.0",
+    "charset_normalizer~=2.0",
```
sethmlarson: Same here.

setup.cfg
nateprewitt: I think we'd intentionally chosen ~=2.0.0 to constrain this to a single minor version. It will be less noise for us if we widen it to majors, but there was originally some concern from Ian while charset_normalizer is still early in its lifecycle.

frenzymadness: According to the [changelog](https://github.com/Ousret/charset_normalizer/blob/master/CHANGELOG.md) they use semantic versioning so version 2 should stay comatible.

nateprewitt (approved):
Looks good to me, thanks @frenzymadness! We'll release this in the next patch release, 2.28.1.

PR #6179, opened 2022-06-26
Update Dependency with `chardet>=3.0.2,<6`

hswong3i:
https://github.com/chardet/chardet/releases/tag/5.0.0 was released on 2022-06-25 with remove of Python 3.6 support, which also align with our Python > 3.6 requirement.

Signed-off-by: Wong Hoi Sing Edison <hswong3i@pantarei-design.com>

Issue #6167, opened 2022-06-19
InvalidHeader exception when trying to add a ":method" header field (breaking change in 2.28.0)


Alex8768765:
My app is making a POST request enforcing some specific header fields. Among them, I was able to explicitly enforce a ":method" header field. But since 2.28.0, this now raises the following error: 

`requests.exceptions.InvalidHeader: Invalid leading whitespace, reserved character(s), or returncharacter(s) in header name: ':method'`

Enforcing this header field is possibly a widespread practice as this is part of the requirements for sending notifications through Apple's push notification servers.

nateprewitt:
Hi @Alex8768765,

Can you please provide the info we originally requested in the issue template? The header you're using isn't valid in the HTTP spec, so it seems unlikely this would be widespread. Python itself rejects sending a header like this in our supported versions.

```python
>>> requests.__version__
'2.27.1'
>>> r = requests.get('https://httpbin.org/get', headers={':method': 'test'})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nateprewitt/.pyenv/versions/3.7.12/lib/python3.7/site-packages/requests/api.py", line 75, in get
    return request('get', url, params=params, **kwargs)
  File "/Users/nateprewitt/.pyenv/versions/3.7.12/lib/python3.7/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/nateprewitt/.pyenv/versions/3.7.12/lib/python3.7/site-packages/requests/sessions.py", line 529, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/nateprewitt/.pyenv/versions/3.7.12/lib/python3.7/site-packages/requests/sessions.py", line 645, in send
    r = adapter.send(request, **kwargs)
  File "/Users/nateprewitt/.pyenv/versions/3.7.12/lib/python3.7/site-packages/requests/adapters.py", line 450, in send
    timeout=timeout
  File "/Users/nateprewitt/.pyenv/versions/3.7.12/lib/python3.7/site-packages/urllib3/connectionpool.py", line 710, in urlopen
    chunked=chunked,
  File "/Users/nateprewitt/.pyenv/versions/3.7.12/lib/python3.7/site-packages/urllib3/connectionpool.py", line 398, in _make_request
    conn.request(method, url, **httplib_request_kw)
[...]

Issue #6159, opened 2022-06-09
Breaking change in 2.28.0 when using string enums as Headers (working in v2.27.1)


bonastreyair:
When using string enums as key values for the `headers` dict parameter with `requests.get()` function in `v2.28.0` it raises an `InvalidHeader` error but using previous version `v2.27.1` works well without any error. It seems there was an unexpected breaking change with that release. 

## Expected Result

Dict keys for the headers parameter when using `requests.get()` are still valid if a string enum is used.

## Actual Result

Instead it raises an `InvalidHeader` error stating that: 

> Header must be of type `str` or `bytes`, not `enum`.

```python
Traceback (most recent call last):
  File "/home/yair/PycharmProjects/tests/requests_error.py", line 9, in <module>
    requests.get("http://URL", headers={CustomEnum.TRACE_ID: "90e85293-afd1-4b48-adf0-fa6daf02359e"})
  File "/home/yair/.local/lib/python3.8/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
  File "/home/yair/.local/lib/python3.8/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/yair/.local/lib/python3.8/site-packages/requests/sessions.py", line 573, in request
    prep = self.prepare_request(req)
  File "/home/yair/.local/lib/python3.8/site-packages/requests/sessions.py", line 484, in prepare_request
    p.prepare(
  File "/home/yair/.local/lib/python3.8/site-packages/requests/models.py", line 369, in prepare
    self.prepare_headers(headers)
  File "/home/yair/.local/lib/python3.8/site-packages/requests/models.py", line 491, in prepare_headers
    check_header_validity(header)
  File "/home/yair/.local/lib/python3.8/site-packages/requests/utils.py", line 1037, in check_header_validity
    raise InvalidHeader(
requests.exceptions.InvalidHeader: Header part (<CustomEnum.TRACE_ID: 'X-B3-TraceId'>) from {<CustomEnum.TRACE_ID: 'X-B3-TraceId'>: '90e85293-afd1-4b48-adf0-fa6daf02359e'} must be of type str or bytes, not <enum 'CustomEnum'>
```

## Reproduction Steps

```python
from enum import Enum
import requests


class CustomEnum(str, Enum):
    TRACE_ID = "X-B3-TraceId"


[...]

PR #6182, opened 2022-06-29
v2.28.1

nateprewitt:
2.28.1 (2022-06-29)
-------------------

**Improvements**

- Speed optimization in `iter_content` with transition to `yield from`. (#6170)

**Dependencies**

- Added support for chardet 5.0.0 (#6179)
- Added support for charset-normalizer 2.1.0 (#6169)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,6 +7,18 @@ dev
 - \[Short description of non-trivial change.\]
 
 
+2.28.1 (2022-06-29)
+-------------------
+
+**Improvements**
+
+- Speed optimization in `iter_content` with transition to `yield from`. (#6170)
+
+**Dependencies**
+
+- Added support for chardet 5.0.0 (#6179)
+- Added support for charset-normalizer 2.1.0 (#6169)
+
 2.28.0 (2022-06-09)
 -------------------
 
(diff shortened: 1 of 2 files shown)


nateprewitt:
I'm deferring the decisions on #6159 and #6167 until we see larger data points. They have 5 and 2 user engagements respectively which seems very low for the current 2.28.x adoption rate. For the time being, there are straight forward workarounds for both.

sethmlarson (approved):
Looks great, let's ship it! :shipit:

afriedma:
> I'm deferring the decisions on #6159 and #6167 until we see larger data points. They have 5 and 2 user engagements respectively which seems very low for the current 2.28.x adoption rate. For the time being, there are straight forward workarounds for both.

Hi, what is the workaround for #6167? I am adding ":authority" and the request fails @nateprewitt

nateprewitt:
@afriedma please keep conversations constrained to relevant issues. I believe this was already [answered](https://github.com/psf/requests/issues/6167#issuecomment-1159781034) in the other issue you commented on. If you're unfamiliar with PreparedRequests, you can find them in our [documentation](https://requests.readthedocs.io/en/latest/user/advanced/#prepared-requests).

PR #6186, opened 2022-06-29
refactor: prefer f-strings to format where fits within black lengths

kkirsche:
Python added f-strings in version 3.6, with [PEP 498](https://www.python.org/dev/peps/pep-0498/). F-strings are a flexible and powerful way to format strings. They make the code shorter and more readable, since the code now looks more like the output.

F-Strings are also more performant than using `.format` (e.g. see [F-String Speed Considerations](https://realpython.com/lessons/f-strings-speed-considerations-performance/), [F-String Benchmarks](https://www.scivision.dev/python-f-string-speed/) for two quick examples) because the python runtime can insert the variables into the string while parsing rather than backtracking to insert them into the placeholders, reducing the number of times the lines need to be processed

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -94,9 +94,10 @@ def _check_cryptography(cryptography_version):
         return
 
     if cryptography_version < [1, 3, 4]:
-        warning = "Old version of cryptography ({}) may cause slowdown.".format(
-            cryptography_version
+        warning = (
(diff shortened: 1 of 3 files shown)


nateprewitt:
I think `ruff` ended up fixing this automatically. I'll close this out since it got lost.

PR #6199, opened 2022-07-19
Fix links in docs to the kenreitz.org domain, which expired.

mila:
When I was studying Contributor’s Guide, I found that links to kenreitz.org are broken.

- "The Future of Python HTTP" was archived by Internet Archive, so I linked there.
- "Be cordial or be on your way" was not archived, so I removed the link.
- "A Kenneth Reitz Project." in a footer is broken too, but I did not touch that link.


nateprewitt:
It appears the links may have been temporarily interrupted but are working presently.

PR #6200, opened 2022-07-19
Rephrase an outdated comment mentioning Python 2.

mila:
As a first-time contributor, I was confused when I run `make test` and saw "This runs all of the tests, on both Python 2 and Python 3." in console output.


nateprewitt (approved):
Thanks @mila

PR #6201, opened 2022-07-19
Suppress context when requests.exceptions.JSONDecodeError is raised.

mila:
When a JSON response cannot be parsed, requests catch json.JSONDecodeError (or simplejson.JSONDecodeError) and raise requests.exceptions.JSONDecodeError instead. Because this happens in an except block, unhandled exceptions were printed in a chain:

```
...
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
During handling of the above exception, another exception occurred:
...
requests.exceptions.JSONDecodeError: [Errno Expecting value] ...
```

This seems unnecessary and it increases cognitive overhead whean examining errors.

As requests.exceptions.JSONDecodeError inherits json.JSONDecodeError and we copy all atrributes from the later to the former, `raise ... from None` seems more approriate than `raise ... from e`.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -965,14 +965,14 @@ class Response:
                     # used.
                     pass
                 except JSONDecodeError as e:
-                    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
+                    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos) from None
 
         try:
             return complexjson.loads(self.text, **kwargs)
         except JSONDecodeError as e:
             # Catch JSON-related errors and raise as requests.JSONDecodeError
             # This aliases json.JSONDecodeError and simplejson.JSONDecodeError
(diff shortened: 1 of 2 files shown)


sigmavirus24:
We don't suppress context anywhere else as far as I remember. I don't understand why it's more valuable here and less helpful elsewhere. (And, no, don't update every exception to suppress context)

mila:
Thank you @sigmavirus24, I see now that the error handing for JSON decoding is consistent with the rest of the library.

> (And, no, don't update every exception to suppress context)

What about updating exceptions to use `raise ... from e`? We would get tracebacks with: "The above exception was the direct cause of the following exception:" instead of "During handling of the above exception, another exception occurred:".

I believe that it would follow intentions in [PEP 3134](https://peps.python.org/pep-3134/). If `__cause__` (instead of `__context__`) was set, it would be more apparent that the exception chain is an expected behavior, not some bug.

PR #6202, opened 2022-07-21
Add response object to JSONDecodeError

bungoume:
Add response object to JSONDecodeError.

We used to use the following code, but since 2.27, errors on json parse failure are now caught as RequestException.
Since there is basically `response` in the `.json()` phase, we would like to add it in JSONDecodeError for convenience.

```python
def my_post(url, payload):
    try:
        res = requests.post(url, json=payload)
        res.raise_for_status()
        return res.json()
    except requests.exceptions.RequestException as e:
        logger.warning("Failed to send data %s", e.response.status_code)
    except Exception as e:
        ...
```
```
AttributeError
'NoneType' object has no attribute 'status_code'
```

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -965,14 +965,14 @@ class Response:
                     # used.
                     pass
                 except JSONDecodeError as e:
-                    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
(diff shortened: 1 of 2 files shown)


sigmavirus24:
No, thank you

PR #6206, opened 2022-07-29
added from_raw class method to PreparedRequest

ToraNova:
Adds a class method to create PreparedRequests from raw HTTP request using Python's builtin BaseHTTPRequestHandler class.

Example usage:

```
>>> from requests import PreparedRequest, Session
>>> prq = PreparedRequest.from_raw('GET / HTTP/1.1\r\nHost: example.com\r\n\r\n', use_https=False)
>>> res = Session().send(prq)
<Response [200]>
```

This would be very useful for someone that wants to use the python requests library but with raw HTTP requests.

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -6,12 +6,13 @@ This module contains the primary objects that power Requests.
 """
 
 import datetime
+from http.server import BaseHTTPRequestHandler
 
 # Import encoding now, to avoid implicit import later.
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Requests is not accepting feature requests or new feature additions at this time. Thank you for putting this together but we won't be accepting it

PR #6207, opened 2022-07-30
Remove `__nonzero__` method from `models.Response`

sobolevn:
`__nonzero__` was used in python2 as we now use `__bool__`. It is not used at all in python3.
And since [`setup.py` says](https://github.com/psf/requests/blob/177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/setup.py#L10) that `reqeusts` supports only `python >= 3.7`, I don't think that it is really required.

I found this while working on `typeshed` types for `requests`: https://github.com/python/typeshed/blob/3fe1f5d6c46b39b8b4632aadc70477a729241d25/stubs/requests/requests/models.pyi#L107

### DIFF
--- a/requests/models.py
+++ b/requests/models.py
@@ -738,16 +738,6 @@ class Response:
         """
         return self.ok
 
-    def __nonzero__(self):
-        """Returns True if :attr:`status_code` is less than 400.
-
-        This attribute checks if the status code of the response is between
-        400 and 600 to see if there was a client error or a server error. If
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @sobolevn,

Thanks for the PR, but I think we’re unlikely to remove this. The use of `super().__nonzero__()` to implement `__bool__` in subclasses was quite popular when the Python 3 migration started. There’s a high chance this will be breaking if we remove the function entirely.

PR #6209, opened 2022-07-31
Added support for TRACE method

cheehong1030:
- Added support for TRACE method
- Added TRACE method in docs
- Fix docs missing OPTIONS method


sigmavirus24:
Please check open and closed issues and pull requests in the future before opening new ones.

We will not be accepting this because we are not adding new areas to the API. This is also possible via `requests.request` or `Session.request`

PR #6214, opened 2022-08-16
Update reference to "cookielib" to "cookiejar" in documentation

ittner:
Docstrings still called this module as it was in python2.7; update
them to use the name from python3. Mostly useful as a reference for
people who need a custom cookiejar.

Code does still use the compatibility layer for a no-more-supported
python2, however.


nateprewitt (approved):
Thanks @ittner!

PR #6216, opened 2022-08-19
updated sha1 hasing to sha256 to avoid collisions

joshmgrant:
Based on a [security review](https://pypi.openrefactory.com/requests/62ff1f9edd30551cbfff8eeb) of the requests repository, it looks like the `sha1` hashing algorithm should be updated here to avoid collisions. 

I have not added any unit tests since this should be a `no-op` change with the expected behaviour after this change identical to the previous behaviour before the change. I'm also happy to help implement/clean up some of the code here as well.

### DIFF
--- a/requests/auth.py
+++ b/requests/auth.py
@@ -202,7 +202,7 @@ class HTTPDigestAuth(AuthBase):
         s += time.ctime().encode("utf-8")
         s += os.urandom(8)
 
-        cnonce = hashlib.sha1(s).hexdigest()[:16]
+        cnonce = hashlib.sha256(s).hexdigest()[:16]
         if _algorithm == "MD5-SESS":
             HA1 = hash_utf8(f"{HA1}:{nonce}:{cnonce}")
 


nateprewitt:
Hi @joshmgrant, for future reference, it's considered best practice to always report issues relating to suspected security problems through private channels. Most projects will have either a [SECURITY.md](https://github.com/psf/requests/blob/main/.github/SECURITY.md) or a [specific issue type](https://github.com/psf/requests/security/policy) with instructions on disclosure. This provides appropriate time for projects to mitigate vulnerabilities.

That said, I don't think this is an issue. The hash being used for the cnonce is already being truncated to 16 bytes, so changing this from SHA1 to SHA256 shouldn't have a meaningful effect on collisions. We also use the cnonce to compute the actual digest hash value which can be MD5, SHA1, SHA256, or SHA512.

Can you clarify the practical attack vector you're observing here? This looks like a generic grep scanner report from what's been provided.

joshmgrant:
@nateprewitt Hello there! Thank you for the feedback on this. I this is my first pull request for a security-related issue (as such) so I was unaware of the instructions for security-related changes and/or vulnerability reporting. 

I was unable to create any sort of (practical) attack related to this change. I completely agree that, as you mention, since this is a cnonce value is used to compute the digest hash, it's not really a meaningful issue to use the SHA1 hash for the conce value. 

My main motivation was that this change was reported by the scanner that I linked to earlier as a _possible_ security vulnerability but without any additional context. After some research and trying to create a meaningful attack, I decided to open the PR first and see what would happen. Now I know :) 

I will close this PR as unmerged but thank you for your patience. Let me know if there's any areas I could look at for contributions (not necessarily security related, I just like Python and Requests). Thank you so much!

PR #6225, opened 2022-08-31
fix order of verify overrides in Sessions

purajit:
Currently, `Session` overrides the request-level `verify` parameter by
reading the environment, and then checks to see if it should choose that
or the session-level settings.

Let's consider the scenario where we've provided `REQUESTS_CA_BUNDLE` in
the environment, and then create a session with `s.verify = 'asdf'`. Now,
the final choice is going to be the environment-provided bundle, rather
than the session-provided setting, which is unexpected behavior.

The correct behavior would be to first do the merge of the request and 
session-level parameters, and then use the environment-provided params
if needed.

### DIFF
--- a/requests/sessions.py
+++ b/requests/sessions.py
@@ -760,20 +760,20 @@ class Session(SessionRedirectMixin):
             for (k, v) in env_proxies.items():
                 proxies.setdefault(k, v)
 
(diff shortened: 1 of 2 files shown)


sigmavirus24:
See #3829 for why we can't just fix this in a bug patch release

PR #6226, opened 2022-09-01
Use urllib3 native chunking ability

shadycuz:
This PR is the same as #5664 but rebased on to the `main` branch to bring it back up to date. This PR is needed to stay compatible  with future changes in urllib3 introduced by https://github.com/urllib3/urllib3/pull/2649. The comments in that PR offer a lot of detail in the changes in urllib3 but this one provided a good overview https://github.com/urllib3/urllib3/pull/2649#issuecomment-1200197578.

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -506,6 +506,10 @@ class HTTPAdapter(BaseAdapter):
 
                 low_conn = conn._get_conn(timeout=DEFAULT_POOL_TIMEOUT)
 
+                if hasattr(conn, 'proxy'):
+                    if conn.proxy is not None and not getattr(low_conn, 'sock', None):
+                        conn._prepare_proxy(low_conn)
+
                 try:
                     skip_host = "Host" in request.headers
                     low_conn.putrequest(
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -485,67 +485,19 @@ class HTTPAdapter(BaseAdapter):
             timeout = TimeoutSauce(connect=timeout, read=timeout)
 
         try:
-            if not chunked:
-                resp = conn.urlopen(
-                    method=request.method,
-                    url=url,
-                    body=request.body,
-                    headers=request.headers,
-                    redirect=False,
-                    assert_same_host=False,
-                    preload_content=False,
-                    decode_content=False,
-                    retries=self.max_retries,
-                    timeout=timeout,
-                )
-
-            # Send the request.
-            else:
-                if hasattr(conn, "proxy_pool"):
-                    conn = conn.proxy_pool
-
-                low_conn = conn._get_conn(timeout=DEFAULT_POOL_TIMEOUT)
-
-                if hasattr(conn, 'proxy'):
-                    if conn.proxy is not None and not getattr(low_conn, 'sock', None):
-                        conn._prepare_proxy(low_conn)
-
-                try:
-                    skip_host = "Host" in request.headers
-                    low_conn.putrequest(
-                        request.method,
-                        url,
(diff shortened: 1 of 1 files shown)


shadycuz:
@sethmlarson @nateprewitt Can you please approve the workflow. I'm curious if the SSL error I had locally still shows up in this pipeline.

shadycuz:
All tests passed except for linting but it needs approval again 🙃

sethmlarson (approved):
LGTM, going to wait for @nateprewitt to merge.

shadycuz:
@sethmlarson I have another linting error, I'll fix real quick

shadycuz:
I ran all the checks locally this time :blush:

```console
$ pre-commit run -a
Check Yaml...............................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
isort....................................................................Passed
black....................................................................Passed
pyupgrade................................................................Passed
flake8...................................................................Passed
```

theGOTOguy:
It is noteworthy that this change, while something I've been wanting for a long time, is a breaking API change.  In particular, the problem is that in the current implementation `urllib3`'s `Retry` is sometimes bypassed when using chunked encoding:  Users have to write their own custom retry logic based on capturing errors when using chunked encoding.  Therefore, this change will quietly break everybody who has written custom error handling for chunked requests versus using `Retry` for non-chunked requests.

shadycuz:
@theGOTOguy is the bypassing of `Retry` something going on here in the request package or is this a bug in urllib3?

theGOTOguy:
No, the issue wasn't with `urllib3`.   The [part of the code you removed that executes if the request is chunked](https://github.com/psf/requests/pull/6226/files) clearly has at least some logic outside `urllib3`, and if an error happened anywhere there then it wouldn't be caught by `Retry` and a non-`urllib3` error would be raised that then had to be handled separately.  I can't seem to find my notes on the specific errors, but in short, if I was dealing with chunked requests and a flaky or overloaded server, I couldn't rely on it being handled in `Retry`.

To be clear, I think that the way you are handling it here is 100% the right thing to do.  My understanding based on the discussion in #5664 was that a change to the errors that are thrown during the course of handling chunked requests is something that won't happen until the next major release, but if it does happen now I would suggest a test simulating chunked requests against a flaky server (#5915, anyone?) and warn about the change in behavior in the changelog.

Edit:  It appears that the errors that weren't caught by `urllib3`'s `Retry` were tied to either timeouts or empty responses from the server.  This would be straightforward to test and verify against a realistic Werkzeug server as opposed to the existing socket-based tests.

shadycuz:
@sethmlarson @nateprewitt We are not getting very far on this =/ 

I have put in an application as PSF managing member. I'm not sure if that will make a difference, but maybe then we could have a call to discuss the path forward. 

It looks like we need to cut a new major release branch for both urlib3 and requests and to start preparing for the next major version?

sethmlarson:
@shadycuz Unfortunately whether or not you're a managing member won't affect this PR being merged faster. People have obligations outside of open source, I've recently bought a new house which is [consuming all my open source time](https://twitter.com/sethmlarson/status/1565039700301283328). I can't speak for Nate but I know with certainty this PR isn't not being worked on for lack of desire to see it succeed. Have patience, promise we will get to this when we have the time.

sethmlarson:
@nateprewitt FYI we're going to skip the failing integration tests for chunking to get that PR merged, this PR will be needed to add full support for urllib3 v2.0 to Requests.

nateprewitt (approved):

PR #6230, opened 2022-09-05
allow string/byte subtypes for header

CarliJoy:
We wrap our secret a in special subclass of string, that prevents leaking confidential information in some circumstances (i.e. stack traces) etc.

The current implementation of the header checks does a `type(val) == str` which is not recommended by PEP8.
Due to this implementation we can't pass this variables as headers. Converting them back to a normal string would defeat the purpose of wrapping them in the first place.

This PR fixes both issues.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -1032,15 +1032,17 @@ def check_header_validity(header):
     """
     name, value = header
 
-    for part in header:
-        if type(part) not in HEADER_VALIDATORS:
-            raise InvalidHeader(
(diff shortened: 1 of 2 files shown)


fstavela (approved):

CarliJoy:
Related to https://github.com/psf/requests/issues/6159 -> there is also another fix defined there.

nateprewitt:
Resolving as I believe this was fixed in Requests 2.29.0.

PR #6234, opened 2022-09-07
Fix CURL_CA_BUNDLE override example

jaap3:
Promote REQUESTS_CA_BUNDLE as it is looked at first


nateprewitt (approved):
Yep, this seems like a reasonable change. Thanks @jaap3!

PR #6237, opened 2022-09-11
Improved documentation of exceptions in api.rst

mhayworth:
Added use cases and inheritance information for the common types of exceptions that are listed in docs/api.rst. Documentation already exists for the main interface, request sessions, and lower-level classes on the website, so those may want to be moved into api.rst as well for consistent formatting. Alternatively, this information could also be chosen to move to the website. I am open to any changes we might want to make to improve this documentation and help users better understand the exceptions their code may be throwing.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4,7 +4,6 @@ Developer Interface
 ===================
 
 .. module:: requests
-
 This part of the documentation covers all the interfaces of Requests. For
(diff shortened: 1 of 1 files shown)


nateprewitt:
I don't think this is something we'd generally add to the docs. Appreciate the PR though.

Issue #6242, opened 2022-09-22
setup.py


lucas-ccc:
## This bug does not affect usage, but it does exist
the bug code is in the "setup.py" file, line 78
here is code:
```python
with open(os.path.join(here, "requests", "__version__.py"), "r", "utf-8") as f:
    exec(f.read(), about)
```
## It's an issue about function open()
I got this from Python Documentation
```python
open(file, mode='r', buffering=- 1, encoding=None, errors=None, newline=None, closefd=True, opener=None)
```
The argument ```"utf-8"``` should be passing by ```keyword```, but actually it's passed by ```position```, 
and the third argument is ```buffering```(a int) . 
So the system throws an ```TypeError```: 'str' object cannot be interpreted as an integer

## Fix
```python
with open(os.path.join(here, "requests", "__version__.py"), "r", encoding="utf-8") as f:
    exec(f.read(), about)
```

## Reproduction Steps
open any file with the code, you will get the TypeError.

## System Information

    TypeError: 'str' object cannot be interpreted as an integer

```python
here = os.path.abspath(os.path.dirname(__file__))
with open(os.path.join(here, "requests", "__version__.py"), "r", "utf-8") as f:
    exec(f.read(), about)
```

Ali-Xoerex:
The function `open()` used in this context was imported from `codecs` and is not the built-in `open()` function in Python itself.
`codecs.open(filename, mode='r', encoding=None, errors='strict', buffering=- 1)` , from [codecs official documentations](https://docs.python.org/3/library/codecs.html#codecs.open)

Ali-Xoerex:
@sigmavirus24
I think this issue can be closed now

PR #6244, opened 2022-09-22
Fix bug, the encoding argument should be passed by keyword in func op…

lucas-ccc:
Issue #6242

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,7 @@ here = os.path.abspath(os.path.dirname(__file__))
 with open(os.path.join(here, "requests", "__version__.py"), "r", encoding="utf-8") as f:
     exec(f.read(), about)
 
-with open("README.md", "r", "utf-8") as f:
+with open("README.md", "r", encoding="utf-8") as f:
     readme = f.read()
 
 setup(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -75,7 +75,7 @@ def run_tests(self):
 
 about = {}
 here = os.path.abspath(os.path.dirname(__file__))
-with open(os.path.join(here, "requests", "__version__.py"), "r", "utf-8") as f:
+with open(os.path.join(here, "requests", "__version__.py"), "r", encoding="utf-8") as f:
     exec(f.read(), about)
 
 with open("README.md", "r", "utf-8") as f:
```
evanstjabadi: @lucas-ccc - L81 may also result in the same error.

```suggestion
with open("README.md", "r", encoding="utf-8") as f:
```

Ali-Xoerex (commented):
This change is not necessary.
`open()` function was imported from `codecs` module and is used correctly. Check out function definition in [codecs official docs](https://docs.python.org/3/library/codecs.html#codecs.open)

lucas-ccc:
> This change is not necessary. `open()` function was imported from `codecs` module and is used correctly. Check out function definition in [codecs official docs](https://docs.python.org/3/library/codecs.html#codecs.open)

Thanks for your comments. I don't check the definition of open function. You're right.

PR #6261, opened 2022-10-20
Allow charset normalizer >=2 and <4

deedy5:
charset-normalizer 3.0.0 has been released.

### DIFF
--- a/requests/__init__.py
+++ b/requests/__init__.py
@@ -80,8 +80,8 @@ def check_compatibility(urllib3_version, chardet_version, charset_normalizer_ver
     elif charset_normalizer_version:
         major, minor, patch = charset_normalizer_version.split(".")[:3]
         major, minor, patch = int(major), int(minor), int(patch)
-        # charset_normalizer >= 2.0.0 < 3.0.0
-        assert (2, 0, 0) <= (major, minor, patch) < (3, 0, 0)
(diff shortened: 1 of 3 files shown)


sigmavirus24 (changes requested):
There's little value to raising the lower bound and much more grief for users

deedy5:
v3.0.0 up to 4x faster than v2.1
https://github.com/Ousret/charset_normalizer/releases/tag/3.0.0
https://charset-normalizer.readthedocs.io/en/latest/community/speedup.html

sigmavirus24:
Yes, and given this isn't critical path for most users it's not critical to force people through a major version change to get that. If they use character detection of any sort and need performance they can force it to upgrade, but that's not our job

sigmavirus24 (approved):

sethmlarson (approved):

nateprewitt (approved):

nateprewitt:
Thanks @deedy5!

arossert:
Any estimation on a new release with this fix?

jimbudarz:
It looks like versions of `charset_normalizer` below 3.0.1 are [no longer available](https://pypi.org/project/charset-normalizer/#history) on pypi, so it's not possible to pip install `requests` until a new release is published.

nateprewitt:
This should now be available on PyPI in Requests 2.28.2.

PR #6266, opened 2022-10-23
Change to SPDX conform license string

elprimato:
I suggest to change the license string in the package information to an [SPDX parsable license expression](https://spdx.org/licenses/).
This makes it easier for downstream users to get the license information directly from the package metadata.


nateprewitt (approved):
Thanks @elprimato!

PR #6274, opened 2022-11-03
pylint and python 3 updates

marksmayo:
some small tweaks to appease pylint and python 3 changes


nateprewitt:
Hi @marksmayo, unfortunately I don't think we can take this PR. We generally don't accept stylistic PR updates. You can find our linting rules in the `pre-commit` file if you have any questions. Thanks

PR #6283, opened 2022-11-11
Single-source required python version in setup.py

MestreLion:
Since there is a `REQUIRED_PYTHON` tuple, use it to fill the `python_requires` argument of `setup()` to avoid hard-coding current "3.7" in multiple places.


sigmavirus24:
It's "hard coded" in one place. It's not a problem and we also don't accept cosmetic changes

PR #6291, opened 2022-11-21
Replace git protocol URL in docs with an HTTPS URL

oliviacrain:
As of March 2022, [GitHub has removed support for cloning repositories using the `git://` protocol](https://github.blog/2021-09-01-improving-git-protocol-security-github/). This project's installation docs still use a URL starting with `git://` when telling users how to clone the source repo. Let's replace that with the HTTPS URL for this repository.


nateprewitt (approved):
Thanks @oliviacrain!

PR #6300, opened 2022-11-30
Specify complete list of all available phony targets

jmarkowski:
The list of phony targets was incomplete. Add a list of the phony targets explicitly at the end of the file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,10 @@
-.PHONY: docs
 init:
 	pip install -r requirements-dev.txt
+
 test:
 	# This runs all of the tests on all supported Python versions.
 	tox -p
+
 ci:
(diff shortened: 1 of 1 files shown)


sigmavirus24:
I don't believe we're using this any longer. Thanks for trying to fix it up but I would just as soon delete it

jmarkowski:
If that's true, then there's a case for removing the Makefile from the repository. No point in keeping it around if any energy put towards maintaining it is dismissed.

If the Makefile is in fact used, and one can make a good argument for maintaining such a file in the repository (as analogously NPM based repositories offer a 'scripts' directive in the package.json file), then I maintain that the proposed changes are relevant.

PR #6302, opened 2022-12-07
Fix the proxy_bypass_registry function all returning true in some cases.

flysee:
With the context code,i think we should avoid this situation with empty strings,thx!

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -97,7 +97,7 @@ if sys.platform == "win32":
         # '<local>' string by the localhost entry and the corresponding
         # canonical entry.
         proxyOverride = proxyOverride.split(";")
-        # filter out empty strings to avoid re.match all true.
+        # filter out empty strings to avoid re.match return true in the following code.
         proxyOverride = filter(None, proxyOverride)
         # now check if we match one of the registry values.
         for test in proxyOverride:
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -97,6 +97,8 @@ if sys.platform == "win32":
         # '<local>' string by the localhost entry and the corresponding
         # canonical entry.
         proxyOverride = proxyOverride.split(";")
+        # filter out empty strings to avoid re.match all true.
+        proxyOverride = filter(None, proxyOverride)
         # now check if we match one of the registry values.
         for test in proxyOverride:
             if test == "<local>":
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -924,3 +924,35 @@ def test_set_environ_raises_exception():
             raise Exception("Expected exception")
 
     assert "Expected exception" in str(exception.value)
+
+
+@pytest.mark.skipif(os.name != "nt", reason="Test only on Windows")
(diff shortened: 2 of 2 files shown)


sigmavirus24:
There's no clear description of a bug, how to reproduce or even tests to show this fixes something. Further you speak of empty strings and are filtering on None which are not the same thing.

flysee:
The value of proxyOverride is controlled by a third-party program on the computer, so it is not controllable.
For example, when I use fiddler on my computer, this value is `<-loopback>;`.(Not all computers)
When proxyOverride splits, it's equal to `['<-loopback>', '']`.
So `re.match('', host, re.I)` will return true,that bypassing the proxy.

And filter(function, iterable) is equivalent to the generator expression (item for item in iterable if item) if function is None,that is, all elements of iterable that are false are removed.
Ref:https://docs.python.org/3/library/functions.html#filter

I think should be avoided for code robustness.

iamxiwang (approved):

ravi-kale (approved):

aaronhua123:
> The value of proxyOverride is controlled by a third-party program on the computer, so it is not controllable. For example, when I use fiddler on my computer, this value is `<-loopback>;`.(Not all computers) When proxyOverride splits, it's equal to `['<-loopback>', '']`. So `re.match('', host, re.I)` will return true,that bypassing the proxy.
> 
> And filter(function, iterable) is equivalent to the generator expression (item for item in iterable if item) if function is None,that is, all elements of iterable that are false are removed. Ref:https://docs.python.org/3/library/functions.html#filter
> 
> I think should be avoided for code robustness.

Yes, I also noticed this situation. This wastes a lot of my time discovering this problem. ProxyOverride sometimes ends with a semicolon and is difficult to reproduce. But it has not yet been merged into the main branch.

flysee:
> > The value of proxyOverride is controlled by a third-party program on the computer, so it is not controllable. For example, when I use fiddler on my computer, this value is `<-loopback>;`.(Not all computers) When proxyOverride splits, it's equal to `['<-loopback>', '']`. So `re.match('', host, re.I)` will return true,that bypassing the proxy.
> > And filter(function, iterable) is equivalent to the generator expression (item for item in iterable if item) if function is None,that is, all elements of iterable that are false are removed. Ref:https://docs.python.org/3/library/functions.html#filter
> > I think should be avoided for code robustness.
> 
> Yes, I also noticed this situation. This wastes a lot of my time discovering this problem. ProxyOverride sometimes ends with a semicolon and is difficult to reproduce. But it has not yet been merged into the main branch.

I'm glad someone else is having the same problem as me, but I also hope that no one will waste time on this issue in the future, haha.

sigmavirus24:
@flysee can you add tests for this change so it does not regress?

flysee:
> @flysee can you add tests for this change so it does not regress?

Of course, test cases have been added.

sigmavirus24 (approved):
@nateprewitt thoughts on getting this into the next release? 2.32 iirc?

nateprewitt (approved):
Yeah, this seems reasonable. I've put it on the 2.32.0 milestone.

PR #6304, opened 2022-12-08
Update `certifi` dependency to patch known vulnerability.

qthequartermasterman:
`certifi>=2017.11.05, < 2022.12.07` has a known security vulnerability, described [here](https://github.com/advisories/GHSA-43fp-rhv2-5gv8). The currently supported `certifi` range `certifi>=2017.4.17` includes these vulnerable versions.

This pull request updates `certifi` to `2022.12.07` which patches the vulnerability.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,7 +4,7 @@ provides-extra =
     socks
     use_chardet_on_py3
 requires-dist =
-    certifi>=2017.4.17
+    certifi>=2022.12.07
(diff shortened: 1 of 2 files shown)


sethmlarson:
Since we're not constraining upgrading to the latest version I think we're okay to keep the version range as-is, users should manage their dependencies and vulnerable versions. Going to close this.

PR #6325, opened 2023-01-06
Switch to Python 3.11 final on CI and add to tox file

browniebroke:


nateprewitt (approved):
Thanks @browniebroke!

PR #6326, opened 2023-01-08
Fix exception causes in utils.py

cool-RR:
More context for this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with


sigmavirus24:
I don't believe we have dropped 2.7 support yet. Even still we don't accept minor refactorings

PR #6330, opened 2023-01-13
Add example and improve docs for RequestException

edward70:
I added a brief example of exception handling for `RequestException` and documented the `request` and `response` attributes. PR does not overlap with #6237


nateprewitt:
Hi @edward70,

Thanks for the contribution! While this is all correct, it's also information that can be understood reading the code. We try to reserve usage comments for areas that are particularly complex or unintuitive, otherwise it can make code harder to follow due to comment clutter. For that reason, I think we'll pass on this PR for the moment.

PR #6333, opened 2023-01-18
fix: Remove '<4' from python_requires

### LINKED ISSUE #6332: Remove '<4' from python_requires
## Summary

As discussed in the discuss.python topic ["Use of “less-than next-major-version” (e.g., `<4`) in `python_requires` (setup.py)"](https://discuss.python.org/t/use-of-less-than-next-major-version-e-g-4-in-python-requires-setup-py/1066) and [other places](https://iscinumpy.dev/post/bound-version-constraints/) there can be some side effects of adding `<4` to a library's `python_requires`. Following these discussions the use of `<4` for `python_requires` was removed from the PyPA's packaging guide in https://github.com/pypa/packaging.python.org/pull/850. While 

https://github.com/psf/requests/blob/61c324da43dd8b775d3930d76265538b3ca27bc1/setup.py#L97

doesn't pose an immediate breaking threat in 2023, it would be good to remove this upper cap and is using `python_requires` in a way that was not intended.

matthewfeickert:
Resolves #6332

The 

https://github.com/psf/requests/blob/61c324da43dd8b775d3930d76265538b3ca27bc1/setup.py#L97

cap was added in PR #6091.

As discussed in https://discuss.python.org/t/use-of-less-than-next-major-version-e-g-4-in-python-requires-setup-py/1066 and [other places](https://iscinumpy.dev/post/bound-version-constraints/) by the PyPA, use of upper bounds with `python_requires` for _future_ versions of Python is unintended use of `python_requires` and actively discouraged.

c.f. https://github.com/pypa/packaging.python.org/pull/850 for further detail.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ setup(
     package_data={"": ["LICENSE", "NOTICE"]},
     package_dir={"requests": "requests"},
     include_package_data=True,
-    python_requires=">=3.7, <4",
+    python_requires=">=3.7",
     install_requires=requires,
     license=about["__license__"],
     zip_safe=False,


matthewfeickert:
This is ready for review.

nateprewitt:
It looks like this is a pretty consistent standard you've helped drive across projects. Given the comments in the discussion from the other PyPA maintainers, I think this is reasonable. Thanks @matthewfeickert!

nateprewitt (approved):

PR #6335, opened 2023-01-19
Add `mime_type` property to response object

MarkusHammer:
Added a `mime_type` property to the response object to allow for quick figuring out of the type of content of a response. This is a tiny QOL improvement, as it makes it easier to extract the mime type from the header without having to clutter up code just to extract it.


nateprewitt:
Hi @MarkusHammer, this is pretty specific behavior that I don't believe has a broadly applicable use case.

The end goal can be accomplished with a few lines of user written code when needed, so I think we'll stick with that as our advisement for now. Thanks!

PR #6336, opened 2023-01-20
Support missing SNIMissingWarning in tests

pquentin:
urllib3 2.0 only supports OpenSSL 1.1.1+ which means SNI is mandatory, and removed the warning altogether.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -2,9 +2,13 @@
 
 import warnings
 
-from urllib3.exceptions import SNIMissingWarning
+try:
+    from urllib3.exceptions import SNIMissingWarning
 
-# urllib3 sets SNIMissingWarning to only go off once,
-# while this test suite requires it to always fire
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
Fair enough, this looks good. Thanks @pquentin!

nateprewitt:
Looks like there's a small linting issue. Could you update the PR using our pre-commit hooks?

Although where it's moving the test import doesn't look right to me 👀. I'll investigate that separately.

pquentin:
Oops, sorry about that. Fixed the lint.

> Although where it's moving the test import doesn't look right to me 👀. I'll investigate that separately.

This is because I used `from tests import X` instead of `from . import X` and isort does not know that `tests` is a [known first party](https://pycqa.github.io/isort/docs/configuration/options.html#known-first-party). I switched back to `from . import X`.

PR #6347, opened 2023-02-07
Use context manager to open files

RichieB2B:
In the documentation for `POST a Multipart-Encoded File` the example code opens the file but never closes it. This leaves the file open until the script ends. It is better to use a context manager even in example code.

Original:
```
>>> url = 'https://httpbin.org/post'
>>> files = {'file': open('report.xls', 'rb')}

>>> r = requests.post(url, files=files)
>>> r.text
```

Replace with:
```
>>> url = 'https://httpbin.org/post'
>>> with open('report.xls', 'rb') as fd:
...     files = {'file': fd, 'rb')}

...     r = requests.post(url, files=files)
...     r.text
```


lpaladin:
I've just got bitten by this example, and was just about to create a PR for this too 😃 

(Maybe
```python
    ...     r = requests.post(url, files=multiple_files)
    ...     r.text
```
could be
```python
    ...     r = requests.post(url, files=multiple_files)
    ... 
    >>> r.text
```

PR #6348, opened 2023-02-08
python 3 language updates

marksmayo:
Fix some f-strings
Remove u-prefix from strings


sigmavirus24:
Thanks for thinking of us but we don't accept cosmetic changes to the project given it's current maintenance status.

PR #6356, opened 2023-02-17
Allow str/bytes subclasses to be used as header parts

sethmlarson:
Closes https://github.com/psf/requests/issues/6159 Adapted from https://github.com/nateprewitt/requests/commit/66fcc9c88cf5a53e85322305f41006032744cbe3 but also supports mixed type tuples.


nateprewitt:
If you can rebase this onto the linting change, I think this is set.

sethmlarson:
@nateprewitt Done, ready for review!

nateprewitt (approved):
🚀

PR #6374, opened 2023-03-09
refactor: Replace text symbol with unicode emoji

JoshuaDRose:
My reasoning is thus: emojis are cool


sigmavirus24:
This isn't a rescue and we don't accept cosmetic changes like this

Issue #6292, opened 2022-11-23
requests.put() does not respect timeout when data is chunked


konlai:
`requests.put(url, data=some_generator, timeout=5.0)` does not timeout if the server accepts the connection but does not send a response.

## Expected Result

`requests.put(url, data=some_generator, timeout=5.0)` would time out after 5 seconds if the server does not send a response.

## Actual Result

`requests.put(url, data=some_generator, timeout=5.0)` can hang indefinitely if the server does not send a response and does not close the connection.

## Reproduction Steps

```python
import requests

def data_chunked():
    data = [b'hello', b'world']
    for chunk in data:
        yield chunk

# This can block forever if the server accepts the connection but does not respond
requests.put("http://localhost:8080/", data=data_chunked(), timeout=5.0)

# This times out as expected
requests.put("http://localhost:8080/", data=b'hello', timeout=5.0)
```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "2.0.10"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.3"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.6.8"
  },
  "platform": {
    "release": "3.10.0-1062.18.1.el7.x86_64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.27.1"
  },
  "system_ssl": {
    "version": "100020bf"
  },
  "urllib3": {
    "version": "1.26.8"
  },
  "using_charset_normalizer": true,
  "using_pyopenssl": false
}
```

nateprewitt:
Thanks for the report, @missing64001. We moved to use urllib3's chunking behavior in 2.30.0 which should have resolved this. I'll close this out along with the PR since it no longer applies to the code base. Please let us know if you're still seeing the issue with the latest release.

PR #6375, opened 2023-03-09
Update adapters.py

missing64001:
requests.put() does not respect timeout when data is chunked


missing64001:
fix issue #6292

nateprewitt:
Resolving as this should be addressed in Requests 2.30.0 and we've removed this code path from Requests.

PR #6379, opened 2023-03-14
fix issue #6377

### LINKED ISSUE #6377: test_zipped_paths_extracted fails if test file has been modified since last run
If you run `python -m pytest tests/test_utils.py -ktest_zipped_paths_extracted` this test will pass normally. If you run it again, it will pass again.

If, however, you make any modifications to the file, the test will now fail and keep failing:

```       assert filecmp.cmp(extracted_path, __file__)
E       AssertionError: assert False
E        +  where False = <function cmp at 0x7f4ea6d4dab0>('/tmp/test_utils.py', '/home/david/project-reducer/scratch/requests/tests/test_utils.py')
E        +    where <function cmp at 0x7f4ea6d4dab0> = filecmp.cmp
```
The reason for this is that the last time the test ran, the contents of `__file__` were written to `/tmp/test_utils.py` and when `extract_zipped_paths` is next called, it doesn't overwrite the existing file, so the filecmp fails because it's now comparing against the old version of the file.

## Expected Result

The test should run successfully when modifications to the file are made.

## Actual Result

Modifications to the test file that are unrelated to the test in question cause it to fail.

## Reproduction Steps

```bash
python -m pytest tests/test_utils.py -ktest_zipped_paths_extracted
echo -n '# modification' >> tests/test_utils.py
python -m pytest tests/test_utils.py -ktest_zipped_paths_extracted
```
## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": "5.1.0"
  },
  "charset_normalizer": {
    "version": "3.0.1"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.4"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.10.8"
  },
  "platform": {
    "release": "5.15.90.1-microsoft-standard-WSL2",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.28.2"
  },
  "system_ssl": {
    "version": "1010106f"
  },
  "urllib3": {
    "version": "1.26.14"
  },
  "using_pyopenssl": false
}
```

Ren-hongchen:
fix #6377 
The reason why the file is not overwritten is that the test file in the TMP directory is not cleared after the test is run for the first time, and the directory and filename of the temporary file are the same after each test run.

This code skips overwriting when it finds that the file already exists:

`if not os.path.exists(extracted_path):` in utils.extract_zipped_paths()

**This means that calling this function twice in a row will not be able to detect changes in compressed files？**

So i thought that judgment might be redundant, so I removed it.

### DIFF
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -290,10 +290,11 @@ def extract_zipped_paths(path):
     # we have a valid zip archive and a valid member of that archive
     tmp = tempfile.gettempdir()
     extracted_path = os.path.join(tmp, member.split("/")[-1])
-    if not os.path.exists(extracted_path):
-        # use read + write to avoid the creating nested folders, we only want the file, avoids mkdir racing condition
-        with atomic_open(extracted_path) as file_handler:
-            file_handler.write(zip_file.read(member))
+
+    # use read + write to avoid the creating nested folders, we only want the file, avoids mkdir racing condition
+    with atomic_open(extracted_path) as file_handler:
+        file_handler.write(zip_file.read(member))
+
     return extracted_path
 
 


sigmavirus24:
This will dramatically kill performance for people using certifi and requests from zip files. We won't be accepting it

Ren-hongchen:
> This will dramatically kill performance for people using certifi and requests from zip files. We won't be accepting it

Sorry, I'm a noob.  @sigmavirus24 
Is it acceptable to use MD5 to compare existing files in a temporary directory with the compressed file to detect if the contents of the file have changed?

```
import hashlib

def extract_zipped_paths():
    ...
    if os.path.exists(extracted_path):
        with open(extracted_path, "rb") as file_handler:
            file_content = file_handler.read()
            file_hash = hashlib.md5(file_content).hexdigest()
        zip_file_hash = hashlib.md5(zip_file.read(member)).hexdigest()
        
        if file_hash == zip_file_hash:
            return extracted_path
        
    # use read + write to avoid the creating nested folders, we only want the file, avoids mkdir racing condition
    with atomic_open(extracted_path) as file_handler:
        file_handler.write(zip_file.read(member))
```

PR #6380, opened 2023-03-15
Clarified Section Header for passing data

jenshb:
This section applies not only to POST request, but all REST methods with data. Furthermore, a POST request with json body ist not complicated, but standard.


nateprewitt:
Thanks for the PR, @jenshb. This may be a bit subjective depending on the reader. We typically won't accept PRs that are making stylistic changes in the docs/code, so I believe we'll pass on this one.

PR #6383, opened 2023-03-19
Clarify docs for multipart file uploads

SyntaxColoring:
This makes some documentation changes that clarify the `files` argument of `requests.post()` et. al.

## Be more precise when we say "multiple files"

The `files` argument can either be a dict:

```python
files = {
    "field_1": file_1,
    "field_2": file_2,
}
```

Or a list of tuples:

```python
files = [
    ("field_1": file_1),
    ("field_2": file_2),
]
```

A few places implied that to "upload multiple files in one request," you had to use the list-of-tuples syntax. But the dict syntax supports multiple files just fine. The added power in the list-of-tuples syntax seems to be that you can upload multiple files *to the same form field.* So, we now say that.

## Update the reference docs

The API reference documentation did not completely describe what values were acceptable for `files`, and in some cases it was misleading.

* It didn't mention the list-of-tuples syntax at all.
* It said you could pass a "dictionary of `'name': file-like-objects`". The pluralization there makes it look like you can do something like `{"name": [file_1, file_2]}`, but that's not correct.
* It didn't describe what you were allowed to provide as a "`fileobj`". (It can be a file-like object, or `str` contents, or `bytes` contents.)
* It said `name` to refer to the value that Requests uses as the form field name. This could be misconstrued as the file name. [RFC 7578](https://www.rfc-editor.org/rfc/rfc7578) and other examples within Requests call it the "field name" or "form field name."

My sources for these changes:

* The existing [POST a Multipart-Encoded File](https://github.com/psf/requests/blob/7f694b79e114c06fac5ec06019cada5a61e5570f/docs/user/quickstart.rst?plain=1#L300) section.
* The existing [POST Multiple Multipart-Encoded Files](https://github.com/psf/requests/blob/7f694b79e114c06fac5ec06019cada5a61e5570f/docs/user/advanced.rst?plain=1#L391) section.
* [Request's type stubs](https://github.com/python/typeshed/blob/8080e491d2fa92b1b1390c87fac23a1a38dcd919/stubs/requests/requests/sessions.pyi#L91).

It's difficult to describe this API concisely because it accepts so many different input shapes, and these additions do make the rendered docs feel a bit crowded.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -386,13 +386,13 @@ parameter of ``None``. If you want to set a maximum size of the chunk,
 you can set a ``chunk_size`` parameter to any integer.
 
 
-.. _multipart:
+.. _multiple_multipart_files_same_field:
 
-POST Multiple Multipart-Encoded Files
--------------------------------------
+POST Multiple Multipart-Encoded Files to the Same Field
+-------------------------------------------------------
 
-You can send multiple files in one request. For example, suppose you want to
-upload image files to an HTML form with a multiple file field 'images'::
+You can send multiple files to the same form field in one request. For example, suppose you want to
+upload image files to an HTML form with a multiple-file field 'images'::
 
     <input type="file" name="images" multiple="true" required="true"/>
 
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -297,6 +297,9 @@ and it will be encoded automatically:
 
 Note, the ``json`` parameter is ignored if either ``data`` or ``files`` is passed.
 
+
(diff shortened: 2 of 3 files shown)


sigmavirus24:
The primary benefit of using list of tuples is that order is preserved which is important for some servers that expect fields in a very specific ordering which is way more than it should be

SyntaxColoring:
@sigmavirus24 Thanks. To be clear, do you think some of this PR needs to change to reflect that, or were you just commenting to lend context?

SyntaxColoring:
Bump. Are any changes needed for this?

PR #6397, opened 2023-03-30
chore: move coverage config to `pyproject.toml`

SauravMaheshkar:
This PR aims to minimize the number of configuration files for the project. Instead of a separate `.coveragerc` for configuring coverage, we now would have the required configuration within the `pyproject.toml` file leading to a overall minimal code structure.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,0 @@
-[run]
-omit = requests/packages/*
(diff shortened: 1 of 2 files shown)


sigmavirus24:
We don't accept cosmetic changes. Thanks for thinking of us though

PR #6404, opened 2023-04-13
Try charset_normalizer first

martinthomson:
The newer `charset_normalizer` is installed as a dependency in a lot of cases, so there isn't much point in going to `chardet` as a first stop.  It might even be possible to switch to an unconditional import.


nateprewitt:
Hi @martinthomson, this is a breaking change. We won't be able to accept this PR. Thanks for the suggestion though.

PR #6415, opened 2023-04-21
Improve documentation of request method in sessions

sodrooome:
Just a small enhancement that listing of all valid HTTP in `request()` method


nateprewitt:
Hi @sodrooome,

Thanks for the contribution! This is currently well documented elsewhere, and we've seen a lot of unnecessary bloat in docstrings over the last couple years. I think we'll have to pass on this one for the time being.

Issue #6419, opened 2023-04-21
pypy 3.7 CI build failing


graingert:
pypy 3.7 fails with a core dump:

```
Fatal Python error: Aborted

Stack (most recent call first, approximate line numbers):
  File "<frozen importlib._bootstrap>", line 220 in _call_with_frames_removed
make: *** [Makefile:8: ci] Aborted (core dumped)
```

graingert:
Looks like it's failing due to the trustme import in the nosan_server fixture

PR #6421, opened 2023-04-22
pin cryptography<40 on pypy 3.7

graingert:
Fixes https://github.com/psf/requests/issues/6419

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,7 +6,7 @@ pytest-mock==2.0.0
 httpbin==0.7.0
 trustme
 wheel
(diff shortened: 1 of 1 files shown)


nateprewitt (approved):

pquentin:
PyPy 3.7 is no longer receiving updates, are you willing to drop it?

nateprewitt:
~~I think we're going to wait until at least June before we get rid of 3.7. Given the adoption level though, we'll likely give 3-6 months of migration time before we fully drop support.~~

~~Is urllib3 moving in that direction sooner?~~

Sorry misread original comment. We'll move PyPy testing to 3.9 next but that's likely still a few months down the road.

PR #6426, opened 2023-04-23
 chore(charset issue): Resolution for issue 6102 

kalingth:
Resolution for the issue https://github.com/psf/requests/issues/6102

Now, If a username and password are passed already encoded, they will not be affected by being expected as a string. In this case, the function will encode the string-formatted attributes into bytes using the latin-1 charset by default due to convention.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,12 +6,14 @@ dev
 
 - \[Short description of non-trivial change.\]
 
-2.32.3 (2024-05-24)
+2.32.3 (2024-05-29)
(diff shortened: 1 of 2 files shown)


kalingth:
Hi, @sigmavirus24 

Thank you for your explanation in my last PR (https://github.com/psf/requests/pull/6422), and you were right. My last PR didn't solve the issue/bug. However, I believe that this PR resolves the bug.

Could you please review the PR and let me know if it's okay?

Respectfully,

PR #6430, opened 2023-04-25
v2.29.0

nateprewitt:
2.29.0 (2023-04-26)
-------------------

**Improvements**

- Requests now defers chunked requests to the urllib3 implementation to improve
  standardization. (#6226)
- Requests relaxes header component requirements to support bytes/str subclasses. (#6356)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,14 @@ dev
 
 - \[Short description of non-trivial change.\]
 
+2.29.0 (2023-04-26)
+-------------------
+
+**Improvements**
+
+- Requests now defers chunked requests to the urllib3 implementation to improve
+  standardization. (#6226)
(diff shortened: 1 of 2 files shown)


sethmlarson:
@nateprewitt Did we want to change the version ranges allowed for urllib3 to allow 2.0.0 without a warning? I figure this is safe now that urllib3 integration test suite passes w/ requests without any patches.

nateprewitt:
Ideally, I think we would leave the pin in place to minimize blast radius. We _shouldn't_ break now that the chunking is merged if users are forcibly upgrading outside the supported range. After we get some prime time with the major version bump, I'm fine cutting another minor version to expand support. I just don't want to opt people into a major version bump without warning.

sethmlarson (approved):

sigmavirus24 (approved):

pquentin:
Thank you!

PR #6449, opened 2023-05-08
add test for httpbin_secure /stream/20

graingert:


graingert:
maybe related to https://github.com/python/cpython/issues/87954 ?

PR #6450, opened 2023-05-08
optimizing the resolve-redirects method in sessions.py

RitvikY:
My attempt at optimizing the resolve redirects method


nateprewitt:
Hi @RitvikY, thanks for the contribution! Unfortunately, I don't believe this is something we'd take at this time. Requests is largely under a feature freeze and we won't often accept refactors that don't solve a specific, wide-spread use case. I don't believe this meets that criteria.

PR #6451, opened 2023-05-08
Adding new tests

abhinavyedla02:
Instead of defining prepare_url as a fixture, it could be defined as a function that is used within the fixture definitions. This would make the code more readable and avoid confusion about what prepare_url is doing.

It might be useful to add more tests that use the httpbin and httpbin_secure fixtures to test different HTTP methods, response codes, and request parameters. This would help ensure that the fixtures are working correctly and that the requests library is handling HTTP requests and responses as expected.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,9 +12,9 @@ import pytest
 from requests.compat import urljoin
 
 
-def prepare_url(value):
+def prepare_url(url):
     # Issue #1483: Make sure the URL always has a trailing slash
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @abhinavyedla02, thanks for the contribution. I'm not sure if this provides a meaningful change to the test suite. It's arguable if readability is improved and moves away from the general intention of [conftest.py](https://docs.pytest.org/en/6.2.x/fixture.html#conftest-py-sharing-fixtures-across-multiple-files). I think we'll pass on this PR.

PR #6454, opened 2023-05-11
Add os.PathLike support for cert files

### LINKED ISSUE #5936: `cert` parameter does not accept `pathlib.Path` objects
requests seems to have special handling for when the `cert` parameter is a single `str`, but does not extend the same special handling to `pathlib.Path`, meaning currently you either have to do `cert=str(path)` or `cert=(path, path)`.

## Expected Result

`Path` objects should be handled just like `str`.

## Actual Result

```
  File "/home/nyuszika7h/test/.venv/lib/python3.9/site-packages/requests/api.py", line 75, in get
  File "/home/nyuszika7h/test/.venv/lib/python3.9/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/nyuszika7h/test/.venv/lib/python3.9/site-packages/requests/sessions.py", line 542, in request
  File "/home/nyuszika7h/test/.venv/lib/python3.9/site-packages/requests/sessions.py", line 655, in send
  File "/home/nyuszika7h/test/.venv/lib/python3.9/site-packages/requests/adapters.py", line 416, in send
    self.cert_verify(conn, request.url, verify, cert)
  File "/home/nyuszika7h/test/.venv/lib/python3.9/site-packages/requests/adapters.py", line 243, in cert_verify
    conn.cert_file = cert[0]
TypeError: 'PosixPath' object is not subscriptable
```

## Reproduction Steps

```python
from pathlib import Path

import requests

requests.get('https://example.com', cert=Path('cert.pem'))
```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "2.0.5"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.2"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.9.2"
  },
  "platform": {
    "release": "5.10.0-8-amd64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.26.0"
  },
  "system_ssl": {
    "version": "101010bf"
  },
  "urllib3": {
    "version": "1.26.6"
  },
  "using_pyopenssl": false
}
```

steveberdy:
## Summary

Adds support for any path-like object, including `pathlib.Path`. See https://github.com/psf/requests/issues/5936#issuecomment-931758148 for more details.
Some things have changed since the linked issue was created, so it may be a good time to add support now.

Fixes #5936

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -276,8 +276,16 @@ class HTTPAdapter(BaseAdapter):
 
         if cert:
             if not isinstance(cert, basestring):
-                conn.cert_file = cert[0]
-                conn.key_file = cert[1]
+                # a subscriptable object
+                if hasattr(cert, '__getitem__'):
+                    conn.cert_file = cert[0]
+                    conn.key_file = cert[1]
+                elif hasattr(cert, '__fspath__'):
+                    # a path-like object implements __fspath__
+                    # see https://docs.python.org/3/library/os.html#os.PathLike
+                    conn.cert_file = cert.__fspath__()
+                else:
+                    conn.cert_file = str(cert)
             else:
                 conn.cert_file = cert
                 conn.key_file = None


sigmavirus24:
My main concern is that if we add support here, folks will want us to support it everywhere else we accept file-like objects and that's not exactly tenable for us to support

steveberdy:
Maybe it doesn't have to be advertised, and we can add in path-like support slowly where it is feasible.

steveberdy:
Should I close this?

Emolin123 (approved):

PR #6456, opened 2023-05-17
add docstring parameter hooks

cpzt:
When I read the source code, I found that the  `request` method misses parameter hooks in docstring


PabloEmidio (commented):
It'd be interesting if followed the `Request` docstring defination, but there are several cases where it's not.

juson12 (approved):

nateprewitt (approved):
Thanks @cpzt!

PR #6458, opened 2023-05-22
I knew we missed a step...

nateprewitt:
Got a bit too far ahead of myself on the release. Running Black over the patch to fix quotes.


sethmlarson (approved):

krishna13052001:
@nateprewitt , what was the problem of having single quotes ?

nateprewitt:
There is no problem with single quotes, but Requests uses [black](https://github.com/psf/black/) for formatting which [prefers double quotes](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings).

PR #6464, opened 2023-05-31
 Updated grammar

puryabzp:
This just fix the grammar in this document when I was reading it and I think it's okay to have a contributed on this project.


puryabzp:
fix grammar

sigmavirus24 (commented):
Thanks for the contribution but none of these look like valid corrections. In fact, they seem to all introduce issues rather than. Fix anything. As a result I'll be closing this

nateprewitt:
Thanks for the PR, @puryabzp. Unfortunately, I don't think we'll be able to accept it.

PR #6465, opened 2023-06-04
Add note about adapter prefix match to docs

ZetaTwo:
This PR adds a note to the documentation highlighting the behaviour of the adapter prefix matching to help users avoid potential mistakes.


nateprewitt (approved):
Thanks @ZetaTwo!

PR #6469, opened 2023-06-07
raise ClientError for 4xx and ServerError for 5xx

Zhong-z:
`raise_for_status()` will now raises 
- ClientError for status code 4xx
- ServerError for status code 5xx

Both ClientError and ServerError are subclass of HTTPError, which should be make them backward compatible for users that catching `HTTPError`

Also kept the error message the same for backward compatibility too in case users are matching the message value.

Ideally the "Client Error" and "Server Error" in the message could be removed given the class name provide the same context.
i.e
f"{self.status_code}: {reason} for url: {self.url}"


sigmavirus24:
Thanks for putting this together! We're under feature freeze now, however, and are not accepting new API additions. As such, we will be closing this

PR #6475, opened 2023-06-26
Specify that Session.headers needs to be set to a OrderedDict in Header Ordering docs

joren485:
### Description.

Small improvement to the docs: `Session.headers` (and not `Session`) should be set to an `OrderedDict` to preserve header order.


nateprewitt (approved):
The `OrderedDict` requirement is a bit redundant now with Python 3, but I don't think we've actually evaluated if we have any special logic around this. For now I think fixing the incorrect link is useful until we consider rewriting this.

Thanks @joren485!

PR #6477, opened 2023-07-04
support adding headers when using proxies

anysoft:
In some cases, when using a proxy, it is necessary to specify specific header content. Therefore, I have implemented this functionality in a simple way.

usage：
https://github.com/anysoft/requests/blob/e6b05c5edba85b6b383b39fdde4b3ab9b514f6db/tests/test_proxy.py
``` python
import requests


def uc_unicom(ip):
    try:
        headers = {
            'user-agent': 'Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/111.0.0.0 Mobile Safari/537.36 Edg/111.0.1661.41'
        }
        proxies = {
            'http': f'http://{ip}',
            'https': f'http://{ip}',
            'headers': {
                'Proxy-Authorization': 'Basic dXNlcjpwd2Q=',
                'Q-GUID': 'XXXX',
            }
        }
        response = requests.get('https://api.ip.sb/ip', headers=headers, proxies=proxies)
        print(response.text)
    except Exception as e:
        print(e)



if __name__ == '__main__':
    uc_unicom('127.0.0.1:8080')

```

### DIFF
--- a/requests/adapters.py
+++ b/requests/adapters.py
@@ -222,7 +222,7 @@ class HTTPAdapter(BaseAdapter):
                 **proxy_kwargs,
             )
         else:
-            proxy_headers = self.proxy_headers(proxy)
+            proxy_headers = self.proxy_headers(proxy, **proxy_kwargs)
             manager = self.proxy_manager[proxy] = proxy_from_url(
                 proxy,
                 proxy_headers=proxy_headers,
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Thanks for putting together a pull request and sending it. Unfortunately, Requests is under feature-freeze and has been for quite a while.

Regardless, there is a clear way to add headers for your proxy. That said, your example test is specifying `Proxy-Authorization` but near the code you altered, that's already supported. If you can't use basic authentication with your proxy, you can change the adapters to do the right thing based off of what your needs are and replace the adapter.

Issue #6070, opened 2022-02-19
Migrate test suite to support Flask>2.0


nateprewitt:
As shown in #5846 and #6069, our test suite is now thoroughly broken from Flask 2.0 onwards. We never did a deep dive to understand what caused the breakage and now changes in the underlying modules (markupsafe, werkzeug) are causing further problems. We need to sit down and work out the issue, unblocking upgrading Flask if possible.

mgorny:
Ping.

nateprewitt:
Resolved in #6496.

PR #6496, opened 2023-08-12
Upgrade to httpbin 0.10.0

nateprewitt:
This PR will upgrade the Requests test suite to use the recently released httpbin 0.10.0 that now supports the latest major version of Flask. This will let us remove our pins on pallets projects and hopefully allow distro maintainers to no longer need to support these older versions.

Thanks @mgorny and @kevin1024 for helping get this unblocked and closing out #6070!

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -26,6 +26,7 @@ jobs:
       uses: actions/setup-python@e9aba2c848f5ebd159c070c61ea2c4e2b122355e # v2.3.4
       with:
         python-version: ${{ matrix.python-version }}
+        cache: 'pip'
     - name: Install dependencies
       run: |
         make
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,11 +3,7 @@ pytest>=2.8.0,<=6.2.5
 pytest-cov
 pytest-httpbin==2.0.0
 pytest-mock==2.0.0
-httpbin==0.7.0
+httpbin==0.10.0
 trustme
 wheel
 cryptography<40.0.0; python_version <= '3.7' and platform_python_implementation == 'PyPy'
-
-# Flask Stack
-Flask>1.0,<2.0
-markupsafe<2.1


nateprewitt:
The PyPy 3.8 Windows build seems to be consistently timing out at 10 min because it's trying to build wheels for a handful of dependencies and installation is taking ~8 min of the run. We may need to either bump the time limit or see if there's a way we can get the installation step sped up.

Edit: It succeeded on the third retry. zope.interfaces, flasgger, and rpds_py seem to be the time intensive builds for wheels. I'm not sure if we can optimize this at all or if this is just a requirement to operate with PyPy. If this becomes a reoccurring issue, I'd be fine dropping back to fewer PyPy builds if there's not a clear solution.

sigmavirus24:
> Edit: It succeeded on the third retry. zope.interfaces, flasgger, and rpds_py seem to be the time intensive builds for wheels. I'm not sure if we can optimize this at all or if this is just a requirement to operate with PyPy. If this becomes a reoccurring issue, I'd be fine dropping back to fewer PyPy builds if there's not a clear solution.

Why can't we cache the wheels that are built so subsequent runs are faster/less likely to time out?

nateprewitt:
TIL [this](https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/) was an option. I've pushed up a change to enable it and we'll see if speed improves.

sigmavirus24 (approved):

nateprewitt:
Looking at the builds, the expensive wheels don't look like they're going to take advantage of the cache. They all have unique ABI for each platform and version of PyPy.

`pp39-pypy39_pp73-win`
`pp38-pypy38_pp73-win`
`pp39-pypy39_pp73-macosx_10_9`
etc.

The issues seem isolated to Windows and showed up on the 3.12-dev builds too. It also was happening with other builds so it's not unique to the httpbin changes. I'll merge the cache change because it appears to have some minor improvements for the other builds.

I also see @mgorny is working on a more involved fix in psf/httpbin to deal with windows. I'd rather forego those for now and bump the timeout to 12 min which seems to be the upper bound for the builds that are currently failing. We can reevaluate if it becomes a nuisance.

PR #6504, opened 2023-08-13
Add minimum PEP518 build-system specification

nateprewitt:
We've been missing the minimum PEP 518 build specification in our pyproject.toml. This PR adds that, along with an explicit backend to remove any ambiguity for alternative build tools.


nateprewitt:
Fully addressed in cb2c800518423e0042789ab09f8f6e719428a444.

PR #6520, opened 2023-09-05
Create python-app.yml

leonardoadame:


nateprewitt:
Hi @leonardoadame, I'm not sure it makes sense to add this to Requests. We already have testing infrastructure set up and it appears this was lifted from another project. We'll go ahead and close this out.

PR #6522, opened 2023-09-07
Response Object Code cleanup suggestions

xxdunedainxx:
Any thoughts here? 

I felt like the exception driven logic of `ok()` wasn't ideal, so I went in an added a new method: `is_raisable_status()`, which basically checks if the response code is 400 or more (could be improved?). 

Also put `reason = self.decode_reason()` reason decoding into its own method.

If theres interest in getting this change pushed thru, I'll put in the effort to get unit tests updated and written, and update some of the method documentation 😂

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -760,11 +760,7 @@ class Response:
         the status code is between 200 and 400, this will return True. This
         is **not** a check to see if the response code is ``200 OK``.
         """
-        try:
-            self.raise_for_status()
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @xxdunedainxx,

I don't think this is something we'd accept as it breaks the established behavior of the API which we wouldn't do in the current minor version.

We'll pass on this for now, but thanks for the contribution.

Issue #4956, opened 2019-02-01
Content-Length is not checked, resulting in short reads with no error


debugmiller:
Currently, requests does not detect when the length of a response body does not match the `Content-Length` header. This behavior results in undetected failures when a TCP connection is closed early (for example) and is contrary to the behavior of other tools (e.g. `curl`) and libraries (e.g. `reqwest`) which fail if a body length does not match `Content-Length` [3]. It is also contrary to the HTTP 1.1 RFC [5]. Frustration with this behavior is well documented ([1][2][3][4]) and was first reported as a bug 5 years ago [5]. In [1] it was agreed that the underlying checking should be done in urllib3. This was implemented in urllib3 v1.17 with the `enforce_content_length` setting [6]. However the setting defaults to False. At the time this was implemented in urllib3, there was a merge on requests:proposed/3.0.0 [7] that set `enforce_content_length` to True so that short reads would be detected (~~it was later reverted~~ edit: it wasn't reverted), but no such merge happened on requests 2.x branch.

Requests master now requires urllib3>=1.21 so we are guaranteed to have enforce_content_length. There is a strong argument to be made that `enforce_content_length=True` should be the default but at the very least it should be possible for a user to opt into this. 

[1] https://github.com/requests/requests/pull/2833
[2] https://github.com/requests/requests/issues/2275
[3] https://blog.petrzemek.net/2018/04/22/on-incomplete-http-reads-and-the-requests-library-in-python/
[4] https://news.ycombinator.com/item?id=16896899
[5] https://github.com/requests/requests/issues/1855
[6] https://github.com/urllib3/urllib3/pull/949
[7] https://github.com/requests/requests/pull/3563.

## Expected Result

HTTP bodies that do not match the HTTP `Content-Length` header should be detected

## Actual Result

HTTP bodies that do not match the HTTP `Content-Length` header are not detected.

## Reproduction Steps

https://github.com/requests/requests/pull/2833#issuecomment-148813870

## System Information

```
$ python -m requests.help
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.7"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.0"
  },
  "platform": {
    "release": "18.2.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.21.0"
  },
[...]

PR #6523, opened 2023-09-09
Pass response_kw to HTTPConnectionPool through HTTPAdapter.send

joren485:
### Summary
This PR adds kwargs arguments to `HTTPAdapter.send`, which it passes to `HTTPConnectionPool.urlopen` in urllib3.

### Description
As discussed in #4956, urrllib3 recently changed the default value of `enforce_content_length` from `False` to `True`. The new default seems like a sane choice, but in some use cases the `Content-Length` header should not be enforced. To change the default behavior, urllib3 allows application code to set the `enforce_content_length` argument. As far as I know, `requests` does not have a way to pass this argument to urllib3.

The `HTTPConnectionPool.urlopen` method in urllib3 has the `**response_kw` kwargs argument to pass extra arguments down to the response parser. This PR adds a similar to argument to `HTTPAdapter.send`.

With this PR, users can override the `HTTPAdapter.send` method to pass extra arguments to `HTTPConnectionPool.urlopen`. For example, this enables users to explicitly set the `enforce_content_length`.

### Example
```Python
import requests

class EnforceContentLengthAdapter(HTTPAdapter):
    def send(self, *args, **kwargs):
        kwargs["enforce_content_length"] = False
        return super().send(*args, **kwargs)

s = requests.Session()
s.mount("http://", EnforceContentLengthAdapter())
s.mount("https://", EnforceContentLengthAdapter())

r = s.get("http://localhost:8080/")
print(r.raw.enforce_content_length) # Returns False
```


sigmavirus24:
Hi @joren485 ,

Requests is under a feature freeze. Further, even though this _may_ seem like a backwards compatible change, it in fact will likely cause issues for a good percentage of users if they are unware. Many people implement customer HTTPAdapters for various reasons and I'd estimate at least 50% copy the exact signature out of requests when they write the code and don't expect to need to update things after the fact. This effectively expands the signature in a way that isn't backwards compatible as now, someone may start using a library that has too permissive of a lower bound on requests with an adapter that requires this and it will cause confusing and unexpected errors in passing too many arguments. This will cost countless people an unnecessary amount of time trying to debug things.

With that in mind, I could see a potential benefit to adding a separate attribute on the HTTPAdapter that is not added to any function signature (i.e., not to `__init__`, `send`, etc.) and is pickled properly that is a dictionary of `urllib3_response_options`, e.g.,

```py
self.urllib3_response_options = {}
```

As the default that could then be updated with `self.urllib3_response_options.setdefault("enforce_content_length", True)`. This could then be used in https://github.com/psf/requests/blob/881281250f74549f560408e5546d95a8cd73ce28/src/requests/adapters.py#L485-L497 where I would create a copy of the dictionary and then override it with what we have in the kwargs to that call, and the explode (`**`) into the `urlopen` call.

Does that make sense to you?

joren485:
@sigmavirus24

That definitely makes sense to me! I pushed a first version of your idea. Is this what you had in mind?

It would work like this:
```Python
import requests

class EnforceContentLengthAdapter(HTTPAdapter):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.urllib3_response_options.setdefault("enforce_content_length", False)


s = requests.Session()
s.mount("http://", EnforceContentLengthAdapter())
s.mount("https://", EnforceContentLengthAdapter())

r = s.get("http://localhost:8080/")
print(r.raw.enforce_content_length)
```

I am aware of the feature freeze, but I dot not consider this change a feature, but rather backwards compatibility as the default behavior of requests has changed. Is this PR allowed?

src/requests/adapters.py
```diff
@@ -455,7 +457,6 @@ def send(
             raise InvalidURL(e, request=request)
 
         self.cert_verify(conn, request.url, verify, cert)
-        url = self.request_url(request, proxies)
```
sigmavirus24: Let's not refactor this here and keep the change as tightly scoped as possible

src/requests/adapters.py
```diff
@@ -154,6 +154,8 @@ def __init__(
 
         self.init_poolmanager(pool_connections, pool_maxsize, block=pool_block)
 
+        self.urllib3_response_options = {}
```
sigmavirus24, line 158: If you want this to survive pickling you need to add it to `__attrs__` above, I suspect people will expect this to be picked appropriately

joren485:
@sigmavirus24 I implemented your suggestions.

I removed the need for `self.urllib3_response_options.copy()` by providing `**self.urllib3_response_options` directly to `urlopen`, because it minimizes the code changes and looks cleaner imo. This way, user code is still not able to override the given arguments (and only provide new arguments like `enforce_content_length`), as providing duplicate arguments throws a `TypeError`. 

However, if you prefer the use of a separate dictionary with `self.urllib3_response_options.copy()`, I will happily change it back.

src/requests/adapters.py
```diff
@@ -494,6 +500,7 @@ def send(
                 retries=self.max_retries,
                 timeout=timeout,
                 chunked=chunked,
+                **self.urllib3_response_options,
```
sigmavirus24: I don't believe a `TypeError` here is actually a good user experience, especially if we want this to be something we get into a relatively soon release because this has never raised a `TypeError` and if it starts to do so now, it is not something users will be prepared to handle - hence why I suggest overriding arguments with what's specified here already.

src/requests/adapters.py
```diff
@@ -130,6 +130,7 @@ class HTTPAdapter(BaseAdapter):
         "_pool_connections",
         "_pool_maxsize",
         "_pool_block",
+        "urllib3_response_options",
```
nateprewitt, line 133: nit; the values in here could be anything for the `urlopen`/`_make_request` methods. Defining them as response_options seems too tightly scoped. I left another comment below, but perhaps `urllib3_urlopen_kwargs` or just `urllib3_send_kwargs` would be more correct?

src/requests/adapters.py
```diff
@@ -481,20 +487,25 @@ def send(
...
+                "redirect": False,
+                "assert_same_host": False,
+                "preload_content": False,
+                "decode_content": False,
+                "retries": self.max_retries,
+                "timeout": timeout,
+                "chunked": chunked,
+            }
+        )
+
```
nateprewitt, line 506: I'm wondering if this should be done in the opposite order (defaults updated by `self.urllib3_response_options`). If we add this feature, we'll almost certainly get future issues with questions like "Why can't I change `preload_content` or `decode_content` with the \<variable name\>?".

If we're going to expose something like this to avoid custom `send` implementations, it should probably be the full set of arguments. We can still keep the same defaults, but whatever users want to pass us, we should just forward.

nateprewitt (commented):
I'm +0 on this. I understand the use case, although it's fairly straightforward to work around. If we're going to add an override hatch directly onto the adapter, it should just override anything in the urlopen kwargs at that point.

joren485:
@nateprewitt Thank you very much for your suggestions, I am happy to implement them. @sigmavirus24 do you agree with the suggested changes?

> I understand the use case, although it's fairly straightforward to work around.

Could you expand on this? The only way to work around the new `enforce_content_length` behavior, that I know of, is to monkey patch `urllib3`. Do you have a better workaround?

src/requests/adapters.py
sigmavirus24: I wanted to be very clear this week only affect one thing because naming things too generically makes others and they should work for other things. Based on your other comments, I'm thinking we should add the word `extra` to the band to be clear these are only extra args

src/requests/adapters.py
sigmavirus24: In my experience if there's something like this designed to help people access something as an escape hatch it will be misunderstood and misused. Copying the dictionary and updating it will avoid footguns. No one will override the headers, method, data, or anything else. I'm not particularly strongly invested in this feature as a whole but the other way I think we can enable this (because overriding send is a nightmare) is breaking out the actual call to urlopen into a different method that can be overridden.

sigmavirus24:
> The only way to work around the new enforce_content_length behavior, that I know of, is to monkey patch urllib3

You can inherit from the adapter and override the send method that you are modifying here.

You could also create a pool manager that sets this parameter and set it on the adapter. Specifically, a Pool manager accepts ConnectionPool kwargs and the Connection Pool accepts response kwargs.

Honestly, it's probably best to update the adapter to pass this to how we create the pool manager now that we're talking about it. It will be far fewer copies and updates here and will look like how folks can already use this today

sigmavirus24:
Sorry for the twists here @joren485. I've been reviewing this entirely from my phone so much harder to review docs etc at the same time and make a better decision

joren485:
> You can inherit from the adapter and override the send method that you are modifying here.

I wanted to prevent the need for such an override with this PR, as `HTTPAdapter.send` is quite a large method and overriding it to change a single call in the middle of the method would require manually copying any upstream changes.
 
> You could also create a pool manager that sets this parameter and set it on the adapter. Specifically, a Pool manager accepts ConnectionPool kwargs and the Connection Pool accepts response kwargs.

I do not see how this would work, as the `connection_pool_kw` kwargs set by `PoolManager._new_pool` are not used in `HTTPConnectionPool.urlopen` (or `HTTPConnectionPool._make_request`). `PoolManager` does not seem to have control over the arguments passed to `HTTPConnectionPool.urlopen`. Please correct me if I am wrong.

PR #6527, opened 2023-09-13
Autoclose specific issue templates

sigmavirus24:
We spend a fair amount of time closing issues because people don't read the template closely or understand what they're being told. Let's take advantage of being able to auto-label an issue based on the template and then trigger a workflow to auto-magically close the issue with a message and lock the issue.

### DIFF
--- a/.github/workflows/close-issues.yml
+++ b/.github/workflows/close-issues.yml
@@ -14,7 +14,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - env:
-          GITHUB_TOKEN: ${{ secrets.MY_TOKEN }}
           ISSUE_URL: ${{ github.event.issue.html_url }}
         run: |
           gh issue close $ISSUE_URL \
@@ -26,7 +25,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - env:
-          GITHUB_TOKEN: ${{ secrets.MY_TOKEN }}
           ISSUE_URL: ${{ github.event.issue.html_url }}
         run: |
           gh issue close $ISSUE_URL \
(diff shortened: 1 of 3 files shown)


.github/workflows/close-issues.yml
```diff
@@ -0,0 +1,35 @@
...
+permissions:
+  issues: write
+
+jobs:
+  close_qa:
+    if: github.event.label.name == 'actions/autoclose-qa'
+    runs-on: ubuntu-latest
+    steps:
+      - env:
+          GITHUB_TOKEN: ${{ secrets.MY_TOKEN }}
```
nateprewitt: I'm not familiar, what is `MY_TOKEN` in this case? Should this be `GITHUB_TOKEN` or were you planning on adding a separate token specifically for this workflow?

.github/workflows/close-issues.yml
```diff
@@ -0,0 +1,35 @@
...
+          ISSUE_URL: ${{ github.event.issue.html_url }}
+        run: |
+          gh issue close $ISSUE_URL \
+            --comment "As described in the template, this is not the appropriate place for questions. Please use Stack Overflow" \
```
nateprewitt: nit on phrasing.
```suggestion
            --comment "As described in the template, we won't be able to answer questions on this issue tracker. Please use [Stack Overflow](https://stackoverflow.com/)" \
```

nateprewitt (commented):
Two minor comments, otherwise looks good.

.github/workflows/close-issues.yml
sigmavirus24: Ah yeah, I based this heavily off the actions docs and didn't look terribly closely at this. I don't think it's necessary honestly

sigmavirus24: Removed those

nateprewitt (approved):

PR #6529, opened 2023-09-14
fix doc: correct models.Response.json() docstring

anupam-arista:
The method is doing json-decode of a response content(if any) to a python object, unlike what the docstring says..


achapkowski:
I know I'm not an admin, but this doc string change is a solid change could it be merged?

anupam-arista:
I had spent like 1 hour because these 2 words were pointing the wrong way, so maintainers please consider this nit.

src/requests/models.py
```diff
@@ -945,7 +945,7 @@ def text(self):
         return content
 
     def json(self, **kwargs):
-        r"""Returns the json-encoded content of a response, if any.
+        r"""Returns the json-decoded dict of a response, if any.
```
sigmavirus24:
```suggestion
        r"""Decodes the JSON response body (if any) as a Python object.

           This may return a dictionary, list, etc. depending on what is in the response.
```

sigmavirus24 (changes requested):

src/requests/models.py
```diff
@@ -945,7 +945,8 @@ def text(self):
         return content
 
     def json(self, **kwargs):
-        r"""Returns the json-encoded content of a response, if any.
+        r"""Decodes the JSON response body (if any) as a Python object.
```
sigmavirus24, line 948:
```suggestion
        r"""Decodes the JSON response body (if any) as a Python object.

```

sigmavirus24 (approved):

nateprewitt (approved):

Issue #6223, opened 2022-08-31
requests.package.chardet* references not assigned correctly?


babyhoo976:
I found that packages.py is assigning sys.modules['requests.package.chardet'] over and over with different modules, is this intentional or a bug?

I looked at the code and this `target` variable confuses me, it is assigned again in loop and placed with itself(so completely?), looks like a name confliction to me. Code is referenced from installed version 2.28.1.
```
target = chardet.__name__
for mod in list(sys.modules):
    if mod == target or mod.startswith(f"{target}."):
        target = target.replace(target, "chardet")
        sys.modules[f"requests.packages.{target}"] = sys.modules[mod]
```

## Expected Result

every chardet.* package maps to requests.packages.chardet.* respectively

## Actual Result

only requests.package.chardet is assigned at last.

## Reproduction Steps

import requests
import sys
print([m for m in sys.modules if name.startswith('requests.packages.chardet')])

amkarn258:
Have proposed a fix here - https://github.com/psf/requests/pull/6562

nateprewitt:
Resolving as it appears #6562 ended up merged without closing this. Thanks @amkarn258!

PR #6562, opened 2023-10-30
every chardet package maps to requests.packages.chardet.* package respectively

amkarn258:
Added a fix for the issue mentioned here - https://github.com/psf/requests/issues/6223

Checked in the code, only one package of chardet was being mapped to requests.packages.chardet despite the for loop. 

Have added the fix so each chardet package maps to it's corresponding requests.packages.chardet.* respectively.


src/requests/packages.py
```diff
@@ -23,6 +23,5 @@
 target = chardet.__name__
 for mod in list(sys.modules):
     if mod == target or mod.startswith(f"{target}."):
-        target = target.replace(target, "chardet")
```
sigmavirus24, line 26: This is unfortunately necessary. This is necessary to support charset-normalizer. The idea here is to import charset if it's installed instead of chardet from this location for users.

This is why we look at `chardet.__name__` (to determine which is most important: https://github.com/psf/requests/blob/839a8edec37c81a18ac8332cfbd44f44e1ae6206/src/requests/packages.py#L8 )

So what we need to do is determine how we want to handle this appropriately.

The idea is that we have two values:

```py
target = "chardet"
target = "charset_normalizer"
```

In the former case if `mod.startswith("chardet.")` then we will be replacing `chardet` with `chardet`. In the latter case, we'd want to replace `charset_normalizer` with `chardet`.

amkarn258: Got it, but still the for loop seems necessary. It'll just be mapping the last package of chardet modules to chardet package of requests. 

If that is the expected behaviour then we can just assign the last package without running the for loop.

sigmavirus24: So I'm talking about the intent behind the line, there's obviously a bug in it. And I was trying to leave that as something for you to see for yourself.

The reality is that what we want is 

```py
sys.modules[f"requests.packages.{mod}"] = sys.modules[mod]
mod = mod.replace(target, "chardet")
sys.modules[f"requests.packages.{mod}"] = sys.modules[mod]
```

Because we may as well add charset normalizer in there but also, we want backwards compatibility with chardet

So it's a tiny bit obvious bug and we'd like the intended behavior fixed

amkarn258:
```
sys.modules[f"requests.packages.{mod}"]` = sys.modules[mod]
mod = mod.replace(target, "chardet")
sys.modules[f"requests.packages.{mod}"] = sys.modules[mod]
```

Doesn't make sense, since then the mod will itself change to chardet and will throw a keyerror. 

Have added the support for both chardet and charset_normalizer in the latest commit, with this change - 

```
sys.modules[f"requests.packages.{mod}"] = sys.modules[mod]
target = target.replace(target, "chardet")
sys.modules[f"requests.packages.{target}"] = sys.modules[mod]
```

src/requests/packages.py
```diff
@@ -24,4 +24,6 @@
 for mod in list(sys.modules):
     if mod == target or mod.startswith(f"{target}."):
         sys.modules[f"requests.packages.{mod}"] = sys.modules[mod]
+        target = target.replace(target, "chardet")
```
sigmavirus24, line 27:
```suggestion
        mod = mod.replace(target, "chardet")
```

src/requests/packages.py
sigmavirus24: Correct, it won't find the module, but it's still very wrong to do `target.replace(target, "chardet")`

src/requests/packages.py
```diff
@@ -23,6 +23,7 @@
 target = chardet.__name__
 for mod in list(sys.modules):
     if mod == target or mod.startswith(f"{target}."):
+        sys.modules[f"requests.packages.{mod}"] = sys.modules[mod]
         target = target.replace(target, "chardet")
         sys.modules[f"requests.packages.{target}"] = sys.modules[mod]
```
sigmavirus24:
```suggestion
        imported_mod = sys.modules[mod]
        sys.modules[f"requests.packages.{mod}"] = imported_mod
        mod = mod.replace(target, "chardet")
        sys.modules[f"requests.packages.{mod}"] = imported_mod
```

sigmavirus24:
Your PR as is will _still_ have the same issue you reported and does not fix it. I've added a suggestion

amkarn258:
Same checks are failing on the main branch - https://github.com/psf/requests/issues/6566

I think the issue is with werkzeug latest version which gets downloaded due to dependency with `httpbin`

amkarn258:
Is there anything more required in this PR to be merged? Or is it on standby until the httpbin issue gets resolved @sigmavirus24

sigmavirus24:
closing & reopening to trigger new build

PR #6563, opened 2023-10-31
Optimized 2 functions + added Doc Strings + fixed formatting issues(non-technical just prettyness issues)

HPD1155:
Optimized the following functions
- `help._implementation()` - Made it faster, doesn't really make an effect but nice to have. `Confirmed to still work on my system`
- `utils.urldefragauth(url)` -  removes the fragment and authentication parts from a URL in a more concise and efficient manner by directly modifying the `netloc` variable and avoiding unnecessary assignments. `Confimed to still work on my system`

Added or modified doc strings:
- `exceptions.JSONDecodeError()` - Capitalized json -> JSON so it matches the rest of the code in that file.
- `status_codes.py` - Modified some of the formatting for the code, so they all match. (No random lines of code that use multiline args )
- Added a few doc strings to functions in `utils.py`
- Added a doc string to `test_lowlevel.test_json_decode_compatibility_for_alt_utf_encodings()

### DIFF
--- a/src/requests/exceptions.py
+++ b/src/requests/exceptions.py
@@ -29,13 +29,13 @@ class InvalidJSONError(RequestException):
 
 
 class JSONDecodeError(InvalidJSONError, CompatJSONDecodeError):
-    """Couldn't decode the text into json"""
+    """Couldn't decode the text into JSON"""
 
     def __init__(self, *args, **kwargs):
         """
         Construct the JSONDecodeError instance first with all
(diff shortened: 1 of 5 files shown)


sigmavirus24:
Thanks for the contribution. Unfortunately we don't accept cleaned charges and changing the names of exception classes is a backwards incompatible change we will not accept. Finally the formatting changes you made conflict with black which we use to format the entire codebase

PR #6574, opened 2023-11-16
Fix missing space in error message

msea1:
InvalidHeader got a smushed together error message due to no trailing space on first line.


nateprewitt (approved):
Thanks, @msea1

PR #6579, opened 2023-11-20
Call url.lower() once inside get_adapter

bfontaine:
This is a re-proposal of https://github.com/psf/requests/pull/5497 three years later. The typical use-case for adaptaters is to create one and then mount it for both `http://` and `https://`. When using https:// URLs, this result in two calls to `url.lower()`, whereas with the optimization by @dbaxa this is computed only once.
IMHO the `prefix.lower()` thing should also be computed only once when we add the adapter, but I don’t know if this could break some code.

### DIFF
--- a/src/requests/sessions.py
+++ b/src/requests/sessions.py
@@ -784,8 +784,9 @@ class Session(SessionRedirectMixin):
 
         :rtype: requests.adapters.BaseAdapter
         """
+        lower_url = url.lower()
         for prefix, adapter in self.adapters.items():
-            if url.lower().startswith(prefix.lower()):
+            if lower_url.startswith(prefix.lower()):
                 return adapter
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
@bfontaine the real issue here is that in the past reasonable, low risk changes like this lead to nonsensical "optimization" PRs that do break things and otherwise look low risk. As with this PR there are no benchmarks showing the actual improvement. And we have a dependency preventing our CI from being reliable. Finally, in general small improvements like this if we start accepting them add up and obfuscate the history of the project which given it's complexity is helpful for Nate and I in our very limited time

bfontaine:
Understood. From my quick benchmark the performance gain would be of 0.05s for 1M requests, which is nothing. Sorry for the noise.

PR #6583, opened 2023-11-24
Fix broken links in docs

EFord36:
Hi,

I noticed a broken link to rfc2988 in the [timeout section of the docs](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts) and thought I would fix it.

While doing that, I thought I would run the sphinx `linkcheck` builder to check for other broken links and fix them.

There were a few, some where I've 'fixed'/updated the link are hopefully uncontroversial. 

Some are more controversial - removing sections written by the original author of requests that now have dead links, removing talks with dead links to slides, removing people's now dead GitHub account links. I wasn't sure what the right thing to do here was, but I went with my intuition, and kept commits pretty discrete so I can drop/amend them if the maintainers want me to do something different. If some are controversial enough to slow down the PR, I can always pull out the uncontroversial ones into their own PR for faster merging.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -8,7 +8,7 @@ Keepers of the Crystals
 
 Previous Keepers of Crystals
 ````````````````````````````
-- Kenneth Reitz <me@kennethreitz.org> `@ken-reitz <https://github.com/ken-reitz>`_, reluctant Keeper of the Master Crystal.
+- Kenneth Reitz <me@kennethreitz.org> `@kennethreitz <https://github.com/kennethreitz>`_, reluctant Keeper of the Master Crystal.
 - Cory Benfield <cory@lukasa.co.uk> `@lukasa <https://github.com/lukasa>`_
 - Ian Cordasco <graffatcolmingov@gmail.com> `@sigmavirus24 <https://github.com/sigmavirus24>`_.
 
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -1099,7 +1099,7 @@ The **connect** timeout is the number of seconds Requests will wait for your
 client to establish a connection to a remote machine (corresponding to the
 `connect()`_) call on the socket. It's a good practice to set connect timeouts
 to slightly larger than a multiple of 3, which is the default `TCP packet
-retransmission window <https://www.hjp.at/doc/rfc/rfc2988.txt>`_.
+retransmission window <https://datatracker.ietf.org/doc/html/rfc2988>`_.
 
 Once your client has connected to the server and sent the HTTP request, the
 **read** timeout is the number of seconds the client will wait for the server
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -291,7 +291,7 @@ versions of Requests.
 For the sake of security we recommend upgrading certifi frequently!
 
 .. _HTTP persistent connection: https://en.wikipedia.org/wiki/HTTP_persistent_connection
-.. _connection pooling: https://urllib3.readthedocs.io/en/latest/reference/index.html#module-urllib3.connectionpool
+.. _connection pooling: https://urllib3.readthedocs.io/en/latest/reference/urllib3.connectionpool.html
 .. _certifi: https://certifiio.readthedocs.io/
 .. _Mozilla trust store: https://hg.mozilla.org/mozilla-central/raw-file/tip/security/nss/lib/ckfw/builtins/certdata.txt
 
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -948,7 +948,7 @@ Link Headers
 Many HTTP APIs feature Link headers. They make APIs more self describing and
 discoverable.
 
-GitHub uses these for `pagination <https://developer.github.com/v3/#pagination>`_
+GitHub uses these for `pagination <https://docs.github.com/en/rest/guides/using-pagination-in-the-rest-api>`_
 in their API, for example::
 
     >>> url = 'https://api.github.com/users/kennethreitz/repos?page=1&per_page=10'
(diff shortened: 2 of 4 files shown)


docs/community/out-there.rst
```diff
@@ -4,7 +4,7 @@ Integrations
 Python for iOS
 --------------
 
-Requests is built into the wonderful `Python for iOS <https://itunes.apple.com/us/app/python-2.7-for-ios/id485729872?mt=Python8>`_ runtime!
+Requests is built into the wonderful `Python for iOS <https://apps.apple.com/us/app/python-2-5-for-ios/id577916777>`_ runtime!
```
EFord36: I believe this is a similar app by the same developer, but it's python 2.5 (long deprecated), and doesn't work on the most recent version of iOS anyway.

I was tempted to just delete the whole 'Python for iOS' section here, but I thought I would see what a maintainer thought before deleting the whole section.

docs/community/out-there.rst
```diff
@@ -15,7 +15,6 @@ Articles & Talks
 ================
 - `Python for the Web <https://www.gun.io/blog/python-for-the-web>`_ teaches how to use Python to interact with the web, using Requests.
 - `Daniel Greenfeld's Review of Requests <https://pydanny.blogspot.com/2011/05/python-http-requests-for-humans.html>`_
-- `Issac Kelly's 'Consuming Web APIs' talk <https://issackelly.github.com/Consuming-Web-APIs-with-Python-Talk/slides/slides.html>`_
```
sigmavirus24: This is actually an easy one to fix: https://issackelly.github.com/Consuming-Web-APIs-with-Python-Talk/slides/slides.html is from the very old GitHub pages and they switched the domain from `github.com` to `github.io`

AUTHORS.rst
```diff
@@ -124,7 +124,7 @@ Patches and Suggestions
 - Bryce Boe <bbzbryce@gmail.com> (`@bboe <https://github.com/bboe>`_)
 - Colin Dunklau <colin.dunklau@gmail.com> (`@cdunklau <https://github.com/cdunklau>`_)
 - Bob Carroll <bob.carroll@alum.rit.edu> (`@rcarz <https://github.com/rcarz>`_)
-- Hugo Osvaldo Barrera <hugo@barrera.io> (`@hobarrera <https://github.com/hobarrera>`_)
```
sigmavirus24: I'd rather not remove these. I don't think there's value in removing these links to folks even if they've deleted their account.

docs/community/out-there.rst
EFord36: ah, can't believe I didn't think of that 🤦 Thanks for finding it!

I've updated the commit to 'fix' it to the correct link rather than removing the line.

AUTHORS.rst
EFord36: Fair enough - I've removed all cases like this from the relevant commit.

I've left in the case where the GitHub user has changed, but let me know if you want me to remove that too.

EFord36:
@sigmavirus24 Thanks for your comments - I think this is ready to look at again whenever you have time. No rush from my side, I'm sure you're busy, just didn't want it to get stuck in process if you thought I was still working on it!

Jamim (approved):
That's great! Thank you, @EFord36!
I believe these changes are ready to be merged.

docs/user/advanced.rst
```diff
@@ -1097,7 +1097,7 @@ The **connect** timeout is the number of seconds Requests will wait for your
 client to establish a connection to a remote machine (corresponding to the
 `connect()`_) call on the socket. It's a good practice to set connect timeouts
 to slightly larger than a multiple of 3, which is the default `TCP packet
-retransmission window <https://www.hjp.at/doc/rfc/rfc2988.txt>`_.
+retransmission window <https://datatracker.ietf.org/doc/html/rfc2988>`_.
```
Jamim, line 1102: > It's a good practice to set connect timeouts to slightly larger than a multiple of 3, which is the default [TCP packet retransmission window](https://datatracker.ietf.org/doc/html/rfc2988).

Despite the fact I already approved this PR, there's still a room for improvement.
[RFC 2988](https://datatracker.ietf.org/doc/html/rfc2988) was obsoleted by [RFC 6298](https://datatracker.ietf.org/doc/html/rfc6298), so this link might be updated even further, but it would require additional changes since newer RFC makes the whole statement not correct due to [this change](https://datatracker.ietf.org/doc/html/rfc6298#section-7).
So this might be handled in an additional PR.

EFord36: Thanks - seems like a sensible comment, agree it makes sense for an additional PR.

EFord36:
Thanks @Jamim for the review!

It looks like GitHub still wants an approval from a reviewer with write access - I'm guessing you don't have write access (just checking in case something has gone wrong with the process)?

I've just rebased against main so this is fresh and to give the GitHub checks another change to pass - looks like the `build` checks for any python >3.7 are failing with this error:

```
ImportError: cannot import name 'parse_authorization_header' from 'werkzeug.http' (/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/werkzeug/http.py)
```

But the 3.7 ones pass, and this error also occurred in the checks for the latest commit to main, so this seems to be an unrelated `wekzeug` dependency-related issue to me.

Jamim:
> I'm guessing you don't have write access

That's true. I just wanted to support your contribution 😃

sigmavirus24:
@EFord36 would you be willing to rebase/resolve the conflicts here?

EFord36:
> @EFord36 would you be willing to rebase/resolve the conflicts here?

Hi,

I've done that now, thanks for the reminder!

sigmavirus24 (approved):

sigmavirus24:
Thank you!

Issue #6586, opened 2023-11-27
The Content-Length header for string `data` counts Unicode characters in the string when it should count encoded bytes


bruceadams:
A call like this:

```python
response = requests.post("https://example.com", data="👍👎")
```

auto sets the `Content-Length` header in the request to `2` when it should be `8`.

I hit this issue was making a request with a JSON body to a service I own (running behind AWS API Gateway) and having the service complain that there was no closing brace `}` in the JSON body. I was passing the JSON body into requests as a string to the `data` argument. It turns out that API Gateway ignores any body bytes beyond the `Content-Length` in the request. Turning up detailed logging on API Gateway, I can see the request headers and realized the value in the `Content-Length` header didn't match the number of bytes in the body.

A quick workaround is to encode the string into bytes before passing it into Requests.

This produces a Content-Length header with the correct value of `8`:

```python
response = requests.post("https://example.com", data="👍👎".encode("utf-8"))
```

## Expected Result

On a server receiving a POST from Requests, I expect the `Content-Length` header value to match the number of bytes in the body of the request. See [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#field.content-length).

## Actual Result

In the specific case where Request's `data` argument is set as a string containing characters which encode into multi-byte UTF-8, the value in the `Content-Length` header is incorrect. Requests appears to be counting the number of Unicode characters in the string instead of the number of bytes that will be sent to the server.

## Reproduction Steps

```python
>>> import requests
>>> thumbs_up_down = "👍👎"
>>> len(thumbs_up_down)
2
>>> len(thumbs_up_down.encode())
8
>>> pending_request = requests.Request("POST", "https://example.com", data=thumbs_up_down)
>>> prepared_request = pending_request.prepare()
>>> prepared_request.headers
{'Content-Length': '2'}
```

I opened a pull request, #6587, that adds a failing unit test that demonstrates this problem.

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "3.3.2"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.9.16"
  },
  "platform": {
    "release": "23.1.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.31.0"
  },
[...]

PR #6589, opened 2023-11-28
Enhance `super_len` to count encoded bytes for str

bruceadams:
This is a possible fix for issue #6586.

I am not at all confident that this is an appropriate change for Requests. All of the unit tests pass locally for me (specifically `make ci`).

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -134,6 +134,9 @@ def super_len(o):
     total_length = None
     current_position = 0
 
+    if isinstance(o, str):
+        o = o.encode("utf-8")
+
     if hasattr(o, "__len__"):
         total_length = len(o)
 
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -1808,6 +1808,23 @@ class TestRequests:
 
         assert p.headers["Content-Length"] == length
 
+    def test_content_length_for_bytes_data(self, httpbin):
+        data = "This is a string containing multi-byte UTF-8 ☃️"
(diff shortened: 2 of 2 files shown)


src/requests/utils.py
```diff
@@ -134,7 +134,10 @@ def super_len(o):
     total_length = None
     current_position = 0
 
-    if hasattr(o, "__len__"):
+    if isinstance(o, str):
```
sigmavirus24, line 137: More simply this can be an if without changing the following conditions and just set `o = o.encode(...)` in the block then let the rest of the logic work. As it will hit another condition which will get it's length

sigmavirus24 (changes requested):

src/requests/utils.py
bruceadams: Sounds fine to me. My usual workflow would be to rewrite my commit and do a force push to my branch (in my fork). Is this ok or would you prefer a separate commit for the change?

sigmavirus24: That's my workflow as well. Feel free to do that

goelbenj:
@sigmavirus24 would you mind taking at these new changes?

sigmavirus24:
closing and re-opening to trigger a new build

nateprewitt (approved):
I'll leave any final comments and merging to @sigmavirus24 but this looks reasonable to me. Thanks @bruceadams!

sigmavirus24 (approved):

PR #6590, opened 2023-11-28
Perform correct length measurement for Python strings in HTTP request `data`

goelbenj:
This PR addresses the problem detailed in #6586. The fix involves measuring the encoded byte-length of the string as per [RFC 9910](https://www.rfc-editor.org/rfc/rfc9110#field.content-length).

This PR includes two tests that verify the fix addresses the problem. Additionally, all tests pass locally, so no regressions have occurred.


sigmavirus24:
Closing as a duplicate of #6589

PR #6594, opened 2023-12-01
Limit werkzeug<3 for CI only, fixes #6541

### LINKED ISSUE #6541: 'parse_authorization_header' not found in werkzeug 3.0.0
I run in to the following error:

`ImportError: cannot import name 'parse_authorization_header' from 'werkzeug.http' (/usr/local/lib/python3.9/site-packages/werkzeug/http.py)`.

## Reproduction Steps

```
pip install -r requirements_dev.txt

```

I'm able to resolve it if I downgrade werkzeug to a lower version (2.3.7). I think this needs to be accounted for in the dependencies or code.

bruceadams:
I believe this change will fix CI run failures, such as [these](https://github.com/psf/requests/actions/runs/7038290331).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,5 +4,6 @@ pytest-cov
 pytest-httpbin==2.0.0
 httpbin==0.10.0
 trustme
+werkzeug<3.0.0
 wheel
 cryptography<40.0.0; python_version <= '3.7' and platform_python_implementation == 'PyPy'


nateprewitt:
Hi @bruceadams, thanks for the PR. This is already fixed in psf/httpbin#29. Once that's released the test suite will function normally again. I don't think we'd merge this when the long term fix is already done.

bruceadams:
Cool! This certainly felt like an upstream problem and I'd figured out that werkzeug was being pulled in via httpbin. I hadn't gone to look at either werkzeug or httpbin to see if they might be dealing with the problem. I'm glad that it's getting fixed!

PR #6606, opened 2023-12-17
Update http status code 103 and 508

jaikishpai:
Added additional definition for `103` status code and added `508` http status code


nateprewitt:
Thanks for the PR @jaikishpai. However, I think we're unlikely to accept this. 103 is still experimental as per RFC 8297 and 508 is a WebDAV specific status code which we won't be supporting.

PR #6609, opened 2023-12-20
refc: remove the if else statement and add redable format.

Krupeshgithub:
`
data = [] if data is None else data
`
This statement is little bit complex to read. It's not readable.

`
data = [] or data
`
In the above statement if data is `None` then data store list but if there anything found in the data so, data  some store value. It is readable format.


nateprewitt:
These are not semantically equivalent statements. This would be a breaking change so we won't be able to accept it.

PR #6603, opened 2023-12-15
Add now mandatory readthedocs config file

nsoranzo:
Docs builds currently fail with:

```
Problem in your project's configuration. No default configuration file found at repository's root. See https://docs.readthedocs.io/en/stable/config-file/
```

See e.g. https://readthedocs.org/projects/requests/builds/22842479/

Issue #6604, opened 2023-12-17
Links at docs site return 404


mishankov:
At https://requests.readthedocs.io/en/latest/ all links under "Usefull links" lead to 404 page

## Expected Result

Open corresponding docs page

## Actual Result

404 page

## Reproduction Steps

- Open https://requests.readthedocs.io/en/latest/
- Click on "[Quickstart](https://requests.readthedocs.io/en/latest/user/quickstart/)"

## System Information
Not applicable

mishankov:
Looks like `.html` at the end of link is what's missing

jaikishpai:
Working on it.

nateprewitt:
Resolved in #6605.

PR #6612, opened 2023-12-23
docs: specify sphinx dirhtml builder, use references instead of absolutes

miketheman:
With the requirement of a configuration file, the default builder of `dirhtml` that RTD used to use is no longer specified. This leads to URLs ending in `.html` now, which
breaks other exisitng references.

Refs: #6603
Refs: https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-builder
Refs: https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.dirhtml.DirectoryHTMLBuilder

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,6 +13,7 @@ build:
 # Build documentation in the "docs/" directory with Sphinx
 sphinx:
   configuration: docs/conf.py
+  builder: "dirhtml"
 
 # Optionally build your docs in additional formats such as PDF and ePub
 formats:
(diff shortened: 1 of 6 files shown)


miketheman:
One example:

https://requests.readthedocs.io/en/latest/user/advanced.html#socks
vs
https://requests--6612.org.readthedocs.build/en/6612/user/advanced/#socks

miketheman:
Probably should revert https://github.com/psf/requests/issues/6604 as well

sigmavirus24:
Please do include a revert of that PR in here. I didn't get a chance to object to it

miketheman:
Revert added

miketheman:
Added references instead of absolutes, so these shouldn't ever need to be changed despite the builder chosen.

sigmavirus24 (approved):
Thank you!

PR #6617, opened 2024-01-04
Closes #6565; Add documentation to `certs.where`.

### LINKED ISSUE #6565: CA-certs bundles and requests.certs.where - Improved documentation
I initially thought `requests.certs.where` (as well as `certifi.where`) returned the location of the cacert file _actually_ being used, and I could use it to verify that the correct file was loaded. I am not the only one to make this mistake: https://stackoverflow.com/a/42982144/735070.
Screenshot from 2023-11-01 05-05-56

I have realized though, that it simply return the path of the bundled cacert file that comes included with the library.

### Demonstration of this behavior:

    #test.py
    import requests, certifi,os
    print('os.environ.get("REQUESTS_CA_BUNDLE") -> ',repr(os.environ.get("REQUESTS_CA_BUNDLE")))
    print('os.environ.get("CURL_CA_BUNDLE")     -> ',repr(os.environ.get("CURL_CA_BUNDLE")))
    print('requests.cert.where())               -> ',repr(requests.certs.where()))
    print('certifi.where())                     -> ',repr(certifi.where()))

Running test.py in shell with REQUESTS_CA_BUNDLE set:

    $ REQUESTS_CA_BUNDLE=/home/jdoe/catest/my_cacert.pem python test.py
    os.environ.get("REQUESTS_CA_BUNDLE") ->  '/home/jdoe/catest/my_cacert.pem'
    os.environ.get("CURL_CA_BUNDLE")     ->  None
    requests.cert.where())               ->  '/home/jdoe/.local/lib/python3.10/site-packages/certifi/cacert.pem'
    certifi.where())                     ->  '/home/jdoe/.local/lib/python3.10/site-packages/certifi/cacert.pem'

### Improve documentation - requests.certs.where-function

Currently `requests.certs.where.__doc__` returns null. How about

_returns the path of the CA-certs bundle that is included with the requests package, ie not necessarily the bundle actually being used._

### Improve documentation - How is CA-bundle chosen

Maybe the simplest way of making this clear is to add a section with a list or decision tree that shows which file (if any) will be used as CA-certs bundle.

### Add some way of getting the bundle _actually_ being used

In my case a static function, `requests.cacerts_loaded()`. I realize this may return a file different from the one being used, if the user specifies `verify` (and maybe its possible to change in other ways too). 

### Demonstration run on

- certifi version: 2023.05.07
- requests version: 2.25.1
- Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux

tanrbobanr:
Closes #6565 by adding documentation to `certs.where`. A wrapper was used so that the `__doc__` attribute of `certifi.where` would be preserved, and to allow for greater intellisense coverage. The second issue brought up in #6565 could not be addressed, as the CA-certs default path is only taken from `certs.where`, and can be overridden on a per-instance level of `Session` instances (for example).

### DIFF
--- a/src/requests/certs.py
+++ b/src/requests/certs.py
@@ -11,7 +11,22 @@ If you are packaging Requests, e.g., for a Linux distribution or a managed
 environment, you can change the definition of where() to return a separately
 packaged CA bundle.
 """
-from certifi import where
+from certifi import where as _where
+
+
+def where() -> str:
+    """Returns the path of the default CA-certs bundle that is included with
+    the requests package. This function is essentially an alias of
+    `certify.where`.
+
+    Note that the default path is not necessarily the one being used, as it can
+    be overridden through ``requests.Session().verify``,
+    ``requests.Session().merge_environment_settings``,
+    ``requests.get(verify=...)``, and others.
+
+    """
+    return _where()
+
 
 if __name__ == "__main__":
-    print(where())
+    print(_where())


sigmavirus24:
This addresses only one sliver of that issue, not the core of what they're asking for.

tanrbobanr:
> This addresses only one sliver of that issue, not the core of what they're asking for. 

Yeah, this is my first PR so I'm kinda new to this. From what I was able to see, the `cacerts_loaded` function would not really be possible unless it was a method of Session, as (to my knowledge) the only way to change the bundle path is through `verify` (a Session argument) and through `merge_environment_settings` (a method of Session), meaning there would be no way to get this information globally (again, to my knowledge).

PR #6623, opened 2024-01-23
handle failure in apparent_encoding gracefully

dgravitate:
Under certain circumstances, we have seen the content of our response come back as a `str` instead of `bytes`.  When that happens, and we use `response.text`, the library is throwing a `TypeError`.  This PR catches that exception, and lets the `text` property gracefully handle coercing the content to a `str`.

In our case, this happened when we received a `ConnectionError`, however, I believe there may be some other situations where this could occur.

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -789,7 +789,12 @@ class Response:
     @property
     def apparent_encoding(self):
         """The apparent encoding, provided by the charset_normalizer or chardet libraries."""
-        return chardet.detect(self.content)["encoding"]
+        try:
+            return chardet.detect(self.content)["encoding"]
+        except TypeError:
+            # Somehow, we received a str here, so return no apparent_encoding, and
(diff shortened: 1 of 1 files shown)


src/requests/models.py
```diff
@@ -915,7 +920,6 @@ def text(self):
         """
 
         # Try charset from content-type
-        content = None
```
dgravitate, line 918: This was just some campground cleanup.  This `content` was never used

sigmavirus24:
Please we need way more information than you have provided before we could merge this including test cases to ensure that _if_ this is a bug that has gone unreported for nearly a decade, that it doesn't regress.

Please look at the issue template and provide as much of that detail as possible

sigmavirus24:
Nearly a month with no more explanation of what's going on here that somehow a string is being returned as content. As a result, I'm closing this.

PR #6634, opened 2024-02-06
feat: ability to custom `HTTPAdapter` instance variables while creating `Session`

scruel:
Ability to custom `HTTPAdapter` instance variables while creating `Session`.

### DIFF
--- a/src/requests/sessions.py
+++ b/src/requests/sessions.py
@@ -387,7 +387,11 @@ class Session(SessionRedirectMixin):
         "max_redirects",
     ]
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This project is under a feature freeze and it's not accepting changes like this. Thank you for thinking of us but we'll be closing this without merging

scruel:
Ok, that's fine, even I do think I will need this, I now have to apply temporary workarounds for this. If you change your mind, feel free to provide your feedback here.

PR #6641, opened 2024-02-20
Fix CI

nateprewitt:
This PR will bring in a few changes to fix our CI suite. The release of `httpbin` 0.10.2 unblocks our test suite with Flask 3.0 which should fix the breakages in the last several months. In that window, our PyPy 3.8 build has broken on Windows due to https://github.com/pyca/cryptography/issues/10371. PyPy has dropped upstream support for both 3.7 and 3.8, so it seems more straight forward to follow them than try to patch around the cryptography issue.

We'll plan to deprecate both EoS PyPy runtimes with the next Requests release alongside CPython 3.7.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,13 +12,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev", "pypy-3.8", "pypy-3.9"]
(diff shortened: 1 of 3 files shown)


nateprewitt:
Updated Sphinx to the latest release as well due to us no longer meeting the minimum required version for readthedocs to do our preview build. I'll spot check the docs once that finishes, they appear to render normally with a local build.

nateprewitt:
Docs appear to be in good shape: https://requests--6641.org.readthedocs.build/en/6641/

sigmavirus24 (approved):

PR #6655, opened 2024-03-03
Use TLS settings in selecting connection pool

sigmavirus24:
Previously, if someone made a request with `verify=False` then made a request where they expected verification to be enabled to the same host, they would potentially reuse a connection where TLS had not been verified.

This fixes that issue.

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -8,6 +8,7 @@ and maintain connections.
 
 import os.path
 import socket  # noqa: F401
+import typing
 
 from urllib3.exceptions import ClosedPoolError, ConnectTimeoutError
 from urllib3.exceptions import HTTPError as _HTTPError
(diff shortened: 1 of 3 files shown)


src/requests/adapters.py
```diff
@@ -61,12 +62,38 @@ def SOCKSProxyManager(*args, **kwargs):
         raise InvalidSchema("Missing dependencies for SOCKS support.")
 
 
+if typing.TYPE_CHECKING:
+    from .models import PreparedRequest
```
nateprewitt, line 66: Not against this, but I think this is the first time we're introducing typing into Requests. I'm curious if we want to start that or push it into typeshed since this will be precedent for future inline typing?

tests/test_requests.py
```diff
@@ -2828,6 +2828,13 @@ def test_status_code_425(self):
         assert r5 == 425
         assert r6 == 425
 
+    def test_different_connection_pool_for_tls_settings(self):
+        s = requests.Session()
+        r1 = s.get("https://invalid.badssl.com", verify=False)
+        assert r1.status_code == 421
+        with pytest.raises(requests.exceptions.SSLError):
+            s.get("https://invalid.badssl.com")
```
nateprewitt, line 2836: There may not be a better way to test this but I don't know if we have other tests that require contacting a live site with TLS disabled. That may have some durability issues and means we're going to take the first response we get back. Probably minor, but figured I'd call it out.

src/requests/adapters.py
sigmavirus24: This is for a private method (that I fully anticipate people abusing) but we're not advertising things are typed and so it's not something I'm concerned with.

tests/test_requests.py
sigmavirus24: There are many alternatives here, but those are all significantly more effort and this shows the behaviour is fixed before and after handily. I'm sure Linux folks will get pissed but I'm not as bothered about finding time later to do this a different way after we have fixed this

nateprewitt (approved):
:shipit:

tests/test_requests.py
sigmavirus24: I'll try to prioritize better (offline) tests soon

Thompson1985 (commented):
Thank you

PR #6661, opened 2024-03-13
chore: change string concatenation way (auth.py)

mrKazzila:
Referring to [Pep8's recommendation](https://peps.python.org/pep-0008/#programming-recommendations), changed the way strings are concatenated (from `+=` to `''.join`)


sigmavirus24:
We don't accept stylistic changes unless they are part of meaningful work. Furthermore, the recommendation against using string addition is only relevant in large enough numbers of concatenation and is likely irrelevant here.

mrKazzila:
Anyway, Thanks for the feedback!

PR #6662, opened 2024-03-13
Add local TLS server

sigmavirus24:
This also adds certificates for testing purposes and files to make it easy to generate/regenerate them.

This also replaces an existing test of how we utilize our pool manager such that we don't connect to badssl.com

Finally, this adds additional context parameters for our pool manager to account for mTLS certificates used by clients to authenticate to a server.


tests/certs/mtls/client/Makefile
```diff
@@ -0,0 +1,16 @@
+.PHONY: all clean
+
+client.key:
+	openssl genrsa -out $@ 2048
+
+client.csr: client.key
+	openssl req -key $< -new -out $@ -config cert.cnf
+
+client.pem: client.csr
+	openssl x509 -req -CA ./ca/ca.crt -CAkey ./ca/ca-private.key -in client.csr -outform PEM -out client.pem -days 730 -CAcreateserial
```
nateprewitt, line 10: nit; any reason we're using 730 days here v 7300 days above in [tests/certs/expired/ca/Makefile](https://github.com/psf/requests/pull/6662/files#diff-44c61b9e0b9f24c770f652b5da6da9ce7c402c4aa1b6dbc47324fbe86cead1c8)?

Not an issue but we may want to keep them consistent or note the variance as significant.

nateprewitt (approved):
minor nit, but otherwise looks good.

tests/certs/mtls/client/Makefile
sigmavirus24: Certificate authorities should last significantly longer (20-25 years) per CAB forum. This is a cert issued by that which could theoretically live longer than the CA if we did the same expiration. I figured ~2 years would be sufficient for this test while helping us make sure the make targets keep working

sigmavirus24:
Ack didn't mean to add that roadmap file. Was a brain dump for years back

sigmavirus24:
> Edit: I guess we've got test failures coming from connection resets.

Yeah I think having a handshake failure causes the socket to be in an unhappy place but I don't feel like making this a perfect server.

.coverage.enoch.2677669.XurTNrNx
nateprewitt, line 1: Was this meant to be included?

tests/certs/mtls/client/Makefile
nateprewitt: TIL, thank you!

.coverage.enoch.2677669.XurTNrNx
sigmavirus24: Ah no. I'll kill this and force push and merge when green

PR #6668, opened 2024-03-22
converted summing of strings to f-strings

igeni:
converted summing of strings to f-strings to avoid potential cases of type's mismatch and make code more clear


sigmavirus24:
We don't accept cosmetic changes that will mask any errors that are purported to be fixed by this change.

PR #6678, opened 2024-04-04
Adds json.JSONDecodeError from standard library to requests.JSONDecodeError

jon-behnken:
Below is a suggestion for slightly enhancing the `JSONDecodeError` behavior.

### Background
The [comments](https://github.com/psf/requests/blob/main/src/requests/models.py#L972) for the `json()` method of `Response` say that `requests.JSONDecodeError` should alias the JSONDecodeError from the standard `json` library (`json.JSONDecodeError`) and `simplejson.JSONDecodeError`. 

### Problem
When `simplejson` is installed, `requests.JSONDecodeError` is not an instance of `json.JSONDecodeError`

### How to reproduce
1. Install `simplejson` in a virtual environment
```
pip install simplejson
```
2. Without the proposed change, [this test](https://github.com/psf/requests/compare/main...jon-behnken:requests:main#diff-05180dda5cdc530f95d400932961d6152bd515a6447af23d22c7f335b390b9d8R2810) should fail:
```
pytest tests/test_requests.py -k test_json_decode_compatibility 
```

### Why
Implementations like this should behave as expected:
```
try:
    response = requests.get("https://www.example.com/data").json()
except json.JSONDecodeError:
    # do something
```
but this exception does not get caught. Currently you must use `requests.exceptions.JSONDecodeError` which is a bit unintuitive.

### DIFF
--- a/src/requests/exceptions.py
+++ b/src/requests/exceptions.py
@@ -7,7 +7,7 @@ This module contains the set of Requests' exceptions.
 from json import JSONDecodeError as StandardJSONDecodeError
 from urllib3.exceptions import HTTPError as BaseHTTPError
 
-from .compat import JSONDecodeError as CompatJSONDecodeError
+from .compat import JSONDecodeError as BaseCompatJSONDecodeError
 
 
 class RequestException(IOError):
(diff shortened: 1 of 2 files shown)


sigmavirus24:
I think the docs aren't sufficiently clear but with this PR you're now doubly inheriting from the stdlib JSONDecodeError if there's no simplejson which seems rather silly. The reason we inherit from whichever library is used is because people had been previously catching that and others reasonably expected to be able to catch an exception from requests. It's not meant such that you can always catch the stdlib exception class

nateprewitt:
I agree with sigmavirus24, the example behavior isn't how this exception was intended to work. It was made to provide a shim allowing code to function regardless of the dependencies available. I'd be happy to review a doc PR to clarify intent of the exception, but don't think this PR is the right path for addressing any confusion.

sigmavirus24:
It's not even how it's documented I think. I think this is purely talking about an inaccurate code comment

PR #6680, opened 2024-04-08
Add rfc9110 HTTP status code names

mbeijen:
RFC 9110 _HTTP Semantics_ obsoletes some earlier RFCs which defined HTTP 1.1. It adds some status codes that were previously only used for WebDAV to HTTP _proper_ after making the names somewhat more generic.

See https://www.rfc-editor.org/rfc/rfc9110.html#name-changes-from-rfc-7231

This commit adds the http status code names from that RFC.

RFC 9110 has been accepted as Internet Standard in June 2022.

In https://github.com/psf/requests/blob/main/docs/user/advanced.rst#compliance  it states:

> Requests is intended to be compliant with all relevant specifications and RFCs where that compliance will not cause difficulties for users. 

So it seems only logical that the status code descriptions found in that RFC are also available in Requests

### DIFF
--- a/src/requests/status_codes.py
+++ b/src/requests/status_codes.py
@@ -24,7 +24,7 @@ _codes = {
     # Informational.
     100: ("continue",),
     101: ("switching_protocols",),
-    102: ("processing",),
+    102: ("processing", "early-hints"),
     103: ("checkpoint",),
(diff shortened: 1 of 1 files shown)


sigmavirus24 (approved):

nateprewitt (approved):
Thanks, @mbeijen!

PR #6684, opened 2024-04-14
Fixed #6683

### LINKED ISSUE #6683: raise_for_status documentation is out of sync
Not a bug per se, more like a discrepancy between documentation and actual behaviour.

## Expected Result

The [documentation](https://3.python-requests.org/user/quickstart/#response-status-codes) says `response.raise_for_status()` should return the response for chaining in trivial use cases.  

Screenshot from 2024-04-14 11-19-57

## Actual Result

The code in the latest version of package doesn't actually allow it, see [raise_for_status in models.py](https://github.com/psf/requests/blob/v2.31.0/requests/models.py#L994)

## Reproduction Steps

```
...
    response.raise_for_status().json()
AttributeError: 'NoneType' object has no attribute 'json'
```

ManiFaridi:
Since the documentation says that it should return a response object in case of no difficulties and the code didn't seem to, I added a statement to return the response object at the end.
[This is the issue](https://github.com/psf/requests/issues/6683)

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -1017,8 +1017,8 @@ class Response:
 
         if http_error_msg:
             raise HTTPError(http_error_msg, response=self)
-        
-        return self
+        else:
+            return self
 
     def close(self):
         """Releases the connection back to the pool. Once this method has been
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -1017,6 +1017,8 @@ class Response:
 
         if http_error_msg:
             raise HTTPError(http_error_msg, response=self)
+        
+        return self
 
     def close(self):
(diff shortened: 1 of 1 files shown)


dennyb87 (approved):
Looks good to me!

dennyb87:
In hindsight perhaps a test wouldn't hurt

ManiFaridi:
My bad, you are absolutely right. I'll get to it.

dennyb87:
All good man :thumbsup: I don't have merge rights so can't really do much here anyway :smile: I raised it just in case :shrug:

nateprewitt:
Thanks for the PR @ManiFaridi. This isn't something we'll likely be adding though, please refer to the original issue for more information.

PR #6685, opened 2024-04-14
Fix raise for status #6683

dennyb87:
Fix `raise_for_status()` behaviour to comply with documentation as per [issue #6683](https://github.com/psf/requests/issues/6683). In particular, makes it return the response to ease chaining in trivial cases.

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -989,7 +989,7 @@ class Response:
 
         return resolved_links
 
-    def raise_for_status(self):
+    def raise_for_status(self) -> "Response":
         """Raises :class:`HTTPError`, if one occurred."""
 
         http_error_msg = ""
@@ -1017,6 +1017,8 @@ class Response:
 
         if http_error_msg:
             raise HTTPError(http_error_msg, response=self)
+        else:
+            return self
 
     def close(self):
         """Releases the connection back to the pool. Once this method has been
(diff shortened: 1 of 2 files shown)


src/requests/models.py
```diff
@@ -989,7 +989,7 @@ def links(self):
 
         return resolved_links
 
-    def raise_for_status(self):
+    def raise_for_status(self) -> "Response":
```
sigmavirus24: This breaks subclassing and those types. Please remove it

sigmavirus24 (changes requested):
I'm not convinced the documentation was ever right it that we should change the code to comport

dennyb87:
> I'm not convinced the documentation was ever right it that we should change the code to comport

I see, if that's the case feel free to close the pull request, fixing the documentation would do as well, although returning the response seemed useful.

sigmavirus24:
Yeah I'd want Nate's input too when he has time. I'm more inclined to fix the docs

nateprewitt:
This looks wrong. The documentation being linked in the original issue is not our documentation, it's Kenneth's 3.x fork. This was a requested feature several years ago to allow for more functional programming with something like `response.raise_for_status().content` which was rejected. If Kenneth added it to his 3.x branch, that's not something we'll be supporting.

I'm not able to find this line in any of our [public docs](https://requests.readthedocs.io/en/latest/user/quickstart/#response-status-codes), but if you are able to, please provide a link. I'm going to close this PR out as won't fix since this isn't expected behavior, we can discuss any instances of this in the official docs in the tracking issue. Thanks!

PR #6693, opened 2024-05-06
fix: enviroment variable PYTHONHTTPSVERIFY is not take in consideration on https requests.

alvieirajr:
In case of the ssl certificate of destination host has expired isn't possible to ignore the cert validation using env PYTHONHTTPSVERIFY=1.

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -522,7 +522,14 @@ class HTTPAdapter(BaseAdapter):
             conn = self._get_connection(request, verify, proxies=proxies, cert=cert)
         except LocationValueError as e:
             raise InvalidURL(e, request=request)
-
+            
+        python_https_verify = '1'       
+        if 'PYTHONHTTPSVERIFY' in os.environ:
+            python_https_verify = os.environ['PYTHONHTTPSVERIFY']        
+            if ((python_https_verify is None) or (python_https_verify == '') or (python_https_verify != '0') or (python_https_verify != '1')):
+                python_https_verify = '0'
+        verify = True if python_https_verify == '1' else False
+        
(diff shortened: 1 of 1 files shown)


sigmavirus24:
We have never supported this: https://requests.readthedocs.io/en/latest/search/?q=PYTHONHTTPSVERIFY

alvieirajr:
- @sigmavirus24: 

If packages that depend on the "request" package encapsulate the get() method in a way that does not give the user the option of determining whether or not to perform SSL validation, this makes it impossible to use the target endpoint if its certificate is expired. An example of this is the hugging face package. Because of this, I advocate the need to empower the user to be able to bypass the SSL check if the package they use does not provide this capability. As it stands, if the package used does not support bypassing validation and the certificate has expired, the project being built stops working until the external endpoint has its certificate updated (if it is updated). Because of this I understand that it would be a case of fixing.

If we agree that this need is valid, we then move on to the strategy of how this should be done in terms of implementation (since in terms of bureaucratic process I am still a noob in the python project, I only have 3 months in this universe). I advocate the use of the PYTHONHTTPSVERIFY environment variable because it has already been used within the Python ecosystem to signal, for example, that pip ignores SSL validation in previous versions.

I suggest using this environment variable for the sake of uniformity but I am open to suggestions.

nateprewitt:
I'm not inclined to add support for this either. Environment variables that automagically shut off TLS for every application using Requests are not something we should be aiming to support. We would consider this case extraordinary in our support model.

The correct option is to fix the certificate. If this is truly a one-off immediate need, you can monkeypatch Session.request to set `verify`. That is never a production solution, but neither is global certificate disabling.

PR #6698, opened 2024-05-09
Update README.md

nikhil-swamix:
update latest figures. higher numbers means more trust.
downloads verified on pepy
dependents verified from github dependency graph.


sigmavirus24:
We don't really update these numbers very often because it's not that important

PR #6702, opened 2024-05-14
Allow optional char detection dependencies in post-packaging

nateprewitt:
This PR is a proposal to relax our requirement to have a character detection library present with all installations of Requests. While we cannot remove our reliance on `charset_normalizer` or `chardet`, there are use cases where Requests is repackaged in applications or other tools that do not need our character detection behavior. These changes will enable them to remove those dependencies from their installation.

The current use case for this shift is to allow `pip` to remove these dependencies entirely. That should slim the package ~2MB in 24.1.0 and avoid the overhead of importing the unneeded dependency.

### DIFF
--- a/src/requests/compat.py
+++ b/src/requests/compat.py
@@ -7,12 +7,27 @@ between Python 2 and Python 3. It remains for backwards
 compatibility until the next major version.
 """
 
-try:
-    import chardet
-except ImportError:
-    import charset_normalizer as chardet
-
+import importlib
 import sys
+
+# -------------------
+# Character Detection
+# -------------------
+
+
+def _resolve_char_detection():
+    """Find supported character detection libraries."""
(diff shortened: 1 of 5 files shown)


sigmavirus24 (approved):

src/requests/compat.py
```diff
@@ -7,13 +7,28 @@
...
 
+# -------------------
+# Character Detection
+# -------------------
+
+
+def _resolve_char_detection():
+    """Find supported character detection libraries."""
+    chardet = None
+    for lib in ("chardet", "charset_normalizer"):
```
nijel, line 21: While this keeps the logic of the earlier code, I think that the library which is required on should be tried first. https://github.com/psf/requests/pull/6714 addresses this.

PR #6709, opened 2024-05-21
Add default timeout

### LINKED ISSUE #3070: Consider making Timeout option required or have a default

sigmavirus24:
This adds a default connect and read timeout value for all usage of Requests. This is to solve a long-standing issue where some systems do not have a sufficiently low default value.

Personally, I'd want these values to be much lower, but a 10 second connection timeout and a 30 second read timeout seem like they should be enough to avoid problems for the edge cases of users while also not being so large that they're basically ineffective.

Closes #3070

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,10 @@ dev
 
 - \[Short description of non-trivial change.\]
 
+**Security**
+- Add a default timeout value to Requests. The default connect timeout is 10.0
+  seconds and the default read timeout is 30.0 seconds. (#3070)
+
 2.32.1 (2024-05-20)
 -------------------
 
--- a/src/requests/sessions.py
+++ b/src/requests/sessions.py
@@ -58,6 +58,10 @@ else:
     preferred_clock = time.time
 
 
+# (connect timeout, read timeout)
+_DEFAULT_TIMEOUT = (10.0, 30.0)
+
+
 def merge_setting(request_setting, session_setting, dict_class=OrderedDict):
     """Determines appropriate setting for a given request, taking into account
     the explicit setting on that request, and the setting in the session. If a
@@ -582,7 +586,7 @@ class Session(SessionRedirectMixin):
 
         # Send the request.
         send_kwargs = {
-            "timeout": timeout,
+            "timeout": timeout if timeout is not None else _DEFAULT_TIMEOUT,
             "allow_redirects": allow_redirects,
         }
         send_kwargs.update(settings)


cmyers009:
If you add a default timeout it may break existing code bases.

For example, if I am using a GET request to download a 1 GB file then I would have to adjust timeout to make my code compatible with this proposed change.

sigmavirus24:
> If you add a default timeout it may break existing code bases.
> 
> For example, if I am using a GET request to download a 1 GB file then I would have to adjust timeout to make my code compatible with this proposed change.

You don't understand timeouts in Python. Please go read any number of past issues and blog posts about this issue before spreading FUD.

neochine:
Consider also adding -1 as timeout value such that it retains wait till infinity as requests currently behaves. That behaviour is desirable
`
ValueError: Attempted to set connect timeout to -1, but the timeout cannot be set to a value less than or equal to 0.
`
```
if timeout is None:
   timeout = _DEFAULT_TIMEOUT
if timeout == -1:
   timeout = None
# causes default timeout if not passed, if -1 is passed it becomes None, else if any other number is passed it becomes timeout after that much seconds
```

ofipify:
> For example, if I am using a GET request to download a 1 GB file then I would have to adjust timeout to make my code compatible with this proposed change.

```python
import requests
response = requests.get(
    "http://ipv4.download.thinkbroadband.com/1GB.zip", timeout=(5, 1)
)
print(f"Status: {response.status_code}, downloaded {len(response.content)} bytes")
```

Why don't you try for yourself? The read timeout (second part) kicks in if the server takes longer than that to return anything, not the overall time the request takes to finish.

Issue #6707, opened 2024-05-20
Requests 2.32.0 Not supported URL scheme http+docker


joshzcold:
Newest version of requests 2.32.0 has an incompatibility with python lib `docker`

Reproducible example

```

INTERNALERROR> Traceback (most recent call last):

INTERNALERROR>   File "/var/lib/jenkins/workspace/Development_sm_master/gravity/.nox/lib/python3.10/site-packages/requests/adapters.py", line 532, in send

INTERNALERROR>     conn = self._get_connection(request, verify, proxies=proxies, cert=cert)

INTERNALERROR>   File "/var/lib/jenkins/workspace/Development_sm_master/gravity/.nox/lib/python3.10/site-packages/requests/adapters.py", line 400, in _get_connection

INTERNALERROR>     conn = self.poolmanager.connection_from_host(

INTERNALERROR>   File "/var/lib/jenkins/workspace/Development_sm_master/gravity/.nox/lib/python3.10/site-packages/urllib3/poolmanager.py", line 304, in connection_from_host

INTERNALERROR>     return self.connection_from_context(request_context)

INTERNALERROR>   File "/var/lib/jenkins/workspace/Development_sm_master/gravity/.nox/lib/python3.10/site-packages/urllib3/poolmanager.py", line 326, in connection_from_context

INTERNALERROR>     raise URLSchemeUnknown(scheme)

INTERNALERROR> urllib3.exceptions.URLSchemeUnknown: Not supported URL scheme http+docker

INTERNALERROR> 

INTERNALERROR> During handling of the above exception, another exception occurred:

INTERNALERROR> 

INTERNALERROR> Traceback (most recent call last):

INTERNALERROR>   File "/var/lib/jenkins/workspace/Development_sm_master/gravity/.nox/lib/python3.10/site-packages/docker/api/client.py", line 213, in _retrieve_server_version

INTERNALERROR>     return self.version(api_version=False)["ApiVersion"]

INTERNALERROR>   File "/var/lib/jenkins/workspace/Development_sm_master/gravity/.nox/lib/python3.10/site-packages/docker/api/daemon.py", line 181, in version

INTERNALERROR>     return self._result(self._get(url), json=True)

INTERNALERROR>   File "/var/lib/jenkins/workspace/Development_sm_master/gravity/.nox/lib/python3.10/site-packages/docker/utils/decorators.py", line 44, in inner

INTERNALERROR>     return f(self, *args, **kwargs)

[...]

PR #6710, opened 2024-05-21
Move _get_connection to get_connection_with_tls_context

nateprewitt:
This PR is meant to provide a new publicly supported api `get_connection_with_tls_context` to better facilitate custom adapters with our recent fix for CVE-2024-35195.

As described [here](https://github.com/psf/requests/issues/6707#issuecomment-2122490651), we will be deprecating the existing `get_connection` API, and custom adapters will need to migrate their existing `get_connection` implementations to use `get_connection_with_tls_context`. Below is a very simple implementation that will get projects relying on a custom `get_connection` implementation unblocked.

However, we _strongly_ recommend maintainers evaluate if their custom `get_connection` is subject to the same issues as the CVE above and make appropriate changes as needed.

**Minimum new code to work as a pass-through**
This will be backwards compatible between versions of Requests.
```python
def get_connection_with_tls_context(self, request, verify, proxies=None, cert=None):
    self.get_connection(request.url, proxies)
```

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -9,6 +9,7 @@ and maintain connections.
 import os.path
 import socket  # noqa: F401
 import typing
+import warnings
 
 from urllib3.exceptions import ClosedPoolError, ConnectTimeoutError
 from urllib3.exceptions import HTTPError as _HTTPError
@@ -425,6 +426,15 @@ class HTTPAdapter(BaseAdapter):
         :param proxies: (optional) A Requests-style dictionary of proxies used on this request.
         :rtype: urllib3.ConnectionPool
         """
+        warnings.warn(
+            (
+                "`get_connection` has been deprecated in favor of "
+                "`get_connection_with_tls_context`. Custom HTTPAdapter subclasses "
+                "will need to migrate for Requests>=2.32.2. Please see "
+                "https://github.com/psf/requests/pull/6710 for more details."
+            ),
+            DeprecationWarning,
+        )
         proxy = select_proxy(url, proxies)
 
         if proxy:
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -374,10 +374,20 @@ class HTTPAdapter(BaseAdapter):
 
         return response
 
-    def _get_connection(self, request, verify, proxies=None, cert=None):
-        # Replace the existing get_connection without breaking things and
-        # ensure that TLS settings are considered when we interact with
-        # urllib3 HTTP Pools
+    def get_connection_with_tls_context(self, request, verify, proxies=None, cert=None):
+        """Returns a urllib3 connection for the given request and TLS settings.
+        This should not be called from user code, and is only exposed for use
+        when subclassing the :class:`HTTPAdapter <requests.adapters.HTTPAdapter>`.
+
+        :param request: The :class:`PreparedRequest <PreparedRequest>` object
+            to be sent over the connection.
+        :param verify: Either a boolean, in which case it controls whether
+            we verify the server's TLS certificate, or a string, in which case it
+            must be a path to a CA bundle to use.
+        :param proxies: (optional) The proxies dictionary to apply to the request.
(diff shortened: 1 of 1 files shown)


nateprewitt:
cc @felixfontein

src/requests/adapters.py
```diff
@@ -404,7 +414,10 @@ def _get_connection(self, request, verify, proxies=None, cert=None):
         return conn
 
     def get_connection(self, url, proxies=None):
-        """Returns a urllib3 connection for the given URL. This should not be
+        """DEPRECATED: Users should move to `get_connection_with_tls_context`
```
sethmlarson, line 418: If this method is deprecated shall we start emitting `DeprecationWarning`?

sethmlarson (approved):
Looks good, only had one suggestion that might require a new test.

src/requests/adapters.py
nateprewitt: I thought about that but it's not reachable in any of our code anymore and if someone is using a custom implementation it won't have the warning.

Happy to be wrong if I'm missing something but it seems like it will just be dead code on arrival.

nateprewitt: I guess it's better to be safe, this should be addressed in 92075b330a30b9883f466a43d3f7566ab849f91b.

kdymov:
A small update about the minimum pass-through code for someone who will stumble upon this - don’t forget to `return` the value, as you will face a different exception if you do 🙂

```python
def get_connection_with_tls_context(self, request, verify, proxies=None, cert=None):
    return self.get_connection(request.url, proxies)
```

nateprewitt:
Thank you @kdymov! I've fixed the initial example as well.

WhyNotHugo:
I'm having some issues upgrading to requests 2.32 which I think are related to this change, but I can't quite figure out how I'd need to upgrade my code.

[My HTTPAdapter subclass basically overrides the TLS context to allow weak DH ciphers](https://github.com/WhyNotHugo/django-afip/blob/3aa3c6e37cc17d5e97b697ee9f91610d15ac1643/django_afip/clients.py#L56-L67):

```py
class AFIPAdapter(HTTPAdapter):
    """An adapter with reduced security so it'll work with AFIP."""

    def init_poolmanager(self, *args, **kwargs) -> PoolManager:
        context = create_urllib3_context(ciphers=CIPHERS)
        kwargs["ssl_context"] = context

    def proxy_manager_for(self, *args, **kwargs) -> ProxyManager:
        context = create_urllib3_context(ciphers=CIPHERS)
        kwargs["ssl_context"] = context
        return super().proxy_manager_for(*args, **kwargs)
```

With `requests==2.31.0` this works, but with `requests==2.32.0` it doesn't. I keep re-reading the changes in this patchset and can't fully grasp how they're affecting me.

Do you have any guidance?

WhyNotHugo:
I'm also a bit confused by `get_connection_with_tls_context`, because it reads `Returns a urllib3 connection for the given request and TLS settings`, but no TLS settings are given.

nateprewitt:
@WhyNotHugo You may be hitting a separate issue with `init_poolmanager`. We just released 2.32.3 to address that problem with custom adapters, you may want to give that a try.

WhyNotHugo:
2.32.3 raises an entirely different exception:

```
requests.exceptions.SSLError: HTTPSConnectionPool(host='wswhomo.afip.gov.ar', port=443): Max retries exceeded with url: /wsfev1/service.asmx (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)')))
```

WhyNotHugo:
Looks like I need to add `context.load_default_certs()` and it works fine. It wasn't required before tho.

nateprewitt:
We may not be loading the cert bundle the same as we were previously. Would you mind opening a new issue with a minimal reproduction (the adapter you have above is likely fine). If you can also include the info we request from help, and if you're working behind a proxy when this happens. That'd be helpful, thanks!

PR #6714, opened 2024-05-22
Prefer charset_normalizer for char detection

nijel:
`charset_normalizer` is a dependency for requests, while `chardet` is not, so it is more likely to be found.

For vendored requests this would still load whatever is available.

The reason I noticed this memory profiling in case both libraries are present and `charset_normalizer` has way lower memory footprint than `chardet`.

### DIFF
--- a/src/requests/compat.py
+++ b/src/requests/compat.py
@@ -18,7 +18,7 @@ import sys
 def _resolve_char_detection():
     """Find supported character detection libraries."""
     chardet = None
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Just don't install chardet

nijel:
I'd love to, but it's a dependency for a third-party library (https://github.com/otsaloma/gaupol/issues/222).

What is actually the reason for trying `chardet` first when it's not dependency for requests? It's just an extra import that is very likely to fail.

PR #6720, opened 2024-05-23
Start testing on 3.13 beta

nateprewitt:
It looks like we're currently blocked on wheels for a few testing dependencies so this isn't mergable currently but we'll use this PR to monitor upstream support from other projects. Once those blockers are moved, we can merge this into our CI to prepare for the upcoming release candidates.

Currently it looks like `flsgger` and `greenlet` are having issues from httpbin. I think we have an open proposal/request to make `flsgger` optional which seems fine. `greenlet` currently has a proposed patch from Victor Stinner so I would assume that's resolved sooner than later.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "pypy-3.9", "pypy-3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13-dev", "pypy-3.9", "pypy-3.10"]
         os: [ubuntu-22.04, macOS-latest, windows-latest]
         # Python 3.8 and 3.9 do not run on macOS-latest which
         # is now using arm64 hardware.


nateprewitt:
Looks like we're still blocked on https://github.com/python-greenlet/greenlet/issues/392. Victor Stinner's patch which appears to fix the issue has been stalled for a couple months. We'll continue to keep an eye on that as we get closer to 3.13 release on October 1st, 2024.

sethmlarson:
@nateprewitt Is greenlet only needed for the gunicorn worker? Could we drop that as well in httpbin in favor of some other worker type? Or some other WSGI implementation?

nateprewitt:
Yeah, we definitely have options in httpbin to resolve the issue. I was going to wait until we're close to 3.13 release before we started pursuing changes there. We honestly don't even need flasgger there, it was a very late add to the project before development stopped. I would bias towards us just moving that to an optional dependency.

graingert:
@nateprewitt greenlet is only needed as a dependency of gevent, it shouldn't be a direct dep https://github.com/psf/httpbin/pull/51

It's currently blocking pytest-httpbin 3.13 support: https://github.com/kevin1024/pytest-httpbin/pull/90

PR #6724, opened 2024-05-28
Don't create default SSLContext if ssl module isn't present

nateprewitt:
This PR is to address a [recent regression in Emscripten support for urllib3](https://github.com/urllib3/urllib3/actions/runs/9253144050/job/25452248794). We began unilaterally creating an SSLContext without consideration for Python versions built without an ssl module. This is handled in urllib3 but missed in our usage of `create_urllib3_context`.

This PR expands on the fix in #6716 by also evaluating both the presence of the `ssl` module before creating and setting the default SSLContext.

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -73,10 +73,18 @@ DEFAULT_POOLSIZE = 10
 DEFAULT_RETRIES = 0
 DEFAULT_POOL_TIMEOUT = None
 
-_preloaded_ssl_context = create_urllib3_context()
-_preloaded_ssl_context.load_verify_locations(
-    extract_zipped_paths(DEFAULT_CA_BUNDLE_PATH)
-)
+
+try:
+    import ssl  # noqa: F401
+
+    _preloaded_ssl_context = create_urllib3_context()
+    _preloaded_ssl_context.load_verify_locations(
+        extract_zipped_paths(DEFAULT_CA_BUNDLE_PATH)
+    )
+except ImportError:
+    # Bypass default SSLContext creation when Python
+    # interpreter isn't built with the ssl module.
+    _preloaded_ssl_context = None
 
 
 def _urllib3_request_context(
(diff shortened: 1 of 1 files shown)


src/requests/adapters.py
```diff
@@ -90,13 +98,19 @@ def _urllib3_request_context(
...
     has_poolmanager_ssl_context = poolmanager_kwargs.get("ssl_context")
+    should_use_default_ssl_context = (
+        _preloaded_ssl_context is not None and not has_poolmanager_ssl_context
+    )
 
     cert_reqs = "CERT_REQUIRED"
     if verify is False:
         cert_reqs = "CERT_NONE"
-    elif verify is True and not has_poolmanager_ssl_context:
+    elif verify is True and should_use_default_ssl_context:
```
sigmavirus24, line 113: Missed this earlier but do we need any logic for verify=True without a condition here?

nateprewitt: Are you asking about the behavior if `verify is True` but `should_use_default_ssl_context` is False? This was a no-op prior to #6655 which fellback to letting the SSLContext be created when fetching the connection from the pool. [`_ssl_wrap_socket_and_match_hostname`](https://github.com/urllib3/urllib3/blob/b07a669bd970d69847801148286b726f0570b625/src/urllib3/connection.py#L755-L765) handles this for us, so I don't know if there's any additional concern from our previous behavior unless I'm missing part of the question.

sigmavirus24: Makes sense. I've a terrible headache so just looking at the branches and concerned about missing something is all.

sigmavirus24 (approved):

src/requests/adapters.py
nateprewitt: Nope, that's a good call out. I'll give it one more look before merging but I think we're ok for that case.

sigmavirus24: Oh, looking again, not on a phone, and without a migraine, I see the `cert_reqs = "CERT_REQUIRED"` line on L110 that I clearly wrote, and which is what I was thinking we may want to be concerned about. So, I'm not nearly as worried.

src/requests/adapters.py
```diff
@@ -90,13 +98,19 @@ def _urllib3_request_context(
...
         pool_kwargs["ssl_context"] = _preloaded_ssl_context
     elif isinstance(verify, str):
         if not os.path.isdir(verify):
```
sigmavirus24, line 116: Looking at this again with something of a performance mindset, do we want to cache somehow lookups to `isdir`? I'm not sure this hurts us at all, but just thinking about things that could slow us down in certain cases now. (Not for this pull request, just putting it out there)

sigmavirus24 (approved):

src/requests/adapters.py
nateprewitt: That seems like a reasonable optimization, I guess we'd need to check how much time we're actually spending on the dir check. I assume we'll get feedback if we have cases where this is a bottleneck.

AdamWill:
We also need to do this for `FileNotFoundError`, which is raised if `DEFAULT_CA_BUNDLE_PATH` isn't there. Sent as https://github.com/psf/requests/pull/6781 .

Pjrich1313 (commented):
Rerun have timestamp

Issue #6730, opened 2024-05-31
Certificate loading regression with HTTPAdapters in 2.32.3


ricellis:
It appears that in version `2.32.3` default certificates are no longer loaded for custom `HTTPAdapter` contexts when they were previously.

I guess this might be a duplicate/related to https://github.com/psf/requests/issues/6726#issuecomment-2138406456.
Also related to https://github.com/psf/requests/pull/6710#issuecomment-2137802782 - adding `load_default_certs()` resolves the issue, but this wasn't required in previous versions and thus makes upgrading to `2.32.3` breaking.

## Expected Result

With the code below using `requests` version `2.32.2` I get the URL content with no error.

## Actual Result

Using `2.32.3` I get:
```
requests.exceptions.SSLError: HTTPSConnectionPool(host='raw.githubusercontent.com', port=443): Max retries exceeded with url: /psf/requests/main/MANIFEST.in (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1006)')))
```

## Reproduction Steps

```python
import requests
import ssl
from requests.adapters import HTTPAdapter, DEFAULT_POOLBLOCK
from urllib3.util.ssl_ import create_urllib3_context

# adapted from https://github.com/IBM/python-sdk-core/blob/1c207385de627df5d12fd0a0ebd04717ce5bb29d/ibm_cloud_sdk_core/utils.py#L34
class SSLHTTPAdapter(HTTPAdapter):
    """Wraps the original HTTP adapter and adds additional SSL context."""

    def init_poolmanager(self, connections, maxsize, block=DEFAULT_POOLBLOCK, **pool_kwargs):
        """Create and use custom SSL configuration."""

        ssl_context = create_urllib3_context()
        ssl_context.minimum_version = ssl.TLSVersion.TLSv1_2
        # ssl_context.load_default_certs() # Adding this resolves the certificate issue but it was not required before

        super().init_poolmanager(connections, maxsize, block, ssl_context=ssl_context, **pool_kwargs)

session = requests.Session()
http_adapter = SSLHTTPAdapter()
session.mount('https://', http_adapter)

print(session.get(url='https://raw.githubusercontent.com/psf/requests/main/MANIFEST.in').text)
```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
[...]

Issue #6726, opened 2024-05-29
Multiple concurrent client certs broken with v2.32.3


jeffreytolar:
We use `requests` with multiple mTLS client certificates - each certificate is signed by the same CA, but they have different subjects - each subject has different permissions. Each distinct client cert is used by a different `requests.Session`
Additionally, we make use of `ThreadPoolExecutor` to make many requests in parallel.
When client certs are in use, urllib3 will [load the cert into the SSL context](https://github.com/urllib3/urllib3/blob/b07a669bd970d69847801148286b726f0570b625/src/urllib3/util/ssl_.py#L454-L458), which, with concurrent requests, will cause the shared SSL context to get modified while it's in use.

The reproducer actually fails with an exception - when we first encountered this, we were seeing the wrong certs get used. (That was with different versions of python and openssl, however, and as mentioned above, they're all signed by the same CA, unlike the reproducer)

## Expected Result

The reproducer below, when run with `requests-2.31.0`:
```
$ ./bin/python3 test.py
/tmp/py/test.py:24: DeprecationWarning: X509Extension support in pyOpenSSL is deprecated. You should use the APIs in cryptography.
  crypto.X509Extension(b"subjectAltName", False, b"DNS:localhost, IP:127.0.0.1")
/tmp/py/test.py:50: DeprecationWarning: ssl.PROTOCOL_TLSv1_2 is deprecated
  self.ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
Server started on port 8443...
  OK client2: CN: client2 / URL: /client2
  OK client3: CN: client3 / URL: /client3
  OK client1: CN: client1 / URL: /client1
```

## Actual Result

When run with `requests-2.32.3`:
```
$ ./bin/python3 test.py
/tmp/py/test.py:24: DeprecationWarning: X509Extension support in pyOpenSSL is deprecated. You should use the APIs in cryptography.
  crypto.X509Extension(b"subjectAltName", False, b"DNS:localhost, IP:127.0.0.1")
/tmp/py/test.py:50: DeprecationWarning: ssl.PROTOCOL_TLSv1_2 is deprecated
  self.ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
Server started on port 8443...
[...]

PR #6731, opened 2024-05-31
Address certificate loading regression

nateprewitt:
## Overview
This PR is intended to address two distinct issues introduced with the default cert optimizations originally introduced in 2.32.0. While we continue to refine the settings considered when opting into our optimized context, we'll no longer use the new default if any custom cert values are supplied. This addresses the concurrency issues raised in #6726.

The second piece of this will be ensuring that when opting out of the default SSLContext, we're still supplying to the default CA Cert bundle correctly. This addresses the problems noticed in https://github.com/psf/requests/pull/6710#issuecomment-2137802782 and #6730.

## Considerations

We're now duplicating a decent chunk of the [logic from cert_verify](https://github.com/psf/requests/blob/0e322af87745eff34caffe4df68456ebc20d9068/src/requests/adapters.py#L324-L357) inside [_urllib3_request_context](https://github.com/psf/requests/blob/0e322af87745eff34caffe4df68456ebc20d9068/src/requests/adapters.py#L115-L128) but without our validation exceptions. That's a potential vector for behavioral shifts in the future. We _could_ consolidate some of this behavior in one place but it's going to require constructing a dict and using `setattr` on our `conn` in `cert_verify` while setting `pool_kwargs` in `_urllib3_request_context`. I started writing that up but it feels clunky. This is probably going to be a tradeoff of risking drift like we have with Session settings and binding the two behaviors together too tightly.

## Testing
I'd like to codify the issues we've encountered through the whole 2.32.x saga in tests to hopefully avoid this in the future. Doing it cleanly without relying on external endpoints is proving to be a bit more involved than I'd like. I think we can harness some of the infrastructure added in #6662, but I haven't had a chance to really dig into that.

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -118,15 +118,23 @@ def _urllib3_request_context(
     poolmanager_kwargs = getattr(poolmanager, "connection_pool_kw", {})
 
     cert_reqs = "CERT_REQUIRED"
+    cert_loc = None
     if verify is False:
         cert_reqs = "CERT_NONE"
     elif _should_use_default_context(verify, client_cert, poolmanager_kwargs):
         pool_kwargs["ssl_context"] = _preloaded_ssl_context
+    elif verify is True:
+        # Set default ca cert location if none provided
+        cert_loc = extract_zipped_paths(DEFAULT_CA_BUNDLE_PATH)
     elif isinstance(verify, str):
-        if not os.path.isdir(verify):
-            pool_kwargs["ca_certs"] = verify
+        cert_loc = verify
+
+    if cert_loc is not None:
+        if not os.path.isdir(cert_loc):
+            pool_kwargs["ca_certs"] = cert_loc
         else:
-            pool_kwargs["ca_cert_dir"] = verify
+            pool_kwargs["ca_cert_dir"] = cert_loc
+
     pool_kwargs["cert_reqs"] = cert_reqs
     if client_cert is not None:
         if isinstance(client_cert, tuple) and len(client_cert) == 2:
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -334,10 +334,8 @@ class HTTPAdapter(BaseAdapter):
         if url.lower().startswith("https") and verify:
             conn.cert_reqs = "CERT_REQUIRED"
 
-            # Only load the CA certificates if 'verify' is a string indicating the CA bundle to use.
-            # Otherwise, if verify is a boolean, we don't load anything since
-            # the connection will be using a context with the default certificates already loaded,
-            # and this avoids a call to the slow load_verify_locations()
+            # Only load the CA certificates if `verify` is a
+            # string indicating the CA bundle to use.
             if verify is not True:
                 # `verify` must be a str with a path then
                 cert_loc = verify
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -87,6 +87,23 @@ except ImportError:
     _preloaded_ssl_context = None
 
 
+def _should_use_default_context(
+    verify: "bool | str | None",
+    client_cert: "typing.Tuple[str, str] | str | None",
+    poolmanager_kwargs: typing.Dict[str, typing.Any],
+) -> bool:
+    # Determine if we have and should use our default SSLContext
+    # to optimize performance on standard requests.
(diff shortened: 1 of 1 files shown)


src/requests/adapters.py
```diff
@@ -98,25 +115,26 @@ def _urllib3_request_context(
...
     cert_reqs = "CERT_REQUIRED"
+    cert_loc = None
     if verify is False:
         cert_reqs = "CERT_NONE"
-    elif verify is True and should_use_default_ssl_context:
+    elif _should_use_default_context(verify, client_cert, poolmanager_kwargs):
         pool_kwargs["ssl_context"] = _preloaded_ssl_context
+    elif verify is True:
+        # Set default ca cert location if none provided
+        cert_loc = extract_zipped_paths(DEFAULT_CA_BUNDLE_PATH)
```
sigmavirus24, line 128: So this is what fixes #6730 right? Because the custome Context + passing this should make their life better, right?

nateprewitt: Yeah, this should address #6730. #6667 assumed we would always be using the default context which lead to removing these [two lines](https://github.com/psf/requests/pull/6667/files#diff-a4598bf8444e0feca489e60704c809c0dbfc47b90c4364216c931b497b3312d7L293-L294). That was "ok" because we were always forcing a load on the default context at import time.

Starting in 2.32.3, we've opted out of the default in the failure cases reported between 2.32.0-2.32.2. These are cases we're taking a custom context from the user but no longer setting the `DEFAULT_CA_BUNDLE` on the connection returned from what was previously `get_connection`. Since we don't know if we have the default SSLContext in `cert_verify`, I ported this into `_urllib3_request_context` to ensure we're passing this as `pool_kwargs` when we've disabled the default.

I validated this against the issue reported in #6726 which was showing this regression after the `cert` fix in 2769cb607d4e696e2fe70802d4246ccc5abd64a8. The other reported issues like #6730 are also passing cleanly now.

Heraldk:
Is there an ETA for this change or a similar change? 2.32.3 doesn't work for us, as it broke our usage of `requests_pkcs12`. It looks like that package was updated with a temporary change [here](https://github.com/m-click/requests_pkcs12/issues/55#issuecomment-2154946453), so if a "proper" fix is not intended to be released anytime soon then we can try that. However I'd be keen to take a proper fix for both if that's expected soon!

stratakis:
The regression is there for some time and the PR is still in draft. Is there some way to move this forward?

ghost:
Maintainers, is there any update here? The latest version of HTTPie is mostly non-functional due to this issue, and we will need to drop it and migrate to other tools if this cannot be addressed.

There does not seem to be any blocking feedback on this PR - can any clarity be provided about what is currently preventing a merge? Is additional contribution required from the community?

kenan-altaki:
A friendly reminder that this issue has been stagnant for 4 months already.

agm-eratosth:
Hi @sigmavirus24  you're listed as reviewer for this. Would you like to review this or assign someone else to do so?

Pjrich1313 (approved):

frenzymadness:
Could we please move this forward somehow? How can the community help?

This makes our lives harder because we don't know whether to wait for this fix or to adapt to the new behavior caused by the security fix.

Pjrich1313 (commented):
Unblock I have the proof

Pjrich1313 (commented):
Thank you

adbuerger:
I found this pull request and one of the associated issues (https://github.com/psf/requests/issues/6730) via the information given in https://github.com/IBM/ibm-cos-sdk-python-core/issues/30. As this seems to be a topic for multiple people, I wanted to ask if it is planned to release this change/fix in some way in the near future. Thank you for your time and for your reply.

adbuerger:
@nateprewitt @Pjrich1313 @sigmavirus24 I hope that I am tagging the right people for this topic since it is unclear to me from the discussion above what aspects are left unresolved regarding this PR. I understand that @sigmavirus24 left comments on the implementations of @nateprewitt and @Pjrich1313 already approved the changes. So are there any aspects that still prevent these changes from being merged? I am not familiar with the details of this topic so I might not understand the remaining issues. Thank you very much for your consideration and for your work on this project.

nateprewitt:
Resolving in favor of #6767.

PR #6732, opened 2024-06-02
Rename lock-issues.yml to 90daylock-issues

Devante7:
d31a23b10d2c721f9ddab7baaafd5869406b1f83


nateprewitt:
Thanks for the PR, @Devante7. It doesn't look like this is a necessary change though. The title is clear on the workflow intention and the specific values are available in the file. Having to rename the file every time a value changes would be unintuitive/disruptive.

PR #6746, opened 2024-06-21
feat:add chunked parameter for request method

LoveEatCandy:
I need `chunked` parameter for uploading big size files, because header `Transfer-Encoding: chunked` is not work.


sigmavirus24:
Please open a bug report with reproducible steps explaining the bug first.

Issue #6734, opened 2024-06-04
requests 2.32.3 & urllib3 1.26.18 issue with unicode put


frenzymadness:
I'm building requests 2.32.3 in Fedora Linux and I have a problem with `test_unicode_header_name` - the test hangs.

It's reproducible - when I use urllib3 at least 2.0.2, the code works fine, with urllib3 1.26.18, it hangs waiting for a response.

## Expected Result

Older urllib3 is still allowed (`urllib3>=1.21.1,<3`) so it should work.

## Actual Result

The call to `requests.put` hangs and if killed, the stacktrace is:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/lbalhar/Software/requests/src/requests/api.py", line 130, in put
    return request("put", url, data=data, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lbalhar/Software/requests/src/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lbalhar/Software/requests/src/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lbalhar/Software/requests/src/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lbalhar/Software/requests/src/requests/adapters.py", line 667, in send
    resp = conn.urlopen(
           ^^^^^^^^^^^^^
  File "/home/lbalhar/.virtualenvs/requests/lib/python3.12/site-packages/urllib3/connectionpool.py", line 715, in urlopen
    httplib_response = self._make_request(
                       ^^^^^^^^^^^^^^^^^^^
  File "/home/lbalhar/.virtualenvs/requests/lib/python3.12/site-packages/urllib3/connectionpool.py", line 467, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/home/lbalhar/.virtualenvs/requests/lib/python3.12/site-packages/urllib3/connectionpool.py", line 462, in _make_request
    httplib_response = conn.getresponse()
[...]

PR #6757, opened 2024-07-02
Test on urllib3 1.26.x

nateprewitt:
First step to address #6734. We'll enable testing on urllib3 1.26.x which was lost when we moved the pin to support 2.x. These will fail for now until the underlying issue is addressed, but this is to ensure we're catching any future breakages.

### DIFF
--- a/src/requests/compat.py
+++ b/src/requests/compat.py
@@ -17,10 +17,10 @@ from urllib3 import __version__ as urllib3_version
 
 # Detect which major version of urllib3 is being used.
 try:
-    is_urllib3_2 = int(urllib3_version.split(".")[0]) == 2
+    is_urllib3_1 = int(urllib3_version.split(".")[0]) == 1
 except (TypeError, AttributeError):
     # If we can't discern a version, prefer old functionality.
-    is_urllib3_2 = False
+    is_urllib3_1 = True
 
 # -------------------
 # Character Detection
--- a/src/requests/compat.py
+++ b/src/requests/compat.py
@@ -10,6 +10,18 @@ compatibility until the next major version.
 import importlib
 import sys
 
+# -------
+# urllib3
+# -------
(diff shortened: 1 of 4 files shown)


sethmlarson (approved):
LGTM! Thanks for testing urllib3 :)

sigmavirus24 (approved):

src/requests/compat.py
```diff
@@ -10,6 +10,14 @@
 import importlib
 import sys
 
+# -------
+# urllib3
+# -------
+from urllib3 import __version__ as urllib3_version
+
+# Detect which major version of urllib3 is being used.
+is_urllib3_2 = urllib3_version.split('.')[0] == 2
```
sethmlarson: I believe this is a string, not an integer. Also, should do a >= check instead of equals.

nateprewitt: Good catch on the `int`. For the equivalence, I don't know if `>=` is what we want if this is specifically scoping the 2.x major version. In the same way I wouldn't want an `is_py3` to include Python 4.x. We could do something like `is_gt_urllib3_1` but that seems like it may be premature forwards-compatibility?

Right now our dependencies are scoped at `urllib3<3` and if we add this check to any other behaviors, I'd rather they stay scoped to the major version. That will let tests fail if we major version again and we can make an informed decision at that point when adding support. Otherwise, we may unintentionally carry forward behaviors that are subtly wrong.

I can see similar risks with both sides, so not a hill I'm going to die on, but that was my initial thought process.

nateprewitt: I've changed the check from checking for urllib3 2.x to check for 1.x. That leaves us open to forward compatibility without the confusing behavior with `is_urllib3_2` including newer major versions.

sigmavirus24 (approved):

Issue #6647, opened 2024-02-23
Wrong CA Certificates file used in requests.Session.send


sunnstix:
Bug in requests.sessions.py:

https://github.com/psf/requests/blob/b8be93a721792deeadd2f498b8f77cf610e7765f/src/requests/sessions.py#L579-L589

session.merge_environment_settings is called in session.request and not in session.send which means if users want to build their own prepared requests (see example below), the correct certificates file will not be used unless explicitly passed to send.

I believe these environment settings should be updated in session.send instead.

## Reproduction Steps

```python
import os
import requests

# Set CA certificates file to custom local file
os.environ["REQUESTS_CA_BUNDLE"] = "/etc/ssl/certs/ca-certificates.crt"

# Set CA certificates file to custom local file
req = requests.Request(
    method="GET", # any method
    url="https://www.example_site.mydomain.com/path" # any url for which the default cert file does not have a cert for
)

with requests.Session() as s:
    respA = s.request( # Success: will use CA certificates from REQUESTS_CA_BUNDLE
        method=req.method,
        url=req.url
    )
    assert respA.status_code == 200
    
with requests.Session() as s:
    respB = s.send( # Success: will use CA certificates from REQUESTS_CA_BUNDLE if explicitly specified
        request=req.prepare(), verify=os.environ["REQUESTS_CA_BUNDLE"]
    )
    
    assert respB.status_code == 200
    
with requests.Session() as s:
    respC = s.send( # Error: Will try to use default CA certificates instead of those specified by REQUESTS_CA_BUNDLE
        request=req.prepare() 
    )
    
    assert respC.status_code == 200
```

## Expected Result

No output

## Actual Result

```
requests.exceptions.SSLError: HTTPSConnectionPool(host='https://www.example_site.mydomain.com/path', port=443): Max retries exceeded with url: /path (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)')))
```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "3.3.2"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.6"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.10.4"
  },
  "platform": {
    "release": "5.4.0-150-generic",
    "system": "Linux"
  },
[...]

PR #6767, opened 2024-07-18
Revert caching a default SSLContext

nateprewitt:
This PR reverts the changes from #6667 to the previous behavior. Due to the number of edge cases and concurrency issues we've encountered with this change, we've decided the benefit doesn't currently outweigh the pain to existing infrastructure. We've iterated on a few tries to keep this functionality in place, but are still receiving reports of novel issues with this behavior.

We may be able to revisit this in a later version of Requests but we'll need a much more comprehensive test plan.


src/requests/adapters.py
```diff
@@ -99,19 +85,9 @@ def _urllib3_request_context(
...
-        _preloaded_ssl_context is not None and not has_poolmanager_ssl_context
-    )
-
     cert_reqs = "CERT_REQUIRED"
     if verify is False:
         cert_reqs = "CERT_NONE"
-    elif verify is True and should_use_default_ssl_context:
-        pool_kwargs["ssl_context"] = _preloaded_ssl_context
     elif isinstance(verify, str):
         if not os.path.isdir(verify):
```
nateprewitt, line 92: This is one piece I left in place from #6667. We were unilaterally considering any `verify` string to be `ca_certs` instead of detecting if it was a directory. That seems like a miss from #6655 unless I'm missing something?

stianjensen:
Is this still planned to be reverted?

Conobi:
I would also suggest to revert it ASAP.
This unsafe caching can make some applications vulnerable to DOS attack when using mTLS authentication, [causing Python to crash](https://github.com/python/cpython/issues/134698).
In its current version, requests cannot be used safely in mTLS scenarios.

sigmavirus24 (approved):

dkliban:
@sigmavirus24 I applied this PR as a patch to my server and it stopped the core dumps i was experiencing. When is this going to be merged and released?

nateprewitt:
@dkliban It should go out with the next release. We'll update once we have a clearer timeline, we'll see what we can do next week.

dkliban:
@nateprewitt Thank you for merging! Looking forward to the release!

dkliban:
@nateprewitt any updates on when this will be released?

rittneje:
@nateprewitt When will this fix be released?

jonyscathe:
I had hoped that this would also fix https://github.com/psf/requests/issues/6647 but on my local testing it doesn't seem that it has. Should this PR have fixed that? Or was that one caused by something else?

PR #6770, opened 2024-07-25
Header-based auth class

vilhelmen:
This adds a short and simple header-based auth backend (and tests!). This allows greater flexibility in auth mechanisms out of the box.

For example, I'm working on on an internal library to interface with OGC services and three providers use three different auth mechanisms. Being able to consolidate this via the auth parameter seems like the correct solution.

It's worth noting that because of when this executes, auth objects override the headers provided both by the session and request call. This has been appropriately documented, and I don't believe any other option would be preferable.

Looks like someone proposed this in 2018 (#4823) but was rejected because of a feature freeze at the time.

If this is acceptable, would the team be interested in something similar like HTTPParamAuth that uses a dict of parameters in a similar fashion? Going back to my library example, one provider uses oauth, which requests supports, one uses headers, and one uses a custom api key parameter. Totally willing to add it to this PR if it simplifies approval :)

### DIFF
--- a/docs/user/authentication.rst
+++ b/docs/user/authentication.rst
@@ -32,6 +32,27 @@ for using it::
 Providing the credentials in a tuple like this is exactly the same as the
 ``HTTPBasicAuth`` example above.
 
+Header Authentication
+--------------------
+
+Some services require authentication data in the header of the request.
+Multiple headers can be added to an authentication object to keep them separate
+from request data::
+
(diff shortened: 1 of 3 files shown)


sigmavirus24:
The simplicity of this adapter combined with the feature freeze requests has maintained for years now means we will be unable to accept this pull request. Thank you for contributing

PR #6771, opened 2024-07-25
Handle multiple paths/cert files in 'REQUESTS_CA_BUNDLE' environment variable

jhess:
An OSError just gets raised saying the cert_loc is an invalid path if you have several CA certificate paths 'REQUESTS_CA_BUNDLE' environment variable separated by the standard ";" delimiter. It should handle this by at the very least by returning an error telling the user to append the certificates to their global trust CA file and just use one certificate instead of saying it is an invalid path, if not support multiple CA certificate files in the environment variable.

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -325,6 +325,16 @@ class HTTPAdapter(BaseAdapter):
                 cert_loc = verify
 
                 if not os.path.exists(cert_loc):
+                    # Need to handle if multiple paths/cert files in 'REQUESTS_CA_BUNDLE' environment variable
+                    if ";" in cert_loc:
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @jhess, could you provide more info on why you would expect chained paths to work in this case? CA Bundle is generally defined as "_CA bundle is **a file** that contains root and intermediate certificates._". We would expect input of a single file for this parameter and that is also the documented behavior of `CURL_CA_BUNDLE` which the parameter was originally based off.

I don't think this is behavior we would generally want to add since it's both surprising to the average user and introduces other parsing complications for a relatively simple interface.

PR #6780, opened 2024-08-01
Add NewModel class to models.py

sarveshdss75:
This PR adds a new class, NewModel, to models.py which includes methods for initialization, display, and updating values


nateprewitt:
Hi @sarveshdss75, it looks like this PR may have been mistakenly opened against the Requests project instead of your personal fork. I'll close this as we won't be able to merge this.

PR #6781, opened 2024-08-01
Don't create default SSLContext if CA bundle isn't present

AdamWill:
Similar to e18879932287c2bf4bcee4ddf6ccb8a69b6fc656 , this also skips creation of the default SSLContext on FileNotFoundError, which is raised if DEFAULT_CA_BUNDLE_PATH does not exist.

See: https://bugzilla.redhat.com/show_bug.cgi?id=2297632

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -81,9 +81,10 @@ try:
     _preloaded_ssl_context.load_verify_locations(
         extract_zipped_paths(DEFAULT_CA_BUNDLE_PATH)
     )
-except ImportError:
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @AdamWill, #6767 is likely the path forward at this point. We've found a number of sharp edges with the default caching behavior to a point it probably doesn't belong in Requests.

AdamWill:
Ah, thanks, didn't see that one. I'll wait for it to be finalized then backport it. Thanks!

PR #6785, opened 2024-08-07
Update conf.py to fix roman year

dpapagiannis:
Very minor fix, which did however catch my eye: Fix for the invalid roman year in the documentation footer.

Same as #6389, which seems to be closed.


sigmavirus24:
We should just switch it to a arabic numeral to avoid confusion. Feel free to update this PR with that or send a new one

PR #6786, opened 2024-08-11
Add some nonstandard HTTP status code

moyanj:


sigmavirus24 (commented):
We're not expanding this and have rejected similar pull requests in the past. Please check for duplicates in the future

Issue #6790, opened 2024-08-15
Import time regression


holmanb:
Between 2.31.0 and 2.32.3, import time increased by ~25%. This appears to have been caused by https://github.com/psf/requests/pull/6667, which changed the expensive `load_verify_locations()` to run at import time (in 2.32.0).

## Reproduction Steps

Compare import time of versions with:
```bash
$ python3 -X importtime -c 'import requests' 2> requests_import.log
$ tuna requests_import.log
```

2.31.0

![image](https://github.com/user-attachments/assets/65d11607-6342-406e-8d59-4d8c934b05cb)

2.32.3:

![image](https://github.com/user-attachments/assets/55d79382-daf2-45e0-8e6c-fcc080683fc6)

donno:
I'm on Windows, with Python 3.11 and requests 2.32.3 and noticed very slow import times and narrowed it down to requests. In my case when importing requests, the `requests.adapters` module made up 4.353 s (85.9%) of the import time with similar command as above.

3CB3EC:
> I'm on Windows, with Python 3.11 and requests 2.32.3 and noticed very slow import times and narrowed it down to requests. In my case when importing requests, the `requests.adapters` module made up 4.353 s (85.9%) of the import time with similar command as above.

Same problem, Windows, with Python 3.13.1 and requests 2.32.3.
I've tried several versions of requests, and I find this problem appear since 2.32.0.

seproDev:
We noticed that this slows down [yt-dlp](https://github.com/yt-dlp/yt-dlp) startup time by ~50%.

![Image](https://github.com/user-attachments/assets/b59b5294-5de5-416c-b28b-be18f665ff7d)

yt-dlp provides its own SSLContext, so the one created by requests on import remains unused.

Is it still planned to revert the caching of a default SSLContext (#6767) or should the SSLContext no longer be created on import as proposed in #6791?

holmanb:
It sounds like this may be fixed in https://github.com/psf/requests/pull/6767, but I will leave this open until released to confirm.

nateprewitt:
That's correct, this was addressed in 2.32.5. There is no longer an upfront import time cost.

PR #6791, opened 2024-08-15
Do not load ssl context during import

holmanb:
Possible fix for https://github.com/psf/requests/issues/6790.

This removes the ssl context load from the import path, and instead caches the context the first time that it is requested.


nateprewitt:
Resolving as this is no longer relevant with #6767.

PR #6801, opened 2024-09-16
Teach users to use with-statements for Multipart-Encoded Files

papb:
Hello, everyone!!

I am proposing a small improvement to `quickstart.rst` in order to teach users to use `with`-statements with `open()` calls when POSTing Multipart-Encoded files.


sigmavirus24:
Changes such as this have been rejected in the past, for good reason. Thanks for spending the time to open this PR. In the future you might consider checking for others that have been rejected

PR #6802, opened 2024-09-18
start testing on 3.13rc

graingert:


graingert:
a remake of https://github.com/psf/requests/pull/6720 with the new pytest-httpbin

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 -e .[socks]
 pytest>=2.8.0,<9
 pytest-cov
-pytest-httpbin==2.0.0
+pytest-httpbin==2.1.0rc1
```
graingert:
```suggestion
pytest-httpbin==2.1.0
```

setup.py
```diff
@@ -40,7 +40,7 @@
     "certifi>=2017.4.17",
 ]
 test_requirements = [
-    "pytest-httpbin==2.0.0",
+    "pytest-httpbin==2.1.0rc1",
```
graingert:
```suggestion
    "pytest-httpbin==2.1.0",
```

.github/workflows/run-tests.yml
```diff
@@ -31,6 +31,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
         cache: 'pip'
+        allow-prereleases: true
```
nateprewitt, line 34: Do we still need this or can it be removed now that `pytest-httpbin` is fully released?

graingert: This is needed for 3.13rc. it doesn't effect pip installs

graingert:
I should probably update the trove classifier

nateprewitt (changes requested):
I think we're also missing an update for the [`tox`](https://github.com/psf/requests/blob/f12ccbef6d6b95564da8d22e280d28c39d53f0e9/tox.ini#L2) file.

nateprewitt (approved):

sigmavirus24 (approved):

PR #6819, opened 2024-10-29
Add !r to error messages for improved clarity on file paths

jjkochan-metro:
If leading and/or trailing spaces are included in the file path, they are clearly shown in the error messages because the file path is shown within quotes. If the file path is not shown within quotes, it may be difficult to identify that leading and/or trailing spaces are the cause of the problem.


sigmavirus24:
If you're willing to fix the tests, I think this change makes good sense to me and would be willing to merge it.

jjkochan-metro:
I think I fixed the tests. Please let me know if there is anything else I need to do, as I am very new to contributing to github projects.

PR #6822, opened 2024-11-08
Add HTTP status group properties for response objects

vengatgs:
## PR Description:

This PR adds properties (`is_2xx_successful`, `is_5xx_server_error`, etc.) to the HTTP response class for easy classification of HTTP status codes, improving code readability.

## Justification:
- **Improved readability**: Using direct range comparisons like `100 <= self.status_code < 200` makes the code clear and intuitive.
- **Maintaining performance**: The range comparison is efficient and keeps the performance trade-offs minimal while avoiding unnecessary complexity.

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -783,28 +783,28 @@ class Response:
 
     @property
     def is_1xx_informational(self):
-        """True if the status code is an informational (1xx)"""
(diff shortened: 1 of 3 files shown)


sigmavirus24:
No thank you. We're not expanding the API of Requests at this time and the freeze is well documented. Sorry you spent your energy on this

PR #6837, opened 2024-12-03
Fix handling of non-latin credentials in HTTPDigestAuth

Sklyvan:
Fixes #6102

Update `HTTPDigestAuth` to handle non-latin credentials correctly.

* Change `_basic_auth_str` in `src/requests/auth.py` to encode `username` and `password` using `utf-8` instead of `latin1`.
* Update `build_digest_header` in `src/requests/auth.py` to correctly format the string for the header with non-latin characters.
* Add tests in `tests/test_requests.py` to verify correct handling of non-latin credentials in `HTTPDigestAuth`.

### DIFF
--- a/src/requests/auth.py
+++ b/src/requests/auth.py
@@ -54,10 +54,10 @@ def _basic_auth_str(username, password):
     # -- End Removal --
 
     if isinstance(username, str):
-        username = username.encode("latin1")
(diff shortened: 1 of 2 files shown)


sigmavirus24:
I think #6772 is the correct way to fix this

PR #6863, opened 2024-12-26
issue resolve - 6653 document uploaded of exceptions from request()

enji-coder:
Detailed document uploaded for exception raise by request() 

This will help for python developers who working upon request() module

### DIFF
--- a/docs/document_exceptions.txt
+++ b/docs/document_exceptions.txt
@@ -1,0 +1,111 @@
+Document which exception request() might raise :: 
+
+Common Exceptions Raised by requests Methods
+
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @enji-coder, thanks for the PR. Unfortunately I don't think we can accept it. We do document the [top-level exception](https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions) which is what we'd generally recommend you use in your code.

If you do need to make granular decisions based on specific exceptions, they're also currently available in the [docs](https://requests.readthedocs.io/en/latest/api/#exceptions).

PR #6865, opened 2025-01-01
Update docstring for Session.verify to include string support (#6859)

### LINKED ISSUE #6859: Docstring:
`Session` docstring for `verify` attribute should note that it supports a string value.

## Actual Result

Currently, the `Session` docstring notes that its `verify` attribute only supports a bool.

However, `session.request` supports a string for its `verify` arg: https://github.com/psf/requests/blob/main/src/requests/sessions.py#L550

and when a `verify` arg isn't passed to `session.request`, then `session.verify` is used instead: 

https://github.com/psf/requests/blob/main/src/requests/sessions.py#L776
https://github.com/psf/requests/blob/main/src/requests/sessions.py#L70

## Reproduction Steps

N/A

## System Information
N/A

M0d3v1:
This pull request updates the docstring for the `Session.verify` attribute to clarify that it supports a string value for custom CA bundle paths.

Closes #6859.

### DIFF
--- a/src/requests/sessions.py
+++ b/src/requests/sessions.py
@@ -420,6 +420,8 @@ class Session(SessionRedirectMixin):
         #: presented by the server, and will ignore hostname mismatches and/or
         #: expired certificates, which will make your application vulnerable to
         #: man-in-the-middle (MitM) attacks.
+        #: If verify is set to a string, it must be the path to a CA bundle file
+        #: that will be used to verify the TLS certificate.
         #: Only set this to `False` for testing.
         self.verify = True
 


src/requests/sessions.py
```diff
@@ -420,6 +420,8 @@ def __init__(self):
         #: presented by the server, and will ignore hostname mismatches and/or
         #: expired certificates, which will make your application vulnerable to
         #: man-in-the-middle (MitM) attacks.
+        #: If verify is set to a string, it must be the path to a CA bundle file
+        #: that will be used to verify the TLS certificate.
         #: Only set this to `False` for testing.
```
nateprewitt: Let's keep the `False` warning attached to the rest of the `False` information, it may be easy to miss if you stop reading at that point.

```suggestion
        #: Only set this to `False` for testing.
        #: If verify is set to a string, it must be the path to a CA bundle file
        #: that will be used to verify the TLS certificate.
```

nateprewitt (changes requested):
Thanks for the PR, @M0d3v1! This seems reasonable to me. I've made a minor suggestion, but we should be ready to merge this.

nateprewitt (approved):

PR #6866, opened 2025-01-03
Adding host details in proxy headers

HarikrishnanTechie:
**Issue:**
The connect request does not include host details when operating in a proxy environment. This can lead to failures in establishing a proper connection, as the host information is essential for routing the request through the proxy.

**Code Explanation:**
To address this issue, the following approach is used:

Check for Host in Proxy Settings:
If the proxy_kwargs dictionary contains a host key, it indicates that host details are available.

Add Host to Proxy Headers:
These host details are added to the proxy_headers dictionary under the Host key. By doing so, the host information is explicitly included in the proxy headers, ensuring that the connect request incorporates the correct routing details.

Impact:
Including the host details in the proxy_headers ensures that the connect request is properly formed and able to route through the proxy environment without any missing information. This resolves the issue and enhances connectivity reliability

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -290,6 +290,8 @@ class HTTPAdapter(BaseAdapter):
             )
         else:
             proxy_headers = self.proxy_headers(proxy)
+            if 'host' in proxy_kwargs and proxy_kwargs['host']:
+                proxy_headers['host'] = f"{proxy_kwargs['host']}"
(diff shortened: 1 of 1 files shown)


sigmavirus24:
We should resolve https://github.com/urllib3/urllib3/pull/3532 first. Both of these PRs lack tests and the correct way to handle the header in question

PR #6879, opened 2025-02-01
docs(exceptions): Add ConnectTimeout and ReadTimeout to documentation

jakobheine:
Comparing `__init__.py` exported exceptions with the documentation revealed that two exceptions were missing from the docs. This commit updates the documentation to include these previously unlisted exceptions, ensuring consistency between the code and the docs.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -37,6 +37,8 @@ Exceptions
 .. autoexception:: requests.ReadTimeout
 .. autoexception:: requests.Timeout
 .. autoexception:: requests.JSONDecodeError
+.. autoexception:: requests.ConnectTimeout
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -37,6 +37,8 @@ Exceptions
 .. autoexception:: requests.ReadTimeout
 .. autoexception:: requests.Timeout
 .. autoexception:: requests.JSONDecodeError
+.. autoexception:: requests.ConnectTimeout
```
nateprewitt, line 40: Not sure if I'm missing something. These already exist on lines 36 and 37. Why are we adding them a second time?

jakobheine: Missed that, thank you. Closing the PR.

PR #6880, opened 2025-02-01
docs(exceptions): Remove unused exception URLRequired from documentation

jakobheine:
The documentation previously listed `requests.URLRequired` as a valid exception, suggesting it would be raised for invalid URLs. However, this exception has been dead code since commit ab27027 (2012) and is never actually raised.

Instead, invalid URLs raise `MissingSchema`, `InvalidSchema`, or `InvalidURL`, none of which were documented. This commit removes `URLRequired` from the documentation to reflect the actual behavior and prevent confusion.


nateprewitt:
Thanks, @jakobheine!

PR #6881, opened 2025-02-01
refactor(URLRequired): Deprecate URLRequired exception with warning

jakobheine:
`requests.URLRequired` has been dead code since ab27027 (2012) and is never raised. This commit adds a `DeprecationWarning` to notify users that it will be removed in a future version. Invalid URLs instead raise `MissingSchema`, `InvalidSchema`, or `InvalidURL`.


The-Compiler:
Note this only emits a warning when the exception is *raised* (which would only ever be done in `requests` code), not when it is caught.

A module-level `__getattr__` would instead emit such a warning whenever it is accessed, see e.g. the removed code in https://github.com/pytest-dev/pytest/commit/477959ef7dc9f8491a80ba4d585d737a68f0a34c for some inspiration.

jakobheine:
Thank you for the review @The-Compiler! I updated the PR accordingly, wdyt now?

The-Compiler:
This doesn't emit any warnings at all, as shown by a simple test with `python -Werror -c "from requests.exceptions import URLRequired"`.

You'll need to rename `URLRequired` to something like `_URLRequired` so that the `__getattr__` actually gets called. Most likely after that, an `import requests` will already emit a warning as *that* does `from .exceptions import URLRequired`. See the pytest example above for how to handle that (with another module-level `__getattr__` in there).

nateprewitt:
I don't think we intend to add the warning at this point. It's part of the public interface and won't be removed prior to a new major version. The impact of leaving it in place is low-to-none. Thanks for the docs PR though, we'll close this and the parent issue out.

PR #6884, opened 2025-02-03
Include invalid type name in error message

tejaschauhan36912:
* Included invalid type name in error message for better debugging.

### DIFF
--- a/src/requests/cookies.py
+++ b/src/requests/cookies.py
@@ -497,7 +497,9 @@ def morsel_to_cookie(morsel):
         try:
             expires = int(time.time() + int(morsel["max-age"]))
         except ValueError:
-            raise TypeError(f"max-age: {morsel['max-age']} must be integer")
+            raise TypeError(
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This doesn't do what you want. What problem are you encountering? Do you have a minimal reproduction test case for it?

sigmavirus24 (changes requested):
This will not do what you want

tejaschauhangit:
> This doesn't do what you want. What problem are you encountering? Do you have a minimal reproduction test case for it?

Hi, @sigmavirus24, Thanks for replying!
I am not facing any problem with this error message in my system,
My suggestion is it's better we don't expose the max-age value in the error message, instead, we can keep the  error message like ,  `
f"max-age must be integer, not {type(morsel['max-age'])}"`.

We can just include the invalid type it's receiving and the valid type it should have in the error message.

sigmavirus24:
You're not making sense and since this isn't solving a real problem, I'm closing this.

PR #6889, opened 2025-02-09
Updated the test for extract_zipped_paths to run correctly if test_utils.py is changed

ShivamBang98:


sigmavirus24:
The tmpdir fixture used by this test generates a random test directory every time it runs and cleans it up after the test is finished. I'm not sure how you're encountering the failure you describe and there's certainly not enough detail here to help debug this further.

ShivamBang98:
I'll add a bit more context. I'm aware that the fixture creates a new temp directory, my assertion here is that the issue is actually in the behaviour of the `extract_zipped_paths` function, and that the function itself was creating a file **outside of the temp dir.** I'll try my best to explain it with screenshots/console logs below:

1. So, when I first ran this test suite, it ran fine. I then changed around some of the other tests, but this one began to fail with this error:

```
AssertionError: assert False
E        +  where False = <function cmp at 0x105cbd3a0>('/var/folders/w7/s_wdvyhn17947ttg6354fs5w0000gp/T/test_utils.py', '/Users/shivambang/practice/requests/tests/test_utils.py')
E        +    where <function cmp at 0x105cbd3a0> = filecmp.cmp

tests/test_utils.py:337: AssertionError
```

2. The directory `/var/folders/w7/s_wdvyhn17947ttg6354fs5w0000gp/T/` turned out to contain my `pytest-of-shivambang` sub-directory, which is the correct expected behaviour. However, surprisingly, it also contained a file called `test_utils.py`.

Screenshot 2025-02-10 at 2 46 28 AM

3. It is my understanding that this file was created by the `extract_zipped_paths` function during the test execution. Since it's already been created once, it wasn't being overwritten on successive test runs. This ultimately caused **any** change to the `test_utils.py` file to fail this particular test case. Manually deleting the file via `rm -rf` fixed the problem, but it had to be done every time a change was made to this file.


Hope this explains it better, happy to share more details!

ShivamBang98:
Steps to reproduce this error:

1. Run the test suite with `pytest tests/test_utils.py`
2. Make some edits to the `test_utils.py` file (any edit is fine, even a comment addition)
3. Re-run the test suite.

@sigmavirus24

sigmavirus24:
I believe the correct fix here is to switch to `tmp_path` https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures since tmpdir is deprecated

ShivamBang98:
Gotcha, I'll give it a shot. `tmpdir` is being used in a few different tests as well, I'll update all of those too and let you know.

ShivamBang98:
Hey @sigmavirus24 , I've updated the tests in this file that were using `tmpdir` to now use `tmp_path`. However, my original fix containing `os.remove` is still needed, since the `extract_zipped_paths` function is extracting files out of the pytest `tmp_path` and directly into the system's temp directory. This can be seen in my attached screenshot, where `tempfile.gettempdir()` is being used to get the ultimate destination of the extracted file:

Screenshot 2025-02-10 at 2 22 57 PM

ShivamBang98:
Hi @sigmavirus24 , did you get time to take a look at this by any chance?

Issue #6896, opened 2025-02-16
Regenerating tests/certs/ with OpenSSL >= 3.2.0 causes test failures


cjwatson:
Regenerating `tests/certs/` using the provided `Makefile`s causes test failures.

## Expected Result

https://bugs.debian.org/1091503 reports that requests 2.32.3 as packaged in Debian fails tests if run on a system whose time has been artificially set to 2028.  It occurred to me that a simple time-trap-free approach to this might be to change the Debian packaging to regenerate the various certificates in `tests/certs/` using the provided `Makefile`s immediately before running the tests, which seems as though it ought to work.

## Actual Result

`tests/test_requests.py::TestPreparingURLs::test_different_connection_pool_for_tls_settings_verify_bundle_unexpired_cert` and `tests/test_requests.py::TestPreparingURLs::test_different_connection_pool_for_mtls_settings` fail with errors indicating an invalid CA certificate.

## Reproduction Steps

```console
$ for cert in expired mtls valid/server; do make -C tests/certs/$cert clean all; done
$ tox -e py313-default -- -k 'test_different_connection_pool_for_tls_settings_verify_bundle_unexpired_cert or test_different_connection_pool_for_mtls_settings'
.pkg: _optional_hooks> python /home/cjwatson/.local/pipx/venvs/tox/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_sdist> python /home/cjwatson/.local/pipx/venvs/tox/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_wheel> python /home/cjwatson/.local/pipx/venvs/tox/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: prepare_metadata_for_build_wheel> python /home/cjwatson/.local/pipx/venvs/tox/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: build_sdist> python /home/cjwatson/.local/pipx/venvs/tox/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
py313-default: install_package> python -I -m pip install --force-reinstall --no-deps /home/cjwatson/src/python/requests/.tox/.tmp/package/7/requests-2.32.3.tar.gz
[...]

PR #6897, opened 2025-02-17
Add CA constraint to test CA

cjwatson:
Otherwise recent versions of OpenSSL reject it as an invalid CA certificate.

Fixes: #6896


cjwatson:
Let me know if you'd rather I dropped the "Regenerate test certificates" commit - but I thought it would be less confusing if the test certificates were consistent with the code that generates them.

sigmavirus24 (approved):

cjwatson:
The pypy-3.9 Windows failure seems unrelated to this change.

PR #6899, opened 2025-02-19
Docs: Add quotes to prevent Zsh wildcard interpretation.

duzhuoshanwai:
```
zsh: no matches found: requests[socks]
```


nateprewitt (approved):
I think this varies based on operating system, Windows will require `"` but since most of our documentation is written assuming a bash based environment, this is probably fine. Thanks @duzhuoshanwai!

PR #6907, opened 2025-02-24
Wrapped _preloaded_ssl_context with function to speed up package import

gregory-shklover:
Wrapped around default SSL context initialization to potentially speed up "import requests" time.

Measurements:

# original
$ time python -c "import requests"
real    0m0.936s
user    0m0.000s
sys     0m0.047s

# modified:
$ time python -c "import requests"
real    0m0.351s
user    0m0.000s
sys     0m0.047s


nateprewitt:
Resolving as this is no longer relevant with #6767.

PR #6909, opened 2025-03-01
reduced cyclomatic complexity in model.py/RequestEncodingMixin._encode_files method

Doughnutik:
I decided to make a research and investigate if there is a high cyclomatic complexity in requests library. I found model.py/RequestEncodingMixin._encode_files method with D(21) value which is not good.
There are parts in code which we can organize in separate functions. After such operations I got new value A(5).
It can be useful because low cyclomatic complexity means that it's easy to test the code and predict its behaviour!

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -148,58 +148,63 @@ class RequestEncodingMixin:
         elif isinstance(data, basestring):
             raise ValueError("Data must not be a string.")
 
-        new_fields = []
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Depending on how you measure cyclomatic complexity, creating functions which are closures actually increases it. Additionally, cognitive complexity is dramatically worse with this change.

Thanks for attempting to help, but we will not be accepting this change

PR #6913, opened 2025-03-17
Minor quirk: more straightforward __iter__ impl. for CaseInsensitiveDict

belliriccardo:
With the actual `CaseInsensitiveDict.__iter__` method, both the key and value are extracted but only the key is used. With this minor change, we would loop directly over the internal dict's keys (which is the default for the dict's `__iter__`). The difference is minimal, just 4 less bytecode instructions: [https://godbolt.org/z/saT7Me3vb](https://godbolt.org/z/saT7Me3vb) (uncomment either `__iter__` implementation).

### DIFF
--- a/src/requests/structures.py
+++ b/src/requests/structures.py
@@ -55,7 +55,7 @@ class CaseInsensitiveDict(MutableMapping):
         del self._store[key.lower()]
 
     def __iter__(self):
-        return (casedkey for casedkey, mappedvalue in self._store.values())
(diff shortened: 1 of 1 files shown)


sigmavirus24:
We are not accepting refactorings, optimizations, or stylistic changes at this time. For more details as to why, please search past closed PRs

belliriccardo:
Oh ok, got it; I knew this could be the outcome since it's such a small, opinionated change. Thanks for your time anyways!

PR #3552, opened 2016-09-05
pysocks 1.5.7 blacklisting, due to IPv6 problems

Trodis:
https://github.com/kennethreitz/requests/issues/3551

PR #6916, opened 2025-03-19
Use a more intuitive PySocks version

bebound:
I accidentally found this in the pip log: `Collecting PySocks!=1.5.7,>=1.5.6 (from requests[socks]`

This seems odd; I'm not sure if it's worth changing to a more intuitive version **at the cost of removing 1.5.6 from compatible version list**.

`!=1.5.7` was introduced in https://github.com/psf/requests/pull/3552
pysocks release history: https://pypi.org/project/PySocks/#history

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ test_requirements = [
     "pytest-cov",
     "pytest-mock",
     "pytest-xdist",
-    "PySocks>=1.5.6, !=1.5.7",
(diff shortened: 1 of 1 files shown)


sigmavirus24:
I don't understand the expected value of this. It's very common for a dependency list to have a lower bound with exclusions for known bad versions.

Issue #6917, opened 2025-03-21
Incorrect Content-Length header with StringIO body


mgraczyk:
When requests is used with an `io.StringIO` as the `data` type, and the body contains characters whose utf-8 encoding is multiple bytes, the Content-Length header is set incorrectly.

## Expected Result

Content-Length should match the number of bytes sent when using io.StringIO

## Actual Result

Content-Length is the length of the string, not the bytes sent.

Looking at the [implementation of `super_len`](https://github.com/psf/requests/blob/1764cc938efc3cc9720188dfa6c3852c45211aa0/src/requests/utils.py#L189-L199), it appears that `io.StringIO` has its length measured using `seek` and `tell`.
It has been implemented that way since June 2016 (af7729f64a97ab35e83a1a7971781e69d124d99e).

It looks like this was fixed for `str` inputs in #6586 in 2023 but was never fixed for io.StringIO

I am happy to send a PR if the implementation is straightforward.
Off the top of my head I don't know how to count the bytes in a utf-8 encoded StringIO without copying, and previous PRs have tried to avoid a copy in `super_len`

## Reproduction Steps

Run the following script, which shows the problem in detail

```python
import io
import requests
from urllib3.connection import HTTPConnection
from requests.utils import super_len
from requests.models import PreparedRequest

# A string that is 1 character but 4 bytes in UTF-8.
# Requests will always send 4 bytes but the Content-Length header depends on the type passed as `data`.
value = "💩"

body_types = [{
  "name": "str",
  "value": value,
}, {
  "name": "bytes",
  "value": value.encode("utf-8"),
}, {
  "name": "io.BytesIO",
  "value": io.BytesIO(value.encode("utf-8")),
},
{
  "name": "io.StringIO",
  "value": io.StringIO(value),
}]

print("## Super Len")
for body_type in body_types:
  print("Body Type:", body_type["name"])
  print("Super Len:", super_len(body_type["value"]))
  p = PreparedRequest()
  p.prepare(
      method="POST",
      url="http://example.com",
      data=body_type["value"],
  )
  print("Prepared Headers:", p.headers)

# Monkey patch to print the data sent.
old_send = HTTPConnection.send
def new_send(self, data):
  print("Sending Data:", data)
  old_send(self, data)
HTTPConnection.send = new_send


print("## Requests")
for body_type in body_types:
  r = requests.post(
    "http://example.com",
    data=body_type["value"],
  )
```

Here is my output:
```
## Super Len
Body Type: str
[...]

PR #6918, opened 2025-03-22
Fix incorrect Content-Length for StringIO with multi-byte characters

hackspaces:
This PR fixes issue #6917 where `io.StringIO` objects containing multi-byte characters report incorrect Content-Length headers.

## The issue
When using `io.StringIO` objects with multi-byte characters (like emoji), the Content-Length header is incorrectly set to the character count instead of the byte count. This happens because `StringIO.tell()` reports character positions, not byte positions.

For example, the emoji "💩" is 1 character but 4 bytes in UTF-8. With the current implementation, the Content-Length header incorrectly shows 1 instead of 4.

## The fix
The solution ensures that `io.StringIO` objects correctly report their UTF-8 encoded byte length by:
1. Detecting when the object is an `io.StringIO` instance
2. Reading the entire content and calculating its UTF-8 encoded length
3. Restoring the original position to maintain the object's state

I've added a test case that reproduces the issue and verifies the fix works with emoji characters.

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -195,6 +195,18 @@ def super_len(o):
                     # seek back to current position to support
                     # partially read file-like objects
                     o.seek(current_position or 0)
+
+                    # Handle StringIO with multi-byte characters by correctly measuring UTF-8 byte length
+                    if isinstance(o, io.StringIO):
+                        # Save current position
+                        current_pos = o.tell()
+                        # Get entire content and calculate byte length after UTF-8 encoding
+                        o.seek(0)
+                        content = o.read()
+                        # Restore original position
+                        o.seek(current_pos)
+                        # Set total_length to the byte length, not character count
+                        total_length = len(content.encode("utf-8"))
                 except OSError:
                     total_length = 0
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,5 +1,6 @@
 import copy
 import filecmp
+import io
 import os
 import tarfile
 import zipfile
@@ -150,6 +151,13 @@ class TestSuperLen:
     def test_super_len_with_no_matches(self):
         """Ensure that objects without any length methods default to 0"""
         assert super_len(object()) == 0
+
+    def test_super_len_with_stringio_containing_multibyte_chars(self):
+        """Ensure StringIO with multi-byte characters reports correct byte length"""
+        # emoji takes 4 bytes in UTF-8 but is 1 character
+        s = io.StringIO("💩")
+        # Super len should return 4 (byte length) not 1 (character length)
+        assert super_len(s) == 4
 
 
 class TestToKeyValList:


sigmavirus24:
Now run benchmarks showing how this works with a StringIO with 4GiB of data in it. People send that much data through an IO object to requests at times and expect it to not create an additional 4GiB of memory while processing the request

mgraczyk:
@sigmavirus24 The proposed changes will not increase memory usage for the situation you described. I just tested it. The high water mark remains the same because `requests` already copies the data. If it did, the memory usage can be mitigated by processing in a loop.

Also, if the user tries to send a 4GiB StringIO that contains characters outside latin-1, the request will likely fail because the Content-Length will be wrong.

For example there is currently a bug in Google's cloud storage library and a workaround in gcloud-aio-storage for this

EDIT: After further testing, this can increase memory usage depending on the size of the request, but I believe this can be mitigated by looping over the StringIO instead of reading it all at once.

hackspaces:
Hey Hey, Thank you so much for looking into this - I was thinking the same if we have the issue of memory - do you think adding a chunking strategy where we set a 1MB chunk(or any arbitrary size) and loop over to read the chunks? Or would that be an overkill in terms of computation?

sigmavirus24:
The fundamental problem folks aren't understanding about this is that you're fixing yet one more _tiny_ part of this problem in ways that are inefficient for the end user.

The correct answer is to _always_ use bytes. Hacks on top of performance-degrading hacks are not the right path here.

We fixed this for `str`. If we did something here, we'd fix it for `StringIO`. Okay. What then about file objects opened like `open(path, "r")`? Those are whatever the filesystem wants - same problem there. Do we loop over a potentially giant file then to determine if there's any chance of there being a multi-byte character? What about someone who passes in `sys.stdin`? We can't consume that without caching the _entire_ thing in memory.

Y'all aren't understanding that this isn't sustainable to keep applying bandaids to a large wound. The library was fundamentally mis-designed a decade or more ago and it's presents to you as a series of tiny paper cuts, but in reality it's a series of large wounds that need stitches and the only way to do something that holistic is to create a backwards incompatible version of it which will create a significant amount of pain for folks like yourself.

hackspaces:
Hey Hey, Thank you so much once again for looking into this and porviding a detailed comment. I do agree - this would not fix the long term strategy that you have put out - thinking from the bigger picture does help here. I appriciate your time on this. I will close this PR.

mgraczyk:
@sigmavirus24 I disagree with the characterization and I think ignoring this problem is a huge mistake for requests.

Other libraries have solved this by throwing when the user passes StringIO. You should do the same if you intend to continue sending incorrect Content-Length headers.

But there is no reason not to fix this. This is a tiny, localized fix that solves the problem with no coupling or interaction. It will not increase maintenance complexity since it only specifically affects StringIO.

As for you point about text mode files, that works! So does stdin! StringIO is unique here, in practice it is the only input type affected.

The reason StringIO is special is that libraries of requests often wrap their str in a StringIO, and those libraries are all currently broken

In your response, you are saying that it's better for requests to be fast than correct. Given the widespread usage of the requests library, that should scare anyone who interacts with the python ecosystem.

sigmavirus24:
@mgraczyk don't put words in my mouth also you seem to consistently miss the point I make. This _only_ fixes StringIO. The only reasonable thing is to fix every file-like object that may be text including stdin. Neither files opened as text today nor stdin are protected from this very same bug. That's my point. 

Just fixing StringIO by using double the memory people expect or by looping and adding another 100ms to minutes of time depending on data size is also surprising. Whether we fix this or not, someone will be surprised. 

It's not performance over correctness as _anyone_ who cares for correctness will just send us bytes.

mgraczyk:
> Neither files opened as text today nor stdin are protected from this very same bug

This is not true, because both text files and stdin are sent using `Transfer-Encoding: chunked`, so the bug does not affect those.

> Just fixing StringIO by .. adding another 100ms to minutes of time depending on data size is also surprising.

This is unequivocally a statement that performance is more important than correctness.

I strongly recommend throwing on `StringIO` inputs or at least fixing the documentation to remove the incorrect claim that "Requests may attempt to provide the Content-Length header for you, and if it does this value will be set to the number of bytes in the file."

---- 

For anyone else reading this by searching "requests StringIO", "requests Content-Length", "cloud storage 502 Bad Gateway" the takeaway is that you should not use `StringIO` with requests and should double check that no libraries you are using do that indirectly.

sigmavirus24:
> This is not true, because both text files and stdin are sent using Transfer-Encoding: chunked, so the bug does not affect those.

You are assuming one very specific use case. Any time we can tell the size of a file we send it with the file size we can determine in the same method of tell, seek, and then tell. 

There's little point continuing this discussion though, you seem intent on a deliberate and bad faith reading of my comments.

mgraczyk:
I am reading carefully and specifically addressing the things you are saying.

You said "files opened as text today are [not] protected from this very same bug"

This is false, files opened as text today are protected from this bug.
Here is a complete repro demonstrating that this claim you just made is not correct:

```
import requests
from urllib3.connection import HTTPConnection
from requests.utils import super_len
from requests.models import PreparedRequest

value = "💩" * 1

# Monkey patch to print the data sent.
old_send = HTTPConnection.send
def new_send(self, data):
  print("Sending Data:", data[:500])
  old_send(self, data)
HTTPConnection.send = new_send

with open("test.txt", "w") as f:
  f.write(value)

with open("test.txt", "r") as f:
  f.seek(0)

  print("Body Type:", "file in text-mode")
  print("Super Len:", super_len(f))
  p = PreparedRequest()
  p.prepare(
      method="POST",
      url="http://example.com",
      data=f,
  )
  print("Prepared Headers:", p.headers)

  f.seek(0)
  r = requests.post(
    "http://example.com",
    data=f,
    stream=False,
  )
  ```
  
  Output:
  ```
Body Type: file in text-mode
Super Len: 4
Prepared Headers: {'Content-Length': '4'}
Sending Data: b'POST / HTTP/1.1\r\nHost: example.com\r\nUser-Agent: python-requests/2.32.3\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\nContent-Length: 4\r\n\r\n'
Sending Data: b'\xf0\x9f\x92\xa9'
```


You next claimed "Any time we can tell the size of a file we send it with the file size we can determine in the same method of tell, seek, and then tell."

This is also false, for text files, you use stat. That is why this works.

It's surprising that requests works for text files and strings, but does not work for StringIO.

I understand your concern that fixing this will use too much memory or time.
IMO it would be much better to add some latency or memory to fix the correctness issue.
It's fine that you disagree, and we can move on.

However I will keep correcting you when you say things that are not true, and telling me I'm responding in "bad faith" by posting code snippets that contradict you does not bother me and won't stop me.

PR #6919, opened 2025-03-22
Improvements for _parse_content_type_header making it more readable a…

perodriguezl:
…nd more robust

## Why this is better at scale:
Avoids full split if no semicolon (fast path).
Does only one split on =, minimizes string churn.
Avoids multiple .strip() calls — chains them efficiently.
Fewer allocations and GC impact.

## Benchmarked using

```
import timeit

# Original version of the function
def parse_content_type_original(header):
    tokens = header.split(";")
    content_type, params = tokens[0].strip(), tokens[1:]
    params_dict = {}
    items_to_strip = "\"' "
    for param in params:
        param = param.strip()
        if param:
            key, value = param, True
            index_of_equals = param.find("=")
            if index_of_equals != -1:
                key = param[:index_of_equals].strip(items_to_strip)
                value = param[index_of_equals + 1:].strip(items_to_strip)
            params_dict[key.lower()] = value
    return content_type, params_dict
# Optimized version
def parse_content_type_optimized(header):
    semicolon_index = header.find(";")
    if semicolon_index == -1:
        return header.strip(), {}
    content_type = header[:semicolon_index].strip()
    params_dict = {}
    rest = header[semicolon_index + 1:]
    for param in rest.split(";"):
        param = param.strip()
        if not param:
            continue
        if "=" in param:
            key, value = param.split("=", 1)
            key = key.strip("\"' ").lower()
            value = value.strip("\"' ")
        else:
            key, value = param.strip("\"' ").lower(), True
        params_dict[key] = value
    return content_type, params_dict
# Sample header for testing
header_sample = "application/json; charset=UTF-8; boundary='abc'; format=pretty"
# Benchmark setup
setup_code = "from __main__ import parse_content_type_original, parse_content_type_optimized, header_sample"
# Run benchmarks
original_time = timeit.timeit("parse_content_type_original(header_sample)", setup=setup_code, number=1_000_000)
optimized_time = timeit.timeit("parse_content_type_optimized(header_sample)", setup=setup_code, number=1_000_000)

print(original_time, optimized_time)
```
Output: 
```
Current                   NEW
1.9832534999586642 1.8716010999633
```

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -514,28 +514,33 @@ def get_encodings_from_content(content):
     )
 
 
-def _parse_content_type_header(header):
-    """Returns content type and parameters from given header
+def _parse_content_type_header(header: str):
+    """Returns content type and parameters from given header.
 
-    :param header: string
-    :return: tuple containing content type and dictionary of
-         parameters
+    :param header: Header string
+    :return: Tuple of content type and parameter dictionary
     """
+    if ";" not in header:
+        return header.strip(), {}
 
-    tokens = header.split(";")
-    content_type, params = tokens[0].strip(), tokens[1:]
+    parts = header.split(";")
+    content_type = parts[0].strip()
     params_dict = {}
-    items_to_strip = "\"' "
 
-    for param in params:
+    for param in parts[1:]:
         param = param.strip()
-        if param:
-            key, value = param, True
-            index_of_equals = param.find("=")
-            if index_of_equals != -1:
-                key = param[:index_of_equals].strip(items_to_strip)
-                value = param[index_of_equals + 1 :].strip(items_to_strip)
-            params_dict[key.lower()] = value
+        if not param:
+            continue
+
+        if "=" in param:
+            key, value = param.split("=", 1)
+            key = key.strip().strip("\"' ")
+            value = value.strip().strip("\"' ")
+        else:
+            key, value = param.strip().strip("\"' "), True
+
+        params_dict[key.lower()] = value
+
     return content_type, params_dict
 
 


nateprewitt:
Hi @perodriguezl,

Thanks for the PR! Unfortunately, we generally won't take stylistic changes for the library. The performance differential from your metrics is ~120 nanoseconds/request which isn't really perceptible on the scale of the request/response lifecycle in Requests and is only realized in specific response scenarios. I think we'll pass on this contribution for now.

perodriguezl:
> Hi @perodriguezl,
> 
> Thanks for the PR! Unfortunately, we generally won't take stylistic changes for the library. The performance differential from your metrics is ~120 nanoseconds/request which isn't really perceptible on the scale of the request/response lifecycle in Requests and is only realized in specific response scenarios. I think we'll pass on this contribution for now.

Thanks for the reply, I actually have another similar contribution coming up in another part of the project but way more impactful, the short story behinf this is: for processing millions of requests (anything over 45) in batch you might get a timeout on some processes while reviewing headers. This is specially meaningful for energy sectors that rely in very basic architecture and were big refactors are not possible due to the high regulated sector policies.

nateprewitt:
I'm not sure I'm following from what's been provided. We're discussing ~120ms per million requests if they're done serially. That's ~70 hours of request time if we're assuming 250ms/request. That's a savings of a fraction of a second for ~3 days of requests. If we're parallelizing them, we're saving fractions of a millisecond off a 250ms request.

We're unlikely to do any significant overhauls of the library to fine-tune performance unless it's broadly applicable and 100% backwards compatible. We generally ask contributors to start with an issue outlining the defect they're seeing in Requests so we can all agree on next steps before reviewing PRs. That helps minimize churn and wasted effort for both parties.

PR #6920, opened 2025-03-24
Fix invalid setuptools config in `setup.cfg`

abravalheri:
In version 78 of setuptools, the invalid configuration in `setup.cfg` will start to fail(there has been a deprecation warning since 3 Mar 2021).

I also noticed that most of the information provided to `setup.cfg` is repeated in `setup.py`, so I removed the duplicated copies.

Alternatively, I can also try to migrate the static configs to `pyproject.toml` if desired.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,9 +1,8 @@
 [metadata]
-license_file = LICENSE
-provides-extra =
-    socks
-    use_chardet_on_py3
-requires-dist =
+license_files = LICENSE
+
+[options]
+install_requires =
     certifi>=2017.4.17
     charset_normalizer>=2,<4
     idna>=2.5,<4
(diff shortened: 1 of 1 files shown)


fogti:
It feels like removing stuff from `setup.cfg` instead of from `setup.py` (which could just parse the `setup.cfg` and get that information from there) is a mistake, simply because it removes information from simple declarative config and makes it only available via imperative config.

abravalheri:
> Alternatively, I can also try to migrate the static configs to `pyproject.toml` if desired (please let me know).

Fair enough, as I mentioned before, if the maintainers are interested:

> Alternatively, I can also try to migrate the static configs to pyproject.toml if desired (please let me know).

In my opinion that would be best case scenario, but I don't know if there is a preference for centralising all the configs in `setup.py` (as it seems to be right now).

nateprewitt:
I don't think this is needed anymore since we moved to pyproject.toml.

PR #6924, opened 2025-03-29
Add key usage extension to test ca.crt

RichieB2B:
When setting `VERIFY_X509_STRICT` in urllib3 (see https://github.com/urllib3/urllib3/pull/3577) the requests tests fail because the key usage extension is missing in ca.crt:
```
% cd tests/certs/valid/server
% openssl verify -CAfile ../ca/ca.crt -x509_strict server.pem
C=US, O=Python Software Foundation, OU=python-requests, CN=Self-Signed Root CA
error 92 at 1 depth lookup: CA cert does not include key usage extension
error server.pem: verification failed
```
This PR fixes the issue.


sigmavirus24:
Pypy 3.9 is long dead, can we first merge a separate pr to drop that and bump to pypy 3.11 instead? I think 3.10 is slated for deprecation soon so no use supporting it

sigmavirus24 (approved):

PR #6926, opened 2025-03-31
Drop pypy 3.9 and add pypy 3.11 support

nateprewitt:
This PR drops support for Pypy 3.9 and adds it for Pypy 3.11 on Linux and macOS. The Windows test runner for Pypy 3.11 is failing because it's unable to install Openssl-sys with the rust bindings for Cryptography. It looks like this may have been an oversight according to https://github.com/pyca/cryptography/issues/12592 and will be released in an upcoming release.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,8 +12,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


sigmavirus24 (approved):

pamelafox:
@nateprewitt I'm confused since your PR title says that this change drops 3.9 support, but I still see 3.9 in the matrix list?

pamelafox:
Ooo, "pypy" is different from python. That took me a while to realize that it said "pypy". Nevermind, thanks!

PR #6936, opened 2025-04-23
test: add two more tests exercising the adapter

allrob23:
This PR adds two test cases for `Session.mount()` to cover the documentation's recommendation about trailing slashes in prefixes (see #6935). 

The first test `test_session_get_adapter_prefix_with_trailing_slash` verifies that prefixes with a trailing / (e.g., https://example.com/) match only the intended hostname. 

The second test `test_session_get_adapter_prefix_without_trailing_slash` verifies that prefixes without a trailing / (e.g., https://example.com) match both the intended hostname and extended hostnames (e.g., https://example.com.other.com), as warned in the docs. 

Together, these tests ensure the longest prefix match behavior is well-documented and stable.

PR #6941, opened 2025-05-03
tests: add trailing-slash mount behavior tests (#6935)

### LINKED ISSUE #6935: tests: add trailing slashes to mount to match docs recommendation?
I was digging into the docs and noticed this:
> The adapter will be chosen based on a longest prefix match. Be mindful prefixes such as http://localhost will also match http://localhost.other.com or http://localhost@other.com. It’s recommended to terminate full hostnames with a /.

[link](https://requests.readthedocs.io/en/latest/user/advanced/)

While checking out some tests that uses the `Session.mount()`, I saw that a few don’t follow this recommendation. For example, in test_session_get_adapter_prefix_matching  (https://github.com/psf/requests/blob/main/tests/test_requests.py#L1620):

```python
prefix = "https://example.com"  # no trailing slash
more_specific_prefix = prefix + "/some/path"  # no trailing slash
...
s.mount(prefix, prefix_adapter)
s.mount(more_specific_prefix, more_specific_prefix_adapter)
```

I know that the tests work great and do their job, but adding trailing slashes (e.g., https://example.com/ and https://example.com/some/path/) would align them with the docs and make the prefixes more precise.

Here are the tests I noticed:
- test_transport_adapter_ordering
- test_session_get_adapter_prefix_matching
- test_session_get_adapter_prefix_matching_mixed_case
- test_session_get_adapter_prefix_matching_is_case_insensitive

Would it be worth opening a PR to update these to include trailing slashes? The tests would stay the same, just following the docs best practice.

internza:
Closes #6935 

**What** Added two new pytest functions to cover: 
- mounting with a trailing slash matches only that exact prefix 
- mounting without a trailing slash still yields the adapter but warns in docs 

**Why** 
Ensures documented “good first issue” behavior is tested, prevents regressions in URL-prefix matching logic.

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -2981,3 +2981,15 @@ def test_json_decode_errors_are_serializable_deserializable():
     )
     deserialized_error = pickle.loads(pickle.dumps(json_decode_error))
     assert repr(json_decode_error) == repr(deserialized_error)
+
+def test_mount_with_trailing_slash_matches_exact_prefix():
+    s = requests.Session()
+    adapter = HTTPAdapter()
+    s.mount("https://example.com/", adapter)
+    assert s.get_adapter("https://example.com/foo").__class__ is HTTPAdapter
+
+def test_mount_without_trailing_slash_shows_warning_behavior():
+    s = requests.Session()
+    adapter = HTTPAdapter()
+    s.mount("https://example.com", adapter)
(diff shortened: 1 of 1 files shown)


sigmavirus24:
#6936 already exists to solve that issue. Thanks for sending this but we'll prioritize that one instead

PR #6953, opened 2025-05-22
fix(headers): content-length shouldnt be set to 0 for all body-less methods

swheaton:
Currently we use `self.method not in ("GET", "HEAD")` to determine "methods that can have a body but don't provide one"

However, there are other [HTTP methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods) that cannot have a body.
Servers can consider it an error if Content-Length=0 when no data is expected. For example, the [Sanic server's position](https://github.com/sanic-org/sanic/issues/2915#issuecomment-2029785788).

MUST have body: PATCH, POST, PUT
MUST NOT have body: GET, HEAD, [TRACE](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.8-4)
SHOULD NOT have body (no official use): [CONNECT](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.6-13), [DELETE](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.5-6), [OPTIONS](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.7-6)

Proposing that Content-Length be set to 0 when there's no data, only for PATCH, POST, PUT methods since those must have content. DELETE should not have content although it seems most common out of the "SHOULD NOT" category so we can grandfather it in.

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -578,12 +578,11 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
                 # to Transfer-Encoding: chunked.
                 self.headers["Content-Length"] = builtin_str(length)
         elif (
-            self.method not in ("CONNECT", "GET", "HEAD", "OPTIONS", "TRACE")
+            self.method in ("DELETE", "PATCH", "POST", "PUT")
             and self.headers.get("Content-Length") is None
         ):
             # Set Content-Length to 0 for methods that can have a body
-            # but don't provide one. (i.e. not CONNECT, GET, HEAD, OPTIONS
-            # or TRACE)
+            # but don't provide one. (i.e. DELETE, PATCH, POST, PUT)
             self.headers["Content-Length"] = "0"
 
     def prepare_auth(self, auth, url=""):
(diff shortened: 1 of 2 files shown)


sigmavirus24:
The motto of the internet is to be liberal in what you accept and conservative in what you send. You're also misinterpreting sanic's seeming position there and reading more into it than what they've said.

They're also reading too much into SHOULD which has very specific meaning in RFCs which is not how they're weaponizing it against their users.

swheaton:
@sigmavirus24 

> The motto of the internet is to be liberal in what you accept and conservative in what you send.

Sure so isn't my proposal being more conservative in what we send? Namely, not including Content-Length=0 unless we know it makes sense.

> You're also misinterpreting sanic's seeming position there and reading more into it than what they've said.

Nah I do not think I am

> They're also reading too much into SHOULD which has very specific meaning in RFCs which is not how they're weaponizing it against their users.

That is probably true - to your statement above, it should be more accepting. However this PR is about `requests` not `sanic`, that was just an example. And in this case not adding a header that the RFC says we [should not send](https://www.rfc-editor.org/rfc/rfc9110.html#field.content-length) is more conservative not less. Yes I do understand the difference between SHOULD and MUST. It means we can deviate for good reason, which would be that we know the server can handle or expects it - but only the caller knows that answer, this library can't infer that for them.

I guess I am not hearing a good reason why OPTIONS should not be treated the same as GET in this case, both of which should not have content.

PR #6961, opened 2025-06-03
Fix type in README

Kanaesuki:


nateprewitt:
Hi @Kanaesuki, I think you may have made your PR mistakenly against the main repo instead of your own. I'll resolve this since we won't be able to accept this change.

PR #6962, opened 2025-06-04
test: Add new test to check netrc auth leak

danigm:
This patch uses the "hostname" attribute from the parsed url to get the host, instead of trying to calculate the host from netloc that can produce errors when "http://username:password@domain.com" format is used.

This should fix the security issue reported here: [CVE-2024-47081: Netrc credential leak in PSF requests library](
https://seclists.org/oss-sec/2025/q2/204)

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -7,6 +7,7 @@ import json
 import os
 import pickle
 import re
+import tempfile
 import threading
 import warnings
 from unittest import mock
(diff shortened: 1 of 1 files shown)


sethmlarson:
@danigm Thanks! The new test is failing on Windows, can you take a peek?

sigmavirus24:
Can you please rebase as well?

danigm:
I've fixed the windows failure and rebased, so now this patch just adds a new test for netrc because the code fix was added in a different pr: https://github.com/psf/requests/pull/6965

Daviey:
An example of this vulnerability is available with this intentionally exploitable site, as documented here, https://www.openwall.com/lists/oss-security/2025/06/03/11

sigmavirus24 (approved):

PR #6963, opened 2025-06-04
fix CVE 2024 47081: manual url parsing leads to netloc credentials leak

awoimbee:
https://seclists.org/fulldisclosure/2025/Jun/2

Honestly I have no idea why this lib used `netloc` + manual parsing instead of `hostname` as I can see references to `hostname` as early as from [the python 2.6 docs](https://docs.python.org/2.6/library/urlparse.html).

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -219,14 +219,7 @@ def get_netrc_auth(url, raise_errors=False):
         netrc_path = None
 
         for f in netrc_locations:
-            try:
-                loc = os.path.expanduser(f)
-            except KeyError:
-                # os.path.expanduser can fail when $HOME is undefined and
-                # getpwuid fails. See https://bugs.python.org/issue20164 &
-                # https://github.com/psf/requests/issues/1846
-                return
-
+            loc = os.path.expanduser(f)
             if os.path.exists(loc):
                 netrc_path = loc
                 break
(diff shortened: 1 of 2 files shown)


src/requests/utils.py
```diff
@@ -236,16 +236,8 @@ def get_netrc_auth(url, raise_errors=False):
...
-        # Strip port numbers from netloc. This weird `if...encode`` dance is
-        # used for Python 3.2, which doesn't support unicode literals.
-        splitstr = b":"
-        if isinstance(url, str):
-            splitstr = splitstr.decode("ascii")
-        host = ri.netloc.split(splitstr)[0]
-
         try:
-            _netrc = netrc(netrc_path).authenticators(host)
+            _netrc = netrc(netrc_path).authenticators(ri.hostname)
```
danigm: This is fixed now in main: https://github.com/psf/requests/commit/96ba401c1296ab1dda74a2365ef36d88f7d144ef

danigm (commented):
About the test, I created another test for the same issue in a [different PR](https://github.com/psf/requests/pull/6962/). So at this point, the issue is fixed and it could be nice to have a test for this case, I'm okay to close my PR if this test is good enough, or keep it if maintainers are okay to have two tests for this case.

sigmavirus24:
This is another way to test this that is less end to end but I'm open to accepting it here as well

awoimbee:
Well OK, there's not much left after merging `main`.
Keeping this open since it contains an actual unit test and removes legacy cruft but feel free to close, I don't care

sigmavirus24 (approved):

PR #6968, opened 2025-06-10
Add Trusted Publishing Release Workflow

sigmavirus24:
Rather than rely on manual releases from a developer laptop, let's use tag pushes to trigger a workflow to publish artifacts to PyPI. This will leverage trusted publishing and upload attestations as well.


nateprewitt (approved):
:shipit:

PR #6971, opened 2025-06-16
Removes simplejson

kingbuzzman:
I don't see a need to keep `simplejson` around, when `json` is builtin to python since 2.6.

### DIFF
--- a/src/requests/compat.py
+++ b/src/requests/compat.py
@@ -54,19 +54,7 @@ is_py2 = _ver[0] == 2
 #: Python 3.x?
 is_py3 = _ver[0] == 3
 
-# json/simplejson module import resolution
-has_simplejson = False
-try:
-    import simplejson as json
-
-    has_simplejson = True
-except ImportError:
-    import json
-
-if has_simplejson:
-    from simplejson import JSONDecodeError
-else:
-    from json import JSONDecodeError
+import json
 
 # Keep OrderedDict for backwards compatibility.
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


kingbuzzman:
[While I wait to get authorized](https://github.com/kingbuzzman/requests/pull/2), CI ran here. All passed.

nateprewitt:
Hi @kingbuzzman, this is also a breaking change. We won't be able to accept this PR.

kingbuzzman:
Breaking what? All the tests pass. I think we have a different definition of "breaking change"

nateprewitt:
@kingbuzzman this is a fun part of open source. I think most individuals would accept if your code stops functioning in the way it did previously with no change on your part, it is potentially breaking. In this specific case, there are fundamental performance and behavioral differences between simplejson and json. We cannot shut off those behaviors unannounced, and if we did regress them it would need serious consideration.

kingbuzzman:
This tracks. Is there a roadmap where this ceases to exist?

nateprewitt:
There is not as we don't have any current intentions to remove simplejson support. It allows users to dynamically opt into simplejson's more performant C-extension and more extensible SerDe APIs. Removing this woule be a net-negative for the library.

kingbuzzman:
All that you've said makes sense.

The part that bothers me about all this: there is not a single tests that ensures anything you've said. Would you be open for me creating said tests?

* Test simplejson
* Test for is_py2/is_py3

PR #6986, opened 2025-07-08
Print post

SeanHgh:
Par rapport à la branche principale, on a rajouté des spécifications ainsi qu'un affichage via print des données poussées sur le serveur via POST.


sigmavirus24:
This seems to be a pull request issued against the wrong repository. If you did in fact intend to send this here, we will not be accepting it

SeanHgh:
Sorry, indeed that was an accident. I am trying to understand how GitHub
works.
All my apologies.
I hope I did not bother you too much.
Sorry again and have a good day.

PR #6987, opened 2025-07-09
Add specifications to the main

SeanHgh:
I added specifications to the requirements of requests using doorstop


SeanHgh:
also added a print to show what we post and its specification

PR #6996, opened 2025-07-18
Remove the unused tox and flake8 sections from Makefile

abhi-jha:


nateprewitt:
Sorry, perhaps I wasn't clear. Let's remove _only_  `flake8` since it's no longer used. We should be leaving `tox`. It is not a core part of the dev cycle, but is still available for developers who want it. Users can install tox if they need it.

nateprewitt:
Let's change `tox -p` to `python -m pytest tests` to avoid any confusion going forward.

Makefile
```diff
@@ -3,16 +3,14 @@ init:
 	python -m pip install -r requirements-dev.txt
 test:
 	# This runs all of the tests on all supported Python versions.
```
nateprewitt:
```suggestion
```

nateprewitt (approved):
Thanks, @abhi-jha

Makefile
abhi-jha: The comment looks relevant to me. Why remove it?

PR #7004, opened 2025-08-05
Fix-#7003: RequestsCookieJar cannot correctly retrieve zero value

### LINKED ISSUE #7003: When using RequestsCookieJar to set a cookie with an empty string ('') or 0 as the value, it cannot be retrieved properly afterward.
When using RequestsCookieJar to set a cookie with an empty string ('') or 0 as the value, it cannot be retrieved properly afterward.

## Expected Result

It can be retrieved properly after being set.

## Actual Result

It cannot be retrieved properly after being set.

## Reproduction Steps

```python
from requests.cookies import RequestsCookieJar

jar = RequestsCookieJar()

jar.set('token', 0, domain='example.com', path='/')

print("Jar contents:", list(jar.items()))  # [('token', 0)]
print("jar.get:", jar.get('token', domain='example.com', path='/'))  # it should return 0 instead of None.

try:
    print("jar['token']:", jar['token'])  # it should return 0 instead of raising an error.
except KeyError as e:
    print("KeyError raised unexpectedly:", e)

```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "2.0.12"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.7"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.6.8"
  },
  "platform": {
    "release": "3.10.0-1160.119.1.el7.x86_64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.27.1"
  },
  "system_ssl": {
    "version": "100020bf"
  },
  "urllib3": {
    "version": "1.26.19"
  },
  "using_pyopenssl": false
}

```

mengxunQAQ:
When using RequestsCookieJar to set a cookie with an empty string ('') or 0 as the value, it cannot be retrieved properly afterward.  (Fix #7003)

### DIFF
--- a/src/requests/cookies.py
+++ b/src/requests/cookies.py
@@ -408,7 +408,7 @@ class RequestsCookieJar(cookielib.CookieJar, MutableMapping):
                         # we will eventually return this as long as no cookie conflict
                         toReturn = cookie.value
 
-        if toReturn:
+        if toReturn is not None:
             return toReturn
         raise KeyError(f"name={name!r}, domain={domain!r}, path={path!r}")
 


nateprewitt:
Resolving as a duplicate of #5950.

james-may2553:
## 🤖 AI Code Review

**src/requests/cookies.py:1**
- Severity: medium
- Issue: Mock review issue for local testing.
- Suggestion: This confirms the pipeline works without calling the API.

PR #7005, opened 2025-08-06
security: update publish workflow to use `artifact-id` instead of `name`

GrantBirki:
## Use `artifact-id` instead of `name` with Actions Artifacts 🔒 

This pull updates the **publish** workflow to use `artifact-id` instead of `name` when downloading artifacts that have been previously published in prior workflow steps. This is important because artifacts produced by GitHub Actions can be completely overwritten by other workflow runs if they use the same `name` under very unique circumstances (like passing the `run-id` value to the download-artifact action to point to an entirely different workflow run - don't do that). To avoid potential TOCTOU issues/vulnerabilities where an artifact might be replaced between upload and download, the new `artifact-ids` input allows you to download artifacts by their specific ID rather than by name. This is safer but also helps lead to more deterministic workflow builds by referencing the artifact you wish to download by its _exact id_.

This PR also hardens the workflow a bit by adding `persist-credentials: false` to the checkout step.

---

I recently did some work to land these exact changes in the [urllib3/urllib3](https://github.com/urllib3/urllib3/pulls?q=is%3Apr+is%3Aclosed+author%3AGrantBirki) and wanted to contribute those same changes here as well!

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,6 +16,7 @@ jobs:
       name: "publish"
     outputs:
       hashes: ${{ steps.hash.outputs.hashes }}
+      artifact-id: ${{ steps.upload-artifact.outputs.artifact-id }}
 
     steps:
       - name: Harden the runner (Audit all outbound calls)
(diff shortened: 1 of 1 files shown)


sigmavirus24 (approved):

sigmavirus24:
🎉 Thank you!

PR #7009, opened 2025-08-10
Add test for RequestsCookieJar zero and empty string value handling

mefi92:
This pull request adds a test that verifies RequestsCookieJar correctly handles cookies with zero and empty string values.
I noticed there are already fixes related to this issue (#7003), but no corresponding test was included.
Therefore, this PR adds the missing test to ensure no regressions happen in the future.

No changes to production code are included, only the test to improve coverage.

### DIFF
--- a/tests/test_cookie_zero.py
+++ b/tests/test_cookie_zero.py
@@ -1,0 +1,10 @@
+from requests.cookies import RequestsCookieJar
+
+
+def test_cookiejar_zero_value():
+    jar = RequestsCookieJar()
+    jar.set('token', 0, domain='example.com', path='/')
+
+    assert ('token', 0) in jar.items()
+    assert jar.get('token', domain='example.com', path='/') == 0
+    assert jar['token'] == 0


nateprewitt:
Hi @mefi92, we'll work through any additional changes in the original PR if it's accepted. We haven't even made a decision on if the original issue is entirely correct. 0 is not a valid value for a cookie, it's always going to be represented as a string or bytes.

We'll follow up on the issue once there's capacity.

mengxunQAQ:
> 0 is not a valid value for a cookie, it's always going to be represented as a string or bytes.

Hi @nateprewitt , I have some thoughts.

- Although integers or floats are not valid cookie values according to the spec, I think it’s reasonable to allow them for greater flexibility. In the end, when they are appended into the cookie, they are effectively converted to strings, so there’s actually no problem in practice.
- Moreover, the current `set` method does accept integers and floats, so if they are accepted during `set`, they should be retrievable afterwards. If `set` no longer accepts integers or floats, it would introduce a backward compatibility issue.
- Even if we don’t consider integers or floats, an empty string is a valid cookie value, and its current behavior is also incorrect.

sigmavirus24:
> Although integers or floats are not valid cookie values according to the spec, I think it’s reasonable to allow them for greater flexibility. In the end, when they are appended into the cookie, they are effectively converted to strings, so there’s actually no problem in practice.

I disagree. The cookie library doesn't guarantee that serialization and it should reject invalid inputs like this. Cookie parsing can be an avenue for attacks. We shouldn't be enabling that.

> Even if we don’t consider integers or floats, an empty string is a valid cookie value, and its current behavior is also incorrect.

Please read Nate's response. There's another PR that addresses this better already. _If_ we accept it, it will solve some of these issues.

PR #7015, opened 2025-08-21
Use EmailMessage for response headers

hramrach:
MockResponse requires EmailMessage for http.CookieJar to be able to
extract cookies from headers but CaseInsensitiveDict is provided.

Use EmailMessage and special-case Set-Cookie per note in
https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2

Fixes: #7014

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -6,6 +6,8 @@ This module contains the primary objects that power Requests.
 """
 
 import datetime
+import email
+import email.policy
 
 # Import encoding now, to avoid implicit import later.
 # Implicit import within threads may cause LookupError when standard library is in a ZIP,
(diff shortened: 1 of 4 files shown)


sigmavirus24:
This is not backwards compatible and will not be accepted

hramrach:
You mean bug-compatible?

src/requests/models.py
```diff
@@ -663,10 +665,10 @@ def __init__(self):
         #: Integer Code of responded HTTP Status, e.g. 404 or 200.
         self.status_code = None
 
-        #: Case-insensitive Dictionary of Response Headers.
+        #: Case-insensitive Dictionary (email.EmailMessage) of Response Headers.
```
sigmavirus24, line 668: This is incorrect for documentation linking (if that's what you're going for)

src/requests/models.py
```diff
@@ -663,10 +665,10 @@ def __init__(self):
...
         #: For example, ``headers['content-encoding']`` will return the
         #: value of a ``'Content-Encoding'`` response header.
-        self.headers = CaseInsensitiveDict()
```
sigmavirus24, line 669: This is specifically backwards-incompatible. `EmailMessage` and the `CaseInsensitiveDict` do not have compatible methods.

sigmavirus24 (commented):
Given your hostility, I've laid it out plainly for you what's wrong and why this won't be accepted. 

Continued hostility will result in moderation being applied to your interactions here

src/requests/models.py
hramrach: I am not particularly familiar with the documentation formatting. It does provide the class name that can be looked up but does not do much else if there are additional things that could be done.

src/requests/models.py
hramrach: Indeed, if somebody wanted to modify the provided object the modification methods are not compatible.

Although such use is not tested if it's possible to do somebody is bound to do it.

What is an option to provide correct headers?

Save the original response headers in another property?

PR #7103, opened 2025-12-01
Fix handling of zero and empty cookie values in RequestsCookieJar

egangi:
Description:
This PR addresses a bug in the _find_no_duplicates method of RequestsCookieJar where cookies with values of 0 or empty strings ('') were not being returned correctly.

Changes included:
Updated _find_no_duplicates to correctly return cookie values that are 0 or ''.
Added and updated test cases in test_cookie.py to cover:
-Cookies with value 0
-Cookies with empty string value
-Cookies with None value
-Cookies with domain and path matching
-Overwriting cookies with zero values
-Persistence of zero values after update

These changes ensure that the get method and other functions depending on _find_no_duplicates behave correctly for all valid cookie values.

Impact:
Existing code that relies on _find_no_duplicates returning cookie.value is preserved.

### DIFF
--- a/src/requests/cookies.py
+++ b/src/requests/cookies.py
@@ -395,22 +395,22 @@ class RequestsCookieJar(cookielib.CookieJar, MutableMapping):
             that match name and optionally domain and path
         :return: cookie.value
         """
-        toReturn = None
+        foundCookie = None
         for cookie in iter(self):
             if cookie.name == name:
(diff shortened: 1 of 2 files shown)


nateprewitt:
I think this is a semantic change to how the API works, this is already addressed in a separate PR so I think we'll pass on this. Thanks for the contribution.

PR #7151, opened 2026-01-09
Fix: Correct Content-Length for StringIO with multi-byte characters

hcthakur2004:
## Summary
Fixes a regression where `Content-Length` is incorrectly calculated for `io.StringIO` objects containing multi-byte characters (like emojis).

## The Issue
When `io.StringIO` is passed as `data`, [super_len](cci:1://file:///c:/Users/Harish/OneDrive/Desktop/webdev/sample%20project/requests_repo/src/requests/utils.py:135:0-203:50) uses character count via [__len__](cci:1://file:///c:/Users/Harish/OneDrive/Desktop/webdev/sample%20project/requests_repo/tests/test_utils.py:76:12-77:24) or [tell()](cci:1://file:///c:/Users/Harish/OneDrive/Desktop/webdev/sample%20project/requests_repo/tests/test_utils.py:89:12-90:29) instead of the UTF-8 encoded byte length. This mismatch causes server errors when the encoded body is larger than the reported length.

## The Fix
Updated [super_len](cci:1://file:///c:/Users/Harish/OneDrive/Desktop/webdev/sample%20project/requests_repo/src/requests/utils.py:135:0-203:50) in [utils.py](cci:7://file:///c:/Users/Harish/OneDrive/Desktop/webdev/sample%20project/requests_repo/tests/utils.py:0:0-0:0) to encode `StringIO` content to UTF-8 before calculating length, matching the behavior for standard strings and `urllib3` expectations.

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -142,6 +142,11 @@ def super_len(o):
         # of latin-1 (iso-8859-1) like http.client.
         o = o.encode("utf-8")
 
+    if isinstance(o, io.StringIO):
+        total_length = len(o.getvalue().encode("utf-8"))
+        current_position = len(o.getvalue()[: o.tell()].encode("utf-8"))
+        return max(0, total_length - current_position)
+
     if hasattr(o, "__len__"):
         total_length = len(o)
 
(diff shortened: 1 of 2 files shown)


maxine-yx (commented):
This is a test review as part of tool exploration. The fix for handling multi-byte characters in StringIO looks reasonable. The approach of encoding to UTF-8 before calculating length is correct.

PR #7181, opened 2026-01-27
docs: clarify gzip decompression behavior when using stream=True

AhamadAlii:
When using stream=True, Requests does not automatically decompress gzip-encoded
responses when accessing Response.raw. This behavior is intentional but not
clearly documented and has caused repeated confusion.

This PR adds a note in the advanced documentation explaining the behavior and
shows how to enable decompression manually:

    r.raw.decode_content = True

This documents the intended behavior discussed in #2155 and related issues.


AhamadAlii:
This documents the behavior discussed in #2155 and related issues.
Happy to adjust wording if needed.

PR #7201, opened 2026-02-10
Fix incorrect Content-Length for StringIO with multi-byte characters

### LINKED ISSUE #6917: Incorrect Content-Length header with StringIO body

veeceey:
## Summary

Fixes #6917.

`super_len()` uses `seek`/`tell` to measure the length of file-like objects such as `StringIO` and `BytesIO`. However, `StringIO.tell()` returns the **character position**, not the byte offset. For strings containing multi-byte UTF-8 characters (e.g. emoji), this produces an incorrect `Content-Length` header that violates [RFC 9110 section 8.6](https://www.rfc-editor.org/rfc/rfc9110#section-8.6).

For example, `io.StringIO("\U0001F4A9")` (a single emoji) previously returned a length of **1** (character count) instead of **4** (UTF-8 byte count), causing the server to receive a `Content-Length: 1` header while 4 bytes are actually sent.

This is the same class of bug that was fixed for plain `str` bodies in #6586 -- `str` is encoded to UTF-8 before measuring, but `StringIO` was not. This PR makes `StringIO` handling consistent with `str` by reading the remaining text, encoding it to UTF-8, and measuring the byte length.

### Before

```
str       → Content-Length: 4  ✓
bytes     → Content-Length: 4  ✓
BytesIO   → Content-Length: 4  ✓
StringIO  → Content-Length: 1  ✗  (character count, not byte count)
```

### After

```
str       → Content-Length: 4  ✓
bytes     → Content-Length: 4  ✓
BytesIO   → Content-Length: 4  ✓
StringIO  → Content-Length: 4  ✓
```

## Changes

- **`src/requests/utils.py`**: In `super_len()`, detect `io.StringIO` and read+encode the remaining text to compute the UTF-8 byte length instead of relying on `tell()`.
- **`tests/test_utils.py`**: Added `test_super_len_stringio_multibyte` covering single emoji, mixed content, partially-read StringIO, and position preservation.

## Test plan

- [x] All existing `TestSuperLen` tests pass (ASCII StringIO, BytesIO, partially-read files, etc.)
- [x] New test verifies correct byte count for multi-byte characters
- [x] New test verifies correct byte count for partially-read StringIO
- [x] New test verifies file position is preserved after `super_len()` call

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -186,13 +186,30 @@ def super_len(o):
             if hasattr(o, "seek") and total_length is None:
                 # StringIO and BytesIO have seek but no usable fileno
                 try:
-                    # seek to end of file
-                    o.seek(0, 2)
-                    total_length = o.tell()
+                    if isinstance(o, io.StringIO):
+                        # StringIO.tell() returns the character
+                        # position, not the byte offset.  Read the
+                        # remaining text and encode it to measure the
+                        # true byte length for Content-Length.
+                        start = current_position or 0
+                        o.seek(start)
+                        total_length = len(o.read().encode("utf-8"))
 
-                    # seek back to current position to support
-                    # partially read file-like objects
-                    o.seek(current_position or 0)
+                        # Reset current_position so the returned value
+                        # is just total_length (the remaining bytes).
+                        current_position = 0
+
+                        # seek back to original position to support
+                        # partially read file-like objects
+                        o.seek(start)
+                    else:
+                        # seek to end of file
+                        o.seek(0, 2)
+                        total_length = o.tell()
+
+                        # seek back to current position to support
+                        # partially read file-like objects
+                        o.seek(current_position or 0)
                 except OSError:
                     total_length = 0
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -142,6 +142,30 @@ class TestSuperLen:
         foo.read(2)
         assert super_len(foo) == 3
 
+    def test_super_len_stringio_multibyte(self):
+        """Ensure StringIO with multi-byte characters returns the UTF-8
+        byte length rather than the character count.  See #6917."""
+        # Single emoji: 1 character, 4 bytes in UTF-8
+        foo = StringIO.StringIO("\U0001F4A9")
+        assert super_len(foo) == 4
+
+        # Mixed ASCII and multi-byte
+        foo = StringIO.StringIO("hello \U0001F4A9 world")
+        assert super_len(foo) == len("hello \U0001F4A9 world".encode("utf-8"))
+
+        # Partially read StringIO with multi-byte characters
+        foo = StringIO.StringIO("hello \U0001F4A9 world")
+        foo.read(6)  # read "hello "
+        remaining_bytes = len("\U0001F4A9 world".encode("utf-8"))
+        assert super_len(foo) == remaining_bytes
+
+        # Position should be preserved after super_len call
+        foo = StringIO.StringIO("hello \U0001F4A9 world")
+        foo.read(3)
+        pos_before = foo.tell()
+        super_len(foo)
+        assert foo.tell() == pos_before
+
     def test_super_len_with_fileno(self):
         with open(__file__, "rb") as f:
             length = super_len(f)


veeceey:
just wanted to follow up and see if this is good to go or needs more work

StantonMatt (commented):
I reproduced the issue and the fix locally.

On current `main`, the reporter's core case still gives `StringIO super_len=1 Content-Length=1` while plain `str` gives `4/4`. On this branch the same `StringIO` case gives `super_len=4 Content-Length=4`, and the cursor-preservation/partially-read cases pass for me:

```console
TOX_WORK_DIR=.codex-tmp/tox/requests-7201 tox -e py312-default -- tests/test_utils.py -q -k super_len
11 passed, 208 deselected
```

One small test gap: since #6917 is user-visible through the prepared request header, I think it would be worth adding a direct `PreparedRequest().prepare(..., data=io.StringIO(...))` assertion for `Content-Length == "4"` as well. The new `super_len` coverage is useful, but a header-level assertion would lock the actual behavior that regressed and make this less dependent on the `prepare_content_length -> super_len` path staying obvious.

PR #7213, opened 2026-02-20
Fix PreparedRequest.copy() sharing hooks reference with original

bysiber:
`PreparedRequest.copy()` currently assigns `p.hooks = self.hooks` which means the copy shares the same hooks dictionary and callback lists with the original request. Modifying hooks on the copy (e.g., registering a new response hook) will unintentionally affect the original as well.

This is inconsistent with how the method handles other mutable attributes — `headers` and `_cookies` are both properly copied via `.copy()` and `_copy_cookie_jar()`, respectively.

```python
import requests

req = requests.Request("GET", "https://example.com").prepare()
copy = req.copy()

copy.hooks["response"].append(lambda r, **kw: print("hook fired"))
print(req.hooks["response"])  # ['<lambda>'] — original was modified
```

The fix creates a new dict with shallow-copied callback lists so that adding or removing hooks on a copy does not affect the original.

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -388,7 +388,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         p.headers = self.headers.copy() if self.headers is not None else None
         p._cookies = _copy_cookie_jar(self._cookies)
         p.body = self.body
-        p.hooks = self.hooks
+        p.hooks = {event: list(callbacks) for event, callbacks in self.hooks.items()}
         p._body_position = self._body_position
         return p
 


src/requests/models.py
```diff
@@ -388,7 +388,7 @@ def copy(self):
         p.headers = self.headers.copy() if self.headers is not None else None
         p._cookies = _copy_cookie_jar(self._cookies)
         p.body = self.body
-        p.hooks = self.hooks
+        p.hooks = {event: list(callbacks) for event, callbacks in self.hooks.items()}
```
sigmavirus24, line 391: `copy.deepcopy` does this better, if I remember correctly and is explicit in what it's doing/achieving

sigmavirus24 (commented):
We likely need to kill the notion that this will avoid mutating the original. A file(-like) object is handled as the as we can't clone that. So we end up mutating that as well.

Either way, there's a better way to do this, if we do it at all

bysiber:
Fair point about file-like objects — copy() can never provide a full deep copy because of those. But I think fixing the hooks case is still worthwhile since hooks is just a plain dict of lists, and the mutation there is unexpected and easy to avoid (unlike file handles). The current inconsistency where headers and cookies are copied but hooks aren't seems like an oversight rather than a deliberate decision.

What approach would you prefer? Happy to adjust.

PR #7217, opened 2026-02-20
Fix #6122: Empty body with None values sends malformed chunked request

rodionsteshenko:
When `data={'foo': None}`, the body encodes to an empty string but `Content-Length` was not set because `0` is falsy in the `if length:` check. This caused the adapter to fall back to `Transfer-Encoding: chunked`, sending a terminating chunk (`0\r\n\r\n`) that servers misinterpreted as a second, malformed request.

## Fix

Changed the `Content-Length` check in `prepare_content_length()` from `if length:` to `if length is not None and "Transfer-Encoding" not in self.headers:`, so that `Content-Length: 0` is correctly set for empty bodies while preserving chunked encoding for streams.

## Test

Added `test_empty_body_from_none_values_sets_content_length_zero` that reproduces the original bug: creates a request with `data={'foo': None}` and verifies that `Content-Length: 0` is set and `Transfer-Encoding` is not present.

Tested locally on macOS ARM (Apple Silicon). All existing tests pass.

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -575,9 +575,13 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
         """Prepare Content-Length header based on request method and body"""
         if body is not None:
             length = super_len(body)
-            if length:
-                # If length exists, set it. Otherwise, we fallback
-                # to Transfer-Encoding: chunked.
+            if length is not None and "Transfer-Encoding" not in self.headers:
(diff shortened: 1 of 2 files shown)


themavik (commented):
nit: skipping Content-Length when Transfer-Encoding is present is right for chunked bodies; worth noting if any code path can leave TE set while the prepared body is empty-sized.

PR #7232, opened 2026-03-01
feat: add RFC 7616 support for non-Latin credentials in HTTPDigestAuth

Jo2234:
## Summary

Implement [RFC 7616](https://www.rfc-editor.org/rfc/rfc7616) extensions to fix `HTTPDigestAuth` failing with non-Latin-1 usernames (e.g., Cyrillic, Czech diacritics).

## Problem

```python
import requests
from requests.auth import HTTPDigestAuth

# This fails — 'ř' cannot be encoded as latin-1
auth = HTTPDigestAuth('Ondřej', 'heslíčko')
# The Digest header produces: username="Ondřej" which is invalid

# Workaround from issue produces garbled bytes:
auth = HTTPDigestAuth('Ondřej'.encode('utf-8'), 'heslíčko')
# Produces: username="b'Ond\\xc5\\x99ej'"  — wrong!
```

## Solution

This PR implements three RFC 7616 features:

### 1. `username*` parameter (RFC 7616 §3.4 + RFC 5987)

When the username cannot be encoded as Latin-1, the `username*` parameter with [RFC 5987](https://www.rfc-editor.org/rfc/rfc5987) encoding is used:

```
# Before (broken):
Digest username="Ondřej", ...

# After (correct):
Digest username*=UTF-8''Ond%C5%99ej, ...
```

### 2. `userhash` support (RFC 7616 §3.4.4)

When the server sends `userhash=true` in the challenge, the username is hashed for privacy:
```
Digest username="<hash of username:realm>", ..., userhash=true
```

### 3. `charset` echo (RFC 7616 §3.3)

When the server advertises `charset=UTF-8`, the client echoes it back.

## Changes

| File | Change |
|------|--------|
| `src/requests/auth.py` | Added `_is_latin1_encodable()`, `_encode_rfc5987()`, and updated `build_digest_header()` with RFC 7616 username handling |
| `tests/test_digest_rfc7616.py` | **12 new tests** covering: Latin-1 detection, RFC 5987 encoding, `username*` parameter, `userhash`, `charset`, and response digest validity |

## Backward Compatibility

- **Latin-1 encodable usernames** (ASCII, Western European): No change in behavior — the standard `username="..." ` parameter is used as before
- **Non-Latin-1 usernames** (Cyrillic, CJK, etc.): Now work correctly instead of failing
- The response digest computation is unchanged — `A1 = username:realm:password` is still hashed with UTF-8

Fixes #6102

### DIFF
--- a/src/requests/auth.py
+++ b/src/requests/auth.py
@@ -12,6 +12,7 @@ import threading
 import time
 import warnings
 from base64 import b64encode
+from urllib.parse import quote
 
 from ._internal_utils import to_native_string
 from .compat import basestring, str, urlparse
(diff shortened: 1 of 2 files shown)


themavik (commented):
nit: digest username* encoding depends on `urllib.parse.quote` plus the custom `safe` set; a regression test for edge chars (e.g. backtick) would lock RFC5987/Digest grammar assumptions.

PR #7272, opened 2026-03-18
Add inline types to Requests

nateprewitt:
# Add inline type annotations

> [!IMPORTANT]
> We want feedback from people who actively maintain projects that depend on Requests or use it heavily. Please share your experience testing this against your code **in the linked issue**.
>
>Comments that are clearly AI-generated will be hidden or removed. This is a library written "for Humans". The conversation is between maintainers and users, not a prompt.

## Overview

This PR adds type annotations throughout Requests to codify the type contract alongside the code. This is a path for what inline typing in Requests _could_ look like. The goals are to establish what APIs are intended to support, make more reasoned decisions about code changes, and make using Requests in a modern development environment easier.

The package passes strict type checking with pyright, with the caveat that a few decisions on internals have been deferred for discussion by the maintainers and broader community. Right now, the intended behavior for those is undefined.

Related issue: [RFC: Adding inline type annotations to Requests](https://github.com/psf/requests/issues/7271)
Currently tracked TODOs: https://gist.github.com/nateprewitt/3e65fc8c3e9db3d5629ec1daf1f993f8

## How to verify locally

```
pip install pyright typing-extensions
python -m pyright src/requests/
```

Should report 0 errors. `typing-extensions` is needed on Python 3.10-3.12.

## Main design decisions

- `cast()` over `assert/isinstance` for type narrowing to avoid breaking downstream code at runtime
- New `_types.py` module centralizing Protocols, type aliases, and TypeVars
- `SupportsRead`/`SupportsItems` as `@runtime_checkable` Protocols replacing `hasattr` checks
- `is_prepared()` as a `TypeIs` guard that's a no-op at runtime (always returns `True`)
- ~100 type suppressions, each documented with an inline PR comment

## Runtime impact

A full catalog of every runtime-impacting change is available [here](https://gist.github.com/nateprewitt/08a622a4bd8895a9c4a7a7047b088648). The behavioral changes and bug fixes are the ones worth scrutinizing. The rest is mechanical.

## Open discussion points

- **`str` vs `bytes` URL handling**: Requests has historically accepted both when Python 2.7 was primary usage. These types currently declare `str` only. This needs a decision on whether to fix existing broken bytes behavior or formalize the `str`-only contract.
- **`yield_requests` API**: Needs documentation and a decision on how this should even work in practice. This API decision came through without review and I don't think there's any realistic usage outside of the library.

## What's NOT in this PR

- **mypy compatibility**: Developed against pyright. If we do add mypy, that will be in a separate follow-up. Public API types should still work for mypy users.
- **Public API changes**: No signature changes, no new exceptions, no runtime assertions.
- **Test type-checking**: Tests are not under strict mode. There are a number of tests that intentionally violate the type contract. We'll address those once we have a final plan here.


src/requests/api.py
```diff
@@ -8,10 +8,20 @@
...
+from .models import Response
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from ._types import DataType, JsonType, ParamsType
 
 
-def request(method, url, **kwargs):
+def request(method: str, url: str, **kwargs: Any) -> Response:
```
cthoyt: I know it would be a lot of work, but annotating the kwargs with an `Unpack[KwargsDict]` where `KwargsDict` is a `typing.TypedDict` would be great!

I hacked this into my own code that interacts with requests code to make my usage more safe: 

- where I defined my own typed dict class https://github.com/cthoyt/pystow/blob/1ca7fa5763c6aa0d2138a164c78fd67955152ebc/src/pystow/utils/download.py#L63-L72
- where I annotate it on a function I wrote, that wraps `requsts.get`: https://github.com/cthoyt/pystow/blob/1ca7fa5763c6aa0d2138a164c78fd67955152ebc/src/pystow/utils/download.py#L87
- where the kwargs get splatted into `requests.get`: https://github.com/cthoyt/pystow/blob/1ca7fa5763c6aa0d2138a164c78fd67955152ebc/src/pystow/utils/download.py#L165

cthoyt: also I am willing to PR this if you think it's a good idea / if you think an external contrib is better/easier than doing it yourself

src/requests/_types.py
```diff
@@ -0,0 +1,133 @@
...
+        method: str  # type: ignore[reportIncompatibleVariableOverride]
+
+    # Type aliases for core API concepts (ordered by request() signature)
+    UriType: TypeAlias = str | bytes
+
+    _ParamsMappingKeyType: TypeAlias = str | bytes | int | float
+    _ParamsMappingValueType: TypeAlias = (
+        str | bytes | int | float | Iterable[str | bytes | int | float] | None
+    )
+    ParamsType: TypeAlias = (
```
cthoyt, line 81: it would be lovely if these type hints were available through a "public" interface, since I've often written code that wraps a requests call, where I would want to use the ParamsType to annotate my own code

I say public in quotes because I _could_ import this, but it feels wrong

src/requests/api.py
nateprewitt: Thanks for comment, @cthoyt! This is a good callout because it is a delta from typeshed that should've been in the overview.

I played with Unpack for this but put that on pause because it got kind of messy. From what I was able to come up with we're going to need several different TypedDicts to accomplish it. Essentially one per signature, it's definitely doable, just clunky.

If you have ideas to simplify the approach, I'd love to hear more. This is something that needs an answer before we ship.

cthoyt: one thing worth noting is that you can have (multiple) inheritance on typed dicts. this can make it easier to factor out similarities between typed dict definitions for several functions

src/requests/_types.py
nateprewitt: I agree having some form of public hints for the top-level APIs is potentially warranted. I started very conservative here because often times new code goes into Requests and immediately comes someone else's critical dependency. I don't want to create a binding contract that people start surfacing in their code, and then "break" when we need to update/tweak it.

I think once we're really happy with the types, the main argument parameters could be surfaced. I've deferred that for now until we can make a more informed decision. Presumably everything calling in is already typed sufficiently.

cthoyt: fair! thank you for doing this and considering carefully :)

src/requests/adapters.py
```diff
@@ -631,29 +683,29 @@ def send(
...
         else:
-            timeout = TimeoutSauce(connect=timeout, read=timeout)
+            resolved_timeout = TimeoutSauce(connect=timeout, read=timeout)
 
         try:
             resp = conn.urlopen(
                 method=request.method,
                 url=url,
-                body=request.body,
+                body=request.body,  # type: ignore[arg-type]  # urllib3 stubs don't accept Iterable[bytes | str]
```
nateprewitt, line 701: Should be addressed with https://github.com/urllib3/urllib3/pull/3799.

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,29 @@
...
+
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+
+    - name: Set up Python
+      uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      with:
+        python-version: "3.10"
```
srittau: You could consider using a version matrix here. This could flag removals, API changes, and deprecations in CPython's stdlib early. Maybe just using the oldest and latest CPython version supported by requests would be enough as a smoke test.

src/requests/_types.py
```diff
@@ -0,0 +1,133 @@
...
+    TypeVar,
+    runtime_checkable,
+)
+
+_T_co = TypeVar("_T_co", covariant=True)
+
+
+@runtime_checkable
+class SupportsRead(Protocol[_T_co]):
+    def read(self, length: int = ...) -> _T_co: ...
```
srittau: You could consider conditionally using `io.Reader` when using Python 3.14+. It is nearly identically defined, with one difference (pos-only length argument) that I would suggest adding:

```suggestion
@runtime_checkable
class SupportsRead(Protocol[_T_co]):
    def read(self, length: int = ..., /) -> _T_co: ...
```

src/requests/_types.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+@runtime_checkable
+class SupportsItems(Protocol):
+    def items(self) -> Iterable[tuple[Any, Any]]: ...
+
+
+# These are needed at runtime for default_hooks() return type
+HookType = Callable[["Response"], Any]
+HooksInputType = Mapping[str, "Iterable[HookType] | HookType"]
```
srittau: Shouldn't these be marked as `TypeAlias`es?

src/requests/_types.py
```diff
@@ -0,0 +1,133 @@
...
+        Mapping[str, _FileSpec] | Iterable[tuple[str, _FileSpec]] | None
+    )
+
+    AuthType: TypeAlias = (
+        tuple[str, str] | AuthBase | Callable[[PreparedRequest], PreparedRequest] | None
+    )
+
+    TimeoutType: TypeAlias = float | tuple[float | None, float | None] | None
+    ProxiesType: TypeAlias = MutableMapping[str, str]
+    HooksType: TypeAlias = dict[str, list["HookType"]] | None
```
srittau: Nit: `HookType` doesn't need to be quotes as it's defined before in this file.

srittau (commented):
Thanks for adding type hints! As a typeshed maintainer, I'm always glad to see annotations being added directly to packages. Less work for me , but more importantly as better user experience.

Also a general note: Some kwargs arguments could probably be accurately annotated using `Unpack[_SomeTypedDict]`. See for example the [ast.pyi module](https://github.com/python/typeshed/blob/main/stdlib/ast.pyi) in typeshed.

You could also consider marking constants as `Final`. This sometimes even makes type annotations easier. For example from below: Instead of 

```python
DEFAULT_PORTS: dict[str, int] = {"http": 80, "https": 443}
```

you could use

```python
DEFAULT_PORTS: Final = {"http": 80, "https": 443}
```

At mypy infers the same type, although in some cases type checkers can even infer literals.

Some specific suggestions below.

src/requests/_types.py
nateprewitt: Good callout, the Hook* types were originally at the bottom of the file, but got lifted just before I posted this.

src/requests/_types.py
nateprewitt: Yep, also a good call out. I'll get those added.

.github/workflows/typecheck.yml
nateprewitt: Yeah, that's fair. I started with the minimum because we've generally tested floors but I agree having something for the latest version also makes sense to catch breakages as we add support.

nateprewitt:
Really appreciate the response and your time looking at this, @srittau! I can get most of those addressed quick. The unpack piece I covered quickly with Chris in https://github.com/psf/requests/pull/7272#discussion_r2954489817. That's definitely a delta from typeshed we'll want to cover. The initial attempt on that got messy pretty quickly so I sidelined it. We'll get that addressed before shipping this though.

nateprewitt:
I've added https://gist.github.com/nateprewitt/3e65fc8c3e9db3d5629ec1daf1f993f8 to track our current checklist before merging this.

src/requests/api.py
nateprewitt: @cthoyt I pushed up another pass at Unpack. Let me know if you're seeing any issues with your setup or have any other feedback. Appreciate it!

.github/workflows/typecheck.yml
nateprewitt: Addressed in 4a6d321f1fff6c7cd829cb4cfc60e54ef8a76b4b.

src/requests/_types.py
nateprewitt: Addressed in https://github.com/psf/requests/pull/7272/commits/5c93c71069a65381a4722dda9df4210e6d381240.

src/requests/_types.py
nateprewitt: Addressed in [f651c9b](https://github.com/psf/requests/pull/7272/commits/f651c9b92d784bb0171939abbf5af778aafa2dc3).

src/requests/_types.py
nateprewitt: Addressed in [f651c9b](https://github.com/psf/requests/pull/7272/commits/f651c9b92d784bb0171939abbf5af778aafa2dc3).

src/requests/structures.py
```diff
@@ -37,63 +44,81 @@ class CaseInsensitiveDict(MutableMapping):
...
         super().__init__()
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"<lookup '{self.name}'>"
 
-    def __getitem__(self, key):
+    def __getattr__(self, key: str) -> _VT | None:
+        # Allow attribute-style access to values
+        return self.__dict__.get(key, None)
```
jorenham: This seems a bit unrelated :thinking:

nateprewitt:
Ok, I think we're pretty close to ready with this. I've resolved the outstanding concerns originally listed. To summarize:

* We've aligned on accepting urls as `str | bytes` because that it what the API has supported _but_ it's always been normalized at the entry point to `str`. `Request`, `PreparedRequest` and many internal APIs/util functions are typed with that constraint. There are many failure points internally if you change the URL to `bytes` in the `PreparedRequest` flow after `prepare()` is called. That usage has no guarantees and the typing reflects that. Acceptance of bytes is purely as a convenience feature for the top-level APIs.

* `yield_requests` is a mess. We're deep enough in Requests, I don't think we're going to solve that now. The ignore is currently honest about the API wart, and I'd be very surprised if someone is using parameter meaningfully.

## For review

There's obviously stylistic choices and naming that I'd welcome comment from other maintainers. Otherwise, I believe the types are largely complete.

### Areas that could use scrutiny/comment
#### Top-level API kwargs
I've added Unpack at the suggestion of @cthoyt and @srittau  (thank you!). They expose the underlying **kwargs options, but the Requests API is also currently defined with only the relevant kwargs being explicit. Now we'll surface thinks like `body` for `get()` which isn't meaningful. This keeps us inline with `typeshed` but may make the current API expectations unclear. I'm somewhat indifferent on what we do here, but want the decision to be clear. We're opting for alignment with existing types and discoverability over trying to convey HTTP semantics. @sigmavirus24 do you have a strong opinon?

#### urllib3
I've fixed 1 of 3 typing issues in urllib3 but the other 2 are likely going to be drawn out fixes. I'm commenting them out with the knowledge we know they work today despite the type inconistencies. We'll revisit those once we get to a conclusion there.

#### Header typing
We've historically said we accept both `bytes` and `str` for header and in fact have noted a preference towards bytes. While we take both as input, header keys are immediately normalized to `str` (and that's consistent with the specification, they need to be `ascii`). The types reflect `Mapping[str, bytes | str]` and `CaseInsensitiveDict[bytes | str]` which is consistent with our current functionality. It's just a subtle difference from what I think we've said in issues previously.

src/requests/structures.py
nateprewitt: It's actually required. I responded earlier on mobile about RequestsCookieJar but couldn't see the surrounding context. I had to go back and dig a bit because this was done during the first pass back in February.

LookupDict is specifically used for [`codes`](https://github.com/psf/requests/blob/6ec76b4a36fba4a32fb72f5f9ba04e632635e477/src/requests/status_codes.py#L23-L106) in Requests which has some ["unique" boostraping](https://github.com/psf/requests/blob/6ec76b4a36fba4a32fb72f5f9ba04e632635e477/src/requests/status_codes.py#L109-L125). The calls to setattr populate the `__dict__` under the hood, allowing attribute access. That behavior isn't clear to mypy or pyright. While this isn't necessary for functionality, the static type analysis fails to find a type without this bridge.

You can remove this change and you'll see errors throughout models.py and session.py like:

```python
  requests/sessions.py:398:36 - error: Type of "found" is unknown (reportUnknownMemberType)
  requests/sessions.py:398:42 - error: Cannot access attribute "found" for class "LookupDict[int]"
    Attribute "found" is unknown (reportAttributeAccessIssue)
  requests/sessions.py:403:36 - error: Type of "moved" is unknown (reportUnknownMemberType)
  requests/sessions.py:403:42 - error: Cannot access attribute "moved" for class "LookupDict[int]"
    Attribute "moved" is unknown (reportAttributeAccessIssue)
```

nateprewitt: I've made a small tweak to this in [fb56915](https://github.com/psf/requests/pull/7272/commits/fb56915c62230fc5e4880e3b576a6d734cd3db9b) and added tests in #7401. This did have some behavioral implications for `getattr`/`hasattr` usage.

I don't think we have any context on how widely the `structures` in Requests are used outside of the context, but this wasn't intended to change behavior.

src/requests/utils.py
```diff
@@ -501,17 +542,18 @@ def get_encodings_from_content(content):
     )
 
 
-def _parse_content_type_header(header):
+def _parse_content_type_header(header: str) -> tuple[str, dict[str, Any]]:
     """Returns content type and parameters from given header.
 
+
```
nateprewitt:
```suggestion
```

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,32 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install dependencies
+      run: |
+        python -m pip install pip==26.0.1
+        python -m pip install -e . --group typecheck
+
+    - name: Run pyright
+      run: python -m pyright src/requests/
```
sigmavirus24, line 32: Why pyright instead of mypy? I've found mypy verified types work fine with pyright but not vice-versa.

src/requests/adapters.py
```diff
@@ -57,16 +60,21 @@
...
-    def SOCKSProxyManager(*args, **kwargs):
+    def SOCKSProxyManager(*args: Any, **kwargs: Any) -> None:
         raise InvalidSchema("Missing dependencies for SOCKS support.")
 
 
 if typing.TYPE_CHECKING:
+    from urllib3.connectionpool import HTTPConnectionPool
+    from urllib3.poolmanager import PoolManager as _PoolManager
+
+    from ._types import CertType, TimeoutType, VerifyType
```
sigmavirus24: Importing these like this will make these "public" right? Should we just import `_types` and use the type aliases directly?

src/requests/api.py
```diff
@@ -8,10 +8,28 @@
...
+from __future__ import annotations
 
+from typing import TYPE_CHECKING
 
-def request(method, url, **kwargs):
+from . import sessions
+from .models import Response
+
+if TYPE_CHECKING:
+    from ._types import (
```
sigmavirus24: Same concern here as above in adapters.py

src/requests/api.py
```diff
@@ -100,7 +120,12 @@ def head(url, **kwargs):
     return request("head", url, **kwargs)
 
 
-def post(url, data=None, json=None, **kwargs):
+def post(
+    url: UriType,
+    data: DataType = None,
+    json: JsonType = None,
+    **kwargs: Unpack[PostKwargs],
```
sigmavirus24: Is this the best way to do this? Would it be better to just do overrides? The permutations are probably too many given the ridiculous number of parameters to these functions

src/requests/auth.py
```diff
@@ -302,13 +341,13 @@ def __call__(self, r):
...
+    def __eq__(self, other: object) -> bool:
         return all(
             [
                 self.username == getattr(other, "username", None),
                 self.password == getattr(other, "password", None),
             ]
         )
 
-    def __ne__(self, other):
+    def __ne__(self, other: object) -> bool:
```
sigmavirus24: `object` or `Any`?

sigmavirus24 (approved):
Nothing blocking

.github/workflows/typecheck.yml
nateprewitt: We could use mypy (or make sure we're fully mypy compliant as well). The first pass at this in 2022, mypy struggled with the "magic" behaviors of Requests. The public API is currently mypy compatible but the internals aren't.

Pyright is mostly because I've worked with it on other projects. Anecdotally, I've found it's handled nuance better. I think the end goal (not this PR) is we do what pydantic does and have it running against both type checkers.

src/requests/adapters.py
nateprewitt: That's a good callout. I don't think they get re-exported in `from requests.adapters import *` but `_types.foo` is probably clearer.

src/requests/api.py
nateprewitt: I'm not sure if overrides would help for this case. I'd originally left this as `Any` to keep the API unchanged. Both @cthoyt and @srittau called this out as a deviation from typeshed. Unpack gives us auto-completion without mangling the signature. If we went the override route, we'd need to explicitly define everything which would essentially just making every api `request()`. At that point, it'd be more straight forward to just remove kwargs for explicit keywords.

I think the only callout I had on this above is **kwargs for every method is the params you generally shouldn't be using (`data` on `get`, `json` on `options`, etc.).

src/requests/adapters.py
nateprewitt: Addressed in [b594765](https://github.com/psf/requests/pull/7272/commits/b594765458375743d843e686607d29521e5f488c).

src/requests/api.py
nateprewitt: Addressed in [b594765](https://github.com/psf/requests/pull/7272/commits/b594765458375743d843e686607d29521e5f488c).

src/requests/auth.py
nateprewitt: Fixed in [4cf9bb5](https://github.com/psf/requests/pull/7272/commits/4cf9bb5cb159ad54165e1808666154a7509da4e6).

src/requests/_types.py
nateprewitt: Just to put a final decision here, everything is being shipped behind a `_t.` prefix for 2.34.0. The types have been tested across a number of downstream projects successfully and we're confident in the bulk of it. Once we get any additional feedback after the initial release, we'll likely provide public API types in a 2.35.x or 2.36.x release.

Issue #4323, opened 2017-10-02
Sessions not safe in multi-process environment


rabbbit:
tldr; It is unsafe to use Session in a multi-process environment - if the fork happens after Session initialisation the underlying connection pool will be shared across both processes, leading to potentially dangerous and hard to debug issues.

I'm thinking either a fix is necessary, or we should update the documentation to state the risk clearly - raising the issue because I'm not sure what is the right approach.

Please case.

## Use case

It is likely to happen if Session is created at module load time, like:

```python
class MyClient(object):
    session = requests.Session()

    def do_things(self, **params):
         self.session.get(**params)
```

or if a 3rd party client is imported at the module level, where it becomes totally invisible:

```python

my_database_client = database.DatabaseClient(**params)

class MyDatabaseWrapper(object):

      def do_things(self, **params):
            my_database_client.update(**params)
```

This is particularly tricky in a **Celery**, where a previously working function might start causing troubles if invoked from Celery. Celery seems like a common Python use case.

I've seen this pattern in 3 different repos written by 3 different developers - it feels common enough for it to be a problem.

## Reason

This is related to https://github.com/shazow/urllib3/issues/850 in urllib3, where it was stated that it's the callers' responsibility to worry about forking - in this case, it's Requests.

## Expected Result

Ideally, a new Session with the same parameters would be started by Requests. 

But since that might be complicated, I think it would be reasonable for Session to explode if it detects a PID change.

## Actual Result

The responses are mixed up - one process might receive a response made for a call it didn't make.

## Reproduction Steps

```python
import os
import sys
import requests

MAX = 20
s = requests.Session()

for n in range(MAX):
    pid = os.fork()
    if pid == 0:
        try:
            # s.mount("http://", requests.adapters.HTTPAdapter())  # uncomment to fix
            r = s.get('http://httpstat.us/300?sleep=100')
        except Exception:
            # ignoring intermittent http errors
            pass
        sys.exit(0)
    else:
        try:
            r = s.get('http://httpstat.us/200')
        except Exception as exc:
            # ignoring intermittent http errors
            pass

        if r.status_code != 200:
            print 'pid: {} Call {}/{}. Wrong status code detected {}'.format(
                os.getpid(),
                n,
                MAX,
                r.status_code
            )
```

```
pid: 30996 Call 7/20. Wrong status code detected 300   
pid: 30996 Call 9/20. Wrong status code detected 300   
pid: 30996 Call 14/20. Wrong status code detected 300  
[...]

PR #7297, opened 2026-03-27
docs: warn about using Session across forked processes

yuchengpersonal:
Context: psf/requests#4323 notes that creating a Session before forking can lead to unsafe behavior because connection pools may be shared between processes.

This PR adds a short warning to the "Session Objects" section of the advanced usage docs to help users avoid this pitfall.

What changed
- Add a warning admonition about fork()/multiprocessing and Session connection pools

How to verify
- Docs build should succeed (RST-only change)

Risk
- Low; documentation-only change.

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -72,6 +72,18 @@ Sessions can also be used as context managers::
 This will make sure the session is closed as soon as the ``with`` block is
 exited, even if unhandled exceptions occurred.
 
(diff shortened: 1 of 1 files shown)


yuchengpersonal:
Hi @nateprewitt, this PR adds a documentation warning about using Session across forked processes. Would appreciate a review when you have time! Thanks for maintaining requests!

PR #7315, opened 2026-03-31
Preserve leading slashes in request path_url

### LINKED ISSUE #6711: breaking aws s3 usage with requests 2.32.0
requests 2.32.0 introduced a change https://github.com/psf/requests/pull/6644 that strips double /. 

This has introduced an issue where generated presigned urls for s3 keys that start with a / can no longer be used. requests now strips that second / meaning that the key is modified and the signature does not match resulting in 403 errors. We can adjust to remove the leading / in our keys but this may be affecting other users or use cases

## Expected Result

To be able to use presigned_urls for s3 keys with leading "/"

## Actual Result

URL that was passed was modified resulting in a 403 error response from aws

## Reproduction Steps

```python
import requests
import boto3
s3 = boto3.client('s3')
bucket = "bucket"
key = "/key_with_leading_slash.txt"
presigned_url = s3.generate_presigned_url("get_object", Params={'Bucket': bucket, 'Key': key})
requests.get(presigned_url)
```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": "5.1.0"
  },
  "charset_normalizer": {
    "version": "3.1.0"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.4"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.9.6"
  },
  "platform": {
    "release": "10",
    "system": "Windows"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.32.1"
  },
  "system_ssl": {
    "version": "101010bf"
  },
  "urllib3": {
    "version": "1.26.15"
  },
  "using_pyopenssl": false
}
```

typhon8:
## Summary
Fixes issue #6711 where S3 presigned URLs with keys starting with '/' were incorrectly modified, breaking URL signatures.
URLs like `https://bucket.s3.amazonaws.com//key_name` now correctly preserve `//key_name` in the path.

## Changes
- Remove URL manipulation that collapsed leading slashes in `request_url()` method
- Add test to verify leading slashes are preserved for S3-style URLs

## Reproduction
```python
import requests

# S3 presigned URL with key starting with '/'
url = "https://bucket.s3.amazonaws.com//key_with_leading_slash.txt"
r = requests.get(url)  # Previously broken, now works correctly
Fixes #6711

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -545,8 +545,6 @@ class HTTPAdapter(BaseAdapter):
             using_socks_proxy = proxy_scheme.startswith("socks")
 
         url = request.path_url
-        if url.startswith("//"):  # Don't confuse urllib3
-            url = f"/{url.lstrip('/')}"
 
         if is_proxied_http_request and not using_socks_proxy:
             url = urldefragauth(request.url)
--- a/tests/test_adapters.py
+++ b/tests/test_adapters.py
@@ -1,8 +1,14 @@
 import requests.adapters
 
 
-def test_request_url_trims_leading_path_separators():
-    """See also https://github.com/psf/requests/issues/6643."""
+def test_request_url_preserves_leading_path_separators():
+    """See also https://github.com/psf/requests/issues/6711.
+    
+    S3 presigned URLs with keys starting with '/' produce paths like
+    '//key_name'. We should preserve leading slashes to avoid breaking signatures.
+    """
     a = requests.adapters.HTTPAdapter()
-    p = requests.Request(method="GET", url="http://127.0.0.1:10000//v:h").prepare()
-    assert "/v:h" == a.request_url(p, {})
+    p = requests.Request(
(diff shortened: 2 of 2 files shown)


nateprewitt:
Hi @typhon8, thanks for the PR. Unfortunately, this is breaking. We specifically changed the behavior to handle ambiguous paths failing in urllib3.

I've added https://github.com/urllib3/urllib3/pull/4952 to urllib3 which should fix the behavior in their next release. At that point we can revisit removing this from Requests. We do not want to remove the test though, that is required to still function before we make any changes.

nateprewitt (approved):
I made the test tweak to go back to the old format. It demonstrates the same thing without adding s3 semantics into the test suite. Once the tests finish, we can merge this and it will go out with 2.34.0.

PR #7365, opened 2026-04-15
Clarify decode_unicode behaviour in iter_{content,lines} docstrings

cjriches:
This corner case around automatic encoding detection is already mentioned in the "Advanced Usage" documentation, but deserves to be in the docstring too.

---

This is the documentation-only version of #7349, as requested.


nateprewitt (approved):
Thanks, @cjriches!

PR #7372, opened 2026-04-19
docs(quickstart): add practical timeout example

tboyila:
## What changed

This updates the Quickstart timeout section to add:

- a practical example using a normal timeout value
- clearer wording before the failure example
- explicit mention that Requests raises `Timeout`

## Why

The section recommends using timeouts in production, so adding a normal usage example makes the guidance easier for readers to apply in real code.


nateprewitt:
Thanks for the PR, @tboyila. I'm not sure we need this though, that already exists in the [Timeout](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts) section and the quickstart guide already shows how to use the feature.

PR #7373, opened 2026-04-19
docs(quickstart): clarify timeout applies to connect and read separately

### LINKED ISSUE #7350: [Docs] Clarify behavior of timeout parameter in requests.get
## Description

The current documentation for the `timeout` parameter in `requests.get()` does not clearly explain that it applies to both connection and read timeouts unless specified otherwise.

This can be confusing for users who expect it to behave as a total request timeout.

## Suggested Improvement

Add clarification in the docs that:
- A single float applies to both connect and read timeouts
- A tuple can be used to specify them separately

Example:
```python
requests.get(url, timeout=(connect_timeout, read_timeout))

CodeLine9:
The quickstart timeout note explains that `timeout` is not a time limit on the entire response download, but does not mention that a single float applies to both connect and read timeouts independently, or that a `(connect_timeout, read_timeout)` tuple can set them separately.

Added a brief clarification to the existing admonition block in `docs/user/quickstart.rst` with a cross-reference to the advanced timeouts section, which already documents this behavior in detail.

Fixes #7350

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -548,6 +548,11 @@ to hang indefinitely::
     received on the underlying socket for ``timeout`` seconds). If no timeout is specified explicitly, requests do
     not time out.
 
+    A single ``timeout`` value applies to both the **connect** and
+    **read** phases separately. To set them independently, pass a
+    ``(connect_timeout, read_timeout)`` tuple. See :ref:`timeouts`
+    for details.
+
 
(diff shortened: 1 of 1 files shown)


alokshukla631 (commented):
Nice clarification. This makes the quickstart timeout note a lot less ambiguous, and the link to the detailed timeouts section is helpful.

Issue #4771, opened 2018-08-26
ChunkedEncodingError isn't explained properly


ron819:
The [docs ](http://docs.python-requests.org/en/master/_modules/requests/exceptions/) says about `ChunkedEncodingError`:
```
class ChunkedEncodingError(RequestException):
    """The server declared chunked encoding but sent an invalid chunk."""
```

While this true it's highly confusing. What actually happens is that the server did not respond in time so the connection was closed. The solution is to retry the code and attempt again:

```
    def get(tries=5):
        for n in range(tries):
            try:
                GET call
                return whatever
            except ChunkedEncodingError as e:
                if n == tries - 1:
                    raise e
```

This has been asked many time over the internet including even here: https://github.com/requests/requests/issues/4484

I recommend adding an **optional**  argument for the `GET `call which will specify the tries number and that will be handled by the get call itself. This will avoid this problem and allow the user more control. 
meaning that the GET will return the exception only when the max retries has been reached.

If my suggestion is not accepted at least edit the docs and suggest the template of the solution to avoid the problem.

nateprewitt:
Hi @ron819, while the exception we're wrapping isn't the clearest, it does convey the connection has been broken with the server. The cause of that could be a multitude of things, but is likely the server forcing the connection closed.
```
requests.exceptions.ChunkedEncodingError: ("Connection broken: error(10054, '')", error(10054, ''))
```

I don't think it's clear that most users would want to retry in the case of a misbehaving server. Our API for Requests is currently frozen, so changes to the `get` method are unlikely unless there's significant oversight in functionality. Your example shows this should be easily implementable in 10-15 lines of code in your application, giving your complete control over how it will work. For those reasons, I don't think we'll be adding anything new to the Requests code base at this time.

PR #7374, opened 2026-04-19
docs(exceptions): clarify ChunkedEncodingError can result from transient network issues

CodeLine9:
The docstring for `ChunkedEncodingError` only mentions "server sent an invalid chunk" but this exception is also raised when the connection drops mid-stream during a chunked transfer (e.g. transient network interruption). The underlying `ProtocolError` from urllib3 covers both cases.

Added a clarifying note to the docstring so users know this error can result from transient issues and that retrying may succeed. This aligns with the discussion in #7341 and #4771.

Fixes #7341

### DIFF
--- a/src/requests/exceptions.py
+++ b/src/requests/exceptions.py
@@ -118,7 +118,12 @@ class InvalidProxyURL(InvalidURL):
 
 
 class ChunkedEncodingError(RequestException):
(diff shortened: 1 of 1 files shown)


alokshukla631 (commented):
This reads well to me. Calling out the transient-network case in the docstring should save people a trip through urllib3 internals.

PR #7380, opened 2026-04-20
feat(security): add SecurityWarning for verify=False with cert

loisccg:
Add a SecurityWarning when verify=False and cert are used together in
HTTPAdapter.send(). This combination disables server certificate
verification while still sending client credentials, which can be
misleading and insecure in production environments.

- Add SecurityWarning class to exceptions.py
- Emit the warning in adapters.py before any network activity
- Expose SecurityWarning in the public API via __init__.py

Add a test ensuring SecurityWarning is emitted when verify=False
and cert are used together in HTTPAdapter.send("http://...", verify=False, cert = ".").

No existing behavior is changed.

### DIFF
--- a/src/requests/__init__.py
+++ b/src/requests/__init__.py
@@ -172,6 +172,7 @@ from .exceptions import (
     Timeout,
     TooManyRedirects,
     URLRequired,
+    SecurityWarning,
 )
 from .models import PreparedRequest, Request, Response
 from .sessions import Session, session
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -44,6 +44,7 @@ from .exceptions import (
     ReadTimeout,
     RetryError,
     SSLError,
+    SecurityWarning,
 )
 from .models import Response
 from .structures import CaseInsensitiveDict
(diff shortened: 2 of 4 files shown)


nateprewitt:
Hi @loisccg, I think this is something we'd be unlikely to add. Urllib3 already raises a warning in this case and it's been a notable complaint from users for the majority of the existence of the library.

Once this is added, we're now going to raise two distinct warnings with mostly overlapping information. Can you provide more about how you encountered this and came to the conclusion a second warning was needed?

loisccg:
Thanks for the clarification, that makes sense regarding duplication.

My intention wasn’t to add a second warning, but rather to see if Requests could handle this case at its own level and potentially suppress the underlying urllib3 warning, so users only see one consistent warning.

From a UX perspective, this might make warning management clearer, since many users don’t interact with urllib3 directly and may find its warnings confusing or inconsistent.

That being said, i understand that it may represent a higher amount of work...

nateprewitt:
I think given the information is already conveyed through an established mechanism and addressing the warning is a quick one-liner, I'm not sure we'd make a change. That would be asking millions of developers to do work to update their code without any real value add in this case. Thanks for the suggestion though!

PR #7393, opened 2026-04-26
Ruff rule UP038 has been removed

DimitriPapadopoulos:
[Removed since 0.13.0](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/)

Fixes this warning while running ruff:
```
warning: The following rules have been removed and ignoring them has no effect:
    - UP038
```


nateprewitt (approved):
Thanks for the PR, @DimitriPapadopoulos!

PR #7394, opened 2026-04-26
PEP 639 compliance

DimitriPapadopoulos:
https://peps.python.org/pep-0639/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,12 +1,13 @@
 [build-system]
-requires = ["setuptools>=61.0"]
+requires = ["setuptools>=77.0.3"]
 build-backend = "setuptools.build_meta"
 
 [project]
 name = "requests"
 description = "Python HTTP for Humans."
 readme = "README.md"
-license = {text = "Apache-2.0"}
+license = "Apache-2.0"
+license-files = ["LICENSE", "NOTICE"]
 authors = [
     { name = "Kenneth Reitz", email = "me@kennethreitz.org" },
 ]
@@ -27,7 +28,6 @@ classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Environment :: Web Environment",
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: Apache Software License",
     "Natural Language :: English",
     "Operating System :: OS Independent",
     "Programming Language :: Python",
@@ -64,9 +64,6 @@ test = [
     "pytest>=3",
     "trustme",
 ]
-
-[tool.setuptools]
-license-files = ["LICENSE", "NOTICE"]
 
 [tool.setuptools.dynamic]
 version = {attr = "requests.__version__.__version__"}


nateprewitt:
Thanks for the PR, @DimitriPapadopoulos. This was an intentional decision recorded [here](https://github.com/psf/requests/pull/7012#discussion_r2748329758), we'll revisit this later.

DimitriPapadopoulos:
It doesn't make sense to keep compatibility with Setuptools < 77. From the [Setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#installation):
> You can install the latest version of `setuptools` using [pip](https://pypi.org/project/pip):
> ```
> pip install --upgrade setuptools[core]
> ```
> Most of the times, however, you don’t have to…
> 
> Instead, when creating new Python packages, it is recommended to use a command line tool called [build](https://pypi.org/project/build). This tool will automatically download `setuptools` and any other build-time dependencies that your project might have. You just need to specify them in a `pyproject.toml` file at the root of your package, as indicated in the [following section](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use).
> 
> You can also [install build](https://build.pypa.io/en/stable/installation.html) using [pip](https://pypi.org/project/pip):
> ```
> pip install --upgrade build
> ```
> This will allow you to run the command: `python -m build`.

nateprewitt:
I think you may be looking at the issue through a narrow lens. Anyone who's installing Requests with pinned dependencies or a static local install can't necessarily "just update" or change their install tooling immediately.

This change is entirely cosmetic for PyPI, there's no real end-user benefit. I don't think that warrants the breakage risk or constraining users to only build tools released in the last year. Almost every other major project has deferred this pin move and we're not interested in pioneering that unnecessarily.

DimitriPapadopoulos:
I wasn't aware of major projects that have deferred this update, but I must admit I'm more interested in scientific Python, where the community might be updating more aggressively.

nateprewitt:
I would take a look at major scientific Python projects because they largely share the same build system ([meson](https://github.com/mesonbuild/meson)) which has their [setuptools pin](https://github.com/mesonbuild/meson/blob/master/pyproject.toml) even lower than ours. Requests is not the outlier here for a project of this size.

DimitriPapadopoulos:
I have already taken a look at major scientific Python projects such as [NumPy](https://github.com/numpy/numpy/blob/bab58ec3127f95d7e63e305711bbd19f57cd0907/pyproject.toml#L51) or [scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/b032031a8390b8b0369bd4e3715a88b18c900287/pyproject.toml#L16).

DimitriPapadopoulos:
But it's true that PEP 639 is especially interesting for projects with multiple licences (using SPDX expressions typically involving `AND` or `OR`).

PR #7395, opened 2026-04-26
Fix typos

DimitriPapadopoulos:
Fix typos found by [typos](https://github.com/crate-ci/typos).

Also, the plural of **status** is **status** or **statuses** - not [**stati**](https://www.oed.com/search/dictionary/?q=stati):
https://github.com/psf/requests/blob/f43f750ee10addbaa13901c1017631b4d70fd6e3/src/requests/models.py#L71-L73
I'd like to fix this too, but isn't `REDIRECT_STATI` pat of the API?

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -401,7 +401,7 @@ class HTTPAdapter(BaseAdapter):
         alter the other keys to ensure the desired behaviour.
 
         :param request:
-            The PreparedReqest being sent over the connection.
+            The PreparedRequest being sent over the connection.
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
We generally don't accept typo PRs but this appears to be the full set of valid typos in the repo running codespell and typos. This seems fine as an exception. Thanks @DimitriPapadopoulos!

nateprewitt:
And that's correct, `REDIRECT_STATI` is part of the public API, so we can't change it now.

PR #7396, opened 2026-04-26
Enforce Ruff-specific rules (RUF)

DimitriPapadopoulos:


nateprewitt:
Thanks for the PR, @DimitriPapadopoulos. We also intentionally did not opt into the RUF specific rules, the set we currently have is what's intended.

PR #7397, opened 2026-04-26
Use `enumerate()` for index variable in `for` loop

DimitriPapadopoulos:


nateprewitt:
Thanks for the PR, @DimitriPapadopoulos. We typically won't accept style nit PRs (especially in tests) because they lead to waves of tiny PRs which have overhead to review and triage. I think we'll pass on this one since the test is correct and this doesn't materially change anything beyond collapsing a couple lines.

PR #7412, opened 2026-05-05
fix: handle None proxies in merge_environment_settings

### LINKED ISSUE #6699: AttributeError: 'NoneType' object has no attribute 'setdefault' on environment proxies
AttributeError: 'NoneType' object has no attribute 'setdefault' on environment proxies

It appears there is a small bug in ‎`Session.merge_environment_settings` where the proxies variable can be `None` but setdefault is called on it anyway when env_proxies are present, causing an AttributeError to be thrown:

https://github.com/psf/requests/blob/main/src/requests/sessions.py#L759-L762

zons-zhaozhy:
Fix AttributeError when proxies is None and trust_env is True.

In merge_environment_settings, when proxies=None and environment
proxy settings exist, calling setdefault on None raises AttributeError.

Fix: initialize proxies to an empty dict when it is None, ensuring
environment proxies are always properly merged.

Fixes #6699

### DIFF
--- a/src/requests/sessions.py
+++ b/src/requests/sessions.py
@@ -839,9 +839,10 @@ class Session(SessionRedirectMixin):
             # Set environment's proxies.
             no_proxy = proxies.get("no_proxy") if proxies is not None else None
             env_proxies = get_environ_proxies(url, no_proxy=no_proxy)
-            if proxies is not None:
-                for k, v in env_proxies.items():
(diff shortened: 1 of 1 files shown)


zons-zhaozhy:
Closing this PR to revisit the fix with additional testing and validation. I'll submit a new PR after thorough verification.

PR #7414, opened 2026-05-05
Fix HTTPDigestAuth non-latin credentials encoding

mitre88:
When HTTPDigestAuth credentials are passed as bytes (e.g. encoded UTF-8), they were being used directly in the digest header without decoding, resulting in headers like `Digest username="b'Ond\xc5\x99ej'"`.

This fix decodes bytes username/password to strings before using them in the digest A1 computation and the username header field.

Fixes #6102

### DIFF
--- a/src/requests/auth.py
+++ b/src/requests/auth.py
@@ -218,7 +218,16 @@ class HTTPDigestAuth(AuthBase):
         if p_parsed.query:
             path += f"?{p_parsed.query}"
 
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @mitre88,

I don't believe this is the right fix. It's broken for `latin-1` which is our standard assumption for bytes input. There are other open PRs that already address this so we'll close this as a duplicate.

PR #7424, opened 2026-05-11
Fix HTTPDigestAuth failing on non-ASCII credentials

CatfishGG:
## Summary

When `HTTPDigestAuth` credentials contain non-ASCII characters (e.g. Czech username `Ondřej`), authentication fails because the digest header encodes the bytes representation rather than the decoded string.

## Root Cause

In `build_digest_header()`, the A1 string was constructed using an f-string directly with `self.username` and `self.password`:

```python
A1 = f"{self.username}:{realm}:{self.password}"
```

When these are `bytes` objects (e.g. `'Ondřej'.encode('utf-8')`), Python's f-string formatting embeds the `repr()` of bytes, producing:

```
Digest username="b'Ond\xc5\x99ej'"
```

instead of the correct:

```
Digest username="Ondřej"
```

This causes the server to reject the credentials as the username doesn't match.

## Fix

Wrap `self.username` and `self.password` with `to_native_string(..., 'utf-8')` when building A1. This function (already imported in auth.py from `_internal_utils`) properly decodes bytes to UTF-8 strings, consistent with how `hash_utf8` handles its inputs.

```python
A1 = f"{to_native_string(self.username, 'utf-8')}:{realm}:{to_native_string(self.password, 'utf-8')}"
```

## Testing

- All 336 existing tests pass (full test suite)
- 6 digest-specific tests confirmed green
- Manual verification: both string (`'Ondřej'`) and bytes (`'Ondřej'.encode('utf-8')`) inputs now produce identical, correct A1 hashes

## Fixes

Fixes #6102

### DIFF
--- a/src/requests/auth.py
+++ b/src/requests/auth.py
@@ -218,7 +218,7 @@ class HTTPDigestAuth(AuthBase):
         if p_parsed.query:
             path += f"?{p_parsed.query}"
 
-        A1 = f"{self.username}:{realm}:{self.password}"
+        A1 = f"{to_native_string(self.username, "utf-8")}:{realm}:{to_native_string(self.password, "utf-8")}"
         A2 = f"{method}:{path}"
 
         HA1 = hash_utf8(A1)


nateprewitt:
Hi @CatfishGG, I think this unfortunately does not address #6102. This breaks the default case for latin-1. There's already other PRs open with alternative approaches so we'll close this as a duplicate.

PR #7425, opened 2026-05-11
sessions: fix hooks type

bastimeyer:
The `hook` keyword in `Session.request()` should be of type `HooksInputType | None`, not `HooksType`.

This is already correctly defined in `models.Request.__init__()`:
- https://github.com/psf/requests/blob/dc9dbdfb3434c6e58d48fd102f93e5342308817e/src/requests/models.py#L283-L333

But `sessions.Session.request()` and the associated `_types.BaseRequestKwargs` are wrong:
- https://github.com/psf/requests/blob/dc9dbdfb3434c6e58d48fd102f93e5342308817e/src/requests/sessions.py#L557-L618
- https://github.com/psf/requests/blob/dc9dbdfb3434c6e58d48fd102f93e5342308817e/src/requests/_types.py#L145-L156

### DIFF
--- a/src/requests/_types.py
+++ b/src/requests/_types.py
@@ -150,7 +150,7 @@ if TYPE_CHECKING:
         timeout: TimeoutType
         allow_redirects: bool
         proxies: dict[str, str] | None
-        hooks: HooksType
+        hooks: HooksInputType | None
         stream: bool | None
         verify: VerifyType | None
         cert: CertType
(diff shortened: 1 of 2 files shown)


nateprewitt:
I'm on mobile but this _seems_ right based on what you described. I'll take a look at some point today. Thanks, @bastimeyer!

nateprewitt (approved):

PR #7428, opened 2026-05-11
v2.34.0

nateprewitt:
2.34.0 (2026-05-11)
-------------------

**Announcements**
- Requests 2.34.0 introduces inline types, replacing those provided by
  typeshed. Public API types should be fully compatible with mypy, pyright,
  and ty. **We believe types are comprehensive but if you find issues, please
  report them to the [pinned tracking issue](https://github.com/psf/requests/issues/7271).**

  Special thanks to @bastimeyer, @cthoyt, @edgarrmondragon, and @srittau for
  helping review and test the types ahead of the release. (#7272)

**Improvements**
- Digest Auth hashing algorithms have added `usedforsecurity=False` to clarify
  security considerations. (#7310)
- Requests added support for Python 3.15 based on beta1. Downstream projects
  should be able to start testing prior to its release in October. (#7422)
- Requests added support for Python 3.14t. (#7419)

**Bugfixes**
- ``Response.history`` no longer contains a reference to itself, preventing
  accidental looping when traversing the history list. (#7328)
- Requests no longer performs greedy matching on no_proxy domains. The
  proxy_bypass implementation has been updated with CPython's fix from
  bpo-39057. (#7427)
- Requests no longer incorrectly strips duplicate leading slashes in
  URI paths. This should address user issues with specific presigned
  URLs. Note the full fix requires urllib3 2.7.0+. (#7315)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,19 +6,31 @@ dev
 
 - \[Short description of non-trivial change.\]
 
-2.34.0 (2026-05-??)
+2.34.0 (2026-05-11)
 -------------------
 
-**Improvements**
-* Requests 2.34.0 introduces inline types, replacing those provided by
+**Announcements**
+- Requests 2.34.0 introduces inline types, replacing those provided by
   typeshed. Public API types should be fully compatible with mypy, pyright,
-  and ty. (#7272)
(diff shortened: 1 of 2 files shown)


cthoyt:
Thanks @nateprewitt !!

sigmavirus24 (approved):

PR #7430, opened 2026-05-11
docs: clarify tuple-based timeouts in advanced usage

Jags-08:
Hello! While reviewing advanced usage for timeouts, I noticed the distinction between connection and read timeouts in a tuple could be slightly more explicit to avoid ReadTimeout confusion. This fast doc patch adds a brief clarification block. Happy to adjust wording!

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -1135,3 +1135,4 @@ coffee.
 
 .. _`wall clock`: https://wiki.php.net/rfc/max_execution_wall_time
 .. _`connect()`: https://linux.die.net/man/2/connect
+`n.. note:: When using a tuple for timeouts, it is structured as (connect_timeout, read_timeout).


nateprewitt:
Hi @Jags-08, I'm not sure this is a change we'd add to the docs. They already clarify this in the paragraph above:

```
 If you specify a single value for the timeout, like this::

    r = requests.get('https://github.com', timeout=5)

The timeout value will be applied to both the ``connect`` and the ``read``
timeouts. Specify a tuple if you would like to set the values separately::

    r = requests.get('https://github.com', timeout=(3.05, 27))
```

as well as the API docstrings:

```
    :param timeout: (optional) How many seconds to wait for the server to send data
        before giving up, as a float, or a :ref:`(connect timeout, read
        timeout) <timeouts>` tuple.
```

PR #7431, opened 2026-05-11
Fix mutability issues with headers input types

nateprewitt:
We already got a helpful email, lack of header mutability in the typing contract is creating friction for some code bases. I think we'd started to go down this route but it got lost in the cleanup. I'm going to stage this as a known issue for a 2.34.1 but we'll likely wait a day or two before cutting another release to see what other feedback we missed.

### DIFF
--- a/src/requests/_types.py
+++ b/src/requests/_types.py
@@ -109,8 +109,7 @@ if TYPE_CHECKING:
         bytes | str | Iterable[bytes | str] | SupportsRead[bytes | str] | None
     )
 
-    HeadersType: TypeAlias = CaseInsensitiveDict[str] | Mapping[str, str | bytes]
-    HeadersUpdateType: TypeAlias = Mapping[str, str | bytes | None]
+    HeadersType: TypeAlias = MutableMapping[str, str | bytes] | None
 
     CookiesType: TypeAlias = RequestsCookieJar | Mapping[str, str]
 
@@ -145,7 +144,7 @@ if TYPE_CHECKING:
     # TypedDicts for Unpack kwargs (PEP 692)
 
     class BaseRequestKwargs(TypedDict, total=False):
-        headers: Mapping[str, str | bytes] | None
+        headers: HeadersType
         cookies: RequestsCookieJar | CookieJar | dict[str, str] | None
         files: FilesType
         auth: AuthType
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -83,6 +83,7 @@ from .utils import (
 )
 
 if TYPE_CHECKING:
+    from collections.abc import MutableMapping
     from http.cookiejar import CookieJar
 
     from typing_extensions import Self
@@ -310,7 +311,7 @@ class Request(RequestHooksMixin):
 
     method: str | None
     url: _t.UriType | None
-    headers: CaseInsensitiveDict[str] | Mapping[str, str | bytes] | None
+    headers: MutableMapping[str, str | bytes]
     files: _t.FilesType
     data: _t.DataType
     json: _t.JsonType
@@ -322,7 +323,7 @@ class Request(RequestHooksMixin):
         self,
         method: str | None = None,
         url: _t.UriType | None = None,
-        headers: Mapping[str, str | bytes] | None = None,
+        headers: _t.HeadersType = None,
         files: _t.FilesType = None,
         data: _t.DataType = None,
         params: _t.ParamsType = None,
--- a/src/requests/sessions.py
+++ b/src/requests/sessions.py
@@ -560,7 +560,7 @@ class Session(SessionRedirectMixin):
         url: _t.UriType,
         params: _t.ParamsType = None,
         data: _t.DataType = None,
-        headers: Mapping[str, str | bytes] | None = None,
+        headers: _t.HeadersType = None,
         cookies: RequestsCookieJar | CookieJar | dict[str, str] | None = None,
         files: _t.FilesType = None,
         auth: _t.AuthType = None,


src/requests/models.py
```diff
@@ -310,7 +311,7 @@ class Request(RequestHooksMixin):
 
     method: str | None
     url: _t.UriType | None
-    headers: CaseInsensitiveDict[str] | Mapping[str, str | bytes] | None
+    headers: MutableMapping[str, str | bytes]
```
sigmavirus24, line 314: Why not headerstype here too?

nateprewitt: None was wrong. We always fallback to {} when you pass None into the constructor. The *Type aliases are all mapping input from the public API, this is representing what we store.

Otherwise you need to both validate  headers is not None and Mutable. I made a similar change for status_code on Response.

nateprewitt: I guess as a tangent to that, `| None` being baked into the types was an opinion. I don't feel super strongly about that, it was just for terseness and keeping the `None` requirement grouped. We could do `HeadersType | None` everywhere it's used now. That applies to most of the *Types.

bastimeyer:
This has caused the same issue as #7426

```py
# foo.py
import requests
requests.get("http://localhost", headers={"X-Foo": "bar"})
```
```sh
$ ty check foo.py
All checks passed!
```

```py
# foo.py
import requests
headers={"X-Foo": "bar"}
requests.get("http://localhost", headers=headers)
```
```sh
$ ty check foo.py
error[invalid-argument-type]: Argument to function `get` is incorrect
 --> foo.py:4:34
  |
4 | requests.get("http://localhost", headers=headers)
  |                                  ^^^^^^^^^^^^^^^ Expected `MutableMapping[str, str | bytes] | None`, found `dict[str, str]`
  |
info: type `dict[str, str]` is not assignable to any element of the union `MutableMapping[str, str | bytes] | None`
info: ├── element `bytes` of union `str | bytes` is not assignable to `str`
info: └── ... omitted 1 union element without additional context
info: Function defined here
  --> /path/to/venv/lib/python3.14/site-packages/requests/api.py:74:5
   |
74 |   def get(
   |  _____^^^-
75 | |     url: _t.UriType, params: _t.ParamsType = None, **kwargs: Unpack[_t.GetKwargs]
76 | | ) -> Response:
   | |_- Parameter declared here
   |

Found 1 diagnostic
```

nateprewitt:
There's some tension here in usage that I'm a bit at a loss for how to address. I saw [this](https://results.pre-commit.ci/run/github/1446467/1778702053.hj4Eo_b5Q6K92VNAKCosPw) in `pip` right before you left this comment which is the same failure mode.

`Mapping` works for the `dict[str, str]` case for the same reasons as #7426 and #7434. `Requests.headers.update()` causes issues for type checkers though. `MutableMapping` covers the modification case, but introduces the covariance issue since header values can be either `str` or `bytes`.

Once we pass `prepare()` it doesn't matter because it all gets turned into a `CaseInsensitiveDict` but the modification window between `Request` creation and `prepare()` is ambiguous.

nateprewitt:
Typeshed got around this issue by just labeling [headers on Request](https://github.com/python/typeshed/blob/536aeeb4396a10a8ba67d15a7acdffdd69444e1f/stubs/requests/requests/models.pyi#L65) as `Incomplete` which is just an alias for `Any`.

We either need to lie about what we accept with `Mapping` to support the input case or lie about what we store to fix the modification case. There's a runtime change of wrapping everything in a dict on intake, but that's certainly going to break someone's custom dict implementation.

The input case is probably where we need to be overly permissive because the new failure is significantly more common than the `update()` call. I'll look at a partial revert of this.

src/requests/_types.py
```diff
@@ -109,8 +109,7 @@ class _ValidatedRequest(PreparedRequest):
         bytes | str | Iterable[bytes | str] | SupportsRead[bytes | str] | None
     )
```
sdm0p, line 110: ⚠️ **HIGH** — *test_coverage*  
**Confidence:** 70%  

New logic without tests: The `HeadersType` has been changed to `MutableMapping[str, str | bytes] | None` without adding corresponding tests.

**Suggested fix:**
```
Add tests to cover the new `HeadersType` logic.
```

src/requests/models.py
```diff
@@ -322,7 +323,7 @@ def __init__(
         self,
         method: str | None = None,
```
sdm0p, line 324: 📌 **MEDIUM** — *test_coverage*  
**Confidence:** 70%  

Modified behaviour without updated assertions: The `Request` class's `__init__` method now uses the new `HeadersType` without updating the corresponding tests.

**Suggested fix:**
```
Update the tests for the `Request` class's `__init__` method to cover the new `HeadersType` logic.
```

sdm0p:
## OpenCodeReview — 🔄 REQUEST_CHANGES

**Score:** 7.5/10  
**Summary:** Changes requested: 1 high, 1 medium — high-severity issues should be addressed.

src/requests/_types.py
sdm0p, line 110: ⚠️ **HIGH** — *test_coverage*  
**Confidence:** 80%  

New logic for handling MutableMapping in Request.headers is not tested.

**Suggested fix:**
```
Add test cases to cover the new logic for handling MutableMapping in Request.headers.
```

sdm0p:
## OpenCodeReview — 🔄 REQUEST_CHANGES

**Score:** 8.0/10  
**Summary:** Changes requested: 1 high — high-severity issues should be addressed.

sdm0p:
## OpenCodeReview — ✅ APPROVE

**Score:** 10.0/10  
**Summary:** No issues found — PR looks clean.

PR #7433, opened 2026-05-12
Fix `prepare_body` stream detection for `__getattr__`-based file wrappers

### LINKED ISSUE #7432: `prepare_body` stream detection regression
Requests 2.34.0 changed the stream detection in PreparedRequest.prepare_body() from `hasattr(data, "__iter__")` to `isinstance(data, Iterable)`. This breaks file-like wrappers that expose `__iter__/read/tell/seek` via `__getattr__ delegation` (e.g. tqdm.wrapattr), because `isinstance(..., Iterable)` only checks the class MRO, not dynamic attribute delegation. As a result, `_body_position` is never recorded and the body cannot be rewound on 307/308 redirects.
## Expected Result

Upload completes successfully through the redirect, as it did in 2.33.1.

## Actual Result

The redirected request sends an empty/consumed body. The server receives no data and the request times out:
```
requests.exceptions.ReadTimeout: HTTPSConnectionPool(...): Read timed out. (read timeout=30)
```

## Reproduction Steps

```python
import os
import requests
from tqdm import tqdm

with open("testfile.txt", "wb") as f:
    f.write(b"hello world " * 1000)

with open("testfile.txt", "rb") as f:
    with tqdm.wrapattr(f, "read", total=os.path.getsize("testfile.txt")) as wrapped:
        # Any endpoint that 307 redirects to a presigned S3 URL
        requests.put("https://your-api.example.com/upload", data=wrapped, timeout=30)

```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "3.4.7"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.13"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.13.5"
  },
  "platform": {
    "release": "24.6.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.34.0"
  },
  "system_ssl": {
    "version": "30000120"
  },
  "urllib3": {
    "version": "2.5.0"
  },
  "using_pyopenssl": false
}
```

k223kim:
Resolves #7432 .

Add `hasattr(data, "__iter__")` as a fallback stream detector:
```python
if (
    isinstance(data, Iterable) or hasattr(data, "__iter__")
) and not isinstance(data, (str, bytes, list, tuple, Mapping)):
```
This catches file-like objects that proxy their interface through `__getattr__`, restoring the 2.33.1 behavior without reverting the isinstance(Iterable) modernization for standard iterables.

### DIFF
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -596,9 +596,7 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             if not isinstance(body, bytes):
                 body = body.encode("utf-8")
 
-        # Also check hasattr("read") for wrappers that proxy file
-        # methods via __getattr__ (e.g. tqdm.wrapattr)
-        if (isinstance(data, Iterable) or hasattr(data, "read")) and not isinstance(
+        if (isinstance(data, Iterable) or hasattr(data, "__iter__")) and not isinstance(
             data, (str, bytes, list, tuple, Mapping)
         ):
             try:
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -596,9 +596,9 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             if not isinstance(body, bytes):
                 body = body.encode("utf-8")
 
-        if (isinstance(data, Iterable) or hasattr(data, "__iter__")) and not isinstance(
-            data, (str, bytes, list, tuple, Mapping)
-        ):
+        # data that proxies attributes to underlying objects needs hasattr
+        is_iterable = isinstance(data, Iterable) or hasattr(data, "__iter__")
+        if is_iterable and not isinstance(data, (str, bytes, list, tuple, Mapping)):
             try:
                 length = super_len(data)
             except (TypeError, AttributeError, UnsupportedOperation):
--- a/src/requests/models.py
+++ b/src/requests/models.py
@@ -596,7 +596,9 @@ class PreparedRequest(RequestEncodingMixin, RequestHooksMixin):
             if not isinstance(body, bytes):
                 body = body.encode("utf-8")
 
-        if isinstance(data, Iterable) and not isinstance(
+        # Also check hasattr("read") for wrappers that proxy file
+        # methods via __getattr__ (e.g. tqdm.wrapattr)
+        if (isinstance(data, Iterable) or hasattr(data, "read")) and not isinstance(
             data, (str, bytes, list, tuple, Mapping)
         ):
             try:
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -81,21 +81,6 @@ except AttributeError:
     HAS_PYOPENSSL = False
 
 
-class AttrProxyFileWrapper:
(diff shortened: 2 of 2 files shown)


nateprewitt:
Hi @k223kim,

This PR has quite a few oddities in it, did you happen to use an LLM to generate it? There are unrelated tests and the change itself is wrong (at least for the issue you reported).

k223kim:
Hey @nateprewitt , thanks for the review!
I did use llm to get help with mimicking the tqdm behavior on the test - but not for the fix itself 😄 The test ensures that correct objects are detected as a stream. + I believe this did fix the issue that I was having with tqdm wrapper + redirect with the latest version of `requests`. Would you explain how this fix is wrong? Happy to adjust as needed.

k223kim:
Perhaps - I can be more precise with the if-statement and just restoring it to `hasattr(data, "__iter__")`. 
i.e.
```
if (
    isinstance(data, Iterable) or hasattr(data, "__iter__")
) and not isinstance(data, (str, bytes, list, tuple, Mapping)):
```
If this sounds better, happy to adjust test accordingly.

nateprewitt:
Yep, it should be `__iter__` instead of `read`, otherwise we're adding new semantics to the check that didn't exist in 2.33.1 either.

The tests are engraining a lot of tqdm/use case specific semantics into the test suite that I'm not sure we want. While this may work, I don't really want to call it a "supported pattern" because that tends to lead to other areas that didn't support this before getting PRs to "fix" support.

I think it's unlikely this changes again in the future, the `__iter__` check was there for ~12 years, but if we want to be really safe, a single test like this is probably sufficient:

```python
 def test_getattr_proxy_stream_follows_redirect(self, httpbin):
     """Ensure stream wrappers that don't implement __iter__ directly are still detected."""
 
     class AttrProxy:
         def __init__(self):
             self._file = io.BytesIO(b"data")
 
         def __getattr__(self, name):
             return getattr(self._file, name)
 
     r = requests.post(httpbin("redirect-to?url=/post&status_code=307"), data=AttrProxy())
     assert r.json()["data"] == "data"
```

k223kim:
@nateprewitt Makes sense. Adjusted the PR 😄 Would appreciate it if you can take a look!

src/requests/models.py
```diff
@@ -596,7 +596,7 @@ def prepare_body(
             if not isinstance(body, bytes):
                 body = body.encode("utf-8")
 
-        if isinstance(data, Iterable) and not isinstance(
+        if (isinstance(data, Iterable) or hasattr(data, "__iter__")) and not isinstance(
```
nateprewitt:
```suggestion
        # data that proxies attributes to underlying objects needs hasattr
        is_iterable = isinstance(data, Iterable) or hasattr(data, "__iter__")
        if is_iterable and not isinstance(
```

nateprewitt (approved):
One minor tweak for readability, but otherwise this looks good. Thanks @k223kim!

PR #7436, opened 2026-05-12
Update JsonType containers to read-based collections

### LINKED ISSUE #7434: Inline annotations in 2.34.0 cause mypy error for Request json payloads
requests version 2.34.0 added inline type annotations. This is causing some issues with mypy, specifically building a `Request` with a `json` argument.

The MR (https://github.com/psf/requests/pull/7272) that introduced the feature  notes that `mypy` support is not currently a goal, but I think it's worth sharing this issue.

## Expected Result
Passing a JSON-compatible dictionary to the `json` argument of `requests.Request` should type-check successfully.

For example, a value typed as:

```python
dict[str, None | bool | int | float | str]
```
or the narrower:
```python
dict[str, str]
```
should be accepted as a valid JSON payload.

## Actual Result
`mypy` reports an incompatible argument type error when a dictionary is passed to the json argument.

## Reproduction Steps
Create a file named `test.py`:

```python
from requests import Request

payload: dict[str, str] = {"test": "test"}

request = Request(json=payload)
```
Then run mypy (used 2.1.0) against the python code:
```bash
python -m mypy test.py
test.py:5: error: Argument "json" to "Request" has incompatible type "dict[str, str]"; expected "JsonType"  [arg-type]
test.py:5: note: "dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
test.py:5: note: Consider using "Mapping" instead, which is covariant in the value type
Found 1 error in 1 file (checked 1 source file)
```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "3.4.7"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.14"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.13.13"
  },
  "platform": {
    "release": "24.6.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.34.0"
  },
  "system_ssl": {
    "version": "30600020"
  },
  "urllib3": {
    "version": "2.7.0"
  },
  "using_pyopenssl": false
}
```

nateprewitt:
The current definition for `JsonType` is constrained to `dict` for map-like objects but that introduces issues when input is explicitly typed.

The type checker is able to infer typing correctly with:

```python
payload = {"test": "test"}
r = requests.post(url, json=payload)
```

but not
```python
payload: dict[str, str] = {"test": "test"}
r = requests.post(url, json=payload)
```

This is due to the varied Value typing on `JsonType` and the mutability of `dict`. Changing it to `Mapping` which only exposes read interfacea allows us to check covariant value types. This works because the `json` input is never touched. It's taken from input and passed directly to `json.dumps` resulting in bytes. There's no access point in the pipeline to modify it later.

We're going to start with this expansion of the type to see if it works broadly. The goal is to avoid `Any` for JsonType if at all possible.

### DIFF
--- a/src/requests/_types.py
+++ b/src/requests/_types.py
@@ -9,7 +9,7 @@ by external code.
 
 from __future__ import annotations
 
-from collections.abc import Callable, Iterable, Mapping, MutableMapping
+from collections.abc import Callable, Iterable, Mapping, MutableMapping, Sequence
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -139,7 +139,13 @@ if TYPE_CHECKING:
     VerifyType: TypeAlias = bool | str
     CertType: TypeAlias = str | tuple[str, str] | None
     JsonType: TypeAlias = (
-        None | bool | int | float | str | list["JsonType"] | Mapping[str, "JsonType"]
+        None
+        | bool
+        | int
+        | float
+        | str
+        | Sequence["JsonType"]
+        | Mapping[str, "JsonType"]
     )
 
     # TypedDicts for Unpack kwargs (PEP 692)
--- a/src/requests/_types.py
+++ b/src/requests/_types.py
@@ -139,7 +139,7 @@ if TYPE_CHECKING:
     VerifyType: TypeAlias = bool | str
     CertType: TypeAlias = str | tuple[str, str] | None
     JsonType: TypeAlias = (
-        None | bool | int | float | str | list["JsonType"] | dict[str, "JsonType"]
+        None | bool | int | float | str | list["JsonType"] | Mapping[str, "JsonType"]
     )
 
     # TypedDicts for Unpack kwargs (PEP 692)


src/requests/_types.py
```diff
@@ -139,7 +139,7 @@ class _ValidatedRequest(PreparedRequest):
     VerifyType: TypeAlias = bool | str
     CertType: TypeAlias = str | tuple[str, str] | None
     JsonType: TypeAlias = (
-        None | bool | int | float | str | list["JsonType"] | dict[str, "JsonType"]
+        None | bool | int | float | str | list["JsonType"] | Mapping[str, "JsonType"]
```
JamMarHer, line 142: Consider making the following update as well:

```suggestion
        None | bool | int | float | str | Sequence["JsonType"] | Mapping[str, "JsonType"]
```

JamMarHer: Just saw your other comment https://github.com/psf/requests/issues/7434#issuecomment-4433151990. 🙇🏼

nateprewitt: Thanks @JamMarHer, I think that was already pushed in [1be9171](https://github.com/psf/requests/pull/7436/commits/1be917108874d28ae47cc311297d30790302b648).

sigmavirus24 (approved):

risicle:
The only issue with this is that the default `JSONEncoder`, used by `requests`, does _not_ cope with arbitrary `Mapping`s and `Sequence`s. So this will happily allow input that will then fail at runtime, e.g. https://github.com/psf/requests/issues/1380

PR #7448, opened 2026-05-16
fix(digest-auth): support non-latin characters in username and password

MD-Mushfiqur123:
## Description

HTTPDigestAuth fails when username or password contain non-latin characters (e.g. `'Ondřej'`, `'Сергей_Ласточкин'`).

### Root Cause

In `build_digest_header()`, `self.username` and `self.password` are used directly in f-strings for both the hash calculation (A1) and the HTTP header value. When users pass `bytes` as a workaround for non-ASCII support:

```python
HTTPDigestAuth('Ondřej'.encode('utf-8'), 'heslíčko')
```

The f-string interpolation produces `b'Ond\xc5\x99ej'` (the bytes repr) instead of the actual decoded string, resulting in incorrect hashes and malformed headers like:

```
Digest username="b'Ond\xc5\x99ej'"
```

### Fix

Normalize `username` and `password` to `str` at the start of `build_digest_header()` — if they're `bytes`, decode them using UTF-8. The normalized string variables are then used consistently for both the digest hash computation and the `Authorization` header.

Fixes #6102

### DIFF
--- a/src/requests/auth.py
+++ b/src/requests/auth.py
@@ -166,6 +166,19 @@ class HTTPDigestAuth(AuthBase):
         opaque = self._thread_local.chal.get("opaque")
         hash_utf8 = None
 
+        # Normalize username and password to strings. If they're passed as bytes
+        # (e.g., to support non-latin characters), decode them using UTF-8.
+        # Using bytes in f-strings or string formatting would produce incorrect
+        # output like ``b'...'`` in both the hash calculation and the header.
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @MD-Mushfiqur123, I'm not sure this PR actually addresses the issue. There's also other PRs already open so I'm going to close this as a duplicate. Please make sure to check open PRs before creating new ones. Thanks!

PR #7449, opened 2026-05-16
Fix JsonType to accept Mapping[str, Any] instead of recursive Mapping[str, "JsonType"]

### LINKED ISSUE #7443: mypy warns about invalid types for json argument
I have warnings about invalid types for the json argument, that didn't occur when using types-requests.

The changes in v2.34.1 changed the errors, but did not fix them.

```
typecheck.py:6: error: Argument "json" to "post" has incompatible type "dict[str, Collection[str]]"; expected "JsonType"  [arg-type]
typecheck.py:9: error: Argument "json" to "post" has incompatible type "dict[str, object]"; expected "JsonType"  [arg-type]
```

## Reproduction Steps

```python
import requests

def fn(d: dict[str, str]) -> None:
    j = {"foo": d, "bar": "hi"}
    requests.post("https://example.com", json=j)

    k = {"foo": d, "bool": True}
    requests.post("https://example.com", json=k)

```

It seems ok if a literal is used without the intermediate variable:

```
    requests.post("https://example.com", json={"foo": d, "bar": "hi"})
    requests.post("https://example.com", json={"foo": d, "bool": True})
```

## System Information

```json
{
  "chardet": {
    "version": null
  },
  "charset_normalizer": {
    "version": "3.4.7"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.15"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.13.5"
  },
  "platform": {
    "release": "6.12.74+deb13+1-amd64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.34.1"
  },
  "system_ssl": {
    "version": "30500050"
  },
  "urllib3": {
    "version": "2.7.0"
  },
  "using_pyopenssl": false
}
```

driphtyio:
Fixes #7443

## Problem
The recursive `JsonType` definition used `Mapping[str, "JsonType"]`, which causes mypy to reject valid dict subtypes like `dict[str, object]` or `dict[str, Collection[str]]` due to mypy's invariance handling of recursive Mapping types.

## Reproduction
```python
def fn(d: dict[str, str]) -> None:
    j = {"foo": d, "bar": "hi"}
    requests.post("https://example.com", json=j)  # mypy error
```

## Fix
Replaced the recursive `Mapping[str, "JsonType"]` with `Mapping[str, Any]`. This matches how Python's `json.dumps` actually accepts any serializable value. `Any` is the standard pattern used by CPython's `json` module stubs for the same purpose.

## Testing
- 407 relevant tests pass (all pre-existing failures are network-dependent)
- No new mypy errors introduced
- `pyright` and `ty` continue to work correctly

### DIFF
--- a/src/requests/_types.py
+++ b/src/requests/_types.py
@@ -144,7 +144,7 @@ if TYPE_CHECKING:
         | float
         | str
         | Sequence["JsonType"]
-        | Mapping[str, "JsonType"]
+        | Mapping[str, Any]
     )
 
     # TypedDicts for Unpack kwargs (PEP 692)


nateprewitt:
Hi @driphtyio, as was said in the issue, we're going to wait to get more feedback before making changes. I don't think we're ready for a PR yet.

PR #7457, opened 2026-05-17
Fix model class runtime type hints

### LINKED ISSUE #7456: typing.get_type_hints() fails on Response, PreparedRequest, and Request after 2.34.0
When preparing an update to 2.34.2 in Fedora Linux, I've found what I think is a bug in requests.

Since 2.34.0, calling typing.get_type_hints() on any of the three public model classes — or on any subclass of them — raises a NameError at runtime. The minimal reproducer is:

```python
import typing
from requests import Response
typing.get_type_hints(Response)
# NameError: name 'RequestsCookieJar' is not defined
```

The same failure occurs for Request and PreparedRequest.

The 2.34.0 inline typing work introduced three things to models.py simultaneously that interact badly:

1. from `__future__` import annotations (PEP 563) — all annotations are now stored as strings instead of being evaluated eagerly.
2. Class-level annotations on the model classes that reference RequestsCookieJar, CookieJar, and HTTPAdapter:
  - Request.cookies: RequestsCookieJar | CookieJar | dict[str, str] | None
  - PreparedRequest._cookies: RequestsCookieJar | CookieJar | None
  - Response.cookies: RequestsCookieJar
  - Response.connection: HTTPAdapter
3. All three of those names (RequestsCookieJar, CookieJar, HTTPAdapter) are imported only inside if TYPE_CHECKING: — so they are absent from the module's runtime namespace.

When typing.get_type_hints() is called on any of these classes, it evaluates the stored annotation strings in requests.models's runtime globals. The names are not there, so a NameError is raised.

Note that sessions.py gets this right: it imports RequestsCookieJar unconditionally (not under TYPE_CHECKING), so get_type_hints(Session) works correctly. models.py needs to do the same.

Any library that subclasses Response, PreparedRequest, or Request and uses a tool that calls get_type_hints() — such as attrs.resolve_types() (used by cattrs for serialization) — is broken. A confirmed real-world case is requests-cache + requests-ratelimiter: caching any response fails with the NameError because cattrs calls resolve_types(CachedResponse) when building the unstructuring function for a class that inherits from Response.

cyphercodes:
## Summary

Fixes `typing.get_type_hints()` for the inline-typed model classes by making the annotation-only names resolvable at runtime.

This keeps the new annotations usable for `requests.Request`, `requests.PreparedRequest`, and `requests.Response` without importing adapters into `models.py` at runtime.

Fixes #7456

## Test plan

- `PYTHONPATH=src python3 -m pytest -q tests/test_requests.py::test_model_class_type_hints_resolve`
- `uv run --group test pytest -q tests/test_requests.py`
- `uv run --group typecheck pyright src/requests`
- `ruff check src/requests/_types.py src/requests/models.py tests/test_requests.py`
- `ruff format --check src/requests/_types.py src/requests/models.py tests/test_requests.py`
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/src/requests/_types.py
+++ b/src/requests/_types.py
@@ -181,3 +181,22 @@ if TYPE_CHECKING:
 
         params: ParamsType
         json: JsonType
+else:
+    UriType: TypeAlias = str | bytes
+    ParamsType: TypeAlias = Any
+    KVDataType: TypeAlias = Any
+    RawDataType: TypeAlias = Any
+    StreamDataType: TypeAlias = SupportsRead[str | bytes]
+    EncodableDataType: TypeAlias = Any
+    DataType: TypeAlias = Any
+    BodyType: TypeAlias = Any
+    HeadersType: TypeAlias = Mapping[str, str | bytes] | None
+    CookiesType: TypeAlias = Any
+    FilesType: TypeAlias = Any
+    AuthType: TypeAlias = Any
+    TimeoutType: TypeAlias = float | tuple[float | None, float | None] | None
+    ProxiesType: TypeAlias = MutableMapping[str, str]
+    HooksType: TypeAlias = dict[str, list[HookType]] | None
+    VerifyType: TypeAlias = bool | str
+    CertType: TypeAlias = str | tuple[str, str] | None
+    JsonType: TypeAlias = Any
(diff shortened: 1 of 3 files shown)


nateprewitt:
Hi @cyphercodes, I don't think this is the approach we intend to take regarding #7456. We'll discuss the issue in the issue before making patches.

Issue #7459, opened 2026-05-18
`Session.get` docstring (added in #7429) is missing indentation, breaks Sphinx builds since v2.34.0


aldbr:
Since v2.34.0, `Session.get`'s docstring has a `:param params:` field whose continuation line isn't indented under the field. docutils treats the field list as ending mid-block, producing:

```
WARNING: Field list ends without a blank line; unexpected unindent. [docutils]
```

Any Sphinx project that documents a subclass of `requests.Session` via `:inherited-members:` (or `automodapi`, etc.) now emits this warning. Projects building docs with `-W` / `fail_on_warning: true` (e.g. the Read the Docs default for many projects) fail outright.

## Expected Result

`sphinx-build` renders `Session.get`'s docstring without docutils warnings, as was the case in v2.33.x.

## Actual Result

```
.../mymodule.py:docstring of requests.sessions.Session.get:5: WARNING: Field list ends without a blank line; unexpected unindent. [docutils]
```

## Reproduction Steps

```bash
pip install 'requests==2.34.2' sphinx
```

`mymodule.py`:
```python
from requests import Session

class MySession(Session):
    pass
```

`conf.py`:
```python
import os, sys
sys.path.insert(0, os.path.abspath("."))
project = "repro"
extensions = ["sphinx.ext.autodoc"]
```

`index.rst`:
```rst
Repro
=====

.. autoclass:: mymodule.MySession
   :members:
   :inherited-members:
```

Then:
```bash
sphinx-build -b html -W . _build
```

Output:
```
.../mymodule.py:docstring of requests.sessions.Session.get:5: WARNING: Field list ends without a blank line; unexpected unindent. [docutils]
build finished with problems, 1 warning (with warnings treated as errors).
```

## System Information

```json
{
  "requests": {"version": "2.34.2"},
  "sphinx": {"version": "9.1.0"},
  "docutils": {"version": "0.22.4"},
  "implementation": {"name": "CPython", "version": "3.12.9"},
  "platform": {"system": "Linux", "release": "6.17.0-20-generic"}
}
```

nateprewitt:
Resolved with #7460.

PR #7460, opened 2026-05-18
docs: indent continuation line in Session.get :param params:

aldbr:
Aims at fixing #7459


nateprewitt (approved):
Yep, it looks like this was missed copy-pasting the comment from api.py. Thanks, @aldbr!

PR #7463, opened 2026-05-18
Fix HTTPDigestAuth UTF-8 username/password encoding

driphtyio:
This fixes an issue where non-latin credentials passed as UTF-8 bytes fail because they're incorrectly encoded to latin1 during Basic Auth header construction.\n\nThe current implementation uses\nif isinstance(username, str):\n    username = username.encode('latin1')\nwhich fails when receiving UTF-8 encoded bytes in . The correct behavior is to use UTF-8 consistently when the input is already bytes.\n\nThis PR changes the encoding to UTF-8 to properly handle international usernames and passwords.

### DIFF
--- a/src/requests/_types.py
+++ b/src/requests/_types.py
@@ -144,7 +144,7 @@ if TYPE_CHECKING:
         | float
         | str
         | Sequence["JsonType"]
-        | Mapping[str, "JsonType"]
+        | Mapping[str, Any]
     )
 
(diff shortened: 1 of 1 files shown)


nateprewitt:
I think we stated in the issue we'd be waiting to get more feedback. Right now there are a few reported instances of this, but they have all been quickly remedied. Given the size of Requests user base, this is looking more like an outlier than a core issue.

PR #7487, opened 2026-05-25
FIX: no_proxy env ignored on 302 redirect (fix #3296)

### LINKED ISSUE #3296: no_proxy env ignored on 302 redirect
Hello,

I have a weird problem using Requests and proxies.
I have https_proxy and no_proxy env variable for a domain and it seems the no_proxy is ignored on a HTTP 302 redirect target.
What I do is basically:
```
s = requests.Session()
r = s.get("https://use-the-proxy.com")
```
The response is 302 Found (Location: https://do-not-use-the-proxy.com) and Requests tries to follow.
Here the problem: even if do-not-use-the-proxy.com is present in no_proxy, Requests still use the proxy for the request following a 302(the 302 "target").
If I directly request https://do-not-use-the-proxy.com, the proxy is not used as expected.

Thanks,

-- 
Nico

JosefVacha:
## Summary
Fix `no_proxy` environment variable being ignored on 302 redirects in `requests`.

## Root Cause
In `resolve_proxies()` function in `utils.py`, when `should_bypass_proxies()` returns `True` (URL matches `no_proxy`), the code skipped setting new proxy but **returned the original `proxies` dict with proxy settings**.

This meant that after a 302 redirect to a `no_proxy` domain, the original proxy was still being used.

## Proposed Changes
- In `resolve_proxies()`: return empty dict `{}` when `should_bypass_proxies()` is `True`
- Previously: returned `new_proxies` (copy of original proxies) even when URL matched `no_proxy`
- Now: correctly returns empty dict to bypass proxy

## Verification
- ✅ All 3 new tests pass
- ✅ `no_proxy` bypass works correctly after redirects
- ✅ Existing proxy behavior preserved for non-`no_proxy` URLs

Fixes #3296.

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -927,9 +927,14 @@ def resolve_proxies(
     url = cast(str, request.url)
     scheme = urlparse(url).scheme
     no_proxy = proxies.get("no_proxy")
+    
+    # If we should bypass proxies (URL matches no_proxy), return empty dict
+    if trust_env and should_bypass_proxies(url, no_proxy=no_proxy):
+        return {}
+    
     new_proxies = proxies.copy()
 
-    if trust_env and not should_bypass_proxies(url, no_proxy=no_proxy):
+    if trust_env:
         environ_proxies = get_environ_proxies(url, no_proxy=no_proxy)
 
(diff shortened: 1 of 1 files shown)


gowdy:
I think I'm seeing this issue. Was this fix not right?

PR #7490, opened 2026-05-31
Add AI Policy

sigmavirus24:
The immense influx of low-quality pull requests, many of which are duplicated tens of times, are creating unnecessary toil. As such, we now need to spell out what used to be basic, decent, and reasonable open source behaviour because many are just using LLMs to spam and reputation farm instead of trying to make meaningful contributions.


nateprewitt (approved):
:shipit:

PR #7496, opened 2026-06-04
Improve static typing of 3rd party imports 

jorenham:
This adds `chardet`, `cryptography`, and `PyOpenSSL` to the `typecheck` dependency group. This way, Pyright is able to resolve these imports, removing the need for a bunch of `# type: ignore` comments. 

I verified that mypy (2.1.0) doesn't report any new errors.

And in case anyone is wondering: I didn't use any AI for this (I'm pretty sure I'm a human).

### DIFF
--- a/src/requests/compat.py
+++ b/src/requests/compat.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import importlib
 import sys
 from types import ModuleType
+from typing import TYPE_CHECKING
 
 # -------
 # urllib3
@@ -46,7 +47,10 @@ def _resolve_char_detection() -> ModuleType | None:
     return chardet
 
 
-chardet = _resolve_char_detection()
+if TYPE_CHECKING:
+    import chardet
+else:
+    chardet = _resolve_char_detection()
 
 # -------
 # Pythons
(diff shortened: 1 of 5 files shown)


sigmavirus24 (approved):

sigmavirus24:
Thank you

src/requests/compat.py
```diff
@@ -46,7 +47,10 @@ def _resolve_char_detection() -> ModuleType | None:
     return chardet
 
 
-chardet = _resolve_char_detection()
+if TYPE_CHECKING:
+    import chardet
+else:
+    chardet = _resolve_char_detection()
```
nateprewitt, line 53: I think this change isn't quite right. We're now type checking against the wrong character detection library and losing type nuance for what actually gets returned by `_resolve_char_detection` at runtime.

That's not the end of the world, because I don't think we're truly making use of the 3p types, but is probably something we want to align with the actual default behavior of Requests in the future.

jorenham: Sharp. But unfortunately it's not possible to express the union of specific module types in Python's type system, i.e.

```py
chardet: typing.Module["chardet"] | typing.Module["charset_normalizer"]
```

isn't possible.

So this is the best we can do here, I'm afraid.

PR #7498, opened 2026-06-04
Add type annotation for `Request.hooks`

jorenham:
The typestats dashboard (https://jorenham.github.io/typestats/dashboard/report/#requests)  showed that this instance attribute wasn't typed yet. And I thought it would be nice if we could get the type coverage % of `requests` to 100%, because it's currently (`requests==2.34.2`) at 97.9% (so close!).


nateprewitt (approved):
Thanks, @jorenham!

PR #7499, opened 2026-06-06
Document pickling custom transport adapters

### LINKED ISSUE #6255: Mention pickling in the docs for creating custom Transport Adapters
While subclassing `HTTPAdapter` as well as `Session` the class variable `__attrs__` needs to be appended with any custom instance variables. Otherwise, the instance breaks during pickling, which can happen in many different scenarios - for example, during Django's `setUpTestData` deep copying/in-memory data isolation. I think it would be nice if this is mentioned in the documentation. 

There's a classic example of a custom adapter for setting a timeout value that is mentioned in multiple blog posts. Unfortunately, all of these examples are essentially broken because of a lack of information about pickling and `__attrs__` class variable. 

And yet the pattern of overriding `__attrs__` exists in many open source libs, for example:

https://github.com/requests/toolbelt/blob/master/requests_toolbelt/adapters/ssl.py#L46
https://github.com/docker/docker-py/blob/main/docker/transport/unixconn.py#L57

If this pattern is so commonly used and needed, it might be a good idea to add it to the docs. Again, this is because so many people can stumble upon this issue.

wayyoungboy:
Fixes #6255.

## Summary
- document that custom Transport Adapters with extra instance state should extend `HTTPAdapter.__attrs__`
- update the SSL adapter example to store `ssl_version` as pickled adapter state

## Tests
- Python assertion check for the new docs text and example syntax
- `git diff --check -- docs/user/advanced.rst`

### DIFF
--- a/docs/user/advanced.rst
+++ b/docs/user/advanced.rst
@@ -1024,10 +1024,22 @@ library to use SSLv3::
     class Ssl3HttpAdapter(HTTPAdapter):
         """"Transport adapter" that allows us to use SSLv3."""
 
+        __attrs__ = HTTPAdapter.__attrs__ + ["ssl_version"]
+
+        def __init__(self, ssl_version=ssl.PROTOCOL_SSLv3, **kwargs):
+            self.ssl_version = ssl_version
+            super().__init__(**kwargs)
+
         def init_poolmanager(self, connections, maxsize, block=False):
             self.poolmanager = PoolManager(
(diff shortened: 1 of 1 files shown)


DTiming24:
Good docs addition. Showing the custom adapter `__attrs__`/`__init__` pattern makes the pickling story much clearer for people extending `HTTPAdapter`.

PR #7502, opened 2026-06-08
Fix `_encode_files` detection for `__getattr__`-based file wrappers

LoicPernod:


LoicPernod:
This is a follow up to https://github.com/psf/requests/pull/7433, that fixes support for NamedTemporaryFile following https://github.com/psf/requests/commit/561e4b6889f53584c39a67f8794c53a414f68481.

When checking out 2.33.1, the tests pass without the source change.
Without this PR, the `test_post_named_tempfile` test fails with:

```TypeError: a bytes-like object is required, not '_TemporaryFileWrapper'```

File: tests/test_requests.py [hunk 1]
```diff
@@ -1087,6 +1087,32 @@ def test_different_encodings_dont_break_post(self, httpbin):
             )
         assert r.status_code == 200
 
+    def test_post_tempfile(self, httpbin):
```
nateprewitt: I'm not sure we need both of these tests. Either class encompasses the behavior we're trying to guard against, and if we truly need both, the test should be parameterized.

I'd bias towards removing the second though.

tests/test_requests.py
```diff
@@ -1087,6 +1087,32 @@ def test_different_encodings_dont_break_post(self, httpbin):
...
+        with tempfile.TemporaryFile(mode="w+") as f:
+            f.write("temp file contents\n")
+            f.seek(0)
+            r = requests.post(
+                httpbin("post"),
+                data={"stuff": json.dumps({"a": 123})},
+                params={"blah": "asdf1234"},
```
nateprewitt: Do we actually need `data` or `params`? The behavior should trigger with just `files`. Let's only use that to keep the tests intent clear.

nateprewitt (changes requested):
Thanks for this, @LoicPernod! I have a couple changes we should make but otherwise this seems reasonable. Also, please make sure you run pre-commit on the change before you push up the next round of changes.

tests/test_requests.py
LoicPernod: In that case, it would be better to parametrize or keep only the second, as that is the one that returns a _TemporaryFileWrapper (except on non Posix systems [source](https://github.com/python/cpython/blob/4fc4e3a5ac30ba4bccdcc7f475ef43a40d904b74/Lib/tempfile.py#L623)), which is the class that causes the issue here.

tests/test_requests.py
LoicPernod: Yes, we could remove them, good call

tests/test_requests.py
nateprewitt: Sure, `NamedTemporaryFile` is fine for the test if that catches the wrapper issue.

nateprewitt (approved):
Thanks again, @LoicPernod!

PR #7506, opened 2026-06-09
`requests.help` static typing improvements

jorenham:
Static type-checkers aren't very good at dealing with `try/except` blocks. Mypy, for example, reported 5 type errors in `help.py` because of this.

By moving these optional imports into an `if TYPE_CHECKING` block, static type-checkers will know how to deal with them. Even if these imports aren't found (because they're not installed), they'll be treated as `Any`, which won't lead to any typing issues (according to the [gradual guarantee](https://typing.python.org/en/latest/spec/concepts.html#the-gradual-guarantee)).

This also fixed another mypy error (related to mypy's greedy dict value type inference) by adding an inline annotation. 

The final score is -6 mypy errors (0 left in `help.py`), and +1.02% type coverage (according to `pyrefly report`) for a grand total of 99.30% type coverage (roughly speaking that's the number of things that *are* typed / the number of things that *can* be typed).

### DIFF
--- a/src/requests/help.py
+++ b/src/requests/help.py
@@ -4,32 +4,39 @@ import json
 import platform
 import ssl
 import sys
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import idna
 import urllib3
 
 from . import __version__ as requests_version
 
-try:
-    import charset_normalizer
-except ImportError:
-    charset_normalizer = None
-
-try:
+if TYPE_CHECKING:
     import chardet
-except ImportError:
-    chardet = None
-
-try:
-    from urllib3.contrib import pyopenssl
-except ImportError:
-    pyopenssl = None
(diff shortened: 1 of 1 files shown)


DTiming24:
Helpful typing cleanup. Narrowing the `try/except` surface in `help.py` should make static analysis a bit happier without changing runtime behavior, which is exactly the kind of low-risk improvement that keeps maintenance cost down.

sigmavirus24 (approved):
@nateprewitt thoughts?

nateprewitt:
I forgot to come back to this one. `help.py` is never intended to be used by other code. My initial impulse was we should ignore typing on this entirely. I don't feel strongly about it though, I'm fine if we want to merge.

sigmavirus24:
I don't feel strongly either

PR #7507, opened 2026-06-11
Use str.(r)partition instead of .(r)split where appropriate

scop:
Microbenchmark'y, (r)partition is much faster than (r)split on first/last only access.

(r)partition might be a bit harder to reason about though.


nateprewitt:
Hi @scop, I'm not sure this is a needed change. These are not hot-path operations and function fine today. A quick benchmark shows <20 ns of perf delta in the context of a 10-100+ ms HTTP request. We're talking about a ~0.00002% delta which is effectively noise.

PR #7513, opened 2026-06-13
Make RequestsCookieJar.popitem() work

### LINKED ISSUE #6190: requests.cookies.popitem() does not work
requests.cookies.popitem() doesn't seem to work even if cookies is not empty.

## Expected Result

remove and return some (key, value) pair[¶](https://requests.readthedocs.io/en/latest/api/#requests.cookies.RequestsCookieJar.popitem)
as a 2-tuple; but raise KeyError if D is empty.

## Actual Result

It always raises KeyError even when it's not empty.

## Reproduction Steps

```python
import requests

r = requests.get('https://google.com')
print(len(r.cookies))

r.cookies.popitem()
```

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": "4.0.0"
  },
  "charset_normalizer": {
    "version": "2.0.10"
  },
  "cryptography": {
    "version": "3.4.8"
  },
  "idna": {
    "version": "2.10"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.9.1"
  },
  "platform": {
    "release": "10",
    "system": "Windows"
  },
  "pyOpenSSL": {
    "openssl_version": "101010cf",
    "version": "20.0.1"
  },
  "requests": {
    "version": "2.28.1"
  },
  "system_ssl": {
    "version": "1010107f"
  },
  "urllib3": {
    "version": "1.26.3"
  },
  "using_pyopenssl": true
}
```

vineethsaivs:
Closes #6190

`RequestsCookieJar` subclasses both `CookieJar` and `MutableMapping[str, str | None]`. Its `__iter__` is inherited from `CookieJar` and yields `Cookie` objects rather than names. The `popitem()` it inherited from `MutableMapping` does roughly:

```python
key = next(iter(self))   # a Cookie object, not a name
value = self[key]        # __getitem__ looks up by name -> KeyError
```

so `popitem()` always raised `KeyError`, even on a non-empty jar:

```python
>>> jar = requests.cookies.RequestsCookieJar()
>>> jar.set("a", "1"); jar.set("b", "2")
>>> jar.popitem()
KeyError: "name=Cookie(version=0, name='a', ...), domain=None, path=None"
```

This overrides `popitem()` to remove and return a `(name, value)` pair using the jar's existing dict-like view (`iteritems()`), and to raise `KeyError` only when the jar is empty, matching the documented behavior and the `MutableMapping[str, str | None]` contract.

Added a regression test (`test_cookie_popitem`) that pops both entries, checks each returned pair is a real `(name, value)` from the jar, and confirms a `KeyError` on the empty jar. It fails on `main` and passes with this change.

### DIFF
--- a/src/requests/cookies.py
+++ b/src/requests/cookies.py
@@ -299,6 +299,20 @@ class RequestsCookieJar(CookieJar, MutableMapping[str, str | None]):  # type: ig
         """
         return list(self.iteritems())
 
+    def popitem(self) -> tuple[str, str | None]:
+        """Dict-like popitem() that removes and returns a (name, value) pair,
+        or raises KeyError if the jar is empty.
+
+        The inherited ``MutableMapping.popitem`` does not work here because
+        ``__iter__`` yields ``Cookie`` objects rather than names.
+        """
+        try:
+            name, value = next(self.iteritems())
+        except StopIteration:
+            raise KeyError("popitem(): cookie jar is empty") from None
+        del self[name]
+        return name, value
+
     def list_domains(self) -> list[str]:
         """Utility method to list all the domains in the jar."""
         domains: list[str] = []
--- a/src/requests/cookies.py
+++ b/src/requests/cookies.py
@@ -307,11 +307,14 @@ class RequestsCookieJar(CookieJar, MutableMapping[str, str | None]):  # type: ig
         ``__iter__`` yields ``Cookie`` objects rather than names.
         """
         try:
-            name, value = next(self.iteritems())
+            cookie = next(iter(self))
         except StopIteration:
             raise KeyError("popitem(): cookie jar is empty") from None
-        del self[name]
-        return name, value
+        # Remove the exact cookie the iterator selected. Deleting by name alone
+        # (``del self[name]``) would drop every cookie sharing that name across
+        # other domains/paths, not just the one returned here.
+        self.clear(cookie.domain, cookie.path, cookie.name)
+        return cookie.name, cookie.value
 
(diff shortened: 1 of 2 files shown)


avinashkamat48 (commented):
`popitem()` removes by cookie name only (`del self[name]`) after reading the first `(name, value)` pair. That can remove more than the single cookie returned when the jar contains multiple cookies with the same name on different domains/paths, or hit the existing duplicate-name conflict behavior. Please add a test with two cookies sharing a name but differing by domain/path and remove the exact cookie object selected by the iterator rather than deleting by name alone.

vineethsaivs:
Good catch, fixed. popitem() now selects the cookie via the iterator and removes that exact object with `clear(cookie.domain, cookie.path, cookie.name)`, so it can no longer drop other cookies that happen to share the name on a different domain or path. Added a regression test (`test_cookie_popitem_removes_only_the_selected_cookie`) that puts two cookies named `dup` on different domains and asserts popitem() removes exactly the one it returns and leaves the other intact.

DTiming24:
Nice to see `RequestsCookieJar.popitem()` behaving like a mapping again. A follow-up test for the empty-jar `KeyError` path might be useful if downstream code depends on that exception.

vineethsaivs:
Good call, added `test_cookie_popitem_on_empty_jar_raises_keyerror` covering exactly that: `popitem()` on a freshly constructed jar raises `KeyError`, matching `dict.popitem()`. (The existing `test_cookie_popitem` reaches the empty case by popping a populated jar to zero; this isolates the never-populated jar.)

PR #7520, opened 2026-06-16
Keep Link header parameter values that contain '='

nyxst4ck:
## Summary

`parse_header_links` splits each `Link` header parameter on `=` with
`key, value = param.split("=")` (no `maxsplit`). RFC 8288 permits quoted
parameter **values** that themselves contain `=` (pagination/cursor tokens,
base64 values, signed URLs, etc.). Such a value produces 3+ parts and raises
`ValueError: too many values to unpack`, which the bare `except ValueError:
break` swallows — silently dropping that parameter **and every parameter after
it** in the link.

```python
>>> import requests
>>> r = requests.Response()
>>> r.headers["Link"] = '<https://api.example.com/items>; rel="next"; title="a=b"'
>>> r.links
{'next': {'url': 'https://api.example.com/items', 'rel': 'next'}}   # 'title' lost
```

## Fix

`param.split("=")` → `param.split("=", 1)` in `parse_header_links`, so the value
keeps its `=` and following parameters are preserved.

## Tests

Added a `test_parse_header_links` parametrization with a value containing `=`
(`title="a=b"` followed by `rel="next"`). It fails on `main` (both `title` and
the trailing `rel` are dropped) and passes with the fix. `tests/test_utils.py::test_parse_header_links`
is green (6 passed); `ruff check`/`ruff format` clean.

No existing issue tracked this; the bug is long-standing.

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -988,7 +988,7 @@ def parse_header_links(value: str) -> list[dict[str, str]]:
 
         for param in params.split(";"):
             try:
-                key, value = param.split("=")
+                key, value = param.split("=", 1)
             except ValueError:
                 break
 
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -962,6 +962,28 @@ def default_headers() -> CaseInsensitiveDict[str]:
     )
 
 
+def _parse_header_links_params(value: str) -> list[str]:
+    """Split Link header parameters without splitting quoted values."""
+    params: list[str] = []
+    start = 0
+    in_quotes = False
+    escaped = False
+
+    for index, char in enumerate(value):
+        if escaped:
+            escaped = False
+        elif char == "\\" and in_quotes:
+            escaped = True
+        elif char == '"':
+            in_quotes = not in_quotes
+        elif char == ";" and not in_quotes:
+            params.append(value[start:index])
+            start = index + 1
+
+    params.append(value[start:])
+    return params
+
+
 def parse_header_links(value: str) -> list[dict[str, str]]:
     """Return a list of parsed link headers proxies.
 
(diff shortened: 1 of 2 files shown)


DTiming24:
Good fix for quoted parameters containing `=`. I like that the test covers the RFC 8288-style case; it might also be worth checking a quoted value that contains both `=` and `;` to prove the split stays local.

nyxst4ck:
Added the quoted semicolon regression and a quote-aware parameter splitter, then rebased onto current main. tests/test_utils.py: 230 passed, 1 skipped. Full suite: 632 passed, 3 skipped; one unrelated Windows TLS test could not find tests/certs/valid/ca/ca.crt.

Sanjays2402:
the inner param split is quote-aware now, but the outer `re.split(", *<", value)` that separates multiple links still isn't — a comma inside a quoted param value splits one link into two. `parse_header_links('<http://x/a>; title="a,<b"; rel="next"')` comes back as two bogus entries. RFC 8288 quoted-strings can contain commas, so it's the same input class you're fixing here.

Samielakkad (commented):
Retracted.

nyxst4ck:
Good catch, thanks — that's a real gap. Fixed by making the entry splitter quote-aware too (same state-machine approach as `_parse_header_links_params`): a `,` immediately followed by `<` no longer splits while inside a quoted value.

Added `'<...front.jpeg>; title="a,<b"; rel="next"'` as a regression case, plus one with a real second link entry after it to prove the split still fires correctly. `tests/test_utils.py -k parse_header_links`: 9/9 passed. Full `test_utils.py`: 232 passed, 1 skipped (unrelated). Ruff check/format clean.

PR #7538, opened 2026-06-24
Preserve username-only URL auth

lin-hongkuan:
## Summary

get_auth_from_url() now keeps a URL username when the URL has no password component instead of dropping the username.

## Why

A username-only URL authority still carries authentication information. Returning an empty auth tuple loses that information and differs from the behavior for URLs that include both username and password.

## Tests

- python -m pytest tests/test_utils.py::test_get_auth_from_url tests/test_utils.py::test_prepend_scheme_if_needed tests/test_utils.py::test_urldefragauth -q
- git diff --check origin/main..HEAD

## AI-assisted

Codex helped prepare this focused change. I reviewed the diff and ran the tests above before submitting.

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -1075,13 +1075,10 @@ def get_auth_from_url(url: str) -> tuple[str, str]:
     """
     parsed = urlparse(url)
 
-    try:
-        # except handles parsed.username/password being None
-        auth = (unquote(parsed.username), unquote(parsed.password))  # type: ignore[arg-type]
-    except (AttributeError, TypeError):
-        auth = ("", "")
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hi @lin-hongkuan, do you have a concrete scenario where you've actually encountered this or is it something that Codex derived?

Given this code has been in place for ~15 years with billions of downloads and no one has hit this issue, we're talking about an extreme edge case. I agree this _could_ be more tolerant but it's an arguably breaking change. Without a strong reason, I don't think we'd accept this in 2.x.

nateprewitt:
Nevermind, I see this is an OpenClaw agent on the loose.

PR #7539, opened 2026-06-24
Inline pre-commit linting

nateprewitt:
This PR moves us off of https://github.com/pre-commit/action to running pre-commit directly. This should avoid the transitive dependency pinning issues from #7536.

I've also reduced our pin update frequency for `pre-commit` to minor versions since we're getting weekly dependabot updates that aren't necessary. This matches our existing constraints for other ecosystems.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,4 +19,6 @@ jobs:
       with:
         python-version: "3.x"
     - name: Run pre-commit
-      uses: pre-commit/action@646c83fcd040023954eafda54b4db0192ce70507 # v3.0.0
+      run: |
+        python -m pip install pre-commit==4.6.0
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -19,4 +19,6 @@ jobs:
       with:
         python-version: "3.x"
     - name: Run pre-commit
-      uses: pre-commit/action@646c83fcd040023954eafda54b4db0192ce70507 # v3.0.0
+      run: |
+        python -m pip install pre-commit==4.6.0
```
nateprewitt, line 23: We could add this to a dependency group or in a requirements file. I'm not sure it's important/needed though, and I'd rather not add another requirements setup to manage.

sigmavirus24 (approved):

PR #7541, opened 2026-06-25
Parse standard cookie Expires dates from Morsel

### LINKED ISSUE #6004: For Morsel cookies requests expects wrong Expires time format
I'm not sure about this, but it looks that requests is expecting invalid date format in Expires section in Cookies passed as Morsel objects.

`morsel_to_cookie` (from `requests/cookie.py`) function parse `expires` attr if there is no `max-age` (which is OK) using format from `time_template`. However this format is set as `'%a, %d-%b-%Y %H:%M:%S GMT'` which is none of allowed expires section date value described in RFC (https://www.rfc-editor.org/rfc/rfc6265#section-4.1.1).

The nearest format is the first choice from https://www.rfc-editor.org/rfc/rfc2616#section-3.3.1. The difference is that between date parts there should be a space, but requests expects dash.

See Reproduction Steps for very simple example which bases on Python std libs only and crashes.

## Expected Result

It's expected that requests properly accepts Morsel cookies when expires follows RFC https://www.rfc-editor.org/rfc/rfc2616#section-3.3.1.

## Actual Result

There is `ValueError` raised by strptime that passed value does not match format.
Example:
`ValueError: time data 'Thu, 01 Jan 1970 00:00:00 GMT' does not match format '%a, %d-%b-%Y %H:%M:%S GMT'`

## Reproduction Steps

```python
    from http.cookies import SimpleCookie
    from requests.cookies import RequestsCookieJar
    cookies = SimpleCookie()
    cookies.load('auth_session=null; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT; httponly; samesite=strict')

    jar = RequestsCookieJar()
    jar.update(cookies)
```

Example above is a simplified case of using [`async_asgi_testclient`](https://github.com/vinissimus/async-asgi-testclient) to test application written with [`Starlette`](https://www.starlette.io/) with `SessionMiddleware`. The `async_asgi_testclient` collects cookies using `SimpleCookie` class from standard `http` lib and then pass them to `requests` lib which is used to perform client test requests.

But this issue is not related to these packages as I reproduced it using only Python standard `http` lib and `requests` as above.

## System Information

    $ python -m requests.help

```json
{
  "chardet": {
    "version": "4.0.0"
  },
  "charset_normalizer": {
    "version": "2.0.9"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "3.3"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.9.9"
  },
  "platform": {
    "release": "5.15.0-1-amd64",
    "system": "Linux"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.26.0"
  },
  "system_ssl": {
    "version": "101010cf"
  },
  "urllib3": {
    "version": "1.26.7"
  },
  "using_pyopenssl": false
}

```

lin-hongkuan:
## Summary
- parse Morsel expires values with http.cookiejar.http2time instead of a single fixed strptime template
- keep invalid expires values rejected when the stdlib parser cannot parse them
- add regression coverage for RequestsCookieJar.update(SimpleCookie(...)) with a standard HTTP-date Expires value

Fixes #6004

## Testing
- python -c "import sys, pytest; sys.path.insert(0, 'src'); raise SystemExit(pytest.main(['tests/test_requests.py::TestMorselToCookieExpires', '-q']))"
- python -m ruff check src/requests/cookies.py tests/test_requests.py
- git diff --check

### DIFF
--- a/src/requests/cookies.py
+++ b/src/requests/cookies.py
@@ -9,7 +9,6 @@ requests.utils imports from here, so be careful with imports.
 
 from __future__ import annotations
 
-import calendar
 import copy
 import time
 from collections.abc import Iterator, MutableMapping
@@ -538,8 +537,9 @@ def morsel_to_cookie(morsel: Morsel[Any]) -> Cookie:
         except ValueError:
             raise TypeError(f"max-age: {morsel['max-age']} must be integer")
     elif morsel["expires"]:
-        time_template = "%a, %d-%b-%Y %H:%M:%S GMT"
-        expires = calendar.timegm(time.strptime(morsel["expires"], time_template))
+        expires = cookielib.http2time(morsel["expires"])
+        if expires is None:
+            raise ValueError(f"expires: {morsel['expires']} must be a valid HTTP date")
     return create_cookie(
         comment=morsel["comment"],
         comment_url=bool(morsel["comment"]),
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -10,6 +10,7 @@ import re
 import tempfile
 import threading
 import warnings
+from http.cookies import SimpleCookie
 from unittest import mock
 
 import pytest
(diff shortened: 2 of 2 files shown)


DTiming24:
This looks like the right compatibility bump. A test with the common RFC 822/850-style expiry forms would be a good guardrail against future regressions in cookie parsing.

PR #7552, opened 2026-06-29
Fix RFC 6265 DQUOTE handling in RequestsCookieJar.set_cookie

NullLabTests:
## Summary

`RequestsCookieJar.set_cookie` mishandles cookie values wrapped in DQUOTEs per [RFC 6265 section 5.2.3](https://datatracker.ietf.org/doc/html/rfc6265#section-5.2.3).

The old code used `value.replace("\\"", "")` which blindly removed all backslash-escaped quote sequences without stripping the outer DQUOTE wrapper. This violates the spec, which requires:
1. Stripping the outer DQUOTE characters
2. Unescaping `\"` → `"` inside

### `src/requests/cookies.py`

| Before | After |
|--------|-------|
| `value.replace("\\"", "")` | `value[1:-1].replace("\\"", "\"")` |

Also added a `len(value) >= 2` guard to avoid IndexError on single-character values.

### `tests/test_requests.py`

Added `test_cookie_quote_escaped` to verify correct RFC 6265 behavior:
- Wire: `Set-Cookie: foo="bar\"baz"` → `Cookie(value="bar"baz")`
- cookielib parses the raw value as `"bar\\"baz"` (outer DQUOTEs + escaped quote)
- After fix: `bar"baz` (outer quotes stripped, `\"` unescaped to `"`)

## Testing

- All 340 tests pass, 1 skipped, 1 xpassed (no regressions)
- Manually verified that the existing `test_cookie_quote_wrapped` (httpbin integration) continues to work correctly — httpbin uses Python SimpleCookie which applies its own quoting layer, and the fix correctly decodes it back to the original value.

### DIFF
--- a/src/requests/cookies.py
+++ b/src/requests/cookies.py
@@ -382,10 +382,11 @@ class RequestsCookieJar(CookieJar, MutableMapping[str, str | None]):  # type: ig
     def set_cookie(self, cookie: Cookie, *args: Any, **kwargs: Any) -> None:
         if (
             (value := cookie.value) is not None
+            and len(value) >= 2
             and value.startswith('"')
             and value.endswith('"')
         ):
-            cookie.value = value.replace('\\"', "")
+            cookie.value = value[1:-1].replace('\\"', '"')
         return super().set_cookie(cookie, *args, **kwargs)
 
     def update(  # type: ignore[override]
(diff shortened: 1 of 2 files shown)


nateprewitt:
Closing as duplicate.

PR #7560, opened 2026-07-05
utils: surface clear ValueError from address_in_network and dotted_netmask for bad input

HrachShah:
`address_in_network()` and `dotted_netmask()` both leak C-level errors for
bad input instead of raising the well-typed `ValueError` users expect from
a public utility.

- `dotted_netmask(33)` raised `ValueError: negative shift count` from
  inside `1 << 32 - mask`, which underflows once the shift count goes
  negative. `dotted_netmask(-1)` raised `struct.error: 'I' format
  requires 0 <= number <= 4294967295` from the subsequent
  `struct.pack(">I", bits)` call. Both are leaky C-level errors whose
  messages give no hint that the input was a network prefix length.

- `address_in_network("192.168.1.1", "192.168.1.0")` (no slash) raised
  `ValueError: not enough values to unpack` from the
  `netaddr, bits = net.split("/")` line. `address_in_network("x", "y/24")`
  raised `OSError: illegal IP address string passed to inet_aton` from
  the C-level `inet_aton`. `address_in_network("x", "y/abc")` raised
  `ValueError: invalid literal for int()` from the `int(bits)` call.
  All of these are unhelpful to a caller that was handed a malformed
  value from `NO_PROXY` or a config file.

Fix:

- `dotted_netmask` now validates the input up front: `TypeError` for a
  non-int (e.g. a string from a config parser), `ValueError` for an
  out-of-range prefix length, with the offending value and the valid
  range in the message.

- `address_in_network` now validates the CIDR shape (`/` required,
  integer prefix, prefix in 0..32) and wraps the `inet_aton` calls in
  a `try/except OSError` that re-raises as a `ValueError` naming both
  the network string and the IP. The function still returns the same
  boolean it did before for the well-formed cases the previous tests
  cover.

12 new tests in `tests/test_utils.py`:

- `TestAddressInNetwork` gets 6 new cases (missing slash, non-integer
  prefix, prefix above 32, negative prefix, invalid IP, invalid network
  address), each asserting the new typed `ValueError` and that the
  message names the offending value.
- A new `TestDottedNetmask` class adds 4 cases (in-range, above 32,
  negative, non-int), each asserting the new typed error.

Verified: pre-fix code raises the leaky C-level error for 8 of the 12
new cases (the 4 `dotted_netmask` cases plus all 4 prefix-length
edge cases for `address_in_network`); the IP/net-address cases already
raised `OSError` and the missing-slash case already raised `ValueError`
from the unpack, but with messages that don't name the input. Post-fix
all 12 cases raise the typed `ValueError`/`TypeError` with the input
named. Pre-existing `test_valid` and `test_invalid` still pass; full
`tests/test_utils.py` is 226 passed, 13 skipped, 0 failed.

### DIFF
--- a/src/requests/utils.py
+++ b/src/requests/utils.py
@@ -731,11 +731,43 @@ def address_in_network(ip: str, net: str) -> bool:
 
     :rtype: bool
     """
-    ipaddr = struct.unpack("=L", socket.inet_aton(ip))[0]
-    netaddr, bits = net.split("/")
-    netmask = struct.unpack("=L", socket.inet_aton(dotted_netmask(int(bits))))[0]
-    network = struct.unpack("=L", socket.inet_aton(netaddr))[0] & netmask
-    return (ipaddr & netmask) == (network & netmask)
+    # Validate the CIDR is well-formed before unpacking; a missing '/' made
+    # `net.split("/")` raise ValueError("not enough values to unpack")
+    # and a negative or >32 prefix made the dotted_netmask call below
+    # raise a confusing ValueError about a "negative shift count". Both
+    # cases surfaced to the user as a stack trace from inside the helper
+    # instead of an explicit error mentioning the bad CIDR.
+    if "/" not in net:
+        raise ValueError(
+            f"network {net!r} is not a valid CIDR: expected '<addr>/<bits>'"
+        )
+    try:
+        netaddr, bits = net.split("/")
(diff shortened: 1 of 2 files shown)


HrachShah:
Closing to reopen against HrachShah/requests; sent to upstream by mistake. Will submit as separate PR after review.

Issue #7574, opened 2026-07-09
Support for HTTP Query Method


yamac-kurtulus:
Hi,

I wanted to ask the if maintainers have plans to support the newly proposed QUERY http method in the near future, as it has been awaited for more than a decade. We plan to use the method in our APIs going forward, albeit most use will be internal. I see requests is matured and in feature freeze at this time, I also understand that we can send arbitrary methods by supplying the name in the "request" method, tho it has a caveat [regarding redirections](https://datatracker.ietf.org/doc/html/rfc10008#name-redirection).

I started some work on a fork. However before opening a PR, I wanted to make sure if you are open to contributions in this direction, as I have seen multiple recent rejections on the subject. (I have to add, rightfully so)

nateprewitt:
Hi @yamac-kurtulus, as you noted Requests already supports arbitrary HTTP Verbs through the `requests.request()` API. The redirection caveat is relatively trivial to handle as well, so Requests should be fully capable of supporting this as is.

QUERY is quite recent (it's only been a proposed standard for a few weeks). I don't think we plan to add an official API or specification specifics until we see how this gets implemented in common server frameworks and the actual adoption rate. Adding a formalized API is a long-lived maintenance contract that if we rush into is a net-negative for the library. We can revisit this topic in the future.

PR #7575, opened 2026-07-09
Support for QUERY http method

yamac-kurtulus:
This PR adds the QUERY http method to the API as defined in [RFC10008](https://datatracker.ietf.org/doc/html/rfc10008).
Albeit we can use `Session.request` method to send arbitrary HTTP verbs, in the case of QUERY, [the redirection rules](https://datatracker.ietf.org/doc/html/rfc10008#name-redirection) are slightly different: 
- 303s follow "Location" header with GET (no change here)
- 301 and 302s do not drop body, and not get turned into GETs

This PR adds the QUERY method, handle the redirection rules and adds tests related to it. During test, httpbin had to be monkeypatched to accept QUERY method on some paths.  Also includes a minor fix to the "AI policy" link in the "Contribution Guidelines".

Relates to issue #7574. I have been bitten by the lack of QUERY method and having to POST a body for complex retrieval operations a bunch of times, so this has been a welcoming change for me, unfortunately adoption is slow.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Before opening any issues or proposing any pull requests, please read
 our [Contributor's Guide](https://requests.readthedocs.io/en/latest/dev/contributing/)
-as well as our [AI Policy](./AI_POLICY.md).
+as well as our [AI Policy](./.github/AI_POLICY.md).
 
 To get the greatest chance of helpful responses, please also observe the
 following additional notes.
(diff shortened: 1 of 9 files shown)


nateprewitt:
I've added a response in the issue you opened. I'll close this since it's not something we'd accept at this time. If you'd like to open a separate PR to fix the link, feel free to do so.

yamac-kurtulus:
> I've added a response in the issue you opened. I'll close this since it's not something we'd accept at this time. If you'd like to open a separate PR to fix the link, feel free to do so.

Sure, why not. Have a good day.

PR #7576, opened 2026-07-09
Fix link to AI Policy in CONTRIBUTING.md

yamac-kurtulus:
The AI Policy is not accessible with a relative link inside .github folder, this PR fixes it.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Before opening any issues or proposing any pull requests, please read
 our [Contributor's Guide](https://requests.readthedocs.io/en/latest/dev/contributing/)
-as well as our [AI Policy](./.github/AI_POLICY.md).
+as well as our [AI Policy](/.github/AI_POLICY.md).
 
 To get the greatest chance of helpful responses, please also observe the
 following additional notes.
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Before opening any issues or proposing any pull requests, please read
 our [Contributor's Guide](https://requests.readthedocs.io/en/latest/dev/contributing/)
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 Before opening any issues or proposing any pull requests, please read
 our [Contributor's Guide](https://requests.readthedocs.io/en/latest/dev/contributing/)
-as well as our [AI Policy](./AI_POLICY.md).
+as well as our [AI Policy](./.github/AI_POLICY.md).
```
nateprewitt: We may need a fully qualified link here. The reason it's broken is because it's rendered at two different points.

It's correct if you're viewing [CONTRIBUTING.md](https://github.com/psf/requests/blob/main/.github/CONTRIBUTING.md) directly, but broken for when you click the [link](https://github.com/psf/requests?tab=contributing-ov-file) for Contributing on the repo.

nateprewitt (changes requested):

.github/CONTRIBUTING.md
yamac-kurtulus: Ah well I just checked it if it works on the tab, not by viewing the file on repo. Good catch.

.github/CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 Before opening any issues or proposing any pull requests, please read
 our [Contributor's Guide](https://requests.readthedocs.io/en/latest/dev/contributing/)
-as well as our [AI Policy](./.github/AI_POLICY.md).
+as well as our [AI Policy](/.github/AI_POLICY.md).
```
yamac-kurtulus, line 5: Tested on my repo, works both from the tab, and the file itself, turns out absolute path from root is sufficient, not a full path.

nateprewitt (approved):
Alright, I think this'll work. Thanks, @yamac-kurtulus!

PR #7582, opened 2026-07-15
Fix #7564: raise FileNotFoundError for missing TLS material

### LINKED ISSUE #7564: raise FileNotFoundError for missing TLS material
Would you accept a PR which changes the existing OSError for missing TLS material to FileNotFoundError?

e.g.:

```
            if conn.cert_file and not os.path.exists(conn.cert_file):
                raise OSError(
                    f"Could not find the TLS certificate file, "
                    f"invalid path: {conn.cert_file}"
                )
```

becomes something like:

```
            if conn.cert_file and not os.path.exists(conn.cert_file):
                from errno import ENOENT
                raise FileNotFoundError(
                    ENOENT,
                    "Could not find the TLS certificate file, invalid path",
                    str(conn.cert_file)
                )
```

Our use case is that there are temporary conditions where the files don't exist and it'd be nicer to handle FileNotFoundError specifically and be able to compare the .filename attribute rather than the more general OSError and test str(excp).endswith(...).  As FileNotFoundError is a subclass of OSError existing try/excepts continue to work but it would change the string representation:

```
>>> str(OSError("Could not find the TLS certificate file, invalid path: filename"))
'Could not find the TLS certificate file, invalid path: filename'
>>> str(FileNotFoundError(errno.ENOENT, "Could not find the TLS certificate file, invalid path", "filename"))
"[Errno 2] Could not find the TLS certificate file, invalid path: 'filename'"
```

Krshs90:
Resolves #7564 by replacing generic OSError with FileNotFoundError for missing TLS materials (certificates and CA bundles).

### DIFF
--- a/src/requests/adapters.py
+++ b/src/requests/adapters.py
@@ -8,6 +8,7 @@ and maintain connections.
 
 from __future__ import annotations
 
+import errno
 import os.path
 import socket  # noqa: F401
 import typing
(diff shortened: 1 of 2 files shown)


Samielakkad (commented):
Retracted.

PR #7583, opened 2026-07-15
refactor: use list() instead of list comprehension

Hamidkarimi-h:
## Summary

Replace the list comprehension used to consume a generator with `list(gen)`.

This change preserves the existing behavior while using the idiomatic way to materialize a generator into a list. No functional changes are intended.


nateprewitt:
Hi @Hamidkarimi-h, we generally won't take stylistic PRs as they're largely subjective and tend to lead to unnecessary commit churn. This won't meaningfully change performance or improve readability of the code.

